PR #367, opened 2013-07-13
[rejected] Add rlm_hmac (similar to rlm_digest), useful for STUN/TURN and possibly DNS TSIG and other

dpocock:
This is an initial cut of rlm_hmac

I've tested it with radclient

We are discussing it in the various open-source TURN server projects

https://groups.google.com/forum/#!topic/turn-server-project-rfc5766-turn-server/A7GefqomAZ0

and would appreciate any feedback from the RADIUS community about how to make this into an official part of RADIUS

### DIFF
--- a/src/tests/Makefile
+++ b/src/tests/Makefile
@@ -19,7 +19,7 @@ export DYLD_LIBRARY_PATH
 RADDB_PATH := $(top_builddir)/raddb
 
 TESTS	= user_password vpiter chap mschapv1 digest-01/digest* \
-	test.example.com wimax
+	test.example.com wimax hmac/hmac-test
 
 PORT	 = 12340
 #PORT	 = 1812
--- a/src/tests/config/test.conf
+++ b/src/tests/config/test.conf
@@ -95,6 +95,7 @@ authorize {
 	chap
 	mschap
 	digest
+	hmac
 	eap
 	pap
 }
@@ -104,6 +105,7 @@ authenticate {
 	chap
 	mschap
 	digest
+	hmac
 	eap
 }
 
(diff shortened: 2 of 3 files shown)


dpocock:
Would you mind re-opening this issue?  It is linked from another project management tool and I don't want people to see it in the closed state and think it is finished.

arr2036:
Have you gotten any closer to getting IANA space attributes allocated?  We now support the extended attribute space so it should be easier to get an allocation.

arr2036:
I've updated the title to make it clear this wasn't merged.

alandekok:
In order to get this under IANA space, you'll have to write an internet-draft, and work it through the RADEXT working group.  That will take 2-3 years.

TBH, just put the data into a vendor-specific attribute.  To be honest... you could just pick an IANA private enterprise code for a vendor that no longer exists, and use that.  No one would notice.

But PLEASE don't publish implementations using the IANA space.  It will cause problems for everyone.

arr2036:
Or just use the FreeRADIUS VSA space... Or look at how Raidator implement it (if they do), and so something similar.

alandekok:
Please don't use the FreeRADIUS VSA space without clearing it with us first.

dpocock:
Instead of having the issue in the "closed" state, could you use a label for pull requests that are not ready for merging?
https://help.github.com/articles/applying-labels-to-issues-and-pull-requests/

You could also use the "feature enhancement" label on this issue to signify that it doesn't represent a bug.

alandekok:
The problem is I don't want to have pull requests open for months.  They should either be merged quickly, or rejected.

In this case, we *cannot* merge it while you're poaching on the IANA attribute space.  Those attributes are not yours to use.  There are already too many implementations which have conflicting definitions for those attributes.  Adding one more is a big problem.

If you update the patch to use VSAs, we'll immediately merge it.

PR #615, opened 2014-04-29
3.0.x - radeapclient - parallel mode (and more stuff)

nchaigne:
I know, the diff will be useless on so many changes, but it wasn't feasible to do small commits on this one.
Anyhow, it works. Still does all what the old radeapclient was doing before, and much more.
I use it mainly for EAP-SIM, but tested also EAP-MD5 (and non EAP requests).
As always, all comments are welcome :)

- Added a "rce_request_t" structure which allows to store a request
context.
- Load all requests from files (or stdin) in memory in a list of
"rce_request_t".
- Use packet list and socket functions provided by FreeRADIUS library.
- Added support for "parallel" (option -p) and "send rate" (option -n).
- Added a "receive/send" loop, sends requests in parallel mode (similar
to radclient).
- Added statistics (option -s).
- Better handling of output according to debug/quiet (options -x and
-q). No more wilds printf's.
- Bug fix in EAP-SIM client state machine (error case).
- Updated attribute handling as from latest radclient code.

### DIFF
--- a/src/modules/rlm_perl/rlm_perl.c
+++ b/src/modules/rlm_perl/rlm_perl.c
@@ -662,7 +662,8 @@ static int pairadd_sv(TALLOC_CTX *ctx, REQUEST *request, VALUE_PAIR **vps, char 
 	VALUE_PAIR      *vp;
 
 	if (SvOK(sv)) {
-		val = SvPV_nolen(sv);
+		STRLEN len;
+		val = SvPV(sv, len);
 		vp = pairmake(ctx, vps, key, NULL, op);
 		if (!vp) {
 		fail:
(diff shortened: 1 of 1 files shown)


arr2036:
Hmm this pull request no longer has any commits :( I guess you reused the branch?

nchaigne:
Yes, I was a total newbie with git at the time and didn't know you had to create branchs for pull requests :)
I still have the source if you want it. Though I've realized since then that a lot of things could be done way better...

arr2036:
If you feel like redoing it (and splitting out the commits this time), it would be appreciated. It'd be even better if you merged the functionality with radclient, so we didn't need to maintain two utilities.

nchaigne:
Maybe I can work on it, if you don't mind it being slow (need to find the time to do it). And yes, I'll make small commits... should be easier now that I know where I want to go :)

PR #755, opened 2014-08-05
dhcpclient - allow to specify interface and send/recv at raw packet level

nchaigne:
This replace the following pull request (which can be closed):
https://github.com/FreeRADIUS/freeradius-server/pull/741

As discussed earlier on the list:
- if the client host has multiple interfaces,
- and at least one of them already has an IP address, Then the source IP
address cannot be 0.0.0.0 (even with "Packet-Src-IP-Address=0.0.0.0").
An actual IP address is automatically used as source.
This is modified by the device driver.

This patch does the following:
- Add an option to dhcpclient allowing to specify which network
interface to use.
- Open a raw socket on the low level packet interface. This allows
packet data to be left unchanged by the device driver.
- Encode Ethernet (send to ff:ff:ff:ff:ff:ff), IP and UDP layers
manually. And let FreeRADIUS do the DHCP stuff, as before.

(This required new specific socket / send / recv functions.)

The existing behaviour of dhcpclient is unchanged, it is used if the new
option -i is not set (or if destination is not broadcast).

### DIFF
--- a/src/include/dhcp.h
+++ b/src/include/dhcp.h
@@ -45,6 +45,13 @@ int fr_dhcp_encode(RADIUS_PACKET *packet);
 ssize_t fr_dhcp_decode_options(VALUE_PAIR **out, TALLOC_CTX *ctx, uint8_t const *data, size_t len);
 int fr_dhcp_decode(RADIUS_PACKET *packet);
 
+#ifdef HAVE_LINUX_IF_PACKET_H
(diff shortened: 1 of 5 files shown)


alandekok:
Closed by commit 7deeac78b.  I've made some minor changes for formatting, etc.  Please double-check.

Sorry for the delay.

PR #848, opened 2014-12-03
Talk to ntlm_auth over socket for mschap

mcnewton:
Using inetd, ntlm_auth can listen on a UNIX socket. This enables the mschap module to talk to this socket for auth, saving an exec. This *seems* to make things faster...

### DIFF
--- a/src/modules/rlm_mschap/auth_wbclient.c
+++ b/src/modules/rlm_mschap/auth_wbclient.c
@@ -1,0 +1,145 @@
+/*
+ *   This program is is free software; you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License, version 2 if the
+ *   License as published by the Free Software Foundation.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, write to the Free Software
+ *   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/**
+ * $Id$
+ * @file auth_wbclient.c
+ * @brief NTLM authentication against the wbclient library
+ *
+ * @copyright 2014  Matthew Newton
+ */
+
+RCSID("$Id$")
+
+#include <freeradius-devel/radiusd.h>
+
+#include <wbclient.h>
+
+#include "rlm_mschap.h"
+#include "mschap.h"
+#include "auth_wbclient.h"
+
+#define NT_LENGTH 24
+
+/*
+ *	Check NTLM authentication using ntlm_auth server via a socket
+ *	Returns -1 for failure and 0 on auth success
+ */
+int do_auth_wbclient(rlm_mschap_t *inst, REQUEST *request,
+		     uint8_t const *challenge, uint8_t const *response,
+		     uint8_t nthashhash[NT_DIGEST_LENGTH])
+{
+	int rcode = -1;
+	struct wbcAuthUserParams authparams;
+	wbcErr err;
+	int len;
+	struct wbcAuthUserInfo *info = NULL;
+	struct wbcAuthErrorInfo *error = NULL;
+	char user_name[500];
+	char domain_name[500];
+	uint8_t resp[NT_LENGTH];
+
+	/*
+	 * Clear the auth parameters - this is important, as
+	 * there are options that will cause wbcAuthenticateUserEx
+	 * to bomb out if not zero.
+	 */
+	memset(&authparams, 0, sizeof(authparams));
+
+
+	/*
+	 * Get the username and domain from the configuration
+	 */
+	len = radius_xlat(user_name, sizeof(user_name), request, inst->ntlm_username, NULL, NULL);
+	if (len < 0) goto done;
+
+	authparams.account_name = user_name;
+
+	if (inst->ntlm_domain) {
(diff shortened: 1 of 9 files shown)


mcnewton:
Right - got to the bottom of this one now.

libwbclient is not thread safe - it stores the file descriptor for the winbind socket in a global variable, which means the threads are always stomping on each other and using the socket in parallel. So everything goes horribly wrong. I believe the patch to libwbclient above is still required, but without it the problems just show sooner.

Putting a mutex around the call to wbcAuthenticateUserEx fixes all the problems - the unprivileged winbind socket is opened once, the name of the privileged socket read, that is closed, then the privileged socket is opened and used for the duration shared amongst all threads. No more memory problems (which were caused by a read getting the data "/var/run/...." and trying to malloc the length made from the integer with chars "/var").

The only problem then being that the whole point of this exercise is to run stuff in parallel, so this doesn't get me very far. libwbclient needs updating to store the fd in a variable local to the thread.

Cheers

Matthew

arr2036:
Simple solution would be to use TLS. It's part of C11, but most compilers support have supported ``_thread`` for a long time. That's if they really wanted to avoid changing the API.

It seems like they really don't have a concept of a connection handle, or at least it's not obvious from reading through ``wbclient.h``, so yeah, TLS would be the only way to do this without huge modifications to all the public functions. I'd prod one of the Samba guys and see what they think about the issue.

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -0,0 +1,145 @@
...
+	 */
+	switch (err) {
+	case WBC_ERR_SUCCESS:
+		rcode = 0;
+		RDEBUG("Authenticated successfully");
+		/* Grab the nthashhash from the result */
+		memcpy(nthashhash, info->user_session_key, NT_DIGEST_LENGTH);
+		break;
+	case WBC_ERR_WINBIND_NOT_AVAILABLE:
+		RERROR("Check that winbind is running and that FreeRADIUS");
```
arr2036: RERROR is considered global error, so this fine

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -0,0 +1,145 @@
...
+		RERROR("has permission to connect to the winbind socket!");
+		break;
+	case WBC_ERR_DOMAIN_NOT_FOUND:
+		RERROR("domain not found");
+		break;
+	case WBC_ERR_AUTH_ERROR:
+		RDEBUG("authentication failed (check domain is correct)");
```
arr2036: But this should be REDEBUG so Module-Failure-Message is still created.

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -0,0 +1,145 @@
...
+		RDEBUG("Authenticated successfully");
+		/* Grab the nthashhash from the result */
+		memcpy(nthashhash, info->user_session_key, NT_DIGEST_LENGTH);
+		break;
+	case WBC_ERR_WINBIND_NOT_AVAILABLE:
+		RERROR("Check that winbind is running and that FreeRADIUS");
+		RERROR("has permission to connect to the winbind socket!");
+		break;
+	case WBC_ERR_DOMAIN_NOT_FOUND:
+		RERROR("domain not found");
```
arr2036: and this should be REDEBUG because it's not a global (server) error

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -0,0 +1,145 @@
...
+		break;
+	default:
+		RDEBUG("other error %d", err);
+		break;
+	}
+
+	if (err != WBC_ERR_SUCCESS) {
+		RDEBUG("authentication failed: wbcErr %d", err);
+		if (error && error->display_string) {
+			RDEBUG("wbcErr %s", error->display_string);
```
arr2036: Should be REDEBUG... I can fix these later though if you don't want to :)

mcnewton: Thanks - I stared at those macros for a while, but obviously not long enough! Need to work out what the difference is and the use cases :-)

I will update this patch with the new Samba API (assuming that gets merged), so will fix these at the same time.

arr2036: > > +
> > +	if (err != WBC_ERR_SUCCESS) {
> > +		RDEBUG("authentication failed: wbcErr %d", err);
> > +		if (error && error->display_string) {
> > +			RDEBUG("wbcErr %s", error->display_string);
> 
> Thanks - I stared at those macros for a while, but obviously not long enough! Need to work out what the difference is and the use cases :-)

http://doc.freeradius.org/log_8h.html

I won't pretend that documentation existed before a few hours ago :)

Arran Cudbard-Bell <a.cudbardb@freeradius.org>
FreeRADIUS development team

FD31 3077 42EC 7FCD 32FE 5EE2 56CF 27F9 30A8 CAA2

alanbuxey: REALLY useful doc! :-)

mcnewton: Agreed! Thanks Arran!

arr2036: No problem. Let me know if there are any other files you want documenting.

mcnewton: All teh filez! We wants documentz for all of teh filez! :)

arr2036: http://doc.freeradius.org/tmpl_8c.html#a78ff836d3f325e5012c64e780de7feac

You can now use PW_TYPE_TMPL in a CONF_PARSER struct, to parse CONF_PAIRs as templates. The quoting of the CONF_PAIR determines what type of template is created.

Get the value in the module with ``tmpl_expand`` or ``tmpl_aexpand``.

mcnewton:
Closing this pull request - I've just done #930, which replaces it.

PR #881, opened 2015-01-20
Update rlm_yubikey to better support decrypt

P-D-D:
This is an enhancement to clarify some of rlm_yubikey setup and code (for people who need things explicitly spelled out like me), add a little support policy, and add "normifying" to the Yubikey-Key / Yubikey-Private-ID attributes.

Specifically:
* The split code now creates the Yubikey-OTP pair if it passes, even if password is zero length.
* If the optional Yubikey-Private-ID and/or Yubikey-Counter attributes exist, then they are checked.
* If the checks pass, then the pairs are created.
* If configuration has normalise set to yes (the default), then the Yubikey-Key and Yubikey-Private-ID values can be stored as a modhex string / hexadecimal string / base64 string / binary blob. Sort of like rlm_pap and passwords.
* The Yubikey-Counter was changed to a 32-bit to a 24-bit value: 16-bits session counter immediately concatenated with the 8-bit use counter.
* As it would seem the in order to reach the authenticate code, the OTP should have already been verified by the authorize code, redoing this check seemed redundant and so I marked it with an #if 0 for potential elimination (it's up to you).
* The suggested policy handles in authorize (after the SQL query) some potential mismatches.

Hopefully you'll find these modifications useful and will add them to the upstream. There is also an attendant branch for a new module rlm_sqlyubikey which updates some of the attributes post-authentication.

### DIFF
--- a/raddb/mods-available/yubikey
+++ b/raddb/mods-available/yubikey
@@ -12,7 +12,7 @@ yubikey {
 
 	#
 	#  If true, the authorize method of rlm_yubikey will attempt to split the
-	#  value of User-Password, into the user's password, and the OTP token.
+	#  value of User-Password, into the user's password and the OTP token.
 	#
 	#  If enabled and successful, the value of User-Password will be truncated
 	#  and request:Yubikey-OTP will be added.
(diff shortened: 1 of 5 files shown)


arr2036:
I'm not going to merge this as is because it's very hard to see what's changed. If you feel the code adds some useful functionality, feel free to redo the patch set, breaking out the changes into individual commits. Please avoid changing variable names and formatting arbitrarily, or at least do them in a commit at the end, which is separate from the rest of the changes.

PR #882, opened 2015-01-20
New rlm_sqlyubikey module

P-D-D:
This is an enhancement to add the ability to update the Yubikey-Counter and/or Yubikey-Timestamp attributes (if they exist). This will hopefully mitigate replay attacks, and lead to session timestamp comparison (where that could be necessary).

Hopefully you'll find these modifications useful and will add them to the upstream. There is also an attendant branch for the rlm_yubikey module which updates some of the capabilities for decrypt.

### DIFF
--- a/raddb/mods-available/sqlyubikey
+++ b/raddb/mods-available/sqlyubikey
@@ -1,0 +1,20 @@
+#
+#  This module updates Yubikey OTP token values in SQL database.
+#  Used in combination with Yubikey module when using decrypt.
+#
+#  This module should be placed in the post-auth section after the logging modules.
+#
+sqlyubikey {
+	sql_module_instance = sql
+
+	#  These attributes available after authentication (if successful):
+	#    * Yubikey-Counter   - The lastest counter value.
+	#    * Yubikey-Timestamp - The lastest timestamp value.
+	#  If these attributes do not exist for the account, they will not be updated
+
+	# These must match SQL module
+	dialect = "sqlite"
+	authcheck_table = "radcheck"
+
+	$INCLUDE ${modconfdir}/sql/yubikey/${dialect}/queries.conf
+}
--- a/raddb/mods-available/yubikey
+++ b/raddb/mods-available/yubikey
@@ -12,7 +12,7 @@ yubikey {
 
 	#
 	#  If true, the authorize method of rlm_yubikey will attempt to split the
-	#  value of User-Password, into the user's password, and the OTP token.
+	#  value of User-Password, into the user's password and the OTP token.
 	#
 	#  If enabled and successful, the value of User-Password will be truncated
 	#  and request:Yubikey-OTP will be added.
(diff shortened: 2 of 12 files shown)


arr2036:
Could you explain why code changes are necessary to use rlm_yubikey with SQL?

P-D-D:
Using the Yubikey-Counter attribute helps mitigate the issue of replaying the OTP. Once the session has been authenticated (post-auth), the new counter value needs to be updated in the database. If the attribute doesn't exist then it won't be checked.

arr2036:
Right, but the user can do that in an unlang policy. Why are *code* changes required?

P-D-D:
The name changes were to hopefully better describe what was happening in the code. I'm sorry if they don't seem to match the style. The Auth-Type setting in the configuration authentication section, or in the code?

P-D-D:
I attempted to use the same comments again, although the flow of the code changed to be more deterministic. This seems to have caused the diff to become somewhat fragmented.

P-D-D:
The intent in rlm_yubikey.c mod_authorize was to eliminate the validation of the OTP in separate places, and if split was configured to do so even if this reduced the User-Password to a zero length string.

P-D-D:
The Auth-Type setting was in the original code. I haven't the foggiest why the diff seems to show it having changed. It would seem by explicitly setting it, then for example PAP (which would in theory follow) would not also validate on only the User-Password portion (and ignore the OTP portion).

arr2036:
I don't think the SQL changes are useful, and i'd advise you to not use them for your own configuration. It's possible to achieve what you did in the policy language without any code changes. The fields from the AES block are broken out into attributes for that reason.

The fact that you thought it required code changes shows an unfamiliarity with how the server operates. I'd advise you to read man unlang, and the policies which ship with the default configuration.

Again if you believe the non SQL changes are useful, please break them out into small commits, and open a new pull request.

P-D-D:
The non SQL changes were initially committed under the yubikey branch so as to purposely separate it from the sqlyubikey changes. I'm not sure how these became mangled, and so I will attempt to commit them as the independent modules they are.

P-D-D:
I was unable to script the commits using unlang, and so following the methods for ippool, etc. I based my module on their implementation. It is not a tremendous piece of work, and so if you believe that this can be done sans code I will attempt again via policy writing.

arr2036:
If you're experiencing a particular issue then feel free to post to the users mailing list and we'll likely be able to help.

P-D-D:
The issue of executing SQL commands (updates specifically) seemed not possible sans C code. Is there an example where unlang does so that you can site for me?

arr2036:
```
update {
    Tmp-String-0 := "%{sql:UPDATE foo WHERE BAR}"
}
```

sql expansion returns number of rows updated.

P-D-D:
I had seen how expressions could be used that way, but has not found the sql module capabilities documented as such. The policy for yubikey will be ammended, and the SQL queries merged with rlm_yubikey (if that is how you would wish to proceed).

arr2036:
The documentation often lags behind server capabilities, especially user contributed documentation, it's best to ask, or even better, read the code https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_sql/rlm_sql.c#L179. I'd create a new policy.d file 'yubikey' with yubikey related policies.

P-D-D:
Done and testing. The modifications to the code will take a little longer, as it will be necessary to make it such the git diff will not match (and thus perhaps make it easier to grok).

PR #883, opened 2015-01-20
Add support for salted SHA2 and PBKDF2 passwords

P-D-D:
This is an enhancement to add to the rlm_pap module the ability to support salted SHA-2 passwords and PBKDF2 passwords. Also there are Perl scripts which should produce the appropriate digests.

Specifically:
* Following the methods used by the SSHA1 code, the SSHA2 code validates the digest of a password appended with a salt of arbitrary content and length. The code supports all 4 full size standard SHA-2 digest lengths stored as a hexadecimal string / base64 string / binary blob.
* The PBKDF2 code also uses the OpenSSL code to validate passwords digested with the HMAC-SHA standard algorithms (SHA-1 / SHA-2). The code supports all SHA-1 and the 4 standard SHA-2 digests stored as a base64 encoded string, including the algorithm (and length) / number of iterations / salt / digest delimited with semi-colons. As mainline OpenSSL does not currently support PBKDF2 operating with SHA-3, this is commented out at this moment.
* The Ssha2Password script will digest the given password using one of the 4 standard SHA-2 lengths, with a optional salt either explicitly given (in character or hexadecimal string format) or a defined cryptographically random size. The output format can be specified as a hexadecimal string / base64 string / binary stream.
* The Pbkdf2 script will digest the given password using one of the standard SHA-1 or 4 standard SHA-2 length or 4 standard SHA-3 length HMAC digests. The salt may be either explicitly given (in character or hexadecimal string format) or a defined cryptographically random size. The iterations may also be defined, as the defaults (although currently acceptable) would probably need to be tuned for the application. Optionally, simple timing of the digest generation duration can be printed for crude estimation of the computational cost. The output format is a base64 encoded string that includes the algorithm (and length) / number of iterations / salt / digest delimited with semi-colons. Although the Perl implementation supports SHA-3 unfortunately the freeradius module does not (at this moment).

Hopefully you will find these useful and include them in upstream.


src/modules/rlm_pap/rlm_pap.c
```diff
@@ -375,11 +401,19 @@ static rlm_rcode_t CC_HINT(nonnull) mod_authorize(void *instance, REQUEST *reque
 
 #ifdef HAVE_OPENSSL_EVP_H
 		case PW_SHA2_PASSWORD:
+		case PW_SSHA224_PASSWORD:
+		case PW_SSHA256_PASSWORD:
+		case PW_SSHA384_PASSWORD:
+		case PW_SSHA512_PASSWORD:
```
arr2036: Um, that's wrong, 28 bytes is the minimum expected size for SHA224, not the rest of them.

P-D-D: Unfortunately the arbitrary length of the appended salt makes precise calculation of the algorithm by the resultant overall size indeterminate. I initially tried to blind detect the hash, but this seemed crude vs. the exact attribute definition.

arr2036: Yeah, but if you know it's SSHA512 (by the format marker in the password) you know at a minimum the normified SSHA string must be at least 64bytes. In fact you probably want to make the check 65 bytes. It's SSHA, it needs at least one byte of salt.

arr2036: You're setting the minimum size to 28 for all SHA2 password types... That's my issue with this code. It will make it significantly more likely that the heuristics will misidentify the encoding of the "known good" password.

src/modules/rlm_pap/rlm_pap.c
```diff
@@ -96,6 +101,27 @@ static const FR_NAME_NUMBER header_names[] = {
...
+	int keylen;
+} rlm_pap_pbkdf2;
+
+static const rlm_pap_pbkdf2 pbkdf2_names[] = {
+	{ "HMACSHA1", EVP_sha1, 20 },
+	{ "HMACSHA2+224", EVP_sha224, 28 },
+	{ "HMACSHA2+256", EVP_sha256, 32 },
+	{ "HMACSHA2+384", EVP_sha384, 48 },
+	{ "HMACSHA2+512", EVP_sha512, 64 },
+//	{ "HMACSHA3+224", EVP_sha3_224, 28 },
```
arr2036: These should be ifdefed out with a check for the relevant OpenSSL macro if they're standardised. If they're not standardised yet (I didn't think SHA3 was), they should be removed.

src/modules/rlm_pap/rlm_pap.c
P-D-D: It would seem that the original SSHA1 code did not place delimiters in the encoding of the hash (I expect as there was one size), and the SHA2 code did not place a length in the string as it was redundant. Unfortunately if you has say an SSHA256 with a 32-byte salt, it would seem to be the size of a SSHA512.

src/modules/rlm_pap/rlm_pap.c
P-D-D: Fair enough. However the attendant Perl will generate these markers, and so I placed them as place holders (in hopes that this will soon change).

src/modules/rlm_pap/rlm_pap.c
P-D-D: Agreed, like the SHA2 code directly above. It is not perfectly optimal, and I can change both if you feel strongly about that.

P-D-D: Is there a standard you would prefer for each of the normalisable formats to place delimiters in binary blobs? It was because of this that I only implemented parsing the PBKDF2 as a base64 encoded string.

arr2036:
Unlike the other two pull requests this code is immediately useful and can be merged into the server with minor changes, @alandekok has offered to do it, but if he doesn't have time or becomes busy with something else, i'll do it in a few hours.

Thanks for implementing SSHA2, is was on the todo list :)

arr2036:
I'm going to merge the SSHA2 stuff, but you have work to do on pap_auth_pbkdf2.

You're using vp->vp_strvalue to access the value of an octets type attribute (which holds binary values). So you're doing an implicit cast on the value buffer from a uint8_t array to a char array, and then passing the value around to functions which expect standard C strings. That's not going to end well. I'm surprised you didn't get any SEGVs during testing (we do *not* \0 terminate octet type value buffers).

P-D-D:
If it pleases you, then I will alter the usage of vp_strvalue to vp_octets (although the contents of the buffer remain characters). As there is no expectation of the buffer being zero-terminated the check towards the end will be eliminated. Other than that, I believe the code stands with no issues of segfaulting.

P-D-D:
Perhaps you have misconstrued the parse of the hash which calculates the length based on the amount str has consumed (offset from the beginning of the password buffer)?

arr2036:
thanks!

P-D-D:
BTW: You may need to modify normify_with_header to a memchr() for the same reason.

arr2036:
Good suggestion, but no, normify_with_header should only be called with a Password-With-Header attribute, which is a string type.

We should add asserts though.

arr2036:
I may have to use this code for the next series of interviews we'll be running.  It's a great example of why peer review is useful.  I'm assuming you've not done that much C or haven't done it in a while? String parsing is an absolute nightmare.

P-D-D:
No, it's the expectation that the string stored is a string when read in. This has been rewritten for octets ... And no, VAXes (and other 7-bit chars) are dead.

arr2036:
I can't merge this pull request because it conflicts with the top of the master branch. Please pull --rebase, resolving conflicts, and squash all commits into a single one to add PBKDF2 support.

It's very difficult to do effective code review, when patches for the same area of code are spread out across multiple commits.

arr2036:
Could you also provide some PBKDF2 test vectors so I can add unit tests?

arr2036:
Please don't add your own test scripts. There's a perfectly good set already. see ``src/tests/keywords``.

Add your test vectors using update blocks.

arr2036:
Don't worry. I have a copy of all your commits for both pull requests, I synced your github repo locally. I'll merge them in the next few weeks.

I do think the changes are useful, but they would have required a lot of attention before they could be merged. Which is why they haven't been, yet.

PR #884, opened 2015-01-21
Updated rcradiusd and fixed radiusd-logrotate

arrad:


arr2036:
We should be using the ``copytruncate`` features of logrotate.

```
copytruncate
Truncate the original log file in place after creating a copy, instead of moving the old log file and optionally creating a new one. It can be used when some program cannot be told to close its logfile and thus might continue writing (appending) to the previous log file forever. Note that there is a very small time slice between copying the file and truncating it, so some logging data might be lost. When this option is used, the create option will have no effect, as the old log file stays in place.
```

arr2036:
Please check both the logrotate scripts, and the init scripts to make sure they function correctly.

PR #890, opened 2015-01-26
add perle dictionary

kokel:


arr2036:
Thanks! Would you recommend them for terminal servers? Been looking for a cheaper replacement for the Digi TS series (which are incredibly expensive for what you get).

kokel:
Yay, I can recommend these console servers but I don't think that these are cheaper than the digi ones, maybe more expensive, actually.

PR #898, opened 2015-02-03
Add --disable-openssl-version-check option

spbnick:
This adds the --disable-openssl-version-check, similarly to v3.0.x, to solve the same problem.
We need this to smooth the RHEL6 upgrade to v2.2.6.


spbnick:
Thanks, Alan!

PR #899, opened 2015-02-04
RPM spec file: add missing files + bug bix

pwdng:


arr2036:
Hm, why would we want rlm_test?

arr2036:
(it's literally a test module for testcases)

pwdng:
Ok, but it that case, it probably shouldn't be in the list of stable modules that get deployed.

arr2036:
Well if it's not in stable it doesn't get built, and then the tests will fail for non development builds.

PR #905, opened 2015-02-10
more attributes definitions

gureedo:
Added some missing attributes. Information taken from http://www.iana.org/assignments/radius-types/radius-types.xhtml

Also I've noted that `rlm_digest` uses old draft
https://tools.ietf.org/html/draft-sterman-aaa-sip-00
and there is a newer version 
https://tools.ietf.org/html/draft-sterman-aaa-sip-01
but it's expired in 2004 year.
Is there need to support old rfc draft? Actual rfc for this module is http://tools.ietf.org/html/rfc5090


arr2036:
radius.h is auto generated... Modify the actual dictionaries not the header file.

alandekok:
Defining more attributes in radius.h is nice.  However, the attributes need to be defined *only* if they're used.  And they're not used by the server right now.

For the digest attributes, all of the NAS vendors implement the draft.  I can't recall any NAS vendor who implements either RFC.  So... we've ignored the RFC.

arr2036:
There's no reason not to add them. Those headers should really be kept in sync with the dictionary files.

alandekok:
On Feb 10, 2015, at 7:45 AM, Arran Cudbard-Bell <notifications@github.com> wrote:

> There's no reason not to add them. Those headers should really be kept in sync with the dictionary files.

  Then they should be auto-generated. :)

gureedo:
When i want to write own module for freeradius-server, i must implement all this declarations by my self. It gets worse when there is more than one module. I think it's good when server offers good api and other declarations for developer.

PR #906, opened 2015-02-11
Move OpenSSL init out of version check

spbnick:
Initialize OpenSSL outside ssl_version_check() to execute even with
disabled version check. Otherwise SSL_CTX_new() returns zero and
FreeRADIUS segfaults in init_tls_ctx with version check disabled.


arr2036:
ah version 2, ok

spbnick:
Thanks, Arran!

arr2036:
np

PR #928, opened 2015-03-18
Abfab policy and trust router changes

hartmans:
Hi.  I'd appreciate it if these changes could be integrated.  This updates the trust router code for exposing errors to the NAS as well as allowing unlang configuration to override the community.  In addition there are significant improvements to the sample abfab configuration to provide better error handling and more functionality.

A bug fix is provided to the sqlite driver.  Currently any database contetion at all will cause a request to fail.  In practice we've found that even under very moderate load this is a significant problem, so sqlite is permitted to use a small busy-wait.

### DIFF
--- a/raddb/mods-config/attr_filter/access_reject
+++ b/raddb/mods-config/attr_filter/access_reject
@@ -15,4 +15,5 @@ DEFAULT
 	Error-Cause =* ANY,
 	Reply-Message =* ANY,
 	MS-CHAP-Error =* ANY,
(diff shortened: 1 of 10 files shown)


arr2036:
Posts going back to 2005 mentioning it, so we're probably safe. I've changed the code to make it user configurable.

spaetow:
As long as this exists in the RHEL version. I suspect you've already checked ;-)

PR #930, opened 2015-03-20
mschap auth directly to winbind via libwbclient

mcnewton:
A rewrite and tidy of pull request 848. This requires Samba master, or should hopefully also be in 4.2.1 (as yet unreleased). Patches apply cleanly to both 3.1.x and 3.0.x, and will apply to master if 993431657b79d6e532426dc3032bac391453aa56 is also applied

I've dropped the ntlm_auth helper mode stuff, as it doesn't seem worthwhile. Therefore config has simplified to two new options - winbind_username and winbind_domain. ntlm_auth overrides these for backwards compatibility.

Cheers,
Matthew


arr2036:
Can't do ``inst = inst`` use ``UNUSED`` macro in front of argument instead.

arr2036:
Looks good, just some very minor changes.

mcnewton:
Just pushed new version with those bits updated.
Cheers
Matthew

arr2036:
Merged to v3.0.x as it's unlikely to break existing installations, and it'll hopefully act as a good incentive for universities to move off 2.2.x.

mcnewton:
Thanks Arran! I did my best to keep it as simple as possible where it touches existing code so as to not affect existing behaviour (and the auth function itself is trivial anyway). So I think it should be safe for 3.0. Agreed, hopefully a good upgrade incentive, though it does still need a local compile of Samba which I'm sure will put some off for the time being.

PR #931, opened 2015-03-23
radeapclient - eap context struct

nchaigne:
Definition of a structure which will hold the EAP context information.
This will also make easier to merge with radclient later on.

As requested, small commits on purpose. If they are too small please let
me know, I will adjust :)

### DIFF
--- a/src/modules/rlm_eap/radeapclient.c
+++ b/src/modules/rlm_eap/radeapclient.c
@@ -148,8 +148,6 @@ static fr_ipaddr_t server_ipaddr;
 static bool server_addr_init = false;
 static uint16_t server_port = 0;
 static int packet_code = PW_CODE_UNDEFINED;
-static fr_ipaddr_t client_ipaddr;
-static uint16_t client_port = 0;
 
 static int rc_map_eap_methods(RADIUS_PACKET *req);
 static void rc_unmap_eap_methods(RADIUS_PACKET *rep);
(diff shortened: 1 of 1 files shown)


arr2036:
A little larger :)

The idea is to group multiple related commits together in a single pull request. The commits themselves should be small, the pull request doesn't have to be.

nchaigne:
Alright, then I'll do more small commits, and we'll try to agree when they are sufficient for a merge :)
I think it would be better for me to wait after each commit, so you can review it before I carry on. That way if you disagree with the changes I won't have lost much time.
What do you think ?

arr2036:
Sure. GitHub also keeps pull requests live. So if you push more commits to a branch a pull request originated on, they appear in the pull request.

Just comment when you want review and we can check over the latest commits.

nchaigne:
I'm done for today.
Can you please take a look before I carry on ?
Thanks!

arr2036:
Looks ok so far. Would still prefer the two utilities to be merged... but oh well...

nchaigne:
Yeah but I thought it would be too complicated to begin with (small steps, remember !). Also, I plan to do the parallelism differently than in radclient, and I didn't want to impact people using this tool.
But maybe later...

nchaigne:
Last commit for today.
Can you please take a look ?
Thanks again :)

(The union is still to be done. But it's not forgotten).

arr2036:
OK :) Yep looks ok. Just fix the issue that broke the build

arr2036:
and it's vp->vp_length (vp_length is a macro)

nchaigne:
Alright, I'm done for today.
I've also checked all the commits on radeapclient since last summer to be sure I didn't break anything changed since then.

Can you review again ?
Thanks!

If you don't have more comments, I believe this is in a good shape for a merge :)

nchaigne:
So... nothing more to add ?
Does this mean everything is ok and that it can be merged into FreeRADIUS ?
(I'm pushing a bit, I don't want this to wait one more year ^^)

PR #939, opened 2015-03-27
Improve errors on winbind authentication failure

mcnewton:
This is very nice, as Samba gives us useful messages like

  Wrong Password [0xC000006A]
  Account expired [0xC0000193]
  Must change password [0xC0000224]
  Account locked out [0xC0000234]

when the authentication fails, which are then pushed into
Module-Failure-Message so the admin knows exactly what happened.

This also now handles password expiry so mschap returns the
"change password" trigger.

### DIFF
--- a/src/modules/rlm_mschap/auth_wbclient.c
+++ b/src/modules/rlm_mschap/auth_wbclient.c
@@ -28,6 +28,7 @@ RCSID("$Id$")
 #include <freeradius-devel/rad_assert.h>
 
 #include <wbclient.h>
+#include <core/ntstatus.h>
 
 #include "rlm_mschap.h"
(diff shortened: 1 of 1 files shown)


src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -28,6 +28,7 @@ RCSID("$Id$")
 #include <freeradius-devel/rad_assert.h>
 
 #include <wbclient.h>
+#include <core/ntstatus.h>
```
mcnewton: Comes as part of Samba, so if wbclient.h is there (tested by configure script), core/ntstatus.h must be as well.

PR #940, opened 2015-03-27
DHCP-Client-FQDN is not a string

nchaigne:
See RFC 4702.
DHCP-Client-FQDN (DHCP option 81) is actually a record composed of:
- first octet: "Flags"
- second octet: "RCODE1"
- third octet: "RCODE2"
- and then "Domain Name" (which is a string)

But I don't think that FreeRADIUS dictionaries support encoding /
decoding such a format, so octets is the best option we have.


nchaigne:
Mmh GitHub seems in very bad shape today, got lots of errors trying to push this commit.
And now Travis failed also because of GitHub... :'(

alandekok:
github was having a DDoS attack.

PR #958, opened 2015-04-16
radeapclient - filters and passwords authentication methods

nchaigne:
Another set of changes to radeapclient 3.1.x.

These changes do not add "new" features, but rather implement things from radclient that are (were) still missing in radeapclient.

The changes are the following (see each commit log for detail):
- Handle PAP, CHAP and MS-CHAP as in radclient.
- Handle filters as in radclient.

I've copied entire functions from radclient, such as mschapv1_encode. But this should be in a library, I think.

Feel free to yell if things are not as you want them to be :)


arr2036:
could you rebase, sorry it took to long to review this.

Issue #830, opened 2014-11-07
Prepared queries in SQL.


alandekok:
This document refers to PostgreSQL, but other databases should be similar.  These notes are rough, but the basic idea is there.

Executing prepared queries: http://www.postgresql.org/docs/9.3/static/libpq-exec.html

Creating prepared queries (syntax): http://www.postgresql.org/docs/9.2/static/sql-prepare.html
## Goal

To have SQL sub-modules automatically create prepared queries and use them.

The queries are per connection, so we update:

`rlm_sql.h`

```
typedef struct rlm_sql_handle {
    void        *conn;  //!< Database specific connection handle.
    rlm_sql_row_t   row;    //!< Row data from the last query.
    rlm_sql_t   *inst;  //!< The rlm_sql instance this connection belongs to.
    rbtree_t        *prepared;  //!< prepared queries
} rlm_sql_handle_t;
```

`prepared` is handled completely by the SQL sub-module.

The source of the queries (config, xlat, etc.) needs to be used as a key for the rbtree.
The contents of the tree are DB-specific prepared queries.

`rlm_sql_module_t`needs to change:

```
sql_rcode_t (*sql_query)(rlm_sql_handle_t *handle, rlm_sql_config_t *config, void const *key, char const *query);
sql_rcode_t (*sql_select_query)(rlm_sql_handle_t *handle, rlm_sql_config_t *config, void const *key, char const *query);
```

where we add `key`.  This is the `CONF_PAIR`, etc.

we should add a mutex to the `rlm_sql` struct, and a hash table which is a list of queries -> xlat, statement mappings.

The code compiling the prepared statement can look up the xlat, instead of re-doing it itself.

If we want to avoid mutexes, the server core can pre-compile all of the queries, via a new sub-module API.

except it has to do that in pass2.

the SQL submodules should have a `prepare` function, which gets passed a query, and which caches the prepared query.

Then in the module sql_query, we do:
- look up key in connection to see if the statement has been prepared
  get name, ptr to prepared
  run it

( if not, look up key in SQL config to get prepared statement
   prepare it
   cache it in the connection

```
   foo = rbtree_find(config->prepared, key);
   if (!foo) {
      // prepare query.

Gresult *PQprepare(PGconn *conn,
                const char *stmtName,
                const char *query,
                int nParams,
                const Oid *paramTypes); // NULL, PQ will assume text and parse

  // SELECT * FROM mytable WHERE x = $1::bigint;
[...]

PR #961, opened 2015-04-21
psql: fix accounting_on query syntax

santigimeno:


arr2036:
Ug and this is why https://github.com/FreeRADIUS/freeradius-server/issues/830 would be so useful. Thanks for the pull request.

santigimeno:
Wow, that was fast! Thanks. Shouldn't it be backported to branch v3.0.x too?

arr2036:
It already has been, but the cherry pick won't show up in the pull request history.

santigimeno:
Ok I see it now. Thanks again!

PR #963, opened 2015-04-21
psql: fix database setup.sql script

santigimeno:


arr2036:
Ug, yes. I remember this from the last PG install I did.

PR #964, opened 2015-04-22
sql: use sql.acct_table1 instead of radacct

santigimeno:
- So the sqlcounter queries honor the sql configuration.


santigimeno:
Disclaimer: I have only tested this changes with the expire_on_login query in postgresql

arr2036:
This isn't quite right. You can't use sql.acct_table, it needs to be the specific instance of rlm_sql the counter module is set to use.

Issue #291, opened 2013-05-12
PyThreadState not set in rlm_python


galves:
I'm trying to use threading.local() to store a Curl client for connecting to a REST service.  The Curl client is itself thread-safe however it cannot be used concurrently in two threads, hence the need for a local instance per thread.

However, every time threading.local() is called, it is from a new python thread state context and so the Curl client is lost and ends up being recreated.

One work around I've found is to use _theading_local.local() which does not use the PyThreadState but there is a very significant performance hit due to the underlying dict storage.  Using a shared connection pool also adds some overhead and complexity which I would like to avoid.

theading.local() does appear to work correctly in some other embedded scenarios (mod_wsgi for example) so I'm optimistic a resolution can be found.

PR #983, opened 2015-04-29
make sure rlm_python does not segfault when using 2 instances and second is called first in a Thread

michael-mri:
First commit is about initializing all instances correctly. I had to make main_thread_state global. Note that only 1 interpreter is used, that means 2 instances are not in separate envs...

Second commit is cleaning of the code as I believe (https://docs.python.org/2/c-api/init.html#non-python-created-threads) PyGILState_* should take care of thread states. This simplifies a bit the code

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -134,12 +134,6 @@ static struct {
 };
 
 /*
- *	This allows us to initialise PyThreadState on a per thread basis
- */
-fr_thread_local_setup(PyThreadState *, local_thread_state)	/* macro */
-
-
-/*
  *	Let assume that radiusd module is only one since we have only
  *	one intepreter
  */
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -54,7 +54,6 @@ struct py_function_def {
 	char const	*function_name;
 };
 
-PyThreadState	*main_thread_state;
 typedef struct rlm_python_t {
 	void		*libpython;
 	char const	*python_path;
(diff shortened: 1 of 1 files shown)


arr2036:
Firstly, in future, please discuss pull requests on GitHub so that all contributors can comment.

Secondly i'm pretty certain this change will break TLS within python. There was a reason thread states were swapped in/out explicitly, and bound to the life of the worker threads.

arr2036:
Using a single master state is probably correct, but we'd need to see evidence of testing to make sure your changes don't break TLS within python.

michael-mri:
On 14/05/2015 17:41, Arran Cudbard-Bell wrote:
>
> Firstly, in future, please discuss pull requests on GitHub so that all 
> contributors can comment.
>
Yes ! that is why I did a pull request... I was really not sure of the 
code cleaning that is why I did it in 2 separate commits. My C is bad 
and does need review by others for sure.
But after 2 weeks with no answer I did mention it to Alan directly... 
next time I will post a comment on the pull request.
>
> Secondly i'm pretty certain this change will break TLS within python. 
> There was a reason thread states were swapped in/out explicitly, and 
> bound to the life of the worker threads.
>
Now I know why it was implemented like that :) A comment in the code 
would have been welcome...

michael-mri:
Any hint on how to test TLS in python ?

arr2036:
The log relating to rlm_python.c will contain comments explaining why it was done that way. It'd probably show up in ``git blame`` too.

Use the thread.local API in python to test TLS. The values should persist across multiple calls to the same worker thread.

michael-mri:
On 15/05/2015 15:12, Arran Cudbard-Bell wrote:
>
> The log relating to rlm_python.c will contain comments explaining why 
> it was done that way. It'd probably show up in |git blame| too.
>
True. It was hidden behind "Sync rlm_python and rlm_ldap with master". I 
should have dug deeper.
>
> Use the thread.local API in python to test TLS. The values should 
> persist across multiple calls to the same worker thread.
>
> —
>

Using the test in issue #291, it showed indeed that my changes break TLS :(

However without the cleanup, it does not. Can you review commit 
043adb8e1a51853648f52d46d17867f6eae63beb and integrate it ?
This solve issue #1015 that I just created so things are done properly.

Next time I will stick to the rule "do not fix what is not broken"...

PR #988, opened 2015-05-01
Try to find mapped values for attributes in rlm_rest/json

herwinw:
Probably not the nicest way to do this, but it's the least intrusive way I could think of.

Imagine changing `Service-Type` to `Login-User`. The generated JSON in `rlm_rest` looks like this:

```
{ ... "Service-Type":{"type":"integer","value":[1]} ... }
```

This lead to code in the Restful daemon like this (using Perl as example language, since the `demo.pl` of `rlm_rest` is also in perl:

```
if ($request->{'Service-Type'}{'value'} == 1) { # Login-User
```
or
```
use constant SERVICE_TYPE_LOGIN_USER => 1;
if ($request->{'Service-Type'}{'value'} == SERVICE_TYPE_LOGIN_USER) {
```

This patch tries to see if there are any mapped values for the attribute (I'm not even sure if mapped values is the correct term here) and send the symbolic value instead, so the JSON looks like this:

```
{ ... "Service-Type":{"type":"mapped","value":["Login-User"]} ... }
```

A possible improvement would be to implement this in `vp_prints_value_json`, which means all users of JSON will use the symbolic values. And maybe add a config option to disable this behaviour in case you want to remain backwards compatible.

(The mappings already work fine when returning values from the Restful daemon)

### DIFF
--- a/src/modules/rlm_rest/rest.c
+++ b/src/modules/rlm_rest/rest.c
@@ -726,12 +726,24 @@ static size_t rest_encode_json(void *out, size_t size, size_t nmemb, void *userd
 		}
 
 		if (ctx->state == READ_STATE_ATTR_BEGIN) {
+			DICT_VALUE *dv;
+
 			/*
 			 *  New attribute, write name, type, and beginning of value array.
 			 */
 			RDEBUG2("Encoding attribute \"%s\"", vp->da->name);
 
-			type = fr_int2str(dict_attr_types, vp->da->type, "<INVALID>");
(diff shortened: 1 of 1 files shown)


arr2036:
I think value aliases is probably the right term, though the structure is a DICT_VALUE

arr2036:
I think implementing it in vp_prints_value_json would be a good idea. But only in v3.1.x as it's a behaviour change.

herwinw:
Closed this, replaced by #1618

Issue #993, opened 2015-05-05
SQL drivers should all support the sql_fields method


arr2036:
Outstanding drivers are:

* [x] rlm_sql_db2
* [x] rlm_sql_firebird
* [x] rlm_sql_freetds
* [x] rlm_sql_iodbc
* [x] rlm_sql_oracle
* [x] rlm_sql_unixodbc

arr2036:
So db2, iodbc, and unixodbc were all identical

arr2036:
We don't want to provide the method for rlm_sql_null, as that would cause it to register a map_proc which'd do nothing. sql_fields doesn't have to be defined for every driver.

arr2036:
Now complete! Thanks @jpereira!

PR #995, opened 2015-05-05
Add sql_fields support for FireBird.

jpereira:
@arr2036 

pull request references for #993

att: I'm fixing problems with oracle build, after fix i'll push.


arr2036:
Thanks!

PR #997, opened 2015-05-05
Add sql_fields for IODBC. refs #993

jpereira:
pull request references for #993

### DIFF
--- a/scripts/bbedit/unlanglaguage.plist
+++ b/scripts/bbedit/unlanglaguage.plist
@@ -29,6 +29,7 @@
 	<key>BBLMKeywordList</key>
 	<array>
 		<string>update</string>
+		<string>map</string>
 		<string>switch</string>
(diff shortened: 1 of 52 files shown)


arr2036:
hmm, i'll redo the iodbc one, that's not right.

jpereira:
About the support for IODBC, this was my solution. 
About the rlm_sql_null, I commited in the same pull request.

arr2036:
IODBC function is actually identical to DB2 :)

arr2036:
There's two functions which you can use to get the column name. I started with the one you used, but it pulls back so much extra crap we don't need. I'll push up the changes and you'll see.

jpereira:
@arr2036  wow !I noted the similarities between ODBC/DB2 and made the patch for IODBC. if you haven't done, i can push up.

PR #999, opened 2015-05-07
Update the Alcatel 7750 dictionary.

jpereira:
Updating the Alcatel 7750 dictionary based on "9304720101_V1_7750 SR-OS RADIUS ATTRIBUTE.pdf" document


alandekok:
This patch is full of changes NOT from you.  This is bad form.

Please re-base your branch onto the v3.1.x branch, add your patch, and do a force push of the branch.  That should show up as only one diff.

jpereira:
Hi @alandekok, 

I had some problems with my tree... btw, follow the new pull request #1000

PR #1001, opened 2015-05-08
Add sql_fields function for freetds refs #993

jpereira:
Hi @arr2036!

Now we have complete support! :+1:


arr2036:
Oh you already did it! Cool! Thanks!

PR #1002, opened 2015-05-08
Add sql_fields function for rlm_sql_null refs #993

jpereira:


arr2036:
Null doesn't need a sql_fields function. The rlm_sql module checks to see if the driver defines sql_fields, and if it doesn't, it skips registering the sql_map_proc function, causing a validation error if someone attempts to use it (which is what we want).

arr2036:
Just need the freetds driver. I can handle it, if you're too busy?

PR #1007, opened 2015-05-14
Removing unused sql_read_naslist()

jpereira:
Hi @arr2036!

This confused me during use of the API!


arr2036:
Yeah this got moved into client.c

PR #1008, opened 2015-05-14
Fix boilerplates after valuepair.c split in 2de33b8.

nward:
This is a cosmetic code change, and has no functional impact.


arr2036:
v3.1.x please

nward:
OK. I have a couple more, would you like me to v3.1.x them as well?

PR #1009, opened 2015-05-14
Fix ${..base_dn} expansion in LDAP config.

nward:
Change single quotes for double quotes to permit expansion in ldap module config.
Partially reverts 34b35e5.


nward:
Actually, this didn't fix what I thought it did. Recalling.

nward:
Sorry, yes it does.. was looking on the wrong server. But, I will leave this closed and re-submit against v3.1.x.

Thanks

PR #1016, opened 2015-05-20
Fixing bug during copy to invalid pointer in the rlm_python

jpereira:


src/modules/rlm_python/rlm_python.c
```diff
@@ -223,9 +223,10 @@ static int mod_init(rlm_python_t *inst)
 	inst->main_thread_state = PyThreadState_Get();	/* We need this for setting up thread local stuff */
 #endif
 	if (inst->python_path) {
-		char *path;
+		char path[4096];
+
+		strncpy(path, inst->python_path, sizeof(path));
```
arr2036: use strlcpy

jpereira:
@arr2036 and @alandekok, done!

PR #1022, opened 2015-05-29
Check if radius is running

hagedose:
post-receive checks if radius is running. If not, it will accept
changes based on the config check, without trying to restart radiusd.


arr2036:
Can you you squash the commits? That's the ``git rebase`` step.

hagedose:
I had actually done that step, but I didn't understand what I had to do in interactive mode. I think I've got it right this time.

PR #1027, opened 2015-05-30
Add support of pre-proxy and post-proxy section to rlm_sqlippool module

geaaru:
Add support of pre-proxy and post-proxy section to rlm_sqlippool module

### DIFF
--- a/src/modules/rlm_sqlippool/rlm_sqlippool.c
+++ b/src/modules/rlm_sqlippool/rlm_sqlippool.c
@@ -49,7 +49,9 @@ typedef struct rlm_sqlippool_t {
 	int		framed_ip_address; 	//!< the attribute number for Framed-IP(v6)-Address
 
 	time_t		last_clear;		//!< So we only do it once a second.
+#ifdef WITH_PROXY
 	time_t		proxy_last_clear;	//!< So we only do it once a second (for proxy section)
+#endif
 	char const	*allocate_begin;	//!< SQL query to begin.
 	char const	*allocate_clear;	//!< SQL query to clear an IP.
 	char const	*allocate_find;		//!< SQL query to find an unused IP.
@@ -83,6 +85,7 @@ typedef struct rlm_sqlippool_t {
 	char const	*off_clear;		//!< SQL query to clear an entire NAS.
 	char const	*off_commit;		//!< SQL query to commit.
 
+#ifdef WITH_PROXY
 						/* Pre proxy sequence */
 	char const	*preproxy_begin;	//<! SQL query to begin
 	char const	*preproxy_clear;	//<! SQL query to clear an IP.
@@ -96,6 +99,7 @@ typedef struct rlm_sqlippool_t {
 	char const	*postproxy_find;	//<! SQL query to find an used IP.
 	char const	*postproxy_update;	//<! SQL query to update an IP entry.
 	char const	*postproxy_commit;	//<! SQL query to commit.
+#endif
 
 						/* Logging Section */
 	char const	*log_exists;		//!< There was an ip address already assigned.
--- a/src/modules/rlm_sqlippool/rlm_sqlippool.c
+++ b/src/modules/rlm_sqlippool/rlm_sqlippool.c
@@ -87,18 +87,18 @@ typedef struct rlm_sqlippool_t {
 
 #ifdef WITH_PROXY
 						/* Pre proxy sequence */
-	char const	*preproxy_begin;	//<! SQL query to begin
-	char const	*preproxy_clear;	//<! SQL query to clear an IP.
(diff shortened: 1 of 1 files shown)


geaaru:
I fix preprocessing define related with attrubute added on rlm_sqlippool_t structure on commit f16bbf816217e9bfc27548d0e75a72df853d22e2 but is not visible on pull request, probably because it is in waiting status.
I'm waiting for yours feedback.
Bye

arr2036:
You've not said why this patch is necessary.

You can perform the IP checks using unlang before calling the module. You can override the module call in pre-proxy with ``sqlippool.authorize``. If you want to control where the IP address is written out to, then yes a patch would be needed for that, and i'm happy to add something in v3.1.x that would allow that.

If there's some element of functionality i've missed then feel free to point it out.

geaaru:
Hi, 
thank you very much for your fast reply.
In my case I need add IP address in post-proxy section, only when I receive Access-Accept, and so it isn't sufficient use authorize section.

If code is ok I'm happy to do same changes to v3.1.x branch.

Probably, could be fine add pre-proxy, post-proxy sections also on rlm_sql module to permit add attributes from database, I will investigate if it easy to do.

alandekok:
On May 30, 2015, at 6:50 PM, geaaru <notifications@github.com> wrote:
> In my case I need add IP address in post-proxy section, only when I receive Access-Accept, and so it isn't sufficient use authorize section.

  The sqlippool module assigns IPs in the post-auth section.  That happens after the post-proxy section is run.

  I think the module already does what you want.

geaaru:
This means that there is an error on Freeradius Tehcnical Guide (figure 4.3) where for Proxy Authentication process is not present Post-Auth section.

And between differences there is also:
* I permit to retrive of Framed-Ip-Address also if Pool-Name is not present (I do this directly with a query by Calling-Station-Id), currently post-auth function check if Pool-Name is present and if it isn't present doesn't execute allocate_find function,
* I remove Framed-Ip-Address Attribute if it is already present and I execute pre/post proxy_find. I know, this could be done through unlang by in my case I think that could be correct that a user on enable sqlippool on post proxy/pre proxy can override exist Framed-Ip-Address attribute.
* I modify check of fmt variable inside sqlippool_query/query1 to avoid xlat_expanded if fmt is NULL or fmt string has len zero and so speed up processing.

So, I think that could be correct add pre-proxy/post-proxy section on sqlippool and probably also in rlm_sql module. The choice is yours.

alandekok:
On May 31, 2015, at 4:44 AM, geaaru <notifications@github.com> wrote:
> This means that there is an error on Freeradius Tehcnical Guide (figure 4.3) where for Proxy Authentication process is not present Post-Auth section.

  We'll fix the diagram.

> And between differences there is also:
> 
> 	� I permit to retrive of Framed-Ip-Address also if Pool-Name is not present (I do this directly with a query by Calling-Station-Id), currently post-auth function check if Pool-Name is present and if it isn't present doesn't execute allocate_find function,

  That is site-specific.  Not everyone wants that functionality.  It can be done via "unlang" checks around the module.

> 	� I remove Framed-Ip-Address Attribute if it is already present and I execute pre/post proxy_find. I know, this could be done through unlang by in my case I think that could be correct that a user on enable sqlippool on post proxy/pre proxy can override exist Framed-Ip-Address attribute.

  The attr_filter module already strips out Framed-IP-Address from post-proxy.

> 	� I modify check of fmt variable inside sqlippool_query/query1 to avoid xlat_expanded if fmt is NULL or fmt string has len zero and so speed up processing.

  That can be done as a minor change.

geaaru:
I will do a new pull request only with minor change.

I'm not agree to add a fake Pool-Name attribute to bypass post-auth check but you are the master.

Thank you very much for your time.

To the next pull request.

Bye
G.

alandekok:
On May 31, 2015, at 1:40 PM, geaaru <notifications@github.com> wrote:
> I'm not agree to add a fake Pool-Name attribute to bypass post-auth check but you are the master.

  You can always do:

post-proxy {
	...
	sqlippool.post-auth
	...
}

geaaru:
Yes, but for this code:

if (pairfind(request->config, PW_POOL_NAME, 0, TAG_ANY) == NULL) {
		RDEBUG("No Pool-Name defined");
                return do_logging(request, inst->log_nopool, RLM_MODULE_NOOP);
}

I need add through unlang a code like this:

post-proxy {

     update reply {
             &Pool-Name = 'FakePoolNotUsed'
     }

     sqlippool.post-auth
     ...
}

And yes, current process is correct: for allocate an ip could be mandatory identify Pool-Name but for performance reason could be also correct retrieve Framed-Ip-Address directly through a single query by Calling-Station-Id.

PR #1028, opened 2015-05-31
Minor changes for speed up rlm_sqlippool module and avoid xlat_expanded if string len is zero

geaaru:
Minor changes for speed up rlm_sqlippool module and avoid xlat_expanded if string len is zero


geaaru:
Perfect.
Thanks.

PR #1029, opened 2015-06-05
Enable MySQL & Postgres tests in Travis + add error log

pwdng:


arr2036:
Reverted, please move the setup for PG into a setup script in the test directory which gets called from travis.yml

arr2036:
... and set it so it doesn't install it if we're not actually doing a make test

PR #1036, opened 2015-06-08
Add the vsa 'Alc-Wlan-SSID-VLAN' for Alcatel 7750

jpereira:
following the document https://infoproducts.alcatel-lucent.com/html/0_add-h-f/93-0088-HTML/7750_SR_RADIUS_Attributes_Reference_Guide/SROS_RADIUS_Attrib.html

@arr2036 Please, replicate in the v3.0.x

### DIFF
--- a/share/dictionary.alcatel.sr
+++ b/share/dictionary.alcatel.sr
@@ -350,4 +350,8 @@ ATTRIBUTE   Alc-Portal-Url                            177 string
 # names longer than the allowed maximum are treated as host setup failures
 ATTRIBUTE   Alc-SLAAC-IPv6-Pool                       181 string
 
+# The VLAN is transparently taken from the UE’s Ethernet layer and can be reflected
+# in both authentication and accounting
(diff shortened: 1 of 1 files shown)


jpereira:
on second thought... @arr2036 can you replicate all the dict share/dictionary.alcatel.sr in the v3.0.x?

arr2036:
do you mean there are attributes in v3.1.x that aren't in v3.0.x? It may have been intentional...

arr2036:
or more, we try not to rename VSAs across branches...

jpereira:
I don't know... i think that the best way it's to have the same dictionaries between the branches. wouldn't you?

arr2036:
No, because then you break people's config when they upgrade the server.

arr2036:
For example if the Alcatel dictionary in v3.0.x had a bunch of attributes like

```
Al-User-Name
Al-VLAN
```

Then they got bought out by Juniper and all the official attribute names changes to

```
JU-User-Name
JU-VLAN
```

and people used the ``AL`` attributes in their configs, the upgrade would break their configs.

PR #1038, opened 2015-06-08
Add support for Aptilo Access Controller

jpereira:
Adding support for "Aptilo Access Controllers" - http://www.aptilo.com/aptilo-access-controller

@arr2036, please replicate in the v3.1.x

Thanks

### DIFF
--- a/share/dictionary
+++ b/share/dictionary
@@ -137,6 +137,7 @@ $INCLUDE dictionary.alcatel-lucent.aaa
 $INCLUDE dictionary.alteon
 $INCLUDE dictionary.alvarion
 $INCLUDE dictionary.apc
+$INCLUDE dictionary.aptilo
 $INCLUDE dictionary.arbor
 $INCLUDE dictionary.aruba
 $INCLUDE dictionary.azaire
--- a/share/dictionary.aptilo
+++ b/share/dictionary.aptilo
@@ -1,0 +1,139 @@
+# -*- text -*-
+# Copyright (C) 2015 The FreeRADIUS Server project and contributors
+#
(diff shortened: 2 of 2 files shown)


alandekok:
What the heck is the internal dictionary?  I don't think that is ever supposed to go j to a radius packet. And if it is, they don't own vendor ID 666.

  that number is VERY likely to me a they're doing something wrong. And they know it.

jpereira:
@alandekok I updated the branch removing the *.internal. Now if you update, you can notice that has only share/dictionary.aptilo

jpereira:
@alandekok I updated the branch removing the *.internal. Now if you update, you can notice that has only share/dictionary.aptilo

arr2036:
What's the Aptilo at the end of every Attribute definition?

arr2036:
Just use ``BEGIN VENDOR Aptilo`` like every other dictionary.

arr2036:
and ``END VENDOR Aptilo`` at the end.

jpereira:
Hi @arr2036 , check again!

arr2036:
The start should be

```
# -*- text -*-
# Copyright (C) 2015 The FreeRADIUS Server project and contributors
#
VENDOR		Aptilo			13209

BEGIN-VENDOR	Aptilo
```

and you should include it in ``share/dictionary`` else it won't get evaluated during the travis test.

jpereira:
@arr2036 done!

arr2036:
Insert the entry after 139 in share/dictionary to maintain lexicographical ordering

jpereira:
@arr2036 done!

PR #1047, opened 2015-06-10
More helpful message during query executions in the rlm_redis

jpereira:
Currently during the query in the rlm_redis we showed:

```
Debug: executing EXPIRE ..
```

Actually, could be more helpful shows the complete query that is executed.

```
Debug: executing the query: EXPIRE 12345 1800
```

@arr2036 agree?


arr2036:
sure

PR #1048, opened 2015-06-10
Unnecessary strlen() in the rlm_redis

jpereira:
Unnecessary strlen() in the rlm_redis


arr2036:
I wouldn't spend too much time on this code. It's all rewritten for 3.1.x.

jpereira:
:+1:

PR #1049, opened 2015-06-11
Module rlm_redis and rlm_rediswho, welcome to stable world!

jpereira:
Hi @arr2036 ,

We have used a lot these modules. was fixed some issues, and i believe that it's the time to bring to stable world in the v3.0.x. agree?


jpereira:
We have hundreds and hundreds of connections per minute, running smoothly.

PR #1053, opened 2015-06-13
Some suggestions for rlm_sql_cassandra

jpereira:
Hi,

 I have tested and worked fine. this patch is only some suggestions.

@linnaea, @arr2036... nice job!

### DIFF
--- a/src/modules/rlm_sql/drivers/rlm_sql_cassandra/rlm_sql_cassandra.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_cassandra/rlm_sql_cassandra.c
@@ -178,10 +178,25 @@ static int _sql_socket_destructor(rlm_sql_cassandra_conn_t *conn)
 {
 	DEBUG2("rlm_sql_cassandra: Socket destructor called, closing socket");
 
-	if (conn->iterator) cass_iterator_free(conn->iterator);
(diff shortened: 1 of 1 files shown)


arr2036:
I'll merge the first one, but I disagree with the second. This behaviour is an anti-pattern for resources being freed. We *want* the server to crash if something attempts to use an object after its freed. That function is a destructor, called when talloc frees the object. If something attempts to use the connection after its been freed it indicates a bug in the code, setting the pointers to NULL may mask the issue and make it harder to track down. It should only be used when a structure needs to be reset to a specific state for re-use.

Generally I disagree with checking the state of a pointer before passing it to the free function too, but in this case there's no documentation to say the libcassandra functions will behave sensibly when passed a NULL pointer

I'll take a sweep through the talloc destructors of the SQL drivers to make sure it's not done elsewhere.

arr2036:
Yes, you can argue that NULL may be more likely to cause a SEGV, except these are pointers being passed to 3rd party libraries, of which many, check the resource pointers being passed to see if they're NULL and return without doing anything, not always triggering errors. For tracking down these sorts of issues using valgrind and memory scribbling features makes much more sense.

jpereira:
@arr2036 good point! Agree with you!

PR #1058, opened 2015-06-15
Add index to be used on accounting off (closes #1037).

### LINKED ISSUE #1037: More indexes for accounting on/off statements
The standard indexes work for the normal auth/acct queries.  But when the server receives an Accouting-On packet, that query devolves to a table scan.

All of the modules in raddb/sql/main/ should have indexes added for accounting on/off.

Postgresql has it:

    -- For use by onoff-
    CREATE INDEX radacct_bulk_close ON radacct (NASIPAddress, AcctStartTime) WHERE AcctStopTime IS NULL;

but the other modules don't have it.  The poor people using MySQL should be using indexes, too.

pwdng:


alandekok:
Any idea about the travis build failures?  The messages aren't overly clear

arr2036:
Looks like there's an error check missing in the boostrap code...

arr2036:
Not really missing, just sqlite isn't consistent with passing back error codes. @pwdng could you rebase from v3.1.x, then force push back to your branch. It should re-run the tests, and use the new error checking code that I just added.

pwdng:
Conditional indexes come with SQLite 3.8.0. Unfortunately, Travis only provides version 3.7.15.
This is why the build is failing. I'd suggest commenting out the conditional index definition and either adding an ordinary index or simply not having one at all.

arr2036:
no index at all is probably the best idea

PR #1066, opened 2015-06-24
Don't leak memory on fr_syserror()

jpereira:
Don't leak memory on each 'No error' messages.


arr2036:
Um, we don't leak memory on fr_syserror. Read the code?

jpereira:
@arr2036  wow! it's true, my bad! please, delete this! :+1:

PR #1067, opened 2015-06-25
Don't repeat code ad nauseum!

jpereira:
could be more nicely...


arr2036:
It's repeated twice. It's about 10 lines. This isn't needed.

PR #1069, opened 2015-06-25
some adjustments for a better error messages.

jpereira:
Hi @arr2036!

Today I had the situation that was necessary to call the 'radiusd' using a directory that the user 'freerad' don't had permissions to read... and was terrible to find  the problem like the below message.

```
......
Wed Jun 24 01:00:55 2015 : Debug:   # Loading module "files" from file /root/teste/mods-enabled/files
Wed Jun 24 01:00:55 2015 : Debug:   files {
Wed Jun 24 01:00:55 2015 : Debug:   	filename = "/root/teste/mods-config/files/authorize"
Wed Jun 24 01:00:55 2015 : Error: Unable to open file "/root/teste/mods-config/files/authorize": 
Wed Jun 24 01:00:55 2015 : Debug:   }
Wed Jun 24 01:00:55 2015 : Error: /root/teste/mods-enabled/files[9]: Invalid configuration for module "files"
......
```

after the adjustments...

```
......
Wed Jun 24 01:00:55 2015 : Debug:   # Loading module "files" from file /root/teste/mods-enabled/files
Wed Jun 24 01:00:55 2015 : Debug:   files {
Wed Jun 24 01:00:55 2015 : Debug:   	filename = "/root/teste/mods-config/files/authorize"
Wed Jun 24 01:00:55 2015 : Error: Unable to open file "/root/teste/mods-config/files/authorize": Permission denied
Wed Jun 24 01:00:55 2015 : Debug:   }
Wed Jun 24 01:00:55 2015 : Error: /root/teste/mods-enabled/files[9]: Invalid configuration for module "files"
......
```

more useful! :+1: 

And I made some others adjustments.

### DIFF
--- a/src/main/conffile.c
+++ b/src/main/conffile.c
@@ -305,7 +305,6 @@ static FILE *cf_file_open(CONF_SECTION *cs, char const *filename)
 	CONF_DATA *cd;
 	CONF_SECTION *top;
 	rbtree_t *tree;
-	int fd;
 	FILE *fp;
 
 	top = cf_top_section(cs);
@@ -321,8 +320,6 @@ static FILE *cf_file_open(CONF_SECTION *cs, char const *filename)
 		return NULL;
 	}
 
-	fd = fileno(fp);
-
 	file = talloc(tree, cf_file_t);
 	if (!file) {
 		fclose(fp);
@@ -333,7 +330,7 @@ static FILE *cf_file_open(CONF_SECTION *cs, char const *filename)
 	file->cs = cs;
 	file->input = true;
 
-	if (fstat(fd, &file->buf) == 0) {
+	if (stat(filename, &file->buf) < 0) {
 #ifdef S_IWOTH
 		if ((file->buf.st_mode & S_IWOTH) != 0) {
 			fclose(fp);
@@ -373,7 +370,6 @@ static bool cf_file_input(CONF_SECTION *cs, char const *filename)
 	if (!cd) return false;
 
 	tree = cd->data;
-
 	file = talloc(tree, cf_file_t);
 	if (!file) return false;
 
--- a/src/main/conffile.c
+++ b/src/main/conffile.c
@@ -1366,6 +1366,7 @@ int cf_item_parse(CONF_SECTION *cs, char const *name, unsigned int type, void *d
 	CONF_PAIR *cp = NULL;
 	fr_ipaddr_t *ipaddr;
 	char buffer[8192];
+	CONF_ITEM *c_item = &cs->item;
 
 	if (!cs) return -1;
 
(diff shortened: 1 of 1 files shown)


arr2036:
Can you print out the current effective user and group as well. That's what I started doing for other file opening errors.

Really helps tracking down why the server works when run in debug mode vs daemon mode.

arr2036:
There's functions already in the server that wrap the euid egid functions.

jpereira:
@arr2036  Good idea! I will push another pull printing the user id!

arr2036:
OK

jpereira:
@arr2036 What do you mean? I think that could be better create a function that return the current 'user' and 'group' name that runs the process. And in this case just calling like print_current_owner(). has suggestions?

jpereira:
Or could be better call the "uid_name/gid_name" from the src/main/mainconfig.c

jpereira:
@arr2036 

I guess that we can have something like:

```
/*
 * return the name (or #number) of the user/group to run radiusd as
 */
const char *fr_get_owner() {
	static char buf[64];
	const char *user;
	const char *group;
	CONF_SECTION *sec;

	sec = cf_section_find("security");
	if (!sec) return NULL;

	user = cf_section_value_find(sec, "user");
	group = cf_section_value_find(sec, "group");

	if (!user || !group) return NULL;

	snprintf(buf, "%s:%s", user, group);

	return &buf;
}
```

And just calling where we need to show. suggestions?

arr2036:
No. That's not helpful. The reason why I said get the euid/egid is because you want the *effective* uid/gid at the time the system call was made. It's not necessarily going to be the ones configured in radiusd.conf.  radiusd usually starts off as root, and may suid up/down.

Running in debug mode means it never suids to the configured uid/gid, so your code would show the wrong output there too.

Use ``rad_prints_uid`` and ``rad_prints_gid`` with ``geteuid`` and ``getegid``. Those functions take all the pain out of getting the usernames/groupnames.

arr2036:
I can't find where I did the other errors though *sigh*. Maybe someone removed them.

jpereira:
@arr2036 exactly! I didn't find. but the idea is a function like as i said? but using the rad_prints_uid! ?

arr2036:
yes, with geteuid and getegid

jpereira:
@arr2036 exactly! I didn't find. but the idea is a function like as i said?

jpereira:
I will put the new function "const char *fr_get_owner()" in the util.c, like it?

arr2036:
It's only four lines of code to get the euid/egid, i'm not sure it's worth writing a function.

jpereira:
@arr2036 I think that could be better put separated, because we have another contexts that can be useful to print the process owner.

jpereira:
@arr2036 I made in another branch, ref #1071

PR #1071, opened 2015-06-25
Improve the report when has problem during reading the file. ref #1069

jpereira:
@arr2036 done! definitely more useful...

```
Thu Jun 25 18:00:22 2015 : Debug:  filename = "/home/jpereira/tmp/freeradius_3.0.9//mods-config/files/authorize"
Thu Jun 25 18:00:22 2015 : Error: Unable to open file "/home/jpereira/tmp/freeradius_3.0.9//mods-config/files/authorize": No error (runs as user=root, group=freerad)
Thu Jun 25 18:00:22 2015 : Debug:   }
Thu Jun 25 18:00:22 2015 : Error: /home/jpereira/tmp/freeradius_3.0.9//mods-enabled/files[9]: Invalid configuration for module "files"
```

### DIFF
--- a/src/main/conffile.c
+++ b/src/main/conffile.c
@@ -380,7 +380,18 @@ static bool cf_file_input(CONF_SECTION *cs, char const *filename)
 	file->input = true;
 
 	if (stat(filename, &file->buf) < 0) {
-		ERROR("Unable to open file \"%s\": %s", filename, fr_syserror(errno));
+		char user[64];
+		char group[64];
+		const char *err_msg;
+
+		err_msg = fr_syserror(errno);
+
+		rad_prints_uid(NULL, user, sizeof(user), geteuid());
+		rad_prints_gid(NULL, group, sizeof(group), getegid());
+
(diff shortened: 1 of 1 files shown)


jpereira:
@arr2036 I made here!

jpereira:
hi @arr2036 , could you merge this?

alandekok:
I'm unsure about this one.  I'm not sure there's a benefit to always printing out the uid/gid when it can't open a file.  There could be other errors, too.

jpereira:
@alandekok I'm too... btw, closed!

arr2036:
It is useful, errors don't always hint as to what's wrong, especially with selinux issues. When there's a delay merging the commits it's likely because Alan and myself are busy.

arr2036:
and please don't immediately delete branches, it's really bad manners. It removes the commits from GitHub so even if we wanted to base something off your commit we can't.

jpereira:
@arr2036 they are back!

jpereira:
for me this is completely helpful!

jpereira:
I agree with you @arr2036, for me this is completely helpful!!

PR #1106, opened 2015-06-30
Add pcap support

pwdng:


jpereira:
This feature sounds great!

jpereira:
@arr2036 This inspires a new feature to be possible listen from a fifo-file.

pwdng:
It's main purpose is for the DHCP server.
It lets us listen to broadcast traffic on BSD systems that won't let us bind to an interface.

arr2036:
As a general rule, any new functions going in should have doxygen headers too. It's a lot of work to go back a fix all the functions without headers (though we are slowly), but it's less work to ensure the problem doesn't get worse.

arr2036:
@pwdng Looks good, but can you squish it into the previous commits so that we can see the code overall, thanks!

PR #1112, opened 2015-07-01
Add dictionary for Arista Networks

hzhongmj:
This adds a dictionary file for Arista Networks.

### DIFF
--- a/share/dictionary
+++ b/share/dictionary
@@ -139,6 +139,7 @@ $INCLUDE dictionary.alvarion
 $INCLUDE dictionary.apc
 $INCLUDE dictionary.aptilo
 $INCLUDE dictionary.arbor
+$INCLUDE dictionary.arista
 $INCLUDE dictionary.aruba
 $INCLUDE dictionary.azaire
 $INCLUDE dictionary.ascend
--- a/share/dictionary.arista
+++ b/share/dictionary.arista
@@ -1,0 +1,14 @@
+# -*- text -*-
+# Copyright (C) 2015 The FreeRADIUS Server project and contributors
+##############################################################################
+#
+#	Arista VSAs
+#
+##############################################################################
+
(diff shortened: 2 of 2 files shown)


arr2036:
Are you really going to go the Cisco route, and just use "<attribute>=<value>" values? You realise there are now complex types in RADIUS that can deal with nesting, and more than 255 attributes, right?

arr2036:
"In December 2014, Cisco filed two lawsuits against Arista alleging extensive intellectual property infringement." was one of them ripping off their RADIUS implementation?

hzhongmj:
Thanks for the comments. I understand more structured attributes could be defined. However, this is what our current releases support (has been shipping for a few years). We'll continue to look at the migrate to a better way of doing this (but we'd support both this and future VSAs in a backward compatible way), but I'd like to merge this first so our existing customers won't have to do it manually.

I did look at some of the existing dictionaries, and many have this type defined already (grep shows about 17 hits). So I hope a progressive approach would be acceptable.

arr2036:
15 out of 209, yes. So 7.17% of vendors are clueless as to how RADIUS is supposed to work. If you look, some of those vendors include other attribute definitions, and have started the migration process.

hzhongmj:
OK. I can certainly come up with a better proposal, but as I explained, we'd have to still keep this little one for backward compatibility. So we have two choices:

1. Merge this first and I'll send another pull requests in the near future.
2. I come back later with a dictionary that is considered good practice, but still including this AVPair.

Kindly let me know what you'd like to see.

arr2036:
I've merged the your pull request to both branches. It'll be included in the initial release of 3.1.0, and 3.0.9. Please could you send another pull request with a dictionary containing properly defined VSAs.

If you have any questions regarding the extended attribute RFCs @alandekok authored them, so he'll likely be able to help.

Issue #1113, opened 2015-07-02
Show the path of object magic number mismatch


jpereira:
Hi,

I think that we can improve the error messages during " like below example.

```
/etc/raddb/mods-enabled/always[35]: Application and rlm_always magic number (version) mismatch.  application: 30008 module: 30004
```

could be better to show the complete path of module rlm_always.so

arr2036:
We don't know the complete path to rlm_always.so?

arr2036:
http://pubs.opengroup.org/onlinepubs/9699919799/

There's no way to get the path of the .so if we didn't specifying it using an absolute path

jpereira:
during the build we have the paths.... we can assume that $prefix/freeradius/lib/ .... something like that. what do you think?

jpereira:
@arr2036 I think that the address handler should be found in /proc/<pid>/map_files I am not sure!

alandekok:
On Jul 2, 2015, at 3:24 PM, Jorge Pereira <notifications@github.com> wrote:
> during the build we have the paths.... we can assume that $prefix/freeradius/lib/ .... something like that. what do you think?

  The dynamic linker searches many paths at run time.  Printing a *wrong* message is bad.  It's better to have *no* message about paths.

arr2036:
Agreed. The best we could do, is if libdir was specified in radiusd.conf, and the dlopen succeeded on the first call (with the full path), we could indicate that, otherwise indicate that the shared object had been loaded using the search path.

Assuming dlopen doesn't do something completely stupid and use its search path anyway, even will a full path.

jpereira:
@alandekok and @arr2036. The name is ```dladdr()```, is the guy that we want to use!

e.g:

```
void say_myname(void) {
    Dl_info dl_info;
    dladdr((void *)say_myname, &dl_info);
    fprintf(stderr, "module %s loaded\n", dl_info.dli_fname);
}
```

arr2036:
It's not POSIX, please add autoconf checks.

alandekok:
It's available on Linux and *BSD.  So that's good enough for now.

jpereira:
in MacOSx also!

arr2036:
Because OSX is BSD :P

PR #1119, opened 2015-07-04
Fix checking dladdr() #1113

jpereira:
hi @alandekok ,

Now the dladdr() will be reached. refs to #1113

### DIFF
--- a/configure
+++ b/configure
@@ -6373,6 +6373,12 @@ _ACEOF
 fi
 
 
+if test "x$ac_cv_func_dladdr" = "xyes"; then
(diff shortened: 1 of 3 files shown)


alandekok:
Not necessary.  The solution is to update autoconf.h.in, which I just did.

jpereira:
@alandekok have you checked? without this doesn't works.

jpereira:
have you checked? without the patch 529bcc5,  this doesn't works.

jpereira:
ah! ok

PR #1120, opened 2015-07-04
move the dladdr() call to dl_get_pathname() to avoid a lot of ifdef()

jpereira:
@alandekok I noticed that dladdr() exist in the main unix systems... so, _dl_get_pathname() always will be used and could be reused!

### DIFF
--- a/src/main/modules.c
+++ b/src/main/modules.c
@@ -98,6 +98,32 @@ const section_type_value_t section_type_value[MOD_COUNT] = {
 #  define LT_SHREXT ".so"
 #endif
 
+/*
+ * Return the Pathname of shared object that contains address
+ */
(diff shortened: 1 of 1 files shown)


jpereira:
@alandekok I noticed that dladdr() exist in the main unix systems... so, ```_dl_get_pathname()``` always will be used and could be reused!

arr2036:
TBH i'd rather use the documented function if we're using non standard functions :)

jpereira:
@arr2036 prefer wihout /* comments */?

jpereira:
@arr2036 done!

jpereira:
@arr2036 I did the changes. but for this work you need to apply this fix https://github.com/FreeRADIUS/freeradius-server/pull/1119

PR #1133, opened 2015-07-10
Add sql_fields function for freetds refs #993

jpereira:


arr2036:
Is this the only one missing from v3.0.x?

jpereira:
@arr2036 exactly!

Issue #1125, opened 2015-07-06
Problems with debian package and proto_dhcp/freeradius-dhcp*.deb


jpereira:
Hi,

We have a conflict between the packages "freeradius-XXX.deb" and freeradius-dhcp-XXX.deb. 

```
[root@jpereira-desktop freeradius]# dpkg -S /usr/lib/freeradius/proto_dhcp.so 
freeradius: /usr/lib/freeradius/proto_dhcp.so
[root@jpereira-desktop freeradius]# dpkg -i freeradius-dhcp_3.0.9+git_amd64.deb 
(Reading database ... 324085 files and directories currently installed.)
Preparing to unpack freeradius-dhcp_3.0.9+git_amd64.deb ...
Unpacking freeradius-dhcp (3.0.9+git) ...
dpkg: error processing archive freeradius-dhcp_3.0.9+git_amd64.deb (--install):
 trying to overwrite '/usr/lib/freeradius/proto_dhcp.so', which is also in package freeradius 3.0.9+git
Errors were encountered while processing:
 freeradius-dhcp_3.0.9+git_amd64.deb
[root@jpereira-desktop freeradius]# 
```

The file "proto_dhcp" is referenced by two packages.

```
[jpereira@jpereira-desktop freeradius-server-3.0.x.git]$ grep "freeradius/proto" debian/*.install
debian/freeradius-dhcp.install:usr/lib/freeradius/proto_dhcp*.so
debian/freeradius.install:usr/lib/freeradius/proto*.so
[jpereira@jpereira-desktop freeradius-server-3.0.x.git]$ 

```

jpereira:
@arr2036 My idea is a simple "freeradius-dhcp.preinst" that removes the file "proto_dhcp.so" if exist before then install.

jpereira:
The problem was caused because exist a double reference for "proto_dhcp.so" into the package "freeradius-XXX.deb" and "freeradius-dhcp-XXX.deb".

And don't make sense to create a separated package "freeradius-dhcp-XXX.deb" and keep/install/enable by default over another package. in this case "freeradius-XXX.deb"

This patch https://github.com/FreeRADIUS/freeradius-server/pull/1126 fixes this.

jpereira:
Below some evidences.

removing everything...
```
[root@jpereira-desktop freeradius]# dpkg -l | awk '/freeradius/ { print $2}' | xargs dpkg -P --force-all
```

installing the basic

```
[...]

PR #1134, opened 2015-07-10
Fix problems with freeradius-dhcp*.deb, refs #1125

jpereira:


jpereira:
@alandekok Now is ok.

jpereira:
Hi @alandekok, could you please merge this?

qnet-herwin:
The lines with `diff --git` and `index` shouldn't be in `radiusd-to-freeradius.diff`. The same thing goes for the appendixes at the line indexes (like `LDFLAGS = $(OPENSSL_LDFLAGS) $(TALLOC_LDFLAGS) @LDFLAGS@`). They clutter the visible changes for this change.

Also, I don't think the file `radiusd-to-freeradius.diff` is the correct one to change DHCP fro a default module, it should belong in a new module.

alandekok:
I agree with herwin here.

jpereira:
The idea of ```radiusd-to-freeradius.diff``` file is to register the differences. remember everyone that the *.diff is out of date. the ```LDFLAGS = $(OPENSSL_LDFLAGS) $(TALLOC_LDFLAGS) @LDFLAGS@``` exist in the Make.inc.in from HEAD but not in the current *.diff

alandekok:
> The idea of radiusd-to-freeradius.diff file is to register the differences.

Which make "radiusd" into the debian "freeradius".

A change with a different purpose should be a different patch.

jpereira:
Ok, In this case the idea is move the changes to another *.diff. right?

alandekok:
Yes.

jpereira:
@alandekok done!

jpereira:
Hi @arr2036 !

have plans to look this? This issue during the *.deb build is very boring.

Issue #1115, opened 2015-07-02
The rlm_rest could return a attr "REST-HTTP-Code" with last http request.


jpereira:
Hi @arr2036,

As we talked, the idea is that rlm_rest set the attribute "REST-HTTP-Status" with the last status of HTTP request. this could be nice to better handler of error messages during rest request.

arr2036:
It could, pleas send a pull request :)

jpereira:
@arr2036 done! :beer:

PR #1135, opened 2015-07-10
rlm_rest: Added support for REST-HTTP-Code, ref #1115

jpereira:


jpereira:
@arr2036 done

arr2036:
Buf should be ``buffer`` and only 4 bytes, seeing as HTTP codes are all three digit numbers.

``pairmake_reply`` will keep adding pairs whereas you need to replace any existing pairs. Using the cursor functions is probably the most efficient way to do that.

qnet-herwin:
> Buf should be `buffer` and only 4 bytes, seeing as HTTP codes are all three digit numbers.

Then shouldn't an integer (or any numeric type) be more suitable?

arr2036:
Yes, but it's being passed into a generic function that takes strings as input and converts them to the correct type for the attribute.

It'd be better to use ``fr_pair_alloc_by_num`` set the value directly on the new attribute, and ``fr_pair_add`` to add it to the reply list.

jpereira:
This was a wrong push. I will up again.

jpereira:
@arr2036 re-check!

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -119,6 +119,17 @@ static int rlm_rest_perform(rlm_rest_t *instance, rlm_rest_section_t *section, v
 	 *  HTTP body data into a single contiguous buffer.
 	 */
 	ret = rest_request_perform(instance, section, request, handle);
+
+	/*
+	 * Set an REST-HTTP-Code with the last code of http request.
```
qnet-herwin: This should be "Set a REST...", an is only used if the next word is a vowel.

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -119,6 +119,17 @@ static int rlm_rest_perform(rlm_rest_t *instance, rlm_rest_section_t *section, v
...
+	 */
+	vp = fr_pair_afrom_num(request->packet, PW_REST_HTTP_CODE, 0);
+	if (vp) {
+		RDEBUG2("Adding REST-HTTP-Code = %d", rest_get_handle_code(handle));
+		vp->vp_integer = rest_get_handle_code(handle);
```
qnet-herwin: I'm not sure how efficient the `rest_get_handle_code` method is, but it feels bad to call it twice. If you reverse the two lines above, the `RDEBUG` line could just get `vp->vp_integer` as the argument, instead of `rest_get_handle_code(handle)`.

jpereira: rest_get_handle_code() is a MACRO to data-struct

Gunisalvo:
I support this suggestion. Using rlm_rest option always require an additional translation step. I think it is a good idea to receive it embedded in the response body.

arr2036:
@jpereira your code was still incorrect. You can't just call ``fr_pair_add``, you need to replace the existing value (if there is one).

The delay in pushing this back was because I was investigating fixing ``fr_pair_make`` to do something sensible.

PR #1143, opened 2015-07-16
dhcp secs is a short

nchaigne:
Properly encode DHCP field "secs" (using htons instead of htonl).

Also I think in this case using "vp_short" is more correct than
"vp_integer" for a short (even though it still works with "vp_integer").

### DIFF
--- a/src/modules/proto_dhcp/dhcp.c
+++ b/src/modules/proto_dhcp/dhcp.c
@@ -1608,6 +1608,7 @@ int fr_dhcp_encode(RADIUS_PACKET *packet)
 	vp_cursor_t	cursor;
 	VALUE_PAIR	*vp;
 	uint32_t	lvalue;
+	uint16_t	svalue;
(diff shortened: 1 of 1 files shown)


alandekok:
the code works only because it's on a little endian CPU. :(

nchaigne:
What about the "bytes" then ? does the following work for big endian ?

	/* DHCP-Hardware-Type */
	if ((vp = fr_pair_find_by_num(packet->vps, 257, DHCP_MAGIC_VENDOR, TAG_ANY))) {
		*p++ = vp->vp_integer & 0xFF;

Or shouldn't this be instead :

		*p++ = vp->vp_byte;

?

alandekok:
Yeah, that's wrong, too.  I've pushed a fix.

  The problem is left over from when byte / short / / integer were all stored in vp_integer.  When that changed, all of the code using byte / short types should have been updated, too.

Issue #1144, opened 2015-07-16
Add radius attribute to the freeradius schema


sammarkhodja:
Hello,

I would like the « radiusMaxDailySession » attribute to be added to the freeradius schema in order to be able to use it for a project.

Thank you.

arr2036:
You're welcome to open a pull request.

PR #1145, opened 2015-07-16
Adding radiusMaxDailySession to LDAP schema

MCMic:
PR as asked in ticket #1144


alandekok:
Why is this needed?  What is the use-case for it?

MCMic:
I don’t know the exact use case as I’m not the author of #1144 , but shouldn’t all options be available through the LDAP backend? What it the rule?

alandekok:
There are about 8000 RADIUS attributes in all of the various dictionaries.  There is *no* need to have all of them in LDAP.

arr2036:
There's a generic ldap attribute that can represent all attributes.  No need for specific ldap schema attributes.

Whoever opened the issue didn't specify *which* schema.  So it was easier to say 'show me'.

MCMic:
What is the generic attribute?

arr2036:
``radiusAttribute``

arr2036:
There there are list specific ones ``radiusControlAttribute`` ``radiusReplyAttribute`` ``radiusRequestAttribute ``

PR #1146, opened 2015-07-16
fixed message "hex string is not even"

nchaigne:
Length of Hex String is not even, got 7 bytes

Is better than:
Length of Hex String is not even, got 18446744073709551615 bytes


arr2036:
I agree!

PR #1147, opened 2015-07-16
rlm_detail: Set maximum value of poll_interval to 300

jpereira:
Give the option to up the maximum interval.

### DIFF
--- a/raddb/sites-available/buffered-sql
+++ b/raddb/sites-available/buffered-sql
@@ -77,7 +77,7 @@ server buffered-sql {
 		#  If the detail file doesn't exist, the server will
 		#  wake up, and poll for it every N seconds.
 		#
-		#  Useful range of values: 1 to 60
+		#  Useful range of values: 1 to 300
 		poll_interval = 1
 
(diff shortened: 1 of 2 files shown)


alandekok:
Why is this change necessary?

jpereira:
I have a lot of unnecessary messages in the log. because, I had some problems during creation/receiving of acct packets using the "buffered-sql"

And after increasing the value, I solve the boring messages.

alandekok:
It's better to fix the messages than to put a band-aid on it.

jpereira:
@alandekok haha! I know.... I Know, but just give us the option to increase the value. Just it! :beer:

alanbuxey:
why stop at 300? why 5 minutes maximum poll?  there are issues with such a
large poll time - the size of the detail file being one(!)  but
what issue were you having that meant a poll time of eg 60 seconds was a
problem?  what aggressiveness value were ou using etc?  if this value is to
be changed
then maybe make 1 hour - 3600 - the maximum value?

alan

PR #1151, opened 2015-07-17
Build without pcap and if/packet

nchaigne:


nchaigne:
Actually I stumbled onto this because HAVE_LINUX_IF_PACKET_H was not defined. But it should have been, I'm on a Linux box...

Maybe there's an issue with the configure tools ?

PR #1154, opened 2015-07-20
Update ldap

alanbuxey:
document the other option (hard) and correct the default value assertion


arr2036:
The default is actually to use libldap's default which depends on the configuration on the system

PR #1156, opened 2015-07-21
Add NULL check in originate-coa

qnet-herwin:
This prevents segfaults that may occur when the COA list could not be created, either by a lack of memory, or a request that is not an Access-Request/Accounting-Request.

Easiest way to reproduce:

Create a CoA request from the coa virtual server
```
server coa {
  recv-coa {
    update coa {
      &NAS-IP-Address = '127.0.0.1'
    }
  }
}
```

Then, send some CoA packet to this server
```
echo 'User-Name = test' | radclient 127.0.0.1:3799 coa testing123
```

The segfault occurs because creating CoA packets is only supported for {Access,Accounting}-Request messages (which is also documented). It may also happen when there is a lack of memory and the talloc call fails, but that is harder to test.


qnet-herwin:
BTW, I haven't tested this with 3.1.x, but I wouldn't be surprised if the same problem occured there too.

PR #1158, opened 2015-07-21
dhcp - interface / raw socket mode - fixup

nchaigne:
At some point this was working (actually on 3.0.x, I'm not sure this
ever built on 3.1.x), but it broke a long time ago. Nobody noticed since
the relevant define was not set by autoconf...

### DIFF
--- a/configure
+++ b/configure
@@ -8088,7 +8088,8 @@ for ac_header in \
   siad.h \
   features.h \
   limits.h \
-  sys/event.h
+  sys/event.h \
+  linux/if_packet.h
 
 do :
   as_ac_Header=`$as_echo "ac_cv_header_$ac_header" | $as_tr_sh`
--- a/src/include/autoconf.h.in
+++ b/src/include/autoconf.h.in
@@ -206,6 +206,9 @@
 /* Define to 1 if you have the <limits.h> header file. */
 #undef HAVE_LIMITS_H
 
+/* Define to 1 if you have the <linux/if_packet> header file. */
+#undef HAVE_LINUX_IF_PACKET_H
+
 /* Define to 1 if you have the `localtime_r' function. */
 #undef HAVE_LOCALTIME_R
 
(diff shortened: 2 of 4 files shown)


nchaigne:
Damn... it doesn't build when pcap and if_packet are both available.
So, let's see, we have to handle:
- no pcap, no if_packet
- pcap, no if_packet
- no pcap, if_packet
- pcap and if_packet

Seems overly complicated :'(

nchaigne:
Now Travis complains about the rest module, I did not have this before.
Not related to dhcp anyway...

src/modules/rlm_rest/rest.c:2014:2: error: the address of ‘timeout’ will always evaluate as ‘true’ [-Werror=address]

arr2036:
Fixed a large number of these just now. Grr. Looks like autoheader/autoconf wasn't run before submitting the pull request, which masked a significant number of issues in the code.  I think all your fixes are covered in the commit I just did, if not, please could you open another PR for the remaining ones.

Cheers!

nchaigne:
It doesn't build if you don't have pcap. The following defines (in src\modules\proto_dhcp\dhcp.c) are required to build with raw sockets:

#ifdef HAVE_PCAP_H
#define ETH_TYPE_IP    0x0800
#define IP_HDR_SIZE    20
#define UDP_HDR_SIZE   8
#define ETH_ADDR_LEN   6
#endif

If you remove the #ifdef / #endif around those, build works.

arr2036:
I guess it's ifdef HAVE_PCAP_H or the linux raw socket header

arr2036:
(clang on OSX complains if there are unused macros now)

arr2036:
Ok, done

nchaigne:
It works again, thanks :)

There's a minor issue with DEBUG: the macro does not add a "\n". Some uses of DEBUG add a "\n", some don't.

I see that radclient's DEBUG macro adds a "\n". Maybe dhcpclient should also.

arr2036:
Yeah in general the macro should add the new line.  If you need multiline output, call the macro multiple times :)

I think the only place where that's ever been an issue is in the SQL drivers, the PG and MySQL client libraries love to produce all kinds of weird intermixed errors, with server errors and client errors and network errors *sigh*.

PR #1174, opened 2015-08-03
misc fixes

leprechau:
* change json.h detection to match rlm_rest
* remove un-needed json.h includes
* fix upstream/map backwards compat check with assignment

### DIFF
--- a/src/modules/rlm_couchbase/config.h.in
+++ b/src/modules/rlm_couchbase/config.h.in
@@ -1,7 +1,13 @@
 /* config.h.in.  Generated from configure.ac by autoheader.  */
 
+/* json.h is at json-c/json.h relative to include dir */
+#undef HAVE_JSONMC_JSON_H
+
 /* Define to 1 if you have the `json_c_version' function. */
 #undef HAVE_JSON_C_VERSION
+
+/* json.h is at json/json.h relative to include dir */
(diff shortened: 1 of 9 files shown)


arr2036:
Ak, you're too quick :) 

I'm re-organising some of the JSON stuff into a separate library which is included by the various things that use json-c (rlm_json, rlm_rest, rlm_couchbase).  It'll have the missing functions, the alternate header locations, and some common marshalling code.

leprechau:
Nice, I saw that pull request.  It would indeed be nice to have that in a common place.  This did clue me into a mismatched int to pointer conversion.  I'll go ahead and fix that.

leprechau:
Okay, that's fixed.  Not sure how that got broken or possibly the timeout value wasn't ever working.  I'll rebase into one commit and leave to your discretion.

arr2036:
Stripped the autoconf stuff, as that'll be fixed when #1173 is merged

arr2036:
@leprechau one thing you might to check in v3.0.x is the use of ``json_object_get_type`` and ``json_object_is_type``. Both will segv with < ``0.10`` when passed a 'null' (represented by NULL pointer).  I wrote wrapper macros for v3.1.x, you may want to backport them if you see any places where an object could be null/NULL.

PR #1180, opened 2015-08-04
added NULL checks

leprechau:
additional null checks per comment on #1174 by @arr2036


jpereira:
This is redundant as found in the json library source[1] and document[2]

[1] https://github.com/json-c/json-c/blob/b594c34f5796ac9b404635e531ea7f73a6e15941/json_object.c
[2] https://json-c.github.io/json-c/json-c-0.10/doc/html/json__object_8h.html#adc5e998b4897b6e450bd0975a58d2f72

arr2036:
We support versions earlier than 0.10

leprechau:
From json_object.c in the 0.9 release as used by Travis:

```c
enum json_type json_object_get_type(struct json_object *jso)
{
  return jso->o_type;
}
```

No null check.

PR #1191, opened 2015-08-15
Added tag to map

plambrechtsen:
Added tag to map_to_vp.


arr2036:
Yeah that function was never intended to be called directly, more as a callback to ``map_to_request`` which *does* perform the fixups.  But I can see a couple of places where it now is, like the generic attribute code in ldap, and I guess the cache creation code... not that that'd have an effect on anything.  You ran into this with LDAP i'm guessing?

plambrechtsen:
Yes I did run into this issue in LDAP. Parsing the Tunnel-* values. 
Many thanks for that. :)

PR #1193, opened 2015-08-17
Develop/rlm sql using client afrom cs dev

arr2036:


arr2036:
This is redundant with the work/design in v4.0.x

PR #1198, opened 2015-08-19
Unnecessary buffer, just call fr_strerror() directly

jpereira:


arr2036:
Thanks! Yes, I added a hack to allow fr_strerror() to be called from snprintf without destroying the buffer contents :)

jpereira:
sure! :+1: 

could be good replicate in the v3.0.x

arr2036:
nope :P

PR #1201, opened 2015-08-26
Added support for Boingo Wi-Fi

jpereira:


arr2036:
Should add it to the main dictionary too?

jpereira:
my bad.. sorry! try again! :+1:

PR #1204, opened 2015-09-01
Fix libs

peruchi:
Include Libs


pwdng:
This change considers rlm_example.so, rlm_idn.so & rlm_smsotp.so to be stable.
The rpm build now fails as these modules are not built by default.

PR #1207, opened 2015-09-02
Fix memory leak in radius_compare_vps()

jpereira:


qnet-herwin:
I guess we'd also like a line `if (preg) talloc_free(preg);` in the part below `regex_error`. It's allocated in `regex_compile`, but has two chances of not being freed.

PR #1211, opened 2015-09-08
Prototype string escaping API for rlm_sql

unikmhz:
I'm currently trying to port our custom GPL'ed FreeRADIUS module to 3.x. In addition to updating it to work on newer versions I tried to convert it from using our own half-assed SQL abstraction layer to rlm_sql's (using rlm_sqlhpwippool as an inspiration). While doing that I noticed that there are no DB-specific mechanisms to escape strings for use in DB queries (sql_escape_func is too strict and doesn't know about DB escaping rules).

So my question is - is it ok for me to extend rlm_sql with DB-specific string escaping API, and if so, what are your suggestions and requirements for such an API?

One thing to consider is that data-string-escaping and identifier-string-escaping follow different rules in several DBs.

This pull request is a prototype implementation of such an API. It consists of:
* New function pointer in ```rlm_sql``` module instance (```sql_string_escape_func```). It has the same signature as ```sql_escape_func```, so it can be used as an xlat escape.
* It defaults to ```sql_escape_func```.
* If a driver implements ```sql_escape_string``` method, it gets used instead.

Please send comments, criticism and flames my way. Thanks.

### DIFF
--- a/src/modules/rlm_sql/drivers/rlm_sql_mysql/rlm_sql_mysql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_mysql/rlm_sql_mysql.c
@@ -712,7 +712,7 @@ static int sql_affected_rows(rlm_sql_handle_t *handle, UNUSED rlm_sql_config_t *
 
 static size_t sql_escape_string(rlm_sql_handle_t *handle,
 			UNUSED rlm_sql_config_t *config, char *out, size_t outlen,
-			char const *in, UNUSED void *arg)
+			char const *in)
 {
 	size_t qlen;
 	rlm_sql_mysql_conn_t *conn = handle->conn;
@@ -720,6 +720,8 @@ static size_t sql_escape_string(rlm_sql_handle_t *handle,
 	/* Check for potential buffer overflow */
 	qlen = strlen(in);
 	if ((qlen * 2 + 1) > outlen) return 0;
+	/* Prevent integer overflow */
+	if ((qlen * 2 + 1) <= qlen) return 0;
 	return mysql_real_escape_string(conn->sock, out, in, qlen);
 }
 
--- a/src/modules/rlm_sql/drivers/rlm_sql_mysql/rlm_sql_mysql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_mysql/rlm_sql_mysql.c
@@ -710,6 +710,19 @@ static int sql_affected_rows(rlm_sql_handle_t *handle, UNUSED rlm_sql_config_t *
 	return mysql_affected_rows(conn->sock);
 }
 
+static size_t sql_escape_string(rlm_sql_handle_t *handle,
+			UNUSED rlm_sql_config_t *config, char *out, size_t outlen,
+			char const *in, UNUSED void *arg)
+{
+	size_t qlen;
+	rlm_sql_mysql_conn_t *conn = handle->conn;
+
+	/* Check for potential buffer overflow */
+	qlen = strlen(in);
+	if ((qlen * 2 + 1) > outlen) return 0;
+	return mysql_real_escape_string(conn->sock, out, in, qlen);
+}
+
 
 /* Exported to rlm_sql */
 extern rlm_sql_module_t rlm_sql_mysql;
@@ -729,5 +742,6 @@ rlm_sql_module_t rlm_sql_mysql = {
 	.sql_free_result		= sql_free_result,
 	.sql_error			= sql_error,
 	.sql_finish_query		= sql_finish_query,
-	.sql_finish_select_query	= sql_finish_query
+	.sql_finish_select_query	= sql_finish_query,
+	.sql_escape_string		= sql_escape_string
 };
--- a/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
@@ -507,7 +507,7 @@ static int sql_affected_rows(rlm_sql_handle_t * handle, UNUSED rlm_sql_config_t 
 
 static size_t sql_escape_string(rlm_sql_handle_t *handle,
 			UNUSED rlm_sql_config_t *config, char *out, size_t outlen,
-			char const *in, UNUSED void *arg)
+			char const *in)
 {
 	int err;
 	size_t qlen;
@@ -516,6 +516,8 @@ static size_t sql_escape_string(rlm_sql_handle_t *handle,
 	/* Check for potential buffer overflow */
 	qlen = strlen(in);
 	if ((qlen * 2 + 1) > outlen) return 0;
+	/* Prevent integer overflow */
+	if ((qlen * 2 + 1) <= qlen) return 0;
 	qlen = PQescapeStringConn(conn->db, out, in, qlen, &err);
 	if (err) return 0;
 	return qlen;
--- a/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
@@ -505,6 +505,22 @@ static int sql_affected_rows(rlm_sql_handle_t * handle, UNUSED rlm_sql_config_t 
 	return conn->affected_rows;
 }
 
+static size_t sql_escape_string(rlm_sql_handle_t *handle,
+			UNUSED rlm_sql_config_t *config, char *out, size_t outlen,
+			char const *in, UNUSED void *arg)
+{
+	int err;
+	size_t qlen;
+	rlm_sql_postgres_conn_t *conn = handle->conn;
+
+	/* Check for potential buffer overflow */
+	qlen = strlen(in);
+	if ((qlen * 2 + 1) > outlen) return 0;
+	qlen = PQescapeStringConn(conn->db, out, in, qlen, &err);
+	if (err) return 0;
+	return qlen;
+}
+
 /* Exported to rlm_sql */
 extern rlm_sql_module_t rlm_sql_postgresql;
 rlm_sql_module_t rlm_sql_postgresql = {
@@ -520,5 +536,6 @@ rlm_sql_module_t rlm_sql_postgresql = {
 	.sql_error			= sql_error,
 	.sql_finish_query		= sql_free_result,
 	.sql_finish_select_query	= sql_free_result,
-	.sql_affected_rows		= sql_affected_rows
+	.sql_affected_rows		= sql_affected_rows,
+	.sql_escape_string		= sql_escape_string
 };
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -701,7 +701,7 @@ static int sql_get_grouplist(rlm_sql_t *inst, rlm_sql_handle_t **handle, REQUEST
 	entry = *phead = NULL;
 
 	if (!inst->config->groupmemb_query || !*inst->config->groupmemb_query) return 0;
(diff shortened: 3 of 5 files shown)


unikmhz:
Travis build borked in ```value_data_snprint``` function. Is it my doing, or is build currently failing in general?

alandekok:
Over all it looks good.  Simple is nice.

I'd prefer to have the finished commits merged in, rather than commit one thing...fix it... commit another.

You can re-write your commit history and do a force push to update the issue.

As for travis, I've pushed fixes.

unikmhz:
Rewritten as 2 commits:
1. Core API and modification of structures.
2. Implementation for mysql/postgres.

Is this ok?

arr2036:
Reserving a connection every time you want to escape an attribute value in not a friendly thing to do.  That'd introduce significantly more contention on the conn pool mutex.  You'd end up reserving and releasing a connection 10-15 times for some of the default queries.

You should modify the calls that do sql xlat expansions to pass the connection handle in as the ``void *arg``, and not reserve a connection handle yourself.

Once you've made that change there's little point in keeping the wrapper function around in the main rlm_sql code, just have each driver provide it's own xlat escape function, with the ones that you've not written proper escape function for, using the current safe chars based one.

unikmhz:
Ok, thanks. I'll update this pull request.

unikmhz:
I preface this with a disclaimer: I'm a long way from being acquainted with FreeRADIUS codebase, so anything that follows might be terribly wrong.

I found two non-trivial bits about this new approach that @arr2036 suggested:
 1. ```xlat_register``` is apparently hard-wired to pass module instance as escape function's ```void *arg``` argument, and additionally can't be dynamic (i.e. can't be dynamically acquired from a connection pool).
 2. Similar problems with ```map_proc``` subsystem (its ```mod_inst``` argument, which is used in several places besides being passed as an escape function's argument, is defined at register time).

A solution I can implement:
 1. Making ```sql_string_escape_func``` not be signature-compatible with ```sql_escape_func``` (it is not used anywhere yet).
 2. Using a default function for a case when driver doesn't implement its own string escaping, that will call ```sql_escape_func``` with ```handle->inst``` as last argument (to DRY, and to maintain being able to customize```allowed_chars```).

An alternative solution would involve rewriting parts of xlat and map_proc subsystems, which I can try, but I'm not feeling very confident about.

arr2036:
Hmm yes, the calling order is wrong for the map_proc and sql xlat functions. I was thinking more for the expansions that get done by rlm_sql itself. I'll think about it some more...

arr2036:
http://dev.mysql.com/doc/refman/5.1/en/c-api-threaded-clients.html mysqlclient docs say the same connection can be shared. Checking PG...

arr2036:
Reading the code for libpq, it should be fine too.  The only reason these functions take a connection is to get the character encoding scheme from the server, and that's done during the initial connection.

```c
size_t
PQescapeStringConn(PGconn *conn,
                                   char *to, const char *from, size_t length,
                                   int *error)
{
        if (!conn)
        {
                /* force empty-string result */
                *to = '\0';
                if (error)
                        *error = 1;
                return 0;
        }
        return PQescapeStringInternal(conn, to, from, length, error,
                                                                  conn->client_encoding,
                                                                  conn->std_strings);
}

/*
 * Escaping arbitrary strings to get valid SQL literal strings.
 *
 * Replaces "'" with "''", and if not std_strings, replaces "\" with "\\".
 *
 * length is the length of the source string.  (Note: if a terminating NUL
 * is encountered sooner, PQescapeString stops short of "length"; the behavior
 * is thus rather like strncpy.)
 *
 * For safety the buffer at "to" must be at least 2*length + 1 bytes long.
 * A terminating NUL character is added to the output string, whether the
 * input is NUL-terminated or not.
 *
 * Returns the actual length of the output (not counting the terminating NUL).
 */
static size_t
PQescapeStringInternal(PGconn *conn,
                                           char *to, const char *from, size_t length,
                                           int *error,
                                           int encoding, bool std_strings)
{
        const char *source = from;
        char       *target = to;
        size_t          remaining = length;

        if (error)
                *error = 0;

        while (remaining > 0 && *source != '\0')
        {
                char            c = *source;
                int                     len;
                int                     i;

                /* Fast path for plain ASCII */
                 if (!IS_HIGHBIT_SET(c))
                {
                        /* Apply quoting if needed */
                        if (SQL_STR_DOUBLE(c, !std_strings))
                                *target++ = c;
                        /* Copy the character */
                        *target++ = c;
                        source++;
                        remaining--;
                        continue;
                }

                /* Slow path for possible multibyte characters */
                len = pg_encoding_mblen(encoding, source);

                /* Copy the character */
                for (i = 0; i < len; i++)
                {
                        if (remaining == 0 || *source == '\0')
                                break;
                        *target++ = *source++;
                        remaining--;
                }

                /*
                 * If we hit premature end of string (ie, incomplete multibyte
                 * character), try to pad out to the correct length with spaces. We
                 * may not be able to pad completely, but we will always be able to
                 * insert at least one pad space (since we'd not have quoted a
                 * multibyte character).  This should be enough to make a string that
                 * the server will error out on.
                 */
                if (i < len)
                {
                        if (error)
                                *error = 1;
                        if (conn)
                                printfPQExpBuffer(&conn->errorMessage,
                                                  libpq_gettext("incomplete multibyte character\n"));
                        for (; i < len; i++)
                        {
                                if (((size_t) (target - to)) / 2 >= length)
                                        break;
                                *target++ = ' ';
                        }
                        break;
                }
        }

        /* Write the terminating NUL character. */
        *target = '\0';

        return target - to;
}
```

unikmhz:
Well, both postgres and mysql have string escape functions which don't need connection handles passed as arguments (```PQescapeString()``` and ```mysql_escape_string()``` respectively), but they both don't respect connection character sets, which can produce weird/incorrect results (as a russian user, I can attest to this). In addition to this, ```PQescapeString()``` is described in documentation as a "security hazard" when used in multi-threaded programs, and ```mysql_escape_string()``` has a note that eloquently says "This function should not be used".

arr2036:
The PG docs are being a bit dramatic.  The main issue I can see is that the default client encoding and string rules are stored in globals, which appear to be set from the last connection opened.

If you have multiple connections to *different* databases, then you'll get some odd behaviour, if you have multiple connections to the same database, it would be fine.  Unfortunately, for us, we allow multiple instances of rlm_sql, so you could have connections to different databases with different encodings.

So for PG at least, your code where you reserved, used and released a connection would work and appears to be the only option.

For MySQL it would be better to create a single connection during driver instantiation, and use that for all escaping.  But that would mean we opened more than the configured max connections which in some situations could cause issues, so again I guess we have to use the pool.

This applies to the escaping function for sql xlat and sql map proc *only*.  For cases where rlm_sql performs string expansion itself, it's much better to pass in the connection as I described above.  Examples of those would be rlm_sql.c:671/831/881/1191/1239/1419/1466/1616/1664.

So if the per driver escape function has the same signature as xlat_escape_t, you can use your wrapper ``sql_string_escape_func`` for xlat_register and map_proc_register, and pass in ``inst->module->sql_escape_string`` as the escape func for calls to radius_axlat et al.

I think that should work for everything, and it's not very much more code than what you have in the current pull request.

unikmhz:
Ok, wil push to this PR and await further comments.

unikmhz:
One more question - is it ok if I apply string escaping (as opposed to identifier escaping or literal escaping) to xlats in rlm_sql? I don't yet know all the ways escaping is applied within xlats, so it may be problematic.

arr2036:
Theoretically identifiers can be specified with expansions too, and there's no way to identify them as such without parsing the query fully and understanding its structure.  If string escaping will produce output that can be used for identifiers and literals, then that's fine.

unikmhz:
Ok, added 2 commits:
1. Rewrites new API to accept connection handle instead of module instance as ```void *arg```.
2. Makes all in-tree explicit SQL xlats use new function and pass connection handles to it.
The not-so-nice thing about this is that escaping for xlats from configuration and from queries will follow different rules. But it's a start.

P.S.: does travis perform runtime tests, or only compile checks? I'm not so sure that rlm_sql will work with current default configuration (missing quotation marks etc.).
P.P.S.: travis is borked again, this time as ```src/main/process.c:1166:5: error: nested extern declaration of ‘debug_log’```

arr2036:
Alan's commit borked it https://github.com/FreeRADIUS/freeradius-server/commit/0ede316ed108bab795371919d0b5d6d1146d562f

arr2036:
Both, but not all modules have runtime tests.

alandekok:
I've pushed fixes

unikmhz:
Ok, I found SQL-related tests, so there's hope for me ;)
Can someone restart travis build for this PR?

arr2036:
no, you need to pull --rebase off upstream-v3.1.x and git push --force

arr2036:
but wait until all the issues are fixed, there's still a nested extern one in process.c

unikmhz:
Yes, sorry, I forgot that I push to my own forked repo.

There is another potential problem I found: there can in theory be an integer overflow in code ```if ((qlen * 2 + 1) > outlen) return 0;``` inside drivers' escape functions. Should I do anything about it, or is it too paranoid?

alandekok:
On Sep 9, 2015, at 10:33 AM, Alex Unigovsky <notifications@github.com> wrote:

> Yes, sorry, I forgot that I push to my own forked repo.
> 
> There is another potential problem I found: there can in theory be an integer overflow in code if ((qlen * 2 + 1) > outlen) return 0; inside drivers' escape functions. Should I do anything about it, or is it too paranoid?

  TBH, don't worry about it.  If the caller allocates a 2G string, there are other problems with the system.

  The server core ensures that strings read from config files are small.

arr2036:
Yeah but not all strings come from the config.  Not that you'd expect to be grabbing a 2G blob from a Rest API call.  @unikmhz if you don't mind fixing then real free.

arr2036:
> The not-so-nice thing about this is that escaping for xlats from configuration and from queries will follow different rules. But it's a start.

@unikmhz Rules should be the same? What did you mean by the above.

unikmhz:
> @unikmhz Rules should be the same? What did you mean by the above.
I mean that escapes from ```xlat_register``` and ```map_proc``` still use original ```sql_escape_func``` function which doesn't use driver-specific escaping.

Ok, added overflow checks. I didn't remove the ```sql_string_escape_func``` proxy function because otherwise driver code would need to mention ```REQUEST```, that is otherwise not found in driver sources. I can eliminate it if you think it is cleaner that way.

unikmhz:
@arr2036, I can add this chunk to beginning of sql_escape_func to make all cases use same escaping rules:
```c
    rlm_sql_handle_t *handle;
    [...snip...]
    if (inst->module->sql_escape_string) {
        handle = fr_connection_get(inst->pool);
        if (!handle) return 0;
        len = inst->module->sql_escape_string(handle, inst->config, out, outlen, in);
        fr_connection_release(inst->pool, handle);
        return len;
    }
```

arr2036:
I've reworked your code to simplify it.  The only remaining issue is that \r \t \n won't be converted to their ascii equivalents with the driver specific functions, whereas they would have been with the original one.

unikmhz:
Thanks @arr2036. Your cleaned up version looks indeed much simpler and more pleasing to the eye.

The thing with "\n" and friends is - we can't guess how to escape non-printable characters, only DB's client library knows that. And at least in the case of ```mysql_real_escape_string``` the library function encodes both ```\n``` and ```\t``` (among other characters).

arr2036:
Ah I guess so, and the various SQL dialects will probably have ``\<char>`` equivalent sequences.  After many years it appears the best way to deal with escape sequences is just to leave them alone, and be as transparent as possible between the producer and consumer. @alandekok went through and fixed them up.

OK I guess this is done then, thanks for the patches!  The next stage is to automatically convert the pre-parsed expansions to stored procedures, which will completely resolve the escaping issue, and fix the acquire/release issue with sql func expansions.

PR #1216, opened 2015-09-09
Optionally send rejects without a delay

qnet-herwin:
Currently there is only one global option to set a delay to every Access-Reject packet: reject_delay. There are use cases where you want certain rejects to have no delay, while others should have a delay. An example might be using 802.1X on Cisco LAN Devices: If a client tries MAC authentication an Access-Reject can force it to switch to 802.1X, this is a reject you want to send without any delay. On the other hand, if the client tries 802.1X with a wrong password, you still want the reject to be delayed.

The config option reject_delay can be bypassed by adding the attribute No-Reject-Delay to control, and setting it to a value larger than 0. Setting it to 0 or not adding it results in the old behaviour.

The implementation is the simplest version I could think of. A more advanced version could read the delay time from the VP, instead of using it as a boolean. This would require that all the sanity checks that are currently in mainconfig.c should be copied here as well (or made more generic).

### DIFF
--- a/src/main/process.c
+++ b/src/main/process.c
@@ -1400,6 +1400,27 @@ static void request_finish(REQUEST *request, int action)
 	    (request->root->reject_delay.tv_sec > 0)) {
 		request->response_delay = request->root->reject_delay;
 
+		vp = fr_pair_find_by_num(request->reply->vps, PW_FREERADIUS_RESPONSE_DELAY, 0, TAG_ANY);
+		if (vp) {
+			if (vp->vp_integer <= 10) {
+				request->response_delay.tv_sec = vp->vp_integer;
+			} else {
+				request->response_delay.tv_sec = 10;
+			}
+			request->response_delay.tv_usec = 0;
+		} else {
+			vp = fr_pair_find_by_num(request->reply->vps, PW_FREERADIUS_RESPONSE_DELAY_USEC, 0, TAG_ANY);
+			if (vp) {
+				if (vp->vp_integer <= 10 * USEC) {
+					request->response_delay.tv_sec = vp->vp_integer / USEC;
+					request->response_delay.tv_usec = vp->vp_integer % USEC;
+				} else {
+					request->response_delay.tv_sec = 10;
+					request->response_delay.tv_usec = 0;
+				}
+			}
+		}
+
 #ifdef WITH_PROXY
 		/*
 		 *	If we timed out a proxy packet, don't delay
--- a/src/main/process.c
+++ b/src/main/process.c
@@ -1160,8 +1160,6 @@ static void request_response_delay(REQUEST *request, int action)
 	case FR_ACTION_TIMER:
 		fr_event_now(el, &now);
 
-		rad_assert(request->response_delay.tv_sec > 0);
-
 		/*
 		 *	See if it's time to send the reply.  If not,
 		 *	we wait some more.


alandekok:
There's already FreeRADIUS-Response-Delay and FreeRADIUS-Response-Delay-USec.

It's better to re-use those.  If the response delay is zero, then send it out immediately.

qnet-herwin:
@alandekok: I'm looking at that now, but I discovered that those delays can't be send with fractional seconds, because of the following assertions:

```C
rad_assert(request->response_delay.tv_usec == 0);
```

What would be the correct way to handle the `FreeRADIUS-Response-Delay-USec` values that are not exact seconds?  I could think of the following options:

* Divide the value by `USEC` to get the seconds, ignore the remainder
* Divide the value by `USEC` to get the seconds, print a log-warning if the remainder is not 0
* Round the value to the closest second (thus: add an extra second if the remainder is greater than or equal to `USEC/2`)
* Disallow the use of `FreeRADIUS-Response-Delay-USec`, only support `FreeRADIUS-Response-Delay`

I think the last option would be the simplest, but it might be confusing that you can use usec in DHCP, but not here.

Another question: `FreeRADIUS-Response-Delay` is currently used in DHCP, where it is read from `request->reply->vps`, should I use that key-value-store as well instead of `request->config`?

alandekok:
<shrug>  Delete the assertions.  And fix the relevant code.  It's only one place.

alandekok:
See commit 1d1c50bb0.  I've pushed a fix.

The code in mainconfig.c ensures that "reject_delay" is not at least 1.0 seconds.  Sub-second delays are forbidden there.

But it's possible in process.c to set response_delay to be 0.1s (for example).  All of the code in process.c handles sub-second response delays without an issue.

The only caveat is you need to be use that "usec" is really microseconds, when you write to it.  i.e. if the attribute `FreeRADIUS-Response-Delay-USec` is more than 1,000,000 then you just set it to zero.  And assume that the administrator did something stupid.

alandekok:
> Another question: FreeRADIUS-Response-Delay is currently used in DHCP, where it is read from request->reply->vps, should I use that key-value-store as well instead of request->config?

Yes.  For consistency with DHCP.

The idea here is that it controls the reply.  So the attribute should be in the reply.

We could have been more consistent with that for other attributes, but oh well...

qnet-herwin:
Well, moving it to reply did have some unexpected side effects:

```
Debug: (0) attr_filter.access_reject: Attribute "FreeRADIUS-Response-Delay" allowed by 0 rules, disallowed by 0 rules
Debug: (0) attr_filter.access_reject: Attribute "FreeRADIUS-Response-Delay-USec" allowed by 0 rules, disallowed by 0 rules
```

So these attributes are removed from the reply before we reach the code that delays it.

alandekok:
I've pushed a fix.  It's possible to update the default configuration if the server behavior changes.

qnet-herwin:
Pushed an update. Just one more thing I don't completely understand:

> The only caveat is you need to be use that "usec" is really microseconds, when you write to it. i.e. if the attribute `FreeRADIUS-Response-Delay-USec` is more than 1,000,000 then you just set it to zero. And assume that the administrator did something stupid.

I copied the logic from [dhcpd.c](https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/proto_dhcp/dhcpd.c#L422-L445). The `-USec` version supports everything from 1 microsecond to 10 seconds, so a value with `1,000,000 < value <= 10,000,000` should be acceptable.

alandekok:
That seems fine.

arr2036:
Why is it seconds and microseconds instead of milliseconds?  Was this related to the high precision proxy timeouts?

alandekok:
@arr2036 yes

PR #1220, opened 2015-09-09
updated variables.rst location in docs..

alanbuxey:
carrying on the earlier work of Matthew.....


mcnewton:
note that it's now variables.rst, not variables.txt... :)

alanbuxey:
doh!  :-)

PR #1231, opened 2015-09-11
small documentation fix/cleanups [ci skip]

mcnewton:
More of this should probably just be totally removed or rewritten...

### DIFF
--- a/doc/README
+++ b/doc/README
@@ -2,7 +2,7 @@
 
   All code in this server was written for this project.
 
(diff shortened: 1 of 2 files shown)


arr2036:
maybe it has to come at the start :(

mcnewton:
There's a travis bug report about it not working properly with pull requests, so it may only with with direct merges. Last try next time will be in the commit message itself (but was trying to avoid contaminating the commit log).

PR #1239, opened 2015-09-15
Fixed bug: It should return CoA-NAK when the home-server is invalid.

jpereira:
When we have a wrong home-server

```
update control {
       &Home-Server-Pool := "coa_pool_wrongset"
}
```

the server currently return

````
Tue Sep 15 19:23:12 2015 : ERROR: (5) Cannot proxy to unknown pool coa_pool_wrongset
Tue Sep 15 19:23:12 2015 : Debug: (5) Empty send-coa section.  Using default return values.
Tue Sep 15 19:23:12 2015 : Debug: (1) Sent CoA-ACK Id 215 from 192.168.1.90:3799 to 10.1.2.128:41165 length 0

```

And after the patch, fix this issue.

```
Tue Sep 15 19:53:35 2015 : ERROR: (1) Cannot proxy to unknown pool coa_pool_wrongset
Tue Sep 15 19:53:35 2015 : Debug: (1) Empty send-coa section.  Using default return values.
Tue Sep 15 19:53:35 2015 : Debug: (1) Sent CoA-NAK Id 225 from 192.168.18.90:3799 to 10.1.2.128:41160 length 0
```

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -23,6 +23,8 @@ FreeRADIUS 3.0.10 Wed 08 Jul 2015 12:00:00 EDT urgency=medium
 	* Documentation fixes from Alan Buxey and Matthew Newton.
 
 	Bug fixes
+	* Fix home-server, It should return CoA-NAK when the Home-Server-Pool
+	  is invalid.
 	* Fix rlm_files so that there are no collisions when loading
 	  10's of 1000's of users.
 	* Fix radclient to use our internal v4/v6 parsing functions.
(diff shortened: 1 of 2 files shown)


jpereira:
almost getting crazy with this issue... a wrong set by capitalize was enough to discover that. :1

alandekok:
I've pushed a fix in 6046d50942.

The issue is that do_proxy() should return -1 on failure, which it didn't before.  A proxy failure means it should reply with an immediate NAK.

A "don't proxy" return means it should reply with an ACK.

A "do proxy" return means it should reply with whatever code was returned by the home server.

jpereira:
the approach is same. ok

PR #1242, opened 2015-09-17
if has a wrong client-settings, don't rise!

jpereira:
Hi,

if you make a mistake in client-settings, the current code let you start the FreeRadius. this is REALLY BAD.... induce to a error! below a example that a client "nas-portal" was pointed to a wrong server...

```
Thu Sep 17 11:43:19 2015 : Debug: rlm_sql (sql): Adding client 192.168.22.11 (nas-portal) to mcare-portal-auth+acct clients list
Thu Sep 17 11:43:19 2015 : Debug: Adding client 192.168.22.11/32 (192.168.22.11) to prefix tree 32
Thu Sep 17 11:43:19 2015 : Error: Failed to find virtual server mcare-portal-auth+acct
Thu Sep 17 11:43:19 2015 : Warning: Failed to add client, possible duplicate?
```

It is bad... really bad if happens in production setup. if something wrong, don't rise! let me know that has something wrong. :)

after patch... more easy to "discover" a wrong set.
```
Thu Sep 17 12:37:56 2015 : Debug: Adding client 192.168.22.11/32 (192.168.22.11) to prefix tree 32
Thu Sep 17 12:37:56 2015 : Error: Failed to find virtual server mcare-portal-auth+acct
Thu Sep 17 12:37:56 2015 : Warning: Failed to add client, possible duplicate?
Thu Sep 17 12:37:56 2015 : Debug: rlm_sql (sql): Released connection (0)
Thu Sep 17 12:37:56 2015 : Error: Failed to load clients from SQL
Thu Sep 17 12:37:56 2015 : Error: /etc/freeradius/mods-enabled/sql[20]: Instantiation failed for module "sql"

### DIFF
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -272,6 +272,7 @@ static int generate_sql_clients(rlm_sql_t *inst)
 	rlm_sql_handle_t *handle;
 	rlm_sql_row_t row;
 	unsigned int i = 0;
+	int ret = 0;
 	RADCLIENT *c;
 
 	DEBUG("rlm_sql (%s): Processing generate_sql_clients",
@@ -340,7 +341,8 @@ static int generate_sql_clients(rlm_sql_t *inst)
 			WARN("Failed to add client, possible duplicate?");
 
 			client_free(c);
-			continue;
+			ret = -1;
+			break;
 		}
 
 		DEBUG("rlm_sql (%s): Client \"%s\" (%s) added", c->longname, c->shortname,
@@ -350,7 +352,7 @@ static int generate_sql_clients(rlm_sql_t *inst)
 	(inst->module->sql_finish_select_query)(handle, inst->config);
 	fr_connection_release(inst->pool, handle);
 
-	return 0;
+	return ret;
 }
 
 


arr2036:
I'm thoroughly against this being merged into v3.0.x.  This is potentially breaking change for users an should NOT go into the STABLE version of FreeRADIUS.

alandekok:
Allowing broken configurations is a bad idea.  The LDAP module refuses to instantiate itself when it fails to read clients.  The SQL module should do the same.  Silently loading invalid configurations is a bad idea.

There have been many, many, other changes in v3.0 which make config file parsing more stringent.  This change is no different.

arr2036:
Yes, but this has a significant chance of breaking working configs on upgrade.  If there are any validation issues with clients, the server will not start.  This is not a bug, it's a behaviour change, it shouldn't go into the stable version of the server, it should go into v3.1.x.

alandekok:
On Sep 17, 2015, at 1:18 PM, Arran Cudbard-Bell <notifications@github.com> wrote:
> Yes, but this has a significant chance of breaking working configs on upgrade.

  As I said, there were many similar changes in v3.0.  Why is this one special?

> If there are any validation issues with clients, the server will not start. This is not a bug, it's a behaviour change, it shouldn't go into the stable version of the server, it should go into v3.1.x.

  Not loading clients is a bug.  *Silently* not loading invalid clients is a bug.

  And yes, it's a behaviour change.  Like many other behaviour changes which happened when we increased the load-time checks of the config files.

  Again, why is this change different from all of the previous ones?

jpereira:
exactly, I wasted a lot of time to discover that the radius rise with a *wrong set* silently.

arr2036:
It didn't *silently* not load clients, it warned about not loading clients.  The reason why this one is special, is because it's very easy to accidentally add a duplicate client into SQL.  Unlike LDAP which will prevent duplicate client objects from being added, at least if they have the same RDN.

The main difference between this and other changes, is that the behaviour change may not become apparent until an arbitrary period after the upgrade.

SQL is a public interface and there may be multiple other systems that use it, they may not check for duplicate entries, because in the past there has been no reason to, it's either not been considered or has been a non issue, because when it did occur, nothing bad happened.

So one day a junior sysadmin decides that they want to do local testing on the server, and seeing that the SQL servers are used for managing all the clients, decides to add a 127.0.0.1 entry.  The next time the servers are restarted, boom, the RADIUS cluster's fucked.

I would rather this be a configurable option for sql, couchbase, and ldap.  If you really want to make this the default, we should do what we did for the other potentially breaking changes, and set the defaults in the CONF_PARSER to match the current behaviour, and change the default values in the default config to be the recommended values.

alanbuxey:
but if you allow duplicates...then which one is used...the entry in SQL or
the entry in clients.conf.... the first entry in SQL or the 2nd entry?
 we've historically had issues with duplicate entries in SQL or one in
clients.conf and another in SQL - the server never liked it...and we've had
to fix it on our server (change, modify or delete the entry) - but the
issue here was virtual-server
didnt exist?  if thats the case should it just be pointed to 'default' ?

arr2036:
First to be defined wins.  Issue was originally an invalid virtual server, but the patch above has side effects that cause the server not to start if there are duplicate clients defined.

arr2036:
We can add any sort of validation we want for v3.1.x, the config items could be temporary.  My main issue is that it's not appropriate to add such a large behaviour change in a point release.

alandekok:
On Sep 19, 2015, at 2:34 PM, Alan Buxey <notifications@github.com> wrote:

> but if you allow duplicates...then which one is used...the entry in SQL or
> the entry in clients.conf.... the first entry in SQL or the 2nd entry?
> we've historically had issues with duplicate entries in SQL or one in
> clients.conf and another in SQL - the server never liked it...and we've had
> to fix it on our server (change, modify or delete the entry) -

  The problem is that the server starts with a configuration that is wrong.  It has a warning buried inside of log messages most people don't see.

  But it starts.  And then does something OTHER than what you wanted it to do.

  Alan DeKok.

PR #1243, opened 2015-09-17
debian: Fixing logrotate script

jpereira:


alandekok:
Please check this in 3.0.10.  I've put patches in which mean you can HUP the server instead of doing a "copytruncate" on the logs.  That would be more efficient.

jpereira:
ah! ok, i will update!

jpereira:
done

jpereira:
done

jpereira:
@alandekok  got it!

jpereira:
@alandekok got it! updated

arr2036:
Copytruncate is the better command to use.  There's no reason to change the logrotation scripts.

arr2036:
killall isn't present in all systems, hup has the side effect of loading config changes.  If you want to make this work reliably, the command socket command is the thing to use, but as not everyone has that enabled, copyrotate is the best way of rotating lots.

alandekok:
Pretty much every other daemon on the planet uses HUP to rotate the logs, and to re-read the configuration.

I don't have a lot of sympathy for people who make random changes to the server and leave them in the raddb directory without (a) testing them, or (b) making them live.

The HUP in 3.0.10 will re-load ONLY those modules which have changed configurations.  Any module or config file which has an error will be ignored, and the old configuration will be remain in use.

jpereira:
I agree with @alandekok

arr2036:
But when you have an alternative which does not have side effects, it's better to use that, surely?

arr2036:
and, yes, all those other daemons were shit and did it wrong too.

PR #1244, opened 2015-09-17
logrotate: killall isn't present on all systems, change to kill

jpereira:


arr2036:
I've reverted the change.  It is *WRONG* to use a mechanism for log rotation that has side effects.  Just because it's been used in the past, it doesn't mean it was right then. IT HAS ALWAYS BEEN WRONG.

PR #1251, opened 2015-09-19
Update sql_postgres

alanbuxey:


alandekok:
Some kind of description would be useful.  "Do stuff" is a bad commit message.

Also, please only supply pull requests for v3.0.x.  We can pull changes between branches.  And it's less work than managing pull requests.

alanbuxey:
sorry! just some basic documentation and example for an SSL connection with
postgresql

PR #1263, opened 2015-09-22
First attempt at a watch file

mans0954:
Provide a debian/watch file and PGP key for identifying new versions and verifying their download.

See https://wiki.debian.org/debian/watch for details.

### DIFF
--- a/debian/upstream/signing-key.asc
+++ b/debian/upstream/signing-key.asc
@@ -1,0 +1,13 @@
+-----BEGIN PGP PUBLIC KEY BLOCK-----
+Version: GnuPG v1.0.6 (GNU/Linux)
+Comment: For info see http://www.gnupg.org
+
+mQCNAzx7wFMAAAEEALq2yahNGENq7Z8xqIaaxlMYPEqdnWme+QQRobX+0mHJ+xjv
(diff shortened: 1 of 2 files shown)


arr2036:
Is that Alan's key?

arr2036:
and thanks for helping fix up the Debian packages!

mans0954:
Yes, downloaded from here: http://freeradius.org/pgp/aland@freeradius.org I suggest you check I haven't tampered with it :)

arr2036:
Ah just we created a new non-developer specific signing key.  But I guess the switchover needs coordinating a bit more.

arr2036:
If you felt like working with the Debian guys, we really need a new package maintainer...

mans0954:
A project signing key sounds sensible. Packaging v3 for Debian is being discussed in Debian bug #797181 https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=797181

alandekok:
Please double-check the v3.0.x branch.  We've pulled over some of your changes from v3.1.x, but not all.

The 3.0.10 release will be taken from the v3.0.x branch, *not* v3.1.x.

alanbuxey:
yes, they need to be aware of the relationship between 3.1.x and 3.0.x - as
Alan says, 3.0.10 is from 3.0.x HEAD with bug fixes, security updates etc
3.1.x HEAD is the development bridgehead with new features, breaking of
older configs etc etc... (there be dragons here! etc etc) :-)

PR #1270, opened 2015-09-24
debian package fails trying to build sqlhpwippool

mcnewton:
Fails trying to run "configure" in src/modules/rlm_sqlhpwippool at make time,
but having disabled it in configure, it all builds fine, including sqlhpwippool...

Probably wants reviewing to make sure this is sane - system here is debian wheezy, so maybe it's fine on jessie, but I can't easily test right now.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -63,6 +63,7 @@ include /usr/share/quilt/quilt.make
 
 config.status: configure
 	dh_testdir
+	dh_testroot
 
 ifeq (config.sub.dist,$(wildcard config.sub.dist))
 	rm config.sub


mcnewton:
Quickly managed to build on jessie and works fine without this patch, but with this patch then _doesn't_ build sqlhpwippool (but otherwise OK), so this might not be the best solution. Not sure what else to suggest though. I don't know why the make stage is running configure.

qnet-herwin:
I have built a lot of Debian-packages, and never had any problem with that module. The source of it doesn't really point to development packages that may be missing.

What exactly is the error that you're getting?

mcnewton:
OK, I've had time to so some more testing on this now. Rather than my previous build machine which could have all sorts of random packages installed, this is in a clean wheezy chroot.

Chroot is built with:

    debootstrap wheezy /tmp/wheezy
    chroot /tmp/wheezy
    apt-get update
    apt-get -y install git build-essential libtalloc-dev libssl-dev debhelper quilt
    apt-get -y install autotools-dev libcurl4-openssl-dev libcap-dev libgdbm-dev
    apt-get -y install libiodbc2-dev libkrb5-dev libldap2-dev libpam0g-dev libpcap-dev
    apt-get -y install libperl-dev libmysqlclient-dev libpq-dev libreadline-dev
    apt-get -y install libsasl2-dev libsqlite3-dev libwbclient-dev libyubikey-dev
    apt-get -y install python-dev libjson0 libjson0-dev
    cd /root
    git clone https://github.com/FreeRADIUS/freeradius-server.git
    cd freeradius-server
    git checkout -b v3.0.x origin/v3.0.x

The following works, and successfully builds the packages:

    ./configure -C
    make deb

However, the following (which is how I usually build debian packages) fails:

    dpkg-buildpackage -us -uc

The last part of the output is as follows:

    ...
    configure: creating ./config.status
    config.status: creating all.mk
    rm config.guess
    mv config.guess.dist config.guess
    rm config.sub
    mv config.sub.dist config.sub
    /usr/bin/make
    make[1]: Entering directory `/root/freeradius'
    CONFIGURE src/modules/rlm_sqlhpwippool/
    configure: WARNING: unrecognized options: --disable-developer, --disable-openssl-version-check, --with-raddbdir, --with-logdir, --with-large-files, --with-udpfromto, --without-rlm_eap_tnc, --with-rlm_sql_postgresql_lib_dir, --with-rlm_sql_postgresql_include_dir, --without-rlm_eap_ikev2, --without-rlm_sql_oracle, --without-rlm_sql_unixodbc
    configure: creating cache config.cache
    checking for gcc... gcc
    checking whether the C compiler works... no
    configure: error: in `/root/freeradius/src/modules/rlm_sqlhpwippool':
    configure: error: C compiler cannot create executables
    See `config.log' for more details
    make[1]: *** [src/modules/rlm_sqlhpwippool/all.mk] Error 77
    make[1]: Leaving directory `/root/freeradius'
    make: *** [build-arch-stamp] Error 2
    dpkg-buildpackage: error: debian/rules build gave error exit status 2

The obvious difference between ./configure && make deb and dpkg-buildpackage is that the former runs configure twice. And interestingly if you run the dpkg-buildpackage command again after it fails then it will successfully build the packages the second time around. So it looks like there may be some sort of dependency problem whereby to build the debian packages you have to run configure twice - it's just that the make deb way you have to do it (to get the Makefile).

This also seems to show that debian/rules clean isn't tidying up properly.

Still investigating...

alandekok:
On Sep 29, 2015, at 11:48 AM, Matthew Newton <notifications@github.com> wrote:

> OK, I've had time to so some more testing on this now. Rather than my previous build machine which could have all sorts of random packages installed, this is in a clean wheezy chroot.
> ...
> mv config.sub.dist config.sub
> /usr/bin/make
> make[1]: Entering directory `/root/freeradius'
> CONFIGURE src/modules/rlm_sqlhpwippool/

  Hmm... that's probably because something touched "configure" or "all.mk.in" in that directory.

  The solution is to not do that.  :(

  The problem is figuring out what piece of the build did that.

  A simple work-around is to do 'rm config.log" after running configure the first time.  That causes the top-level Makefile to *not* check for the above dependencies, and to *not* re-run configure.

> This also seems to show that debian/rules clean isn't tidying up properly.

  It should remove config.log.


  The key part of the Makefile is this:

#  If we've already run configure, then add rules which cause the
#  module-specific "all.mk" files to depend on the mk.in files, and on
#  the configure script.
#
ifneq "$(wildcard config.log)" ""
CONFIGURE_ARGS	   := $(shell head -10 config.log | grep '^  \$$' | sed 's/^....//;s:.*configure ::')

src/%all.mk: src/%all.mk.in src/%configure
	@echo CONFIGURE $(dir $@)
	@rm -f ./config.cache $(dir $<)/config.cache
	@cd $(dir $<) && ./configure $(CONFIGURE_ARGS)
endif


  Remove that rule, and it won't re-run configure.

mcnewton:
Utterly weird - it turns out that building with fakeroot (which is done by make deb, and sbuild) everything is fine. Without fakeroot get this problem.

Lack of fakeroot generally will cause permission problems in the final package, not make to call configure again, so it looks like fakeroot is somehow hiding another issue.

So the problem can be worked around by adding dh_testroot to the configure target of the Makefile - update patch for this.

The even more weird thing is that with dh_testroot in there, running without fakeroot now doesn't bomb out as I'd expect - it actually builds the packages.

Completely lost for words now... but new patch should safely fix the issue.

PR #1276, opened 2015-09-28
Exclude rlm_cache_redis.so from freeradius package

mans0954:
Exclude rlm_cache_redis.so from freeradius package to avoid conflict with freeradius-redis

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -169,7 +169,7 @@ install-arch: build-arch-stamp
 	done
 
 	dh_install --sourcedir=$(freeradius_dir) -p freeradius-utils
-	dh_install --sourcedir=$(freeradius_dir) -p freeradius
+	dh_install --sourcedir=$(freeradius_dir) -p freeradius -Xusr/lib/freeradius/rlm_cache_redis.so
 
 	dh_strip -a --dbg-package=freeradius-dbg
 


arr2036:
Could you explain why this is necessary? Ideally all the redis related modules would be installed together by a redis package, with libhiredis as a dependency.

mans0954:
The way the debian/rules file has been implemented, make install 'installs' the whole of FreeRADIUS under debian/tmp. A loop then operates over the module list and calls dh_install for each module, to copy the relevant files as specified by the debian/*.install files. In most cases, the only binary file for a module is rlm_$mod.so, and this is then deleted from debian/tmp by the for loop. However, in the case of redis, libfreeradius-redis*.so, rlm_rediswho*.so and rlm_cache_redis*.so are also installed to the freeradius-redis package and left behind in debian/tmp.

After this, usr/lib/freeradius/rlm*.so (amongst other things) is copied from debian/tmp into the freeradius package. Thus, as was, rlm_cache_redis*.so was getting copied into both the freeradius-redis and freeradius packages. Since two packages aren't allowed to install the same file[0], this creates a conflict and prevents freeradius-redis being installed with freeradius.

As I didn't do the original packaging, I may have mis-understood the author's intention, but the patch  is my understanding of how to resolve the problem. An alternative approach would be to explicitly list the files to be copied to  the freeradius package.

Christopher

[0] https://www.debian.org/doc/debian-policy/ch-files.html

arr2036:
OK, that's fine, but shouldn't there be similar statements for libfreeradius-redis and rlm_rediswho?  They all need to be installed by the same package, and excluded from the main FreeRADIUS package.

mans0954:
Looking at this more closely, the for loop actually deletes rlm_$$mod*.so - so this matches rlm_rediswho.so. libfreeradius-redis isn't matched by any of the lines in freeradius.install. Thus, installable packages should be built without needing to explicitly exclude either libfreeradius-redis or rlm_rediswho.

So, this is a bit convoluted, but I think it works.

arr2036:
But is rlm_rediswho.so explicitly included in the redis package?  I'm keen on making sure that all the redis components get included in the redis package...

PR #1283, opened 2015-09-30
refresh quilt patches

mcnewton:
so they apply cleanly without hunk offset


arr2036:
Can you do the same for v3.1.x too. Thanks!

mcnewton:
Will do. Trying to make sure things are nice and tidy as possible for 3.0.10...

PR #1285, opened 2015-10-01
debian packaging updates

mcnewton:
* build redis and memcached packages
* rlm_sql_iodbc is broken in jessie as isql.h moved - fixed
* add snmp as a suggests to freeradius pkg because the build suggests it
* add openssl as a depends on freeradius-config so it will install on its own
* move yubikey to separate package and build with ybclient

tested build, install and run successfully on wheezy, jessie and sid using sbuild.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -23,6 +23,8 @@ Build-Depends: debhelper (>= 9),
  libtalloc-dev,
  libwbclient-dev,
  libyubikey-dev,
+ libmemcached-dev,
+ libhiredis-dev,
  python-dev
 Section: net
 Priority: optional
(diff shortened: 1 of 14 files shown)


arr2036:
Please do equiv for v3.1.x too.  I'm merging the ones that cherry-pick across cleanly, but this one doesn't...

mcnewton:
Righto, I'll take a look.

Was wondering on whether the better way for 3.1.x might be to drop all the -ldap -postgres etc packags and just have freeradius-extras which contains All The Other Things. Strongswan has this in debian, and it's much easier rather than having one package per module.

Thoughts? If that's OK I'll work towards that instead on 3.1.x.

arr2036:
I'd rather go the other route and package everything with an external dependency. Each module has its own set of dependencies, so if we did have an extras package it'd pull in many many other libraries.

I doubt Strongswan, requires the range of libraries we do :(

mcnewton:
Maybe - it just means a lot of packages, and that will just grow. Exim is a better example really - exim4-daemon-light: small package not too many dependencies, exim4-daemon-heavy: full features, has a lot of dependencies.

Just thought it might make things a bit simpler on the package front. Most dependencies are just client libraries, so aren't really a problem having on a system (it's not like freeradius-redis actually installs redis, or freeradius-postgresql installs the full database).

Maybe rather than freeradius-extra, break it down into freeradius-database-modules (mysql, redis, etc), freeradius-proto (dhcp), and freeradius-extra-modules (anything else) to slightly reduce the number of dependencies per module.

Dunno, was just wondering. 17 packages seems like quite a lot. Happy to sort it either way for now.

arr2036:
I just don't really see the point.  If you're going to be using MySQL then you probably won't be using Redis.  The dependencies for the modules don't really change, so it's not massive amounts of work to maintain them... I was planning on doing the same thing for the .spec files when I got some free time.

qnet-herwin:
I have to agree with @arr2036 here: I'm not really fond of the idea of installing libraries for redis, memcache, yubikey, odbc and countless others, just because I require the postgres libs.

You could always simplify the installation by creating a virtual package (that may be called freeradius-extra) that just has dependencies on the several modules.

mcnewton:
Maybe a freeradius-extra (or -modules etc) metapackage might be the best idea, but there's no point really doing that unless it's in the deban repositories because they'll currently just have built debs from the source anyway, so can dpkg -i *.deb instead. No point adding yet another package when it's not needed!

Looks like it stays as is then. Though to note if you're worried about useless dependencies, the build warns of several being added that aren't currently needed anyway, so you've potentially got a lot of unnecessary dependencies installed already :-P

arr2036:
I know *sigh* Make.inc.in and configure.ac need some tidying.

PR #1286, opened 2015-10-01
Fixed build for systems without support for 128 bit integers

qnet-herwin:
The build still fails, but at least it's a step in the right direction.


arr2036:
Thanks!

PR #1290, opened 2015-10-01
revert recent dh_testroot change

mcnewton:
Successfully enforced fakeroot in normal build, but then annoyingly fails in
sbuild which calls each build target separately and doesn't use fakeroot for
the build target.


mcnewton:
Will have to hack on that a bit more to find the real underlying cause it seems. Or just ignore it and hope everyone just uses fakeroot as per the docs... :)

PR #1296, opened 2015-10-02
Compile 128bit

herwinw:
I haven't tested this in any way except for compiling, so a review of the changes wouldn't be a bad idea.


arr2036:
Looks good, thanks!

Issue #1298, opened 2015-10-04
Problems with Packet-Dst-Port and "proxy-request" list


jpereira:
Hi,

We have a problem with Packet-Dst-Port and "proxy-request" as below attempt.

```
Sun Oct  4 23:17:14 2015 : Debug: (0) Received CoA-Request Id 176 from 10.1.2.128:34084 to 192.168.56.90:3799 length 33
Sun Oct  4 23:17:14 2015 : Debug: (0)   Acct-Session-Id = "12345"
Sun Oct  4 23:17:14 2015 : Debug: (0)   Session-Timeout = 180
Sun Oct  4 23:17:14 2015 : Debug: (0) # Executing section recv-coa from file /etc/freeradius/sites-enabled/mcare-nas-test
Sun Oct  4 23:17:14 2015 : Debug: (0)   recv-coa {
Sun Oct  4 23:17:14 2015 : Debug: (0)     update control {
Sun Oct  4 23:17:14 2015 : Debug: (0)       &Home-Server-Pool := "coa_pool_wifilabs"
Sun Oct  4 23:17:14 2015 : Debug: (0)     } # update control = noop
Sun Oct  4 23:17:14 2015 : Debug: (0)   } # recv-coa = noop
Sun Oct  4 23:17:14 2015 : Debug: (0) server mcare-nas-radius-wifilabs {
Sun Oct  4 23:17:14 2015 : Debug: (0)   # Executing section pre-proxy from file /etc/freeradius/sites-enabled/mcare-nas-radius-wifilabs
Sun Oct  4 23:17:14 2015 : Debug: (0)     pre-proxy {
Sun Oct  4 23:17:14 2015 : Debug: (0)       update proxy-request {
Sun Oct  4 23:17:14 2015 : Debug: (0)         &Packet-Dst-Port := 1812
Sun Oct  4 23:17:14 2015 : Debug: (0)       } # update proxy-request = noop
Sun Oct  4 23:17:14 2015 : Debug: (0)     } # pre-proxy = noop
Sun Oct  4 23:17:14 2015 : Debug: (0) }
Sun Oct  4 23:17:14 2015 : Debug: (0) proxy: Trying to allocate ID (0/2)
Sun Oct  4 23:17:14 2015 : Debug: (0) proxy: request is now in proxy hash
Sun Oct  4 23:17:14 2015 : Debug: (0) proxy: allocating destination 10.11.1.2 port 3799 - Id 35
[...]

PR #1299, opened 2015-10-05
Fix problem with Packet-Dst-Port and proxy-request list #1298

jpereira:
Hi,

Fix the question about the Packet-Dst-Port in proxy-request list. below, some evidences about that.

The setting...
```
.......................
server mcare-nas-radius-wifilabs {
     pre-proxy {
          update proxy-request {
              &Packet-Dst-Port := 1812
          }
     }
}
....................
```

if 'Packet-Dst-Port' exist in proxy-request, we will have a new message.
```
Sun Oct  4 23:58:06 2015 : Debug: (0) proxy: Found 'Packet-Dst-Port = 1812' in &proxy-request
Sun Oct  4 23:58:06 2015 : Debug: (0) proxy: Replacing the destination port 3799 by 1812
```

```
Sun Oct  4 23:58:06 2015 : Debug: (0) proxy: Trying to allocate ID (0/2)
Sun Oct  4 23:58:06 2015 : Debug: (0) proxy: request is now in proxy hash
Sun Oct  4 23:58:06 2015 : Debug: (0) proxy: allocating destination 10.11.19.2 port 1812 - Id 252
Sun Oct  4 23:58:06 2015 : Debug: (0) Proxying request to home server 10.11.19.2 port 1812 timeout 30.000000
Sun Oct  4 23:58:06 2015 : Debug: (0) Sent CoA-Request Id 252 from 0.0.0.0:54946 to 10.11.19.2:1812 length 38
Sun Oct  4 23:58:06 2015 : Debug: (0)   Acct-Session-Id = "12345"
Sun Oct  4 23:58:06 2015 : Debug: (0)   Session-Timeout = 180
Sun Oct  4 23:58:06 2015 : Debug: (0)   Proxy-State = 0x313338
Sun Oct  4 23:58:06 2015 : Debug: Waking up in 0.3 seconds.
Sun Oct  4 23:58:07 2015 : Proxy: (0) Marking home server 10.11.19.2 port 1812 alive
Sun Oct  4 23:58:07 2015 : Debug: (0) Clearing existing &reply: attributes
Sun Oct  4 23:58:07 2015 : Debug: (0) Received CoA-ACK Id 252 from 10.11.19.2:1812 to 192.168.56.90:54946 length 80
Sun Oct  4 23:58:07 2015 : Debug: (0)   Reply-Message = "AAA->NOKIA() listen::type=coa pack-type=(CoA-Request)"
Sun Oct  4 23:58:07 2015 : Debug: (0)   Proxy-State = 0x313338
Sun Oct  4 23:58:07 2015 : Debug: (0) server mcare-nas-radius-wifilabs {
Sun Oct  4 23:58:07 2015 : Debug: (0)   Empty post-proxy section in virtual server "mcare-nas-radius-wifilabs".  Using default return values.
Sun Oct  4 23:58:07 2015 : Debug: (0) }
Sun Oct  4 23:58:07 2015 : Debug: (0) Empty send-coa section in virtual server "mcare-nas-test".  Using default return values.
Sun Oct  4 23:58:07 2015 : Debug: (0) Sent CoA-ACK Id 138 from 192.168.56.90:3799 to 10.1.2.128:40378 length 0
Sun Oct  4 23:58:07 2015 : Debug: (0)   Reply-Message = "AAA->NOKIA() listen::type=coa pack-type=(CoA-Request)"
Sun Oct  4 23:58:07 2015 : Debug: (0) Finished request
Sun Oct  4 23:58:07 2015 : Debug: Waking up in 4.9 seconds.
```

Works very well and will help a lot the integrations with other Radius vendors that is out of standards! :+1:

### DIFF
--- a/src/main/process.c
+++ b/src/main/process.c
@@ -2178,6 +2178,7 @@ static int insert_into_proxy_hash(REQUEST *request)
 	int tries;
 	bool success = false;
 	void *proxy_listener;
+	VALUE_PAIR *vp;
 
 	VERIFY_REQUEST(request);
 
@@ -2190,6 +2191,14 @@ static int insert_into_proxy_hash(REQUEST *request)
 	proxy_listener = NULL;
 	request->num_proxied_requests = 1;
 	request->num_proxied_responses = 0;
+
+	vp = fr_pair_find_by_num(request->proxy->vps, PW_PACKET_DST_PORT, 0, TAG_ANY);
+	if (vp) {
+		RDEBUG3("proxy: Found 'Packet-Dst-Port', Overriding the destination port %d by %d",
+						request->proxy->dst_port,
+						vp->vp_integer);
+		request->proxy->dst_port = vp->vp_integer;
+	}
 
 	for (tries = 0; tries < 2; tries++) {
 		rad_listen_t *this;
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -90,6 +90,8 @@ FreeRADIUS 3.0.10 Wed 08 Jul 2015 12:00:00 EDT urgency=medium
 	* Fix off by one error that caused some MSCHAP-Error messages to
 	  be sent without the password change version (V=3) and the textual
 	  message component (M=).
+	* Fix proxy to Packet-Dst-Port, so that it uses the correct
+	  destination port. Patch from Jorge Pereira
 
 FreeRADIUS 3.0.9 Wed 08 Jul 2015 12:00:00 EDT urgency=medium
 	Feature improvements


src/main/process.c
```diff
@@ -2191,6 +2192,14 @@ static int insert_into_proxy_hash(REQUEST *request)
 	request->num_proxied_requests = 1;
 	request->num_proxied_responses = 0;
 
+	vp = fr_pair_find_by_num(request->proxy->vps, PW_PACKET_DST_PORT, 0, TAG_ANY);
+	if (vp) {
+		RDEBUG3("proxy: Found 'Packet-Dst-Port = %d' in &proxy-request", vp->vp_integer);
+		RDEBUG3("proxy: Replacing the destionation port %d by %d", request->proxy->dst_port,
```
qnet-herwin: A small typo here: destionation should be destination

jpereira: haha! Sorry, tks @qnet-herwin

jpereira:
@qnet-herwin I changed the text, now looks better.

alandekok:
This is the wrong place to put the patch.  It completely bypasses all checks which verify that the home server exists.

The problem in 3.0 is that for listeners, `type = auth+acct` means "accept both packets on the lister port".  But for home servers, "type = auth+acct` means "create two home servers, one for each port".

So this can't be easily fixed for 3.0.

In 3.1, we should move to a new syntax:

    type = "auth,acct,coa"

Or more verbose, but more descriptive:

  allowed_packets = "Access-Request, Accounting-Request, CoA-Request, ..."

jpereira:
@alandekok I believe that if we have the use of Packet-Dst-Port in others places, can't be any problem to override the value.

we have anothers examples.

```
4081     } else if (!coa->home_server) {
4082         uint16_t port = PW_COA_UDP_PORT;
4083 
4084         vp = fr_pair_find_by_num(coa->proxy->vps, PW_PACKET_DST_PORT, 0, TAG_ANY);                                                
4085         if (vp) port = vp->vp_integer;
4086 
4087         coa->home_server = home_server_find(&ipaddr, port, IPPROTO_UDP);
4088         if (!coa->home_server) {
4089             RWDEBUG2("Unknown destination %s:%d for CoA request.",
4090                    inet_ntop(ipaddr.af, &ipaddr.ipaddr,
4091                      buffer, sizeof(buffer)), port);
4092             goto fail;
4093         }
4094     }
```

and
```
2881         vp = fr_pair_find_by_num(request->config, PW_PACKET_DST_PORT, 0, TAG_ANY);                                                
2882         if (!vp) {
2883             if (request->packet->code == PW_CODE_ACCESS_REQUEST) {
2884                 dst_port = PW_AUTH_UDP_PORT;
2885 
2886 #ifdef WITH_ACCOUNTING
2887             } else if (request->packet->code == PW_CODE_ACCOUNTING_REQUEST) {
2888                 dst_port = PW_ACCT_UDP_PORT;
2889 #endif
2890 
2891 #ifdef WITH_COA
2892             } else if ((request->packet->code == PW_CODE_COA_REQUEST) ||
2893                    (request->packet->code == PW_CODE_DISCONNECT_REQUEST)) {
2894                 dst_port = PW_COA_UDP_PORT;
2895 #endif
2896             } else { /* shouldn't happen for RADIUS... */
2897                 return 0;
2898             }
2899 
2900         } else {
2901             dst_port = vp->vp_integer;
2902         }
```

My suggestions is based in other that's already exist.

jpereira:
I can't figure out your point, because if we have other places that can allow the override the port. Can't see any problem to give the option to the user do what is necessary.

alandekok:
Those places are setting the port BEFORE it looks up the home server.  You changes are AFTER it looks up the home server.

With the current code, if you set Packet-Dst-Port = 12345, it will fail to find the home server, and refuse to send the packet anywhere.

With your patches, if you set Packet-Dst-Port = 12345, it will find the home server using the *old* port (say 1812), but then send it to the home server IP, but dst port 12345.  Which doesn't have a RADIUS server.

alandekok:
The real underlying problem is that you can't mark a home server as "auth + acct + coa".  If you could, you could just proxy the packet to the home server, and all of the existing logic would work without fail.

Making the existing logic more complex is bad.  It means there's more work-arounds in the code, which have to be tested and maintained.

fixing the underlying issue is preferable.  And it has no negative side-effects, like your patch does.  It's just fixing the parsing code, and changing nothing of the state machine

jpereira:
The problems is because are setting before look up. And if change the patch to setting the dst_port after! Its better?

```
update control {
    &Packet-Dst-Port := 1234
    &Home-Server-Pool := myhs
}
```

better? My point is because I can understand your point, but I believe that could give the option to the user do what is necessary. We can't be completely *hard*!

alandekok:
That isn't really any better.  The code has to set the port *before* it looks up the home server.  And there's already code to do that.

You'd be better off with a patch that changes request_will_proxy() to look for the home server *type* from a VALUE_PAIR, instead of from request->packet->code.  That way you can create one home server (`type = auth`).  And the for CoA packets say "I know it's coa, but I want to find a home server of type `auth`, not of type `coa`.

alandekok:
And in fact in v3.0.x, you can proxy CoA packets to any IP / port you want:

		update control {
		       Packet-Dst-IP-Address := 127.0.0.1
		       Packet-Dst-Port := 18120
		}

when that IP / port is a home_server of `type = auth`.  You just can't use a `home_server_pool`, or a `realm`.

alandekok:
similar code should work for originating coa requests, too.  I just checked that.  just set an IP/port, and it should work.

PR #1308, opened 2015-10-07
Correct SIGINT and SIGQUIT termination in debug mode

EllezDi:
If server starts without -m  option, then SIGINT and SIGQUIT signal handlers doesn't init.

The -m option for memory cleanup then correctly processed in sig_fatal function.

### DIFF
--- a/src/main/radiusd.c
+++ b/src/main/radiusd.c
@@ -476,15 +476,13 @@ int main(int argc, char *argv[])
 	 *  immediately.  Use SIGTERM to shut down the server cleanly in
 	 *  that case.
 	 */
-	if (main_config.debug_memory || (rad_debug_lvl == 0)) {
-		if ((fr_set_signal(SIGINT, sig_fatal) < 0)
+	if ((fr_set_signal(SIGINT, sig_fatal) < 0)
 #ifdef SIGQUIT
-		|| (fr_set_signal(SIGQUIT, sig_fatal) < 0)
+	|| (fr_set_signal(SIGQUIT, sig_fatal) < 0)
 #endif
-		) {
-			ERROR("%s", fr_strerror());
-			exit(EXIT_FAILURE);
-		}
+	) {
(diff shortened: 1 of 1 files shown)


alandekok:
Why is this "correct"?  The SIGINT and SIGQUIT handlers shouldn't be set when -m isn't used.

EllezDi:
On my debian 8.2 radiusd -XXX didn't terminate on Ctrl-C (terminates with radiusd -mXXX).
So, when I explicitly sets the fr_set_signal(SIGINT, sig_fatal); - it does.
My thoughts - this is because no one handles the SIGINT signal.
Maybe I'm wrong ?

alandekok:
If there is no signal handler for CTRL-C, then the process should quit.  It works fine on my system when using `radiusd -X` and CTRL-C.

EllezDi:
I can't believe. Magic! After your post it works !! (recompiled distro 2 times). =) Thanks =)

arr2036:
I've noticed some modules like rlm_ruby, register their own signal handlers which can cause this to happen.

EllezDi:
I found it. When rlm_sql_oracle loads - it does not terminates on SIGINT. Only with SIGQUIT. 
Without rlm_sql - works fine for both signals.

arr2036:
We should probably try to remove any registered signal (that we didn't add) handlers after module instantiation

alandekok:
While I thing it's stupid for libraries to register signal handlers, I'm not sure it's worth adding more code to the server to work around stupid crap.

arr2036:
@EllezDi can you confirm this fixes the issue? Then i'll merge to v3.0.x?

EllezDi:
@arr2036  on 3.0.10 - your patch didn't fixes the issue. Oracle client 12.1 x64 (db 10 or 11)
export BEQUEATH_DETACH=YES - still didn't work
As I see in docs, bequeathed connections are all local. So for Net8 it may not work.

PR #1333, opened 2015-10-20
When cannot encode, please say the attr name

jpereira:


alandekok:
Fixed via commit a9d44129.  The `vp->da->name` is always available, so there's no need to check it.

PR #1340, opened 2015-10-23
Cross-build fixes

alllexx88:
Some fixes for better cross-build support

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1246,6 +1246,9 @@ if test "x$WITH_OPENSSL" = xyes; then
       [
         AC_MSG_RESULT(no)
         AC_MSG_FAILURE([OpenSSL library version does not match header version])
+      ],
+      [
+        AC_MSG_RESULT([cross-compiling (assuming yes)])
       ]
     )
     CPPFLAGS="$old_CPPFLAGS"
--- a/scripts/boiler.mk
+++ b/scripts/boiler.mk
@@ -128,7 +128,7 @@ endef
 #
 ifeq "${CPP_MAKEDEPEND}" "yes"
 define ADD_OBJECT_RULE
-$${BUILD_DIR}/objs/%.${OBJ_EXT} $${BUILD_DIR}/objs/%.d: ${1} ${JLIBTOOL}
+$${BUILD_DIR}/objs/%.${OBJ_EXT} $${BUILD_DIR}/objs/%.d: ${1} headers ${JLIBTOOL}
 	${2}
 	$${CPP} $${CPPFLAGS} $$(addprefix -I,$${SRC_INCDIRS}) $${SRC_DEFS} $$< | sed \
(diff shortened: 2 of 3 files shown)


alanbuxey:
what is this to achieve?  building FreeRADIUS against a different OpenSSL
version to that against which is will run on a real server is a VERY bad
idea
as the OpenSSL API changes all over the place...even on same release track!

alllexx88:
@alanbuxey 

    what is this to achieve?  building FreeRADIUS against a different OpenSSL
    version to that against which is will run on a real server is a VERY bad
    idea
    as the OpenSSL API changes all over the place...even on same release track!

Perhaps, you are getting the idea of cross-compilation wrong. Say, you want to run a freeradius server on your ARM router, which is powerful enough to run it, but not very fit to build it. The solution would be to cross-compile on an i686/amd64 PC, since it is *tons* faster. Normally, you use the same dependencies versions you compiled freeradius against on the target device. In my case, I cross-compile freeradius as part of Optware-ng project to provide feeds for different archs. I don't update OpenSSL often, and I can always rebuild freeradius against updated version if issues are reported by users.

alandekok:
I've pushed some changes which should help avoid the build dependency on jlibtool.  Please check that.

But I still think that the other patch isn't necessary.  If the OpenSSL consistency check fails, it's because the build environment is wrong.

alllexx88:
@alandekok OpenSSL consistency check *always* fails during a cross-build. Even with a perfectly sane build environment, smiply because AC_RUN_IFELSE is used, which tries to run a compiled binary, which is impossible during a cross build. Thus I do still think the other patch is needed

arr2036:
The test binary needs to be compiled for the build system not the target system.  If you can fix that then we'd be happy to merge that change.

alandekok:
There is no other way to check that the OpenSSL version and library are consistent.  We *must* build a binary, and we *must* run it.

You can work around the checks with a `configure` flag, as you noticed.

Adding the patch means that the version checks will *always* be ignored on *all* platforms.  That is not acceptable.

alllexx88:
@alandekok You are actually wrong. Se [here](https://www.gnu.org/software/autoconf/manual/autoconf-2.63/html_node/Runtime.html) for details on AC_RUN_IFELSE. The third parameter I've added is *only* used when cross-compiling, so it will have no effect on *normal* native builds: the check on *all* native platforms will be performed.

alllexx88:
@arr2036 It doesn't make sense to build test binary for the build system, since it will use *build* machine OpenSSL library, it just can't be linked with a library of different architecture (target one), and hence the test result will be misleading. The only workaround I see would be to use some sort of qemu emulation, but I believe it is an overkill, and that a person who goes cross-compiling should also make sure that correct staged headers and libraries are used anyway, otherwise SEGFAULTs are inevitable.

PR #1341, opened 2015-10-23
Fixed usage of Post-Auth-Type REJECT

qnet-herwin:
The values had been shifted in the dictionaries in commit d9a9d594e666d70b9762c879c2c69d9a9bb38a64, but this piece of code used a hardcoded magic value instead of the autogenerated constant from include/attributes.h. Fix this by using the constant.


arr2036:
Thanks!

PR #1342, opened 2015-10-23
added Meraki dictionary file

alanbuxey:


alanbuxey:
added meraki dictionary (and listed it in the main dictionary file)

PR #1345, opened 2015-10-23
trust, magic values aren't fun

jpereira:


arr2036:
Hm where are those defined? Should be MACRO STYLE LIKE THIS

jpereira:
my bad.... wrong regex! just 1 sec..

Issue #1339, opened 2015-10-23
Some fields isn't updated in 'radmin> stats home_server acct'


jpereira:
Hi,

   I noticed that always when i receive the message

```
Fri Oct 23 05:25:42 2015 : ERROR: (463) ERROR: Failing proxied request for user "ec:88:92:86:06:12", due to lack of any response from home server 10.1.2.4 port 1
```

The field 'invalid or malformed' isn't updated in "radmin stats".

```
# radmin
radmin> stats home_server acct
requests        865
responses       857
dup             0
invalid         0
malformed       0
bad_authenticator       0
dropped         0
unknown_types   0
timeouts        0
last_packet     1445578085
elapsed.1us     0
elapsed.10us    0
elapsed.100us   0
elapsed.1ms     853
elapsed.10ms    62
elapsed.100ms   1
elapsed.1s      0
elapsed.10s     0
#
```

arr2036:
Why would you expect it to be?

jpereira:
I noticed that don'`t exist this code... Only to: 

```
src/main/listen.c:	FR_STATS_TYPE_INC(client->auth.total_requests);
src/main/listen.c:	FR_STATS_TYPE_INC(client->auth.total_requests);
src/main/listen.c:	FR_STATS_TYPE_INC(client->acct.total_requests);
.......
```

jpereira:
@arr2036 It is helpful to collect metrics about "who" home_server has X invalid/... packets.

jpereira:
@arr2036 It is helpful to collect metrics about "who" home_server has X invalid/... packets. And if something wrong happens... Like timeout/invalid/... during the Proxy over home_server.... I think that is necessary to increase some field.

Or we can add a new field like "failed". What do you think?

jpereira:
In my example you can notice that has a gap between requests and response.

```
requests        865
responses       857
dup             0
invalid         0
malformed       0
bad_authenticator       0
dropped         0
unknown_types   0
timeouts        0
```

After check the code, I noticed that could be better to add a new field "failed".

alandekok:
> On Oct 23, 2015, at 1:30 AM, Jorge Pereira <notifications@github.com> wrote:
> 
> Hi,
> 
> I noticed that always when i receive the message
> 
[...]

PR #1346, opened 2015-10-23
Bugfix - increase a wrong list in stats #1339

jpereira:


jpereira:
@arr2036 when possible, bring to v3.1.x

PR #1347, opened 2015-10-23
trust, magic values aren't fun

jpereira:


alandekok:
It's better to manage the `#define` by hand.  I've pushed a fix for that.

Please update, remove the defines, and I can pull in the rest of the patch.

jpereira:
@alandekok  done!

PR #1348, opened 2015-10-23
Add src/include/freeradius.h to .gitignore

jpereira:


jpereira:
@arr2036 I have added in wrong .gitignore. now is correct.

PR #1349, opened 2015-10-25
magic values aren't fun

jpereira:


jpereira:
@alandekok maybe is better to cherry-pick in v3.1.x

PR #1361, opened 2015-10-29
It's better to keep the option to show the big packets.

jpereira:


arr2036:
It's really, really not.  The only thing this message is useful for is showing if you've screwed up calculating the link layer header size (this is sometimes complex due to VLANS and QinQ).

Lots of things seem to put trailing junk after the UDP packet data, there's probably a line in an RFC somewhere that says adding data after the UDP packet body is allowed and shouldn't be processed.

Also that patch would cause the process to ignore packets which trailing garbage, if the debug level was high, which is not what we want.

PR #1367, opened 2015-11-03
remove redundant "name" config option

mcnewton:
note also that "name" has vanished from the config in 3.1.x, but isn't marked deprecated in the code, so not sure if the config needs re-adding or the code updating (in which case the config should list it as deprecated as with the other options).


arr2036:
Yes, there should be a config item for that added

mcnewton:
PR incoming if you've not got there already :)

PR #1377, opened 2015-11-10
Mix virtual server into session-state rbtree index key

skids:
This gives each virtual server its own session-state even
if a packet with the same State attribute ends up being handled
by more than one virtual server in the same process.

### DIFF
--- a/src/main/state.c
+++ b/src/main/state.c
@@ -184,7 +184,7 @@ void fr_state_delete(fr_state_t *state)
 /*
  *	Create a new entry.  Called with the mutex held.
  */
-static state_entry_t *fr_state_create(fr_state_t *state, RADIUS_PACKET *packet, state_entry_t *old)
+static state_entry_t *fr_state_create(fr_state_t *state, const char *server, RADIUS_PACKET *packet, state_entry_t *old)
 {
 	size_t i;
 	uint32_t x;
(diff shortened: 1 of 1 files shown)


skids:
The preveious and current systems both do some pre-emptive deletion of state.
Without this, any setup that hairpins through a third party server (acting both
as a front-end and back-end for it) loses state spuriously.  Unless there is a use
case for different virtual servers having the same session-state in this situation,
this is the cleanest solution I could think of.

arr2036:
I guess that works.  Just considering the alternatives... Proxy-State won't work as it's opaque, and servers aren't required to add or forward Proxy-State attributes,  client won't work because of proxy-failover.  I guess you could see each virtual server as owning the session-state, in which case, yes it'd be good to qualify state values with virtual server information.  I'll let Alan decide whether to merge this in v3.0.x, but my preference for v.3.1.x would be to maintain per virtual server trees.

skids:
WRT the hash I could see that being necessary if people might try to pass their own string generated by other means to the internal function, or if multiple copies of the string wth different pointers can end up in request->server, but since the pointers are guaranteed to be unique, so their XORs against State are guaranteed to be unique for any given State, which is why I used the pointer.

alandekok:
You should not use the pointer.  There is no guarantee that the same servers string "foo" is always pointed to from the same pointer.  You can have

    char const *x = "foo";
    char const *y = "foo"

Where you have `x != y`.  You MUST use the contents of the string in the hash.

skids:
I'll read that as saying there are ways for request->server to pick up a different copy of that string.  Before I code up the hash-of-server-name-string, are there any other immutable unique IDs possessed by virtual servers accessible through a request object?  That might be a more efficient option.

alandekok:
There are no other unique IDs possessed by a virtual server.

arr2036:
Could you develop an equivalent patch for v3.1.x? Else this functionality will be lost.

skids:
I'm kinda in a bad situation both time-wise and test-environment-wise at the moment, but I'll try to remember to do it when after we've rolled 3.0 out, probably in a few months from now.

PR #1379, opened 2015-11-10
Fixing wrong set with $raddb/sites-available/abfab-tr-idp

jpereira:
please replicate in v3.1.x


jpereira:
please replicate in v3.1.x

PR #1381, opened 2015-11-10
Fixing potential invalid free()

jpereira:


arr2036:
As a convention, functions should set output pointers to NULL even if they error out as a protection against this, so I don't think the patch is required.

PR #1392, opened 2015-11-17
fixup configure script

mcnewton:
didn't rerun autoconf for this... The GNU-Force isn't with me today.

### DIFF
--- a/configure
+++ b/configure
@@ -9980,7 +9980,7 @@ for ac_func in \
   strsignal \
   unlinkat \
   bindat \
-  dladdr
+  dladdr \
   vdprintf \
   vsnprintf
 
--- a/configure.ac
+++ b/configure.ac
@@ -1476,7 +1476,7 @@ AC_CHECK_FUNCS( \
   strsignal \
   unlinkat \
   bindat \
-  dladdr
+  dladdr \
   vdprintf \
   vsnprintf
 )


arr2036:
They were actually dups caused by a merge conflict, so I removed them.  Thanks pointing out the issue anyway.

mcnewton:
OK cool. Doing a final build before pushing fr3 live this afternoon (yay) and it failed ;-)

arr2036:
Yaaaaaay!  So you will be giving that talk next year ;)?

mcnewton:
hah! This isn't for eduroam, though - it's for managed devices. But it's a start :)

I've got at least three talks on different topics I could easily do (and probably a couple more on FR as well), but can't plan 6 months ahead to submit them. So they never get done. Paper deadline in October for conference at Easter is crazy IMO. I don't even know if I can go until the new year.

arr2036:
I think @alanbuxey was semiserious about an upgrade talk, and seeing is it's after October...

We need people with success stories else no one is going to upgrade :(

mcnewton:
Paper deadline was 9 October, not submit them after October :(. But maybe they need more talks. Problem is I actually need to upgrade the eduroam servers beforehand, really.

Success so far is a segfault... somewhere in eap or ocsp. coredumps enabled and waiting for it to happen again. But aside from that it's working. :)

arr2036:
oops :(

PR #1394, opened 2015-11-24
Fixing problem with radmin> stats detail <filename>

jpereira:
Current problem.
```
Tue Nov 24 00:50:53 2015 : Debug:  ... new connection request on command socket
Tue Nov 24 00:50:53 2015 : Debug: Listening on command file /opt/radius/var/run/freeradius/freeradius.sock
Tue Nov 24 00:50:53 2015 : Info: Ready to process requests
Tue Nov 24 00:50:57 2015 : Debug: radmin> stats detail /somewhere/detail
Segmentation fault (core dumped)
```

After patch
```
radmin> stats detail /somewhere/detail
ERROR: No detail file listener
radmin> stats detail /var/log/freeradius/radacct/buffered/detail
state	no-reply
packets	1
tries	15
offset	213
size	852
radmin> 
```


jpereira:
@arr2036 This is necessary to ask the FR about which line is being read.

jpereira:
@alandekok could check this?

PR #1403, opened 2015-11-26
The field 'last_packet' in 'radmin>stats client' now is shown like a date

jpereira:
Currently the field 'last_packet' is shown like a any number.

```
radmin> stats client acct
requests	116
responses	116
dup		0
invalid		0
malformed	0
bad_authenticator	0
dropped		0
unknown_types	0
last_packet	1448563105
elapsed.1us	0
elapsed.10us	0
elapsed.100us	102
elapsed.1ms	10
elapsed.10ms	3
elapsed.100ms	1
elapsed.1s	0
elapsed.10s	0
radmin> 
```

The value shown like a date is completely more helpful than a simple number!

```
radmin> stats client acct
requests	116
responses	116
dup		0
invalid		0
malformed	0
bad_authenticator	0
dropped		0
unknown_types	0
last_packet	1448563105	(2015-11-26 16:38:25)
elapsed.1us	0
elapsed.10us	0
elapsed.100us	102
elapsed.1ms	10
elapsed.10ms	3
elapsed.100ms	1
elapsed.1s	0
elapsed.10s	0
radmin> 
```

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -13,6 +13,8 @@ FreeRADIUS 3.0.11 Mon 05 Oct 2015 15:00:00 EDT urgency=medium
 	* Portability fixes for Solaris.
 	* More errors from ntlm_auth gets passed to MS-CHAP.
 	* Update abfab-tr-idp virtual server.
+	* The field 'last_packet' in 'radmin> stats client' now is
+	  shown like a date.
 
 	Bug fixes
 	* Fix issue where field nas_type would not be accessible via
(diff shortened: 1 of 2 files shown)


arr2036:
No, you'll break integration with things that consume radmin output like munin

alanbuxey:
such tools/scripts need to be updated (they are probably broken anyway if
they arent programmed to ingest data specifically.

might I suggest they use XML format?  ;-)

arr2036:
Yes it needs to use an interface not designed for humans.  We can alter the radmin output once we've migrated to SNMP based stats queries.

PR #1411, opened 2015-12-01
Fixing wrong sizeof() in rlm_sql/rlm_sql_print_error()

jpereira:


arr2036:
So explain why the original code was wrong?

jpereira:
Wow!! I saw wrong, my bad!

PR #1414, opened 2015-12-03
Fixing missing headers

jpereira:


jpereira:
@arr2036 / @alandekok, I fix the issues related to missing headers & include-guards.

jpereira:
@arr2036 / @alandekok, I fixed the issues related to missing headers & include-guards. Now the headers will be installed.

jpereira:
Some evidences.

```
[jpereira@jpereira-desktop sample]$ cat -n radius-sample1.c
     1  #include <freeradius-devel/libradius.h>
     2
     3  static char     const *radius_dir = RADDBDIR;
     4  static char     const *dict_dir = DICTDIR;
     5
     6  int main() {
     7          printf("Using libfreeradius-radius.so\n");
     8
     9          if (fr_check_lib_magic(RADIUSD_MAGIC_NUMBER) < 0) {
    10                  fr_perror("radclient");
    11                  return 1;
    12          }
    13
    14          if (dict_init(dict_dir, RADIUS_DICTIONARY) < 0) {
    15                  fr_perror("radclient");
    16                  return 1;
    17          }
    18
    19          if (dict_read(radius_dir, RADIUS_DICTIONARY) == -1) {
    20                  fr_perror("radclient");
    21                  return 1;
    22          }
    23          printf("done\n");
    24          return 0;
    25  }
    26
cc -I/opt/radius/include -DHAVE_PTHREAD_H -o radius-sample1 radius-sample1.c -L/opt/radius/lib -Wl,-R/opt/radius/lib -L/opt/radius/lib/freeradius -Wl,-R/opt/radius/lib/freeradius -lfreeradius-radius
```

arr2036:
Not merging into v3.0.x

arr2036:
If I see any non bug fix pull requests against v3.0.x, unless it's really simple changes i'm going to close it

jpereira:
The latest version of FreeRADIUS is the v3.0.x, the v3.1.x is unstable. right?

Currently, It's impossible to use the freeradius sdk, we have many missing headers during install and another issues about wrong includes.

are you sure that isn't a bug?

arr2036:
It's not an SDK, that's the point.  You want to develop against libfreeradius-radius, it needs to be fixed up so that it's independent from the FreeRADIUS build system as time of install.

There is absolute ZERO value in fixing this in v3.0.x,  you may introduce weird build system issues, and most of the functions have been renamed for v3.1.x.  v3.1.x is the first release where we will have an official, supported, and semi-stable API.

@alandekok is going through now and cleaning up v3.1.x and organising the includes and source files to allow this.

PR #1415, opened 2015-12-03
Fixed the capability to use libfreeradius-radius with distributed headers

jpereira:
Using the below sample

```
[jpereira@jpereira-desktop sample]$ cat -n radius-sample1.c
     1  #include <freeradius-devel/libradius.h>
     2
     3  static char     const *radius_dir = RADDBDIR;
     4  static char     const *dict_dir = DICTDIR;
     5  fr_dict_t       *dict = NULL;
     6
     7  int main() {
     8          printf("Using libfreeradius-radius.so %s\n", RADIUSD_VERSION_STRING);
     9
    10          if (fr_check_lib_magic(RADIUSD_MAGIC_NUMBER) < 0) {
    11                  fr_perror("radclient");
    12                  return 1;
    13          }
    14
    15          if (fr_dict_init(NULL, &dict, dict_dir, RADIUS_DICTIONARY, "radius") < 0) {
    16                  fr_perror("radclient");
    17                  return 1;
    18          }
    19
    20          if (fr_dict_read(dict, radius_dir, RADIUS_DICTIONARY) == -1) {
    21                  fr_perror("radclient");
    22                  return 1;
    23          }
    24          printf("done\n");
    25          return 0;
    26  }
    27
cc -I/opt/radius-3.1.x/include -DHAVE_PTHREAD_H -o radius-sample1 radius-sample1.c -L/opt/radius-3.1.x/lib -Wl,-R/opt/radius-3.1.x/lib -L/opt/radius-3.1.x/lib/freeradius -Wl,-R/opt/radius-3.1.x/lib/freeradius -lfreeradius-radius
[jpereira@jpereira-desktop sample]$ ./radius-sample1
Using libfreeradius-radius.so 3.1.0
done
```

p.s: Was necessary to create apply a "duck-tape"

```
# cd /opt/radius-3.1.x/include
# ln -fs freeradius freeradius-devel
```

Because whole os codes have references to

```
#include <freeradius-devel/name.h>
```

### DIFF
--- a/src/include/all.mk
+++ b/src/include/all.mk
@@ -90,9 +90,11 @@ src/include/%.h: share/dictionary.% share/dictionary.vqp
 #
 src/include/features.h: src/include/features-h src/include/autoconf.h
 	@$(ECHO) HEADER $@
-	@cp $< $@
+	@echo "#ifndef FR_FEATURES_H_ /* auto generated */" > $@
+	@cat $< >> $@
 	@grep "^#define[ ]*WITH_" src/include/autoconf.h >> $@
 	@grep "^#define[ ]*RADIUSD_VERSION" src/include/autoconf.h >> $@
+	@echo "#endif /*FR_FEATURES_H_ */" >> $@
 
 #
 #  Use the SED script we built earlier to make permanent substitutions
--- a/src/include/all.mk
+++ b/src/include/all.mk
@@ -42,6 +42,13 @@ HEADERS	= \
 	base64.h \
 	map.h \
 	udp.h \
+	tcp.h \
+	threads.h \
+	regex.h \
+	inet.h \
+	dict.h \
+	pair.h \
+	proto.h \
 	$(HEADERS_DY)
 
 #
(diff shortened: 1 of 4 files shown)


alandekok:
Your code should just include <freeradius/foo.h>. 

  The headers in the build tree include <freeradius-devel/foo.h>. 

  When the headers are installed, the "-devel" text is removed before the header is installed. 

If an installed header uses -devel, it needs to be fixed.

Otherwise, fix your application to use the installed headers instead of -devel

arr2036:
Where in the build system is this meant to be happening?

jpereira:
@alandekok 

```
[jpereira@jpereira-desktop sample]$ grep ".*devel" -r /opt/radius-3.1.x/include/
Binary file /opt/radius-3.1.x/include/.git/index matches
/opt/radius-3.1.x/include/freeradius/libradius.h:#  include <freeradius-devel/tcp.h>
/opt/radius-3.1.x/include/freeradius/radiusd.h:#  include <freeradius-devel/tls.h>
```

I fixed the sed/regex in https://github.com/FreeRADIUS/freeradius-server/pull/1421

PR #1418, opened 2015-12-03
Souldn't include the autoconf.h

jpereira:


arr2036:
So you chose to ignore the comment?

jpereira:
We can share during installation?

arr2036:
No, you can't, because your linking program might also be using autoconf, and include its autoconf header file, and then it'll throw a bunch of macro redefinition errors.  That is why we went to the effort of copying values from the autoconf headers into the installed headers.

jpereira:
so, the best way is to define the ```-DLITTLE_ENDIAN``` or ```-DBIG_ENDIAN``` in configure.ac (CFLAGS/CPPFLAGS) instead of autoconf.h . right?

arr2036:
Do you understand how build.h is used?

jpereira:
yes, the only thing the build.h use from autoconf.h is

```
124 /*
125  *  Try and determine endianness of the target system.
126  *
127  *  Other projects seem to use endian.h and variants, but these are
128  *  in non standard locations, and may mess up cross compiling.
129  *
130  *  Here at least the endianess can be set explicitly with
131  *  -DLITTLE_ENDIAN or -DBIG_ENDIAN.
132  */
133 #if !defined(FR_LITTLE_ENDIAN) && !defined(FR_BIG_ENDIAN)
134 #  if defined(__LITTLE_ENDIAN__) || \
135       (defined(__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
136 #    define FR_LITTLE_ENDIAN 1
137 #  elif defined(__BIG_ENDIAN__) || \
138       (defined(__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_BIG_ENDIAN__))
139 #    define FR_BIG_ENDIAN 1
140 #  else
141 #    error Failed determining endianness of system
142 #  endif
143 #endif
```

And the configs is setted in configure.ac

```
dnl #
dnl #  check for system bytesex
dnl #  AC_DEFINES WORDS_BIGENDIAN
dnl #
AC_C_BIGENDIAN(
    [AC_DEFINE(FR_BIG_ENDIAN, 1, [Define if your processor stores words with the most significant byte first])],
    [AC_DEFINE(FR_LITTLE_ENDIAN, 1, [Define if your processor stores words with the least significant byte first])]  
)
```

And The idea would be to append in CFLAGS

```
dnl #
dnl #  check for system bytesex
dnl #  AC_DEFINES WORDS_BIGENDIAN
dnl #
AC_C_BIGENDIAN(
        CFLAGS="${CFLAGS} -DFR_BIG_ENDIAN=1,
        CFLAGS="${CFLAGS} -DFR_LITTLE_ENDIAN=1
)
```

arr2036:
and where is autoconf.h included?

jpereira:
What do you think about this proposal? https://github.com/FreeRADIUS/freeradius-server/compare/v3.1.x...jpereira:fix/build1?expand=1

arr2036:
No, absolutely not, we are not installing autoconf.h and I told you why just a few comments above.

arr2036:
If you need something from autoconf.h it needs to be copied over to another header file.

arr2036:
build.h is not included in any public headers, it's passed to the compiler by the build system.

arr2036:
I'm not even sure what problem you're trying to solve... maybe if you told us that we could come up with solutions...

jpereira:
Simple: Try to build a simple code using the current headers and libfreeradius-radius.so

arr2036:
I know what you're trying to do.  You've not said why you can't do it.

jpereira:
The current code
```C
#include <freeradius/libradius.h>
#include <freeradius/radpaths.h>
#include <freeradius/conf.h>

static char     const *radius_dir = RADDBDIR;
static char     const *dict_dir = DICTDIR;
fr_dict_t       *dict = NULL;

int main() {
        printf("Using libfreeradius-radius.so %s\n", RADIUSD_VERSION_STRING);

        if (fr_check_lib_magic(RADIUSD_MAGIC_NUMBER) < 0) {
                fr_perror("radclient");
                return 1;
        }

        if (fr_dict_init(NULL, &dict, dict_dir, RADIUS_DICTIONARY, "radius") < 0) {
                fr_perror("radclient");
                return 1;
        }

        if (fr_dict_read(dict, radius_dir, RADIUS_DICTIONARY) == -1) {
                fr_perror("radclient");
                return 1;
        }
        printf("done\n");
        return 0;
}
```

try to build

```
# cc -I/opt/radius-3.1.x/include -DHAVE_PTHREAD_H -o radius-sample1 radius-sample1.c -L/opt/radius-3.1.x/lib -Wl,-R/opt/radius-3.1.x/lib -L/opt/radius-3.1.x/lib/freeradius -Wl,-R/opt/radius-3.1.x/lib/freeradius -lfreeradius-radius
```

you can't! 

1) Missing autoconf.h (this discussion)
2) Need to fix the order of macro vs include #1422 
3) Add missing include-guard #1419
4) Could be better to add missing headers in libradius.h #1417

arr2036:
Why can't you, what are the errors?

jpereira:
this is the current problem related to #1422 

```
[jpereira@jpereira-desktop sample]$ make 2>&1 | head -10
cc -I/opt/radius-3.1.x/include -DHAVE_PTHREAD_H -o radius-sample1 radius-sample1.c -L/opt/radius-3.1.x/lib -Wl,-R/opt/radius-3.1.x/lib -L/opt/radius-3.1.x/lib/freeradius -Wl,-R/opt/radius-3.1.x/lib/freeradius -lfreeradius-radius 
In file included from radius-sample1.c:1:0:
/opt/radius-3.1.x/include/freeradius/libradius.h:22:21: error: expected ‘)’ before string constant
 RCSIDH(libradius_h, "$Id: d356b97f983fba4421176d6b637bdd907ad36279 $")
                     ^
In file included from /usr/include/talloc.h:28:0,
                 from /opt/radius-3.1.x/include/freeradius/libradius.h:60,
                 from radius-sample1.c:1:
/usr/include/stdlib.h:139:8: error: unknown type name ‘size_t’
 extern size_t __ctype_get_mb_cur_max (void) __THROW __wur;

```

jpereira:
after fix that with #1422. the error is 

```
cc -I/opt/radius-3.1.x/include -DHAVE_PTHREAD_H -o radius-sample1 radius-sample1.c -L/opt/radius-3.1.x/lib -Wl,-R/opt/radius-3.1.x/lib -L/opt/radius-3.1.x/lib/freeradius -Wl,-R/opt/radius-3.1.x/lib/freeradius -lfreeradius-radius 
In file included from /opt/radius-3.1.x/include/freeradius/libradius.h:22:0,
                 from radius-sample1.c:1:
/opt/radius-3.1.x/include/freeradius/build.h:13:64: fatal error: freeradius/autoconf.h: No such file or directory
compilation terminated.
Makefile:6: recipe for target 'all' failed
make: *** [all] Error 1
```

That for solve this I have a suggestion like used in libcurl: https://github.com/FreeRADIUS/freeradius-server/compare/v3.1.x...jpereira:fix/build1?expand=1

arr2036:
OK, so build.h needs to not be included in libradius.h and any substitutions or macros that dependended on it should be fixed, i'll take a look.

jpereira:
isn't only this. When possible, take a look in another fixes. (#1417,#1419 and #1422)

arr2036:
Fixed some issues, please re-evaluate and determine what else needs fixing

PR #1420, opened 2015-12-03
Adding include-guards to src/include/features.h

jpereira:


alandekok:
This is better done by updating the original `features-h` file

jpereira:
isn't better, because is necessary to append this

```
	@grep "^#define[ ]*WITH_" src/include/autoconf.h >> $@
 	@grep "^#define[ ]*RADIUSD_VERSION" src/include/autoconf.h >> $@
```

arr2036:
Yeah the lines being concatenated on at the end get added after the 'endif'

arr2036:
Original patch missed #define which made the guard useless.  Macro name inconsistent with other macro names.

arr2036:
All guard macros should really be in the form ``_FR_<header name>_H``.  They need a prefix so they don't conflict with other libraries, and the first underscore marks them as 'private' ish.  It doesn't really do anything, but it lets you know they're not really useful for specifying values.

Feel free to update the headers.

PR #1421, opened 2015-12-03
Fixed correct strip of #include <freeradius-devel

jpereira:
The current regex doesn't works with spaces.

```
[jpereira@jpereira-desktop sample]$ grep ".*devel" -r /opt/radius-3.1.x/include/
/opt/radius-3.1.x/include/freeradius/libradius.h:#  include <freeradius-devel/tcp.h>
/opt/radius-3.1.x/include/freeradius/radiusd.h:#  include <freeradius-devel/tls.h>
```

Keep some headers with -devel


jpereira:
@arr2036 updated with your suggestion.

arr2036:
Commited

PR #1426, opened 2015-12-07
Allow verify cb to be skipped if ocsp response was ok

pwdng:


alandekok:
I've pushed a minimal patch without the whitespace changes.

PR #1430, opened 2015-12-07
Fixed memory leak with fr_dict_init()

jpereira:
dict->value_fixup was allocated with malloc(), so need to be released


arr2036:
Lets just convert it to talloc :)

PR #1437, opened 2015-12-08
Fix - added missing Access-Challenge in realm/pool set

jpereira:


alandekok:
The server never receives Access-Challenge messages.  It only sends them.

PR #1440, opened 2015-12-09
Replaced array function parameters with pointers

qnet-herwin:
This was already done implicitely, but writing it as an array suggests that it is passed by value instead, and that you can call things like `sizeof` on it. Some explanation and arguments can be found at https://lkml.org/lkml/2015/9/3/428 (although you have to read between the rants, as usual with texts of Linus).

I've only tested it with a local users backend, not with the winbind-client (I don't have a build system with samba 4.2 at hand)


arr2036:
I disagree, it's useful to hint at the size of the array in the signature.

PR #1441, opened 2015-12-10
Fix build failure when --disable-openssl-version-check is set.

TheMysteriousX:
Minor fix, as mentioned on the devel list.

4f24d4c mostly corrected the behaviour, however mainconfig.allow_vulnerable_ssl still had a dependency on ENABLE_OPENSSL_VERSION_CHECK.

### DIFF
--- a/src/include/radiusd.h
+++ b/src/include/radiusd.h
@@ -360,7 +360,7 @@ typedef struct main_config_t {
 	int		proxy_requests;
 	int		reject_delay;
 	int		status_server;
-#ifdef ENABLE_OPENSSL_VERSION_CHECK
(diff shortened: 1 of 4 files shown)


arr2036:
Could you change it so that it checks both macros,  thanks!

TheMysteriousX:
Checking both macros in mainconfig.c/radiusd.h would make the build still fail (as it's the resultant undefined reference to mainconfig.allow_vulnerable_ssl in radiusd.c that fails), so I've handled it by splitting up ssl_version_check().

There's a shorter fix here, but sticking in another preprocessor condition didn't seem as clean: https://github.com/TheMysteriousX/freeradius-server/commit/e24a1c2678df25f22df734abfb4a86ec4cbc95f1

PR #1445, opened 2015-12-10
Parsing error on debian/control: Fixes #1442

### LINKED ISSUE #1442: Debian: freeradius-config is missing make
Steps to reproduce: 
- Build the debian packages on Debian (jessie)
- Install them on another other, but minimal system where no make is present
- Postinst fails on freeradius-config where it wants to execute make to create the initial certificates.

The folowing change could fix it:
```
--- a/debian/control
+++ b/debian/control
@@ -61,7 +61,7 @@ Description: FreeRADIUS common files

 Package: freeradius-config
 Architecture: any
-Depends: freeradius-common (>= 3), ${misc:Depends}, openssl
+Depends: freeradius-common (>= 3), ${misc:Depends}, openssl, make
 Breaks: freeradius-config
 Description: FreeRADIUS default config files
  This package should be used as a base for a site local packages
```

matsimon:
Without the comma dpkg-buildpackage doesn't like the conrol file.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -61,7 +61,7 @@ Description: FreeRADIUS common files
 
 Package: freeradius-config
 Architecture: any
-Depends: freeradius-common (>= 3), ${misc:Depends}, openssl make
+Depends: freeradius-common (>= 3), ${misc:Depends}, openssl, make
 Breaks: freeradius-config
 Description: FreeRADIUS default config files
  This package should be used as a base for a site local packages


matsimon:
Shit happens... ;-)

Now this was build-tested on a Ubuntu box with no issues and the proper dependency handling.

arr2036:
It's annoying, we used to build that as part of CI, but they didn't provide all the packages in the containerised infrastructure, grrr.

PR #1448, opened 2015-12-11
Allow to set source interface (rlm_rest)

nchaigne:
Sometimes it is required to be able to set the source interface (or IP
address).
This small patch adds a configuration item and sets the appropriate
libcurl option.

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -46,6 +46,12 @@ rest {
 #	connect_proxy = "socks://127.0.0.1"
 
 	#
+	#  Set a specific source interface or IP address for outgoing traffic.
+	#  (Refer to libcurl CURLOPT_INTERFACE option.)
+	#
+#	src_interface = ""
+
+	#
 	#  The following config items can be used in each of the sections.
 	#  The sections themselves reflect the sections in the server.
 	#  For example if you list rest in the authorize section of a virtual server,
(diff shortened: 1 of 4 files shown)


alandekok:
You're missing the "src_interface" in the configuration struct.  Travis says:

    src/modules/rlm_rest/rlm_rest.c:85:64: error: no member named 'src_interface' in
      'struct rlm_rest_t'

nchaigne:
Doh. I botched this one :'(
I'll push a fix.

nchaigne:
OK got it. Compilation worked because I didn't have libcurl-devel installed, and consequently rlm_rest was not built.

arr2036:
So can you explain the need for this?  There are plenty of other modules that send traffic to external servers are none of them take a src_interface.

alandekok:
To be pedantic, home servers do take src_ipaddr.  That's because the source IP matters 100% of the time for RADIUS.

As a more general option, it's less useful.  Picking the correct source IP is generally better left to the OS routing stack.

nchaigne:
This is a requirement from a customer who asked to use separate source addresses for different kind of traffic. Probably it could be useful for the other modules, but I'm focusing on the modules which we do use.

arr2036:
Is this for authentication of the traffic?

jpereira:
This it's a good one! :+1:

arr2036:
You could replicate the same behaviour by setting up NAT rules on the local machine, and sending the packets to different local IP addresses.

There's a need for consistency when it comes to networking between the modules,  that's why they all use the same connection pool, and all have connection timeouts abstracted as the same config item.  This breaks that consistency, adding a one off hack to satisfy some bizarre requirement from one of your customers.  This isn't perl, we don't just tack stuff on randomly.  You're welcome to maintain your patches locally, but they're not suitable for inclusion into the server.

nchaigne:
I'm not an expert in networking, but what you suggest seems more complicated and dirty than just allowing the module to use whatever source address is wanted.

But it's alright. I have my local patches anyway. I just thought it might be useful.
I understand if you prefer not to merge this. :)

(and leave that poor perl alone, it didn't do anything wrong this time !)

PR #1456, opened 2015-12-15
After removing a vp, if removed element was cursor->found it should b…

pwdng:
…e set to null, not current

### DIFF
--- a/src/lib/cursor.c
+++ b/src/lib/cursor.c
@@ -469,7 +469,7 @@ fixup:
 	/*
 	 *	Fixup cursor->found if we removed the VP it was referring to
 	 */
-	if (vp == cursor->found) cursor->found = cursor->current;
+	if (vp == cursor->found) cursor->found = NULL;
 
 	/*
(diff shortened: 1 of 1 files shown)


arr2036:
No, that's wrong.  That would mean you started searching from the start of the cursor, why would you want that behaviour?

pwdng:
What I'm trying to solve is when fr_cursor_next_by_xxx gets called after fr_cursor_remove, cursor->found points to the vp after the one that was deleted and in fr_cursor_next_by_xxx, if cursor->found exists, we start iterating from cursor->found->next. The vp at cursor->found therefore gets skipped.
Where would it cause the search to restart from the cursor start?

arr2036:
In the case you remove any attribute whilst you're iterating with fr_cursor_next_by_xxx.

arr2036:
cursor->found should be set to the attribute before the one removed, or NULL if no attributes exist

arr2036:
``if (vp == cursor->found) cursor->found = before;``

Actually, looking through the code your original patch would work, but it's confusing.

Issue #1451, opened 2015-12-15
Python modules do not receive configuration parameters


alejandro-perez:
There is no way to provide configuration parameters to python modules. This is inconvenient as it requires users to modify the python module script, that might be installed in a standard PYTHON PATH. Also it makes difficult to maintain configuration when a new version is installed.

I understand that the dynamic nature of python vs. the static nature of the FR configuration conflict, but a trade-off solution should be provided. Here is my suggestion:
* Add a new configuration parameter to the rlm_python module, called "module_configuration". This is a string.
* Make this string somehow available to the python module (perhaps as a module constant). The format of this parameter is not standardize, so one python module might use CSV, whereas other might use JSON.

alejandro-perez:
In case you are wondering, yes I have a tentative patch :)
```
diff --git a/src/modules/rlm_python/rlm_python.c b/src/modules/rlm_python/rlm_python.c
index 22456a0..aadb0e1 100644
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -58,6 +58,7 @@ typedef struct rlm_python_t {
 	void		*libpython;
 	PyThreadState	*main_thread_state;
 	char const	*python_path;
+	char const  *module_configuration;
 
 	struct py_function_def
 	instantiate,
@@ -102,6 +103,7 @@ static CONF_PARSER module_config[] = {
 #undef A
 
 	{ "python_path", FR_CONF_OFFSET(PW_TYPE_STRING, rlm_python_t, python_path), NULL },
+	{ "module_configuration", FR_CONF_OFFSET(PW_TYPE_STRING, rlm_python_t, module_configuration), NULL },
 	CONF_PARSER_TERMINATOR
 };
 
@@ -239,6 +241,14 @@ static int mod_init(rlm_python_t *inst)
 		}
 	}
 
+	/*
+     * add module configuration as a constant
+	 */
+	if ((PyModule_AddStringConstant(radiusd_module, "module_configuration",
+					     inst->module_configuration)) < 0){
+		goto failed;
+	}
+
 #ifdef HAVE_PTHREAD_H
 	PyThreadState_Swap(NULL);	/* We have to swap out the current thread else we get deadlocks */
 	PyEval_ReleaseLock();		/* Drop lock grabbed by InitThreads */
```

jpereira:
Hi @alejandro-perez,

This is a good idea! I completed your suggestion here in #1453

[...]

PR #1459, opened 2015-12-17
Rlm python fixes

herwinw:
A number of changes to the python module to make it a lot more usable than it was. It started when I was trying to understand how rlm_python works due to https://github.com/FreeRADIUS/freeradius-server/issues/1451#issuecomment-164894510, and there were a lot of easy fixes to make it more mature than it is now. All these fixes can be backported to 3.0.10, because they preserve backwards compatibility. It was mainly as an excercise for myself to get some experience with embedding python.

As a bonus, I also updated the (almost completely empty) wiki page at http://wiki.freeradius.org/modules/Rlm_python.

### DIFF
--- a/src/modules/rlm_python/prepaid.py
+++ b/src/modules/rlm_python/prepaid.py
@@ -161,15 +161,14 @@ def authorize(authData):
   # If you want to use different operators
   # you can do
   # return (radiusd.RLM_MODULE_UPDATED,
-  #         radiusd.resolve(
-  #            'Session-Timeout := %s' % str(sessionTimeout),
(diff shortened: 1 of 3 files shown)


arr2036:
The most useful fix would be to make it pseudo multithreaded again, if you wanted to look at that

herwinw:
> The most useful fix would be to make it pseudo multithreaded again, if you wanted to look at that

Maybe. I have to admit that I have zero experience with Python embedding except the "monkey see, monkey do" style I have used for these few fixes. Nevertheless, I think the multithreading fixes are pretty much distinct from this branch.

PR #1460, opened 2015-12-18
Fix acctinputoctets & acctoutputoctets been reset to 0 when accounting stopped with no attributes of traffic.

moontide:
An example log of "Acct-Status-Type = Stop" with no attributes of traffic (no Acct-Input-Gigawords, no Acct-Input-Octets, no Acct-Output-Gigawords, no Acct-Output-Octets).

	Fri Dec 18 14:00:12 2015
			Acct-Status-Type = Stop
			Acct-Authentic = RADIUS
			User-Name = "**********"
			Called-Station-Id = "00-0C-43-**-**-**:\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**\\x**"
			NAS-Port-Type = Wireless-802.11
			NAS-Port = 4
			Calling-Station-Id = "9C-FC-01-**-**-**"
			Connect-Info = "CONNECT 54Mbps 802.11g"
			Acct-Session-Id = "5670F442-00000330"
			WLAN-Pairwise-Cipher = 1027076
			WLAN-Group-Cipher = 1027076
			WLAN-AKM-Suite = 1027073
			Acct-Session-Time = 811
			Event-Timestamp = "Dec 18 2015 14:00:12 CST"
			NAS-IP-Address = 192.168.201.20
			Acct-Unique-Session-Id = "0c82627b41241813"
			Timestamp = 1450418412

This fix is for v3.0.x, v3.1.x. For v2.x.x, it can be simpler:

	 acctinputoctets    = %{%{Acct-Input-Gigawords}:-(acctinputoctets >> 32)} << 32 | %{%{Acct-Input-Octets}:-(acctinputoctets & 0xFFFFFFFF)}, \
	 acctoutputoctets   = %{%{Acct-Output-Gigawords}:-(acctoutputoctets >> 32)} << 32 | %{%{Acct-Output-Octets}:-(acctoutputoctets & 0xFFFFFFFF)}, \

### DIFF
--- a/raddb/mods-config/sql/main/mysql/queries.conf
+++ b/raddb/mods-config/sql/main/mysql/queries.conf
@@ -292,10 +292,26 @@ accounting {
 						UNIX_TIMESTAMP(@acctupdatetime_old), \
 					framedipaddress = '%{Framed-IP-Address}', \
 					acctsessiontime = %{%{Acct-Session-Time}:-NULL}, \
-					acctinputoctets = '%{%{Acct-Input-Gigawords}:-0}' \
-						<< 32 | '%{%{Acct-Input-Octets}:-0}', \
-					acctoutputoctets = '%{%{Acct-Output-Gigawords}:-0}' \
-						<< 32 | '%{%{Acct-Output-Octets}:-0}' \
+					acctinputoctets	= \
+						CASE \
+							WHEN '%{Acct-Input-Gigawords}'='' THEN (acctinputoctets >> 32) \
+							ELSE '%{Acct-Input-Gigawords}' \
+						END << 32 \
+						| \
+						CASE \
+							WHEN '%{Acct-Input-Octets}'='' THEN (acctinputoctets & 0xFFFFFFFF) \
+							ELSE '%{Acct-Input-Octets}' \
+						END, \
+					acctoutputoctets = \
+						CASE \
+							WHEN '%{Acct-Output-Gigawords}'='' THEN (acctoutputoctets >> 32) \
+							ELSE '%{Acct-Output-Gigawords}' \
+						END << 32 \
+						| \
+						CASE \
+							WHEN '%{Acct-Output-Octets}'='' THEN (acctoutputoctets & 0xFFFFFFFF) \
+							ELSE '%{Acct-Output-Octets}' \
+						END \
 				WHERE AcctUniqueId = '%{Acct-Unique-Session-Id}'"
 
 			#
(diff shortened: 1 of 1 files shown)


alandekok:
Why is this necessary?  Please explain.

    acctinputoctets    = %{%{Acct-Input-Gigawords}:-(acctinputoctets >> 32)} << 32 | %{%{Acct-Input-Octets}:-(acctinputoctets & 0xFFFFFFFF)}

That string expansion works for v3, too.

And is your NAS really sending accounting "alive" packets with `Acct-Input-Gigawords`, and then a "stop" packet with no `Acct-Input-Gigawords`?  If that's the case, your NAS is broken.  Throw it in the garbage and buy one that works.

If you look at all of the accounting queries, all of them just store the data from the packet in the table.  None try to "merge" the data.  Because merging the data is wrong.

moontide:
@alandekok , I have two FreeRADIUS servers and two hostapd servers:
1. FreeRADIUS 3.0.10, hostapd 2.4. Both running in same Fedora 23 OS
2. FreeRADIUS 2.2.6 running in CentOS 6.7, wpad_2015-03-25 running in OpenWRT 15.05 or 14.07 OS in a HiWiFi-J2 router

In server 1 (FreeRADIUS 3.0.10), the '(' and ')' characters are expanded to '=28' and '=29', so it will caused updating SQL failed to execute.

In server 2 (FreeRADIUS 2.2.6), the '(' and ')' character are not expanded, so the update SQL will successfully executed.

----

No matter where I added `radius_acct_interim_interval=NNN` to hostapd.conf or not (both hostapd in Fedora or wpad in OpenWRT), I got some "Acct-Status-Type = Stop" log with no Acct-Input-Gigawords/Acct-Input-Octets/Acct-Output-Gigawords/Acct-Output-Octets, and of course it will reset `acctinputoctets` to zero, and I can't use `MaxDailyTraffic` counter to restrict users access WiFi AP.

alandekok:
OK, the brackets are escaped because they shouldn't be here.  You could just as well do:

    ... (%{%{Acct-Input-Octets}:-acctinputoctets} & 0xFFFFFFFF) ...

    No matter where I added `radius_acct_interim_interval=NNN` to hostapd.conf ...

Then that software is broken.  I'll push a patch to the hostap people.

Adding hacks to FreeRADIUS to work around broken vendor software is *very* problematic.

moontide:
> If you look at all of the accounting queries, all of them just store the data from the packet in the table. None try to "merge" the data. Because merging the data is wrong.

@alandekok, I just checked [rfc2866](http://freeradius.org/rfc/rfc2866.html), it said the statistics data are optional. so, this workaround can tolerate such situation.

> ## 2. Operation
> ...
> At the end of service delivery the client will
>   generate an Accounting Stop packet describing the type of service
>   that was delivered **and optionally statistics such as elapsed time,
>   input and output octets, or input and output packets**.
>  ...

alandekok:
I know what RFC 2866 says.  There are a LOT of things in the RFCs that are allowed, or that the RFCs are silent on.  That doesn't make them good ideas.

This is the first time in 20 years I've seen a NAS behave this way.  It's wrong.  The NAS should be fixed.

arr2036:
I agree with @alandekok NAS is broken.  Sending stats for only *some* accounting types is plain stupid.  For Acct-Status-Type = Stop you *DEFINITELY* want the accounting stats, as that's what the counters were at the end of the session.

If you want to fix this for your deployment, with your broken hostapd daemon, then that's fine, but it's not something that's useful for 99% of other deployments.

PR #1462, opened 2015-12-18
Add systemd support for Debian Jessie

mcnewton:
Apparently it is moving a step forward to break convenience.
So systemd users will have to manually update the system to use
/usr/local/etc/freeradius themselves if that is where their
config is located.

Without this, the freeradius package refuses to start on jessie, and systemd can't cope with shell code in the default file. But this patch doesn't change functionality when building on wheezy.

### DIFF
--- a/debian/freeradius.default
+++ b/debian/freeradius.default
@@ -1,6 +1,15 @@
-# Options for the FreeRADIUS deamon.
+# Options passed to the FreeRADIUS deamon.
+#
+FREERADIUS_OPTIONS=""
+
+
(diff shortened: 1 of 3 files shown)


arr2036:
Wow, ok :(

mcnewton:
Yeah - and it's not something broken on my system either; it's a completely clean VM install :(

Looks like there may be a hack that would get it working so the local raddb is picked up if it exists, but I poked out my eyes looking at it so I can't see any more.

PR #1463, opened 2015-12-18
Add systemd support for Debian Jessie

mcnewton:
Apparently it is moving a step forward to break convenience.
So systemd users will have to manually update the system to use
/usr/local/etc/freeradius themselves if that is where their
config is located.

cherry-pick for v3.1.x.

### DIFF
--- a/debian/freeradius.default
+++ b/debian/freeradius.default
@@ -1,6 +1,15 @@
-# Options for the FreeRADIUS deamon.
+# Options passed to the FreeRADIUS deamon.
+#
+FREERADIUS_OPTIONS=""
+
+
+# If FreeRADIUS is being used on a SysVinit system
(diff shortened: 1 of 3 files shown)


arr2036:
Do all the patches still apply cleanly for you?  I was having an issue with the monit configuration file last night.  Couldn't determine if it was just the machine I was building on...

mcnewton:
Everything builds cleanly with sbuild from v3.0.x on both wheezy and jessie chroots. Haven't tested v3.1.x - will try now.

mcnewton:
For both of wheezy and jessie: All patches apply cleanly on 3.0.x; one hunk offset by 1 line on 3.1.x. monit patch applies cleanly on all. Your machine probably does not have super cow powers.

PR #1465, opened 2015-12-20
Debian path cleanups

herwinw:
A few changes to reflect the actual config path in debian (`/etc/freeradius`) instead of the default of freeradius (`/etc/raddb`). Based on https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=775281

### DIFF
--- a/debian/patches/radiusd-to-freeradius.diff
+++ b/debian/patches/radiusd-to-freeradius.diff
@@ -90,3 +90,436 @@ index 1ae66ec..ee1a711 100644
 
 	/*
 	 *	Don't put output anywhere until we get told a little
+diff --git a/man/man1/radclient.1 b/man/man1/radclient.1
+index 6305411..8902db6 100644
+--- a/man/man1/radclient.1
++++ b/man/man1/radclient.1
+@@ -56,7 +56,7 @@ Use IPv6
+ Send each packet \fIcount\fP times.
+ .IP \-d\ \fIraddb_directory\fP
+ The directory that contains the user dictionary file. Defaults to
+-\fI/etc/raddb\fP.
++\fI/etc/freeradius\fP.
+ .IP \-D\ \fIdictionary_directory\fP
+ The directory that contains the main dictionary file. Defaults to
+ \fI/usr/share/freeradius\fP.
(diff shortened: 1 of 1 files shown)


arr2036:
Ugh this is ridiculous, is there any other way to modify the sources, using sed perhaps?

mcnewton:
Patch is a good idea (to make it correct for Debian), but yeah, this is just starting to get nasty :-(

Could it be made more generic? The man pages are also wrong if FreeRADIUS is installed in /opt/... or /usr/local/... etc.

I think each man page should be turned into e.g. rlm_mschap.5.in and processed by the standard automake type stuff. They could use the @xxxDIR@ type expansions and then would have the correct locations in them *wherever* FreeRADIUS is installed, not just for Debian.

arr2036:
Sounds like a good plan.

alandekok:
Honestly, the whole autoconf nonsense is starting to irritate me.  It's not rocket science to write:

    sed 's,/etc/raddb,$RADDBDIR,g' < rlm_mschap.5 > $MANPATH/rlm_mschap.5

I'd rather have one-line sed scripts in a `Makefile` than butcher every single `man` file.

The autoconf replacement stuff is good for updating 30 things... like radiusd.conf.  But for references to `/etc/raddb/`, sed scripts are a lot better.

mcnewton:
I've been taking a look, and just doing sed in makefiles seems the easiest approach. That's what autoconf is going to do anyway, so it seems rather pointless to use it...

mcnewton:
OK, generic version in bc96405cce8ad3bf0be2e8b534c9a2688c52e647

@herwinw thanks - it's possible that the generic install sed misses one or two things (unusual paths in man pages, or not raddbdir or datarootdir) so feel free to open a new PR if you see any bits that need fixing up. Also I guess it's possible some of the debian patches can be cleaned up now.

arr2036:
If you want to be super-efficient you can stick ``Closes #1465`` in the commit, then it'll close the issue, and create a link from the issue to the commit :)

mcnewton:
Ah, cool :). I mistakenly assumed that would close an issue rather than a PR.

PR #1469, opened 2015-12-22
Allow config section in python config

herwinw:
They will recursively be turned into a dict, and added to radiusd.config. Behaviour is identical to the same config option in rlm_perl.

This is the extended version of #1451/#1453. It has *not* been addressed by #1459.

This patch can be backported to v3.0.x, it just adds another attribute to the `radiusd` module, script that don't use it just keep not using it ;)

### DIFF
--- a/raddb/mods-available/python
+++ b/raddb/mods-available/python
@@ -50,4 +50,15 @@ python {
 
 	mod_send_coa = ${.module}
 #	func_send_coa = send_coa
+
+	# You can define configuration items (and nested sub-sections) in python "config" section.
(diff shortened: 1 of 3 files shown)


herwinw:
@alejandro-perez: You are the one who originally requested this feature, care to try/test it?

alejandro-perez:
Sure. Will try it in a few days, though. I'm on holidays now :)

alejandro-perez:
It works perfectly fine for the tests I've done. Thanks! (the most difficult part was figuring out that it was a different repository the one I had to clone :))

alejandro-perez:
Please, accept this request.

PR #1476, opened 2015-12-30
Fix policyd filter

nicklowe:
Update policy.d/filter

Improve the regular expressions in filter_username to mimise the use of .* where possible.

Add support for User-Name values that are realm\user delimited.

Correct the filter_inner_identity implementation to:
1) Permit the case that no realm is given in an anonymous outer identity where the inner identity has a realm.
2) Require the inner identity to not be anonymized.
3) Support case insensitive operation, default to permitting this.
4) Make the check for "anonymous" stricter by requiring anonymous or anon, not just any stripped user-name that begins with anon.
5) Do not allow subrealms in the inner identity where the outer identity is not anonymized.
6) Improve the module failure messages that are returned.

### DIFF
--- a/raddb/policy.d/filter
+++ b/raddb/policy.d/filter
@@ -3,27 +3,39 @@
 #	realms.
 #
 deny_realms {
-	if (&User-Name && (&User-Name =~ /@|\\/)) {
+	#
+	#  No User-Name, skip.
+	#
+	if (!&User-Name || &User-Name == "") {
+		noop
+	}
+
+	if (&User-Name =~ /@|\\/) {
+		update reply {
+			&Reply-Message += 'Rejected: User-Name contains a realm'
+		}
 		reject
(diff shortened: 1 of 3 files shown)


raddb/policy.d/filter
```diff
@@ -3,27 +3,39 @@
...
-	if (!&User-Name) {
+	#
+	#  No User-Name, skip.
+	#
+	if (!&User-Name || &User-Name == "") {
 		noop
 	}
 
 	#
 	#  reject mixed case e.g. "UseRNaMe"
```
herwinw: Maybe not directly related to this pull request, but this comment is not completely valid. `User-Name := "BOB"` is not mixed case, but would still be rejected below.

nicklowe: I'll fix that.

PR #1478, opened 2015-12-30
ability to fetch attributes from sql module to control list

gigabytetmn:
I have cisco ASR1000 for PPPoE subscriber termination.
I need automatic service activation/deactivation, so I need CoA queries, maybe on accounting update.
This patch allows fetch on some cases (maybe dependent on check in database logic) attributes from sql server and place in control list.
After that freeradius check flag and build CoA request.
This patch tested on Cisco ASR1000 version 15.5, after some time I will test this functionality on Mikrotik RouterOS >= 6.33.

Example in attached file.

[configs.zip](https://github.com/FreeRADIUS/freeradius-server/files/75163/configs.zip)

### DIFF
--- a/share/dictionary.rfc4072
+++ b/share/dictionary.rfc4072
@@ -2,7 +2,7 @@
 # Copyright (C) 2015 The FreeRADIUS Server project and contributors
 #
 #	Attributes and values defined in RFC 4072
-#	http://www.ietf.org/rfc/4072.txt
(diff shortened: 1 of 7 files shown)


arr2036:
Duplicate functionality.  The same thing can be achieved by calling 'sql.authorize' in an accounting section.

gigabytetmn:
hm, yes it works.
please tell me where it is documented.

PR #1480, opened 2015-12-30
Fix policyd filter

nicklowe:
Update policy.d/filter

Improve the regular expressions in filter_username to minimise the use of .* where possible.

Add support for User-Name values that are realm\user delimited.

Correct the filter_inner_identity implementation to:
1) Permit the case that no realm is given in an anonymous outer identity where the inner identity has a realm.
2) Require the inner identity to not be anonymized.
3) Support case insensitive operation, default to permitting this.
4) Make the check for "anonymous" stricter by requiring anonymous or anon, not just any stripped user-name that begins with anon.
5) Do not allow subrealms in the inner identity where the outer identity is not anonymized.
6) Improve the module failure messages that are returned.

### DIFF
--- a/raddb/policy.d/filter
+++ b/raddb/policy.d/filter
@@ -3,16 +3,10 @@
 #	realms.
 #
 deny_realms {
-	#
-	#  No User-Name, skip.
-	#
-	if (!&User-Name || &User-Name == "") {
-		noop
-	}
-
-	if (&User-Name =~ /@|\\/) {
+	if (&User-Name && &User-Name =~ /@/) {
+	#if (&User-Name && &User-Name =~ /@|\\/) {
 		update reply {
-			&Reply-Message += 'Rejected: User-Name contains a realm'
+			Module-Failure-Message = 'Rejected: User-Name contains a realm'
 		}
 		reject
 	}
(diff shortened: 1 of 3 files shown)


alandekok:
Please don't open / close / open / close pull requests.  You can just force push changes to your branch, and it will update the existing pull request.

nicklowe:
Sorry, I didn't realise github would behave that way.

nicklowe:
Isn't anon or anonymous acceptable? I really don't think allowing $anon.* is a good idea.

alandekok:
It's not about what's a good idea.  It's about what terrible things that vendors do.

We *cannot* break existing networks.  Even if vendors do dumb, but reasonably secure things.

nicklowe:
Do we have any data that people do send something other than anon or anonymous?

nicklowe:
You are not going to break existing networks because this is a change to the default configuration, so I don't think the concern applies?

nicklowe:
Due to the branch conflicts, I closing this.

alanbuxey:
why remove the "Reply-Message" in the filter too - helps with federated
systems so they can understand why some remotely authenticated visitor
didnt get auth'd

alandekok:
> On Jan 3, 2016, at 11:54 AM, Alan Buxey <notifications@github.com> wrote:
> 
> why remove the "Reply-Message" in the filter too - helps with federated
> systems so they can understand why some remotely authenticated visitor
> didnt get auth'd

  It can be added back.

  The goal was to have messages which are useful to the local administrator.

  We can always copy the Module-Failure-Message attributes to Reply-Message in Access-Reject.

PR #1481, opened 2015-12-31
Policyd filter fix

nicklowe:
Update policy.d/filter

Rename the filter_inner_identity implementation to filter_inner_outer_identity and...

1) Make the check for "anonymous" absolute and deterministic by looking for anonymous or anon, not just any stripped user-name that begins with anon. This is incorrect, poor behaviour as it would match any genuine identities that begin with anon.

This change should be acceptable, and is sensible and reasonable, on balance, as the changes being made here are to the default configuration only. They do not apply to existing installations so there is no risk of breaking a current deployment. The behaviour could be modified as a new installation is configured if this ever became necessary.

Additionally, without doing this, we are not going to see feedback about what this would break.

A strict, absolute default is desirable as there is no ambiguity in what is considered an anonymized identity and what is not.

2) This strict definition then allows us to require the inner identity to not be anonymized, having defined that "anonymous" and "anon" are anonymized.

3) Change the logic to not allow subrealms in the inner identity where the outer identity is not anonymized: bob@secure.example.com (inner) and bob@example.com (outer) should fail. Presently, this passes.

4) Make the module failure messages more descriptive.

### DIFF
--- a/raddb/policy.d/filter
+++ b/raddb/policy.d/filter
@@ -3,13 +3,22 @@
 #	realms.
 #
 deny_realms {
-	if (&User-Name && (&User-Name =~ /@|\\/)) {
+	#
+	#  reject realms
+	#  e.g. "user@example.com"
+	#  e.g. "example.com\user" (disabled by default)
+	#
+	if (&User-Name && &User-Name =~ /@/) {
+	#if (&User-Name && &User-Name =~ /@|\\/) {
+		update request {
+			&Module-Failure-Message += 'User-Name contains a realm'
+		}
 		reject
 	}
 }
 
 #
-#	Filter the username
+#	Filter the User-Name
 #
 #  Force some sanity on User-Name. This helps to avoid issues
 #  issues where the back-end database is "forgiving" about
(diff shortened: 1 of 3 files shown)


nicklowe:
The patch does contain a change to share/dictionary.freeradius.internal

nicklowe:
anon(ymous)? is different to anon where there is that $

My point is that we are not going to understand the impact of that change without enforcing this strongly, enforcing it strongly also allows us to ensure that the inner identity is not anonymised by having that strict definition.

We aren't forcing this as it is a change to the default configuration, it doesn't affect **existing** deployments therefore. Any **new** deployment would be subjected and this should be caught in the testing phase for that if there are any issues.

This is also only enabled by default in FreeRADIUS 3.1. Respectfully, I disagree here and strongly believe we should noot take on a future technical debt by not doing the correctly upfront.

Cheers,

Nick

nicklowe:
Agreed, but the other two attributes were only added a few days ago so it's surely moot in this case? They've never been in an official release, only in work in progress software.

It seemed cleaner to put the stripped component before to the realm to me.

nicklowe:
The reason for adding the Inner and Outer stripped username was simplification of other checks without having to use temporary attributes.

nicklowe:
If somebody disables filter_username, I am of the opinion that we should code this robustly enough that it stands on its own merit.

nicklowe:
I am withdrawing this to rework it to log on the aspects Alan is unhappy about, but not enforce.

PR #1484, opened 2016-01-06
Rlm python fixes

### LINKED ISSUE #1467: OP codes in rlm_python radiusd module do not match those from token.h
They are out of sync, thus produce mistaken results. In particular, the '++' operator was missing, hence everything after that value has shifted.

They should be 
```
OP={       '{':2,   '}':3,   '(':4,   ')':5,   ',':6,   ';':7,  '++': 8, '+=':9,  '-=':10,  ':=':11,
  '=':12, '!=':13, '>=':14,  '>':15, '<=':16,  '<':17, '=~':18, '!~':19, '=*':20, '!*':21,
 '==':22 , '#':23 }
```

herwinw:
Backport of  #1459 to release 3.0, fixes #1467

### DIFF
--- a/src/modules/rlm_python/prepaid.py
+++ b/src/modules/rlm_python/prepaid.py
@@ -161,15 +161,14 @@ def authorize(authData):
   # If you want to use different operators
   # you can do
   # return (radiusd.RLM_MODULE_UPDATED,
-  #         radiusd.resolve(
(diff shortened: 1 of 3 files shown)


herwinw:
Travis generates errors, but those are the same as in v3.0.x. Building it locally works fine.

PR #1494, opened 2016-01-14
Fix build on wheezy

matsimon:
ntstatus.h file is present in samba-dev, however before jessie it
was called samba4-dev. Ubuntu precise 12.04LTS is another derived
distribution release with this package name.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -27,7 +27,7 @@ Build-Depends: debhelper (>= 9),
  libmemcached-dev,
  libhiredis-dev,
  python-dev,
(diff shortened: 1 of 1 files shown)


matsimon:
Maybe the change by @herwinw that added samba-dev should be applied to the branch 3.1.x as well but not samba4-dev as here. When 3.1 will be released these distributions will be much older than they are already now.

Anything older than that, be it Debian or Ubuntu doesn't have Samba 4 packages with ntstatus.h anyway.

PR #1499, opened 2016-01-18
add note about elasticsearch versions

mcnewton:
Last minute check shows it probably won't work with elasticsearch 2 (dots in field names) so warn anyone about this. Fix in logstash at the moment isn't entirely trivial AFAICT, and I don't like messy configs to work around things if it can be helped...


mcnewton:
Sorry, another one incoming - should fix up the problems with es2, though I've not tested it with that version. Should be the last update on this...

PR #1521, opened 2016-02-02
Removed references to SQL_DOWN in an outdated README

qnet-herwin:
There were no references to SQL_DOWN in the code, so this part of the README only provided false information.

The remainder of the file didn't look that useful too, but I left it there because it included some authors information. I can probably be removed because that info can be found in the git logs, but I'm not really the one to make that call.


arr2036:
You can probably just delete the file.  Any code documentation should be in doxygen.

qnet-herwin:
Updated PR to just remove the file completely

PR #1526, opened 2016-02-05
[3.1.x] DHCP strings decoding

nchaigne:
Header fields of type "string" are null terminated strings, decode them
as such


alandekok:
I've pulled this over to 3.0, too.

nchaigne:
Nice. Thanks !

arr2036:
Looks good, thanks for submitting this.  I noticed we were getting lots of zeros in the debug output, but thought it was just an artefact of working with fixed length fields :)

Issue #1524, opened 2016-02-04
Reply attributes lost when reply from perl contains one attribute with no value


bt4:
Hi

I noticed that when the perl module returns a reply attribute with no value, this somehow corrupts all the reply attributes in the reply. I have reproduced this on a test server running 3.0.11 with a simple setup as follows:

Perl script:
```
use strict;
use warnings;
use Data::Dumper;

# Bring the global hashes into the package scope
our (%RAD_REQUEST, %RAD_REPLY, %RAD_CHECK);

#
# This the remapping of return values
#
use constant {
	RLM_MODULE_REJECT   => 0, # immediately reject the request
	RLM_MODULE_OK       => 2, # the module is OK, continue
	RLM_MODULE_HANDLED  => 3, # the module handled the request, so stop
	RLM_MODULE_INVALID  => 4, # the module considers the request invalid
	RLM_MODULE_USERLOCK => 5, # reject the request (user is locked out)
	RLM_MODULE_NOTFOUND => 6, # user not found
	RLM_MODULE_NOOP     => 7, # module succeeded without doing anything
	RLM_MODULE_UPDATED  => 8, # OK (pairs modified)
	RLM_MODULE_NUMCODES => 9  # How many return codes there are
};

# Same as src/include/radiusd.h
use constant	L_DBG=>   1;
use constant	L_AUTH=>  2;
use constant	L_INFO=>  3;
use constant	L_ERR=>   4;
use constant	L_PROXY=> 5;
use constant	L_ACCT=>  6;

# Function to handle post_auth
sub post_auth {
    if ($RAD_REQUEST{'DHCP-Client-Hardware-Address'} eq "00:26:24:a3:9f:d6") {
	$RAD_REPLY{'DHCP-Your-IP-Address'} = "10.180.1.2";
        $RAD_REPLY{'DHCP-Subnet-Mask'} = "255.255.255.0";
	$RAD_REPLY{'DHCP-Router-Address'} = "10.180.1.1";
    }

    # INSERT DUFF ATTRIBUTE!!!!
    $RAD_REPLY{'DHCP-Relay-Remote-Id'} = ();

    &log_reply_attributes;
    return RLM_MODULE_OK;
}

sub log_reply_attributes {
	for (keys %RAD_REPLY) {
		&radiusd::radlog(L_DBG, "XXXXXXXXXXXXXXXXXXXXXX RAD_REPLY: $_ = $RAD_REPLY{$_}");
	}
}
```

[...]

PR #1527, opened 2016-02-06
Made rlm_perl less error-prone

herwinw:
- If parsing an attribute failed when updating the list, don't ignore
  the other updates to that list.
- Print an error when passing an undef value in the return list, instead
  of silently discarding it.

This fixes the problems seen in #1524

### DIFF
--- a/src/modules/rlm_perl/rlm_perl.c
+++ b/src/modules/rlm_perl/rlm_perl.c
@@ -688,50 +688,46 @@ static void perl_store_vps(UNUSED TALLOC_CTX *ctx, REQUEST *request, VALUE_PAIR 
  *     Value Pair Format
  *
  */
-static int pairadd_sv(TALLOC_CTX *ctx, REQUEST *request, VALUE_PAIR **vps, char *key, SV *sv, FR_TOKEN op,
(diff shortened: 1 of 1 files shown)


arr2036:
I know it seems like cruel and unusual punishment, but could you add a few test cases around this?  They're pretty easy to do with the new framework, and hopefully it'll encourage others to start doing the same.

See ``src/tests/modules``. Thanks!

herwinw:
Well, I often blame my colleagues for not adding test, so no blame taken here. On the other hand: I can't promise that I'll have time to do that this week.

lytboris:
Btw this pull request is not backported to 3.1.x branch and may be the reason #1697 appeared.

qnet-herwin:
> Btw this pull request is not backported to 3.1.x branch and may be the reason #1697 appeared.

I doubt that, since the full debug log shows "freeradius-server : 3.0.12"

lytboris:
There are two separate statements.
1) this pull request is not ported to 3.1 and up so it will stay with 3.0.x forever
2) since it is merged into 3.0.x, it may be the cause of issue I mentioned.

Hope this clears my point up.

PR #1529, opened 2016-02-08
Update rlm_python.c

alejandro-perez:
The mod_instantiate() function just forwards the return code the Python instantiate(p) call gives. But in case of error, the mod_error() and mod_instantiate_clear() functions are not called. 

The immediate consequence of this is that, even in the presence of an error in the module, FR continues executing and therefore, the module can be called again and again, even though it claimed to not being  able to instantiate itself.

The patch makes sure that if the do_python() call returns a value <0, the failed: label is executed.

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -770,7 +770,7 @@ static void python_parse_config(CONF_SECTION *cs, int lvl, PyObject *dict)
 static int mod_instantiate(CONF_SECTION *conf, void *instance)
 {
 	rlm_python_t *inst = instance;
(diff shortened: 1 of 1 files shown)


arr2036:
Can you rebase these into one commit, thanks!

alejandro-perez:
Done :)

arr2036:
Thanks!

alejandro-perez:
Umm, apparently I missed another modification to make rlm_python fail when the Python code says so in the instantiate() method (by returning -1). Should I create a new Pull Request with that additional modification, or I'd better modify this one and resubmit?

PR #1530, opened 2016-02-09
Clarify sqlippool operation and queries, change defaults

abatie:
Comments updated to describe operation of sqlippool and the supporting queries.

Also changed default queries to preserve user info, allowing reuse of ip addresses after disconnects or outages.

### DIFF
--- a/raddb/mods-config/sql/ippool/mysql/queries.conf
+++ b/raddb/mods-config/sql/ippool/mysql/queries.conf
@@ -3,9 +3,43 @@
 #  ippool/mysql/queries.conf -- MySQL queries for rlm_sqlippool
 #
 #  $Id$
+#
+#  sqlippool (rlm_sqlippool) is called in the postauth section to allocate an IP address:
+#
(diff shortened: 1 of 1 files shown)


arr2036:
Comments are useful, but why are the clear queries commented out?

arr2036:
Ah that's the preserving user info part?  I'd request those queries not be changed unless that's the default for other languages.

abatie:
On 2/8/16 4:22 PM, Arran Cudbard-Bell wrote:
> Comments are useful, but why are the clear queries commented out?

Clearing all the entries for a NAS on startup/shutdown removes the
information needed to allow users to get the same ip address they had
before.  I should probably split this into two requests, but the
comments were sorta tied to the specific queries...

abatie:
On 2/8/16 4:25 PM, Arran Cudbard-Bell wrote:
> Ah that's the preserving user info part? I'd request those queries not
> be changed unless that's the default for other languages.

Yes, that makes sense.

PR #1536, opened 2016-02-14
Adds missing error codes to rlm_python

alejandro-perez:


qnet-herwin:
It's got my blessing :)

PR #1540, opened 2016-02-16
Make do_python to return the indicated code even when there is no request

alejandro-perez:
current implementation of do_python() always returns SUCESS when there is no request. Such is the case of instantiate(). Since the
```
if (!request)		
		goto finish;
```
code seems to be there to protect the access to request-> elements later, we add the protection in 
```
if (PyTuple_CheckExact(pRet) && request != NULL) {
 		PyObject *pTupleInt;		 		PyObject *pTupleInt;
```
that is the only place in the remaining code of the method request is used.

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -546,9 +546,6 @@ static rlm_rcode_t do_python(rlm_python_t *inst, REQUEST *request, PyObject *pFu
 		goto finish;
 	}
 
-	if (!request)
-		goto finish;
-
 	/*
 	 *	The function returns either:
 	 *  1. (returnvalue, replyTuple, configTuple), where
@@ -562,7 +559,7 @@ static rlm_rcode_t do_python(rlm_python_t *inst, REQUEST *request, PyObject *pFu
 	 *
 	 * xxx This code is messy!
 	 */
-	if (PyTuple_CheckExact(pRet)) {
+	if (PyTuple_CheckExact(pRet) && request != NULL) {
 		PyObject *pTupleInt;
 
 		if (PyTuple_GET_SIZE(pRet) != 3) {


qnet-herwin:
```
def instantiate(request):
    return radiusd.RLM_MODULE_NOOP

def instantiate(request);
    return (radiusd.RLM_MODULE_NOOP, (), ())
```

With this patch the behaviour of these two returns suddenly differs: the first really returns noop, the second returns the default value (ok).

But honestly, to indicate that something went wrong with instantiation or detaching, I guess an exception would be more suitable than a return value.

alejandro-perez:
El 16/02/16 a las 13:22, Herwin escribió:
> |def instantiate(request): return radiusd.RLM_MODULE_NOOP def 
> instantiate(request); return (radiusd.RLM_MODULE_NOOP, (), ()) |
>
> With this patch the behaviour of these two returns suddenly differs: 
> the first really returns noop, the second returns the default value (ok).
>
> But honestly, to indicate that something went wrong with instantiation 
> or detaching, I guess an exception would be more suitable than a 
> return value.
>

That would work for me. The API of rlm_python says that you must return 
-1 in case of error.
I agree an exception would be better.

alejandro-perez:
Ok, but I see where the problem is.

do_python returns rlm_rcode_t, but mod_instantiate does not. So when Python's instantiate() method fails (e.g. exception), RLM_MODULE_FAIL is returned. But that does not have any impact on whoever is handling mod_instantiate() as RLM_MODULE_FAIL != -1 (in particular, all RLM_MODULE_XXX are > 0).

So, perhaps the change to do is something as follows (in modification of my previous PR):
```
        if (code == RLM_MODULE_OK) {
                return code;
        }
```
And then indicate that in instantiate() you have to raise an exception if you want to abort() server initiation.

alejandro-perez:
I forgot to mention that I have tested it and it works in my configuration. Let me create a new PR

alejandro-perez:
https://github.com/FreeRADIUS/freeradius-server/pull/1541

PR #1541, opened 2016-02-16
Make instantiate() return -1 when an exception occurs

alejandro-perez:
As do_python returns RLM_MODULE_FAIL when an exception occurs, and RLM_MODULE_OK in any other situation, just compare the code value and return RLM_MODULE_OK (in any case != -1) when success, and reach the failuer label when RLM_MODULE_FAIL (that makes -1 be returned)

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -816,7 +816,7 @@ static int mod_instantiate(CONF_SECTION *conf, void *instance)
 	 *	return value if successful. Otherwise, continue to the failed label
 	 */
 	code = do_python(inst, NULL, inst->instantiate.function, "instantiate", false);
-	if (code >= 0) {
+	if (code == RLM_MODULE_OK) {
 		return code;
 	}
 failed:


arr2036:
So your reading of the Python examples suggests that the python function should return a module return code from the instantiate function?

alejandro-perez:
Umm, let me explain it a little better. If I'm wrong, please, forgive my ignorance. Just trying to help.

"do_python" is used for forwarding all the module callbacks (instantiate, authorize, authenticate....). It returns rlm_module_t, being RLM_MODULE_OK by default. If an exception is raised, RLM_MODULE_FAIL is returned instead.

mod_instantiate (before my last merged PR) called to do_python and returned whatever it gave. That means that mod_instantiate was returning RLM_MODULE_OK or RLM_MODULE_FAIL, both >= 0. However, in the FreeRadius side, -1 is considered a failure, and >= 0 considered a success for this function. 

So, for mod_instantiate, it is needed a "mapping", which makes that for a return code of RLM_MODULE_OK gives a 0, whereas a return code of RLM_MODULE_FAIL gives a -1.

The reason why returning -1 directly from the instantiate() python function does not work is because do_python does not even checks the return value when request == NULL (such is the case of the instantiate call).

Hope this makes my point clearer.

mcnewton:
(tldr; nak)

This patch makes no sense, and also breaks the default config.

As standard, the python instantiate function is not defined. do_python returns RLM_MODULE_NOOP if the function is not defined. So this patch then fails instantiation and the server won't start up, just because no function has been set.

Arguably, as stands

    return code;

is pointless as well; the only thing that matters for the instantiate function is that it returns <0 for error. So that should really just be "return 0".

There's no neat solution here (for example arguably "RLM_MODULE_HANDLED" is a stupid response from an instantiate function) but the most sensible to me would be to define RLM_MODULE_FAIL as the only negative response, and do

    if (code != RLM_MODULE_FAIL) {
        return 0;
    }

(As it stands, rlm_python instantiation will never fail because RLM_MODULE_* are enum starting at 0.)

However, even that won't work, because do_python says

    if (!request)
        goto finish;

when ret=RLM_MODULE_OK. At instantiation time there is no request, so the return code from the instantiation function is never checked and it always returns OK.

If you really want to be able to bomb out at python instantiation time, the code (around the comment "xxx This code is messy!") needs hacking on in do_python.

mcnewton:
See #1546 which should just get allow the return code to be passed back from do_python when there is no request, and then returns success (0) from mod_instantiate if this is >= 0.

alejandro-perez:
Thanks!

PR #1542, opened 2016-02-19
Increase to varchar(128), from 32, for cui in DB schemas.

knutid:
CUI from NASes that I have worked with can be at least 80 characters long. Hence lifting it to 128, from 32.


alanbuxey:
in theory a CUI could fill a whole TLV of 255 bytes...

PR #1545, opened 2016-02-22
Correctly copy vendor specific attributes in copy_request_to_tunnel

mcnewton:
Checking just vp->da->attr is not enough; we also need to ensure that a vendor is not set. Specifically Cisco-AVPair is attribute 1, the same as User-Name...


alandekok:
I'll pull over the patch from v3.0.x by hand.

PR #1546, opened 2016-02-23
Allow rlm_python to fail instantiation

mcnewton:
See comments on alejandro-perez's PR #1541


src/modules/rlm_python/rlm_python.c
```diff
@@ -817,7 +824,7 @@ static int mod_instantiate(CONF_SECTION *conf, void *instance)
 	 */
 	code = do_python(inst, NULL, inst->instantiate.function, "instantiate", false);
 	if (code >= 0) {
-		return code;
+		return 0;
```
mcnewton: No point returning code here; it won't currently be used, so safer to just return 0.

PR #1557, opened 2016-03-09
remove rest of -m option

mcnewton:


arr2036:
Ugh, there's so much of it.

mcnewton:
Doesn't surprise me. Small thing like that have a habit of sneakily getting everywhere. AFAICT it was always pretty much meaningless anyway.

arr2036:
It explicitly freed various things.  Without it connections weren't shut down properly, and some exec triggers didn't fire.

mcnewton:
Ah, OK. Makes sense to do by default then anyway.

PR #1563, opened 2016-03-12
Fix the 'at least one dot' filter to actually allow more than one dot

grawity:
Surely I'm not the first person to use realms like "test.example.com"?...

### DIFF
--- a/raddb/policy.d/filter
+++ b/raddb/policy.d/filter
@@ -61,7 +61,7 @@ filter_username {
 		#  must have at least 1 string-dot-string after @
 		#  e.g. "user@site.com"
 		#
-		if ((&User-Name =~ /@/) && (&User-Name !~ /@([^.]+)\.([^.]+)$/))  {
+		if ((&User-Name =~ /@/) && (&User-Name !~ /@[^.]+(\.[^.]+)+$/))  {
 			update request {
 				&Module-Failure-Message += 'Realm does not have at least one dot separator'
 			}


alanbuxey:
ummm, you got rid of the anchor ($) - why?

alanbuxey:
its 'at least one dot' - so test.example.com  would work..... are you sure
you havent got a trailing space in your realm?

grawity:
No, I'm quite sure. Just take a look at the regex: both the first and second `[^.]+` allow _anything but_ a dot. So you have:

 - `@` – a literal "at" sign
 - `([^.]+)` – anything _except_ a dot
 - `\.` – a literal dot
 - `([^.]+)` – anything _except_ a dot
 - `$` – the end-of-string anchor

_How_ could `@test.example.com` or `@one.two.three` possibly match that?

Try it out with _egrep_, or perhaps at <https://regex101.com/r/oY0nJ7/1>, if you still insist that the regex is right.

arr2036:
@grawity you're right, but you still shouldn't remove the anchor. You need to verify the string doesn't contain '..'.

grawity:
Hmm, in that case... would `@([^.]+\.)+[^.]+$` work? (I'm a bit wary of nested +'s/*'s.)

arr2036:
```
(&User-Name !~ /@([^.]+)(\.([^.]+))+$/)
```

grawity:
Yeah, pretty much same thing just reversed.

Though I don't think you need so many (capture groups) there; I mean, they aren't actually referenced. So `/@[^.]+(\.[^.]+)+$/` would do the same.

arr2036:
Yes, just noticed that after i'd removed the capture groups :)

arr2036:
Yeah, I'm not sure why there are so many capture groups in the stock expressions.  They do have a performance penalty.

grawity:
Hmm, actually, I'm looking at the whole `policy.d/filter` file again, and I just noticed that there _already is_ a separate check to reject multiple dots **before** even reaching this rule.

So maybe this "improvement" wasn't even necessary?

arr2036:
Maybe the multiple dots check isn't necessary ;)

arr2036:
Regular expressions are pretty expensive, the fewer we use the better.

alanbuxey:
its not checking for '..' - thats covered by another rule to catch such a
typoe. its checking that theres at least one dot seperator in the realm
component
ie  @test.example.com  or @example.com  - both of those are fine   @com or
@test is NOT fine.

PR #1571, opened 2016-03-28
tidy rlm_mschap mod_authenticate

mcnewton:
mod_authenticate is soooo long :( and I can't contemplate doing anything else with it until it's been tidied up somewhat. Move some chunks of code out to separate functions.

### DIFF
--- a/src/modules/rlm_mschap/rlm_mschap.c
+++ b/src/modules/rlm_mschap/rlm_mschap.c
@@ -517,7 +517,7 @@ static int _mod_conn_free(struct wbcContext **wb_ctx)
 /*
  *	Create connection pool winbind context
  */
-static void *mod_conn_create(TALLOC_CTX *ctx, UNUSED void *instance, UNUSED struct timeval *timeout)
+static void *mod_conn_create(TALLOC_CTX *ctx, UNUSED void *instance, UNUSED struct timeval const *timeout)
 {
 	struct wbcContext **wb_ctx;
 
--- a/src/modules/rlm_mschap/rlm_mschap.c
+++ b/src/modules/rlm_mschap/rlm_mschap.c
@@ -1478,6 +1478,143 @@ static rlm_rcode_t mschap_error(rlm_mschap_t *inst, REQUEST *request, unsigned c
 	return rcode;
 }
 
+
+/*
+ *	find_nt_password() - try and find a correct NT-Password
+ *	attribute, or calculate one if possible.
+ */
+static bool CC_HINT(nonnull (1, 2, 4)) find_nt_password(rlm_mschap_t *inst,
+							REQUEST *request,
+							VALUE_PAIR *password,
+							VALUE_PAIR **ntpw)
+{
+	VALUE_PAIR *nt_password;
+
+	/*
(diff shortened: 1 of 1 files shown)


src/modules/rlm_mschap/rlm_mschap.c
```diff
@@ -1483,10 +1483,10 @@ static rlm_rcode_t mschap_error(rlm_mschap_t *inst, REQUEST *request, unsigned c
  *	find_nt_password() - try and find a correct NT-Password
  *	attribute, or calculate one if possible.
  */
-static bool CC_HINT(nonnull (1, 2, 4)) find_nt_password(rlm_mschap_t *inst,
-							REQUEST *request,
-							VALUE_PAIR *password,
-							VALUE_PAIR **ntpw)
+static inline bool CC_HINT(nonnull (1, 2, 4)) find_nt_password(rlm_mschap_t *inst,
```
alandekok: inline shouldn't be used.  C compilers are generally smart enough to do the right thing.

alandekok:
The "inline" patch should just be removed.  It isn't necesssary.  We use "inline" only rarely... if at all.

mcnewton:
OK, thanks - done.

arr2036:
Compiler will only inline things with optimization turned on, which means there is no inlining for developer builds unless it's explicitly declared, but that won't really make a difference.

Is there a specific argument against inlining?

src/modules/rlm_mschap/rlm_mschap.c
```diff
@@ -1641,29 +1585,132 @@ static rlm_rcode_t CC_HINT(nonnull) mod_authenticate(void *instance, REQUEST *re
...
+ *	attributes and configuration.
+ *	We will try to find out password in configuration
+ *	or in configured passwd file.
+ *	If one is found we will check paraneters given by NAS.
+ *
+ *	If PW_SMB_ACCOUNT_CTRL is not set to ACB_PWNOTREQ we must have
+ *	one of:
+ *		PAP:      PW_USER_PASSWORD or
+ *		MS-CHAP:  PW_MSCHAP_CHALLENGE and PW_MSCHAP_RESPONSE or
+ *		MS-CHAP2: PW_MSCHAP_CHALLENGE and PW_MSCHAP2_RESPONSE
```
arr2036: Could this be doxygenised?

src/modules/rlm_mschap/rlm_mschap.c
```diff
@@ -1599,23 +1529,37 @@ static rlm_rcode_t CC_HINT(nonnull) mod_authenticate(void *instance, REQUEST *re
...
 
-	/*
-	 *	Or an LM-Password.
-	 */
+	*ntpw = nt_password;
+	return true;
+}
+
+
+/*
```
arr2036: Same here

src/modules/rlm_mschap/rlm_mschap.c
mcnewton: Probably... that's not my comment ;) (pretty much none of it is mine - I just moved stuff around).

I'll take a look if I get a moment.

PR #1573, opened 2016-03-31
Regex fr_pair_make fix

0xbad0c0d3:
Why we ignore value when expecting regex? If value ignored then fr_pair_cmp() will fail in line 1988 because a->vp_strvalue has NULL value

### DIFF
--- a/src/lib/pair.c
+++ b/src/lib/pair.c
@@ -1519,8 +1519,8 @@ VALUE_PAIR *fr_pair_make(TALLOC_CTX *ctx, VALUE_PAIR **vps,
 			talloc_free(vp);
 			return NULL;
 		}
-
-		value = NULL;	/* ignore it */
+		/* Ignore? But what about radclient checking oO We use strvalue but if we ignore it - all checks fails!*/
+		//value = NULL;	/* ignore it */
 		break;
 #endif
 	}


alandekok:
I'm not sure why this is necessary.  The `value` variable is local to the function, and can be set to `NULL` without affecting fr_pair_cmp().

Do you have a test case that shows a problem?  i.e. it should work, but doesn't?

0xbad0c0d3:
Yes, i have test it. You right, it is local variable, but... Look at line 1590 if local value is NULL than fr_pair_value_from_str() doesn't call, if it's doesn't call - value_data_from_str() from value.c doesn't call too and strvalue can't be allocated.

alandekok:
You're not answering my question.

The `pair_mark_xlat()` function copies the contents of `value` into the `VALUE_PAIR`.  That's what it's supposed to do.

The `vp->vp_strvalue` field is used for STRINGS.  Not regular expressions.

Again, *what problem does this cause?*  I can read the code and see what it's doing.  But so far as I can tell it's working as intended.

Please show that this causes a real world problem.  If you can't, I'll close the bug as a misunderstanding of how the server is supposed to work.

0xbad0c0d3:
ok. if 
> The vp->vp_strvalue field is used for STRINGS. Not regular expressions.

why in fr_pair_cmp() for regex (T_OP_REG_NE and T_OP_REG_EQ) used vp_strvalue nor a->value->xlat? and
`#define vp_strvalue	data.strvalue`

If strval doesn't used for regex current PR - wrong and solution in modification fr_pair_cmp() to use right field - xlat

alandekok:
That is interesting, but again... does this cause a problem?

Github issues are intended for people to submit bug reports (with USEFUL information).  They're not intended to ask questions about how the server works.  If you have questions, please use the freeradius-devel mailing list.

0xbad0c0d3:
No, if we should use xlat - current PR cause problem, but in a wrong way.

alandekok:
So... you can't say where this causes real-world problems.

PR #1577, opened 2016-04-06
Re-shape SSL_CTL storage, force OpenSSL to reload CApath contents periodically

lytboris:
OpenSSL does not reload contents of ca_path dir over time.
That means that if check_crl is enabled and CRLs are loaded from ca_path dir, at some point CRLs will expire and RADIUSd will stop authenticating users. If ca_path_reload_interval is non-zero, it will force OpenSSL to reload all data from ca_path periodically.

Commit  9bca727 replaces SSL_CTL *\* with a struct that can hold all per-context additional fields.
Commit  062f066 actually adds CA path reloading feature

### DIFF
--- a/raddb/mods-available/eap
+++ b/raddb/mods-available/eap
@@ -283,6 +283,16 @@ eap {
 
 		ca_path = ${cadir}
 
+		# OpenSSL does not reload contents of ca_path dir over time.
+		# That means that if check_crl is enabled and CRLs are loaded
+		# from ca_path dir, at some point CRLs will expire and
+		# RADIUSd will stop authenticating users.
+		# If ca_path_reload_interval is non-zero, it will force OpenSSL
+		# to reload all data from ca_path periodically
+		#
+		# Flush ca_path each hour
+		ca_path_reload_interval = 3600
+
 		#
 		#  If check_cert_issuer is set, the value will be checked
 		#  against the DN of the issuer in the client certificate.  If
--- a/raddb/sites-available/tls
+++ b/raddb/sites-available/tls
@@ -162,6 +162,16 @@ listen {
 	#	check_crl = yes
 		ca_path = ${cadir}
 
+		# OpenSSL does not reload contents of ca_path dir over time.
+		# That means that if check_crl is enabled and CRLs are loaded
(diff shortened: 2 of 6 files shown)


lytboris:
Additionally I have a patch to backport d9c4e68 into 3.0.x

arr2036:
With OpenSSL 1.1.0 this patch should increase performance even more for the X509_STORE objects.  I have no major issues with what you've done, just need to review it properly, then I'll pull the commits across.

lytboris:
>This branch has conflicts that must be resolved
Heh. Should I spend some time to fix conflicts now or more changes are coming?

arr2036:
No more changes.  It was to support OCSP stapling.  The other OCSP stapling X509_STORE might benefit from the same treatment.

lytboris:
Cleaned up & merged

lytboris:
Any news?

arr2036:
The reason why this is on hold, is because the restructuring in 4.0.x.  We're likely moving to thread dependent SSL contexts which makes this significantly easier to implement.

lytboris:
Any ideas when this can be pulled in? Should I make a separate pull request for 3.0.x while 3.1 & 4.0 settle down?

alandekok:
I need to find time to review it.  I'd strongly prefer to leave 3.0 alone, unless the feature is hugely useful.  This one seems to fit that requirement, tho.

If this patch can be applied to 4.0, it can just be pulled over via a cherry-pick.  This section of code isn't changing much in 4.0, for at least the next few months.

alanbuxey:
I'd go for 4.x for this - its a change of behaviour that might be an issue
on some systems but it also provides a good reason for
certain installs to be looking at 4.x installs in the future rather than
camp out on 3.x for the next 5 or 10 years :/

lytboris:
> its a change of behaviour
That can be disabled by default and enabled at-will.

nwkrad:

nwkrad:

arr2036:
test this please

mattrose:
retest this please

mattrose:
retest this please

Issue #1572, opened 2016-03-31
Undefined reference errors in tls.c during compilation


tdebaets:
When I try to compile branch 3.0.x on CentOS 5, the compilation fails with these linker errors:

```
build/objs/src/main/tls.o: In function `tls_application_data':
/root/freeradius-server-3.0.x/src/main/tls.c:3295: undefined reference to `tls_error_io_log'
build/objs/src/main/tls.o: In function `tls_init_ctx':
/root/freeradius-server-3.0.x/src/main/tls.c:2654: undefined reference to `tls_error_log'
/root/freeradius-server-3.0.x/src/main/tls.c:2672: undefined reference to `tls_error_log'
/root/freeradius-server-3.0.x/src/main/tls.c:2839: undefined reference to `tls_error_log'
/root/freeradius-server-3.0.x/src/main/tls.c:2849: undefined reference to `tls_error_log'
/root/freeradius-server-3.0.x/src/main/tls.c:2663: undefined reference to `tls_error_log'
build/objs/src/main/tls.o:/root/freeradius-server-3.0.x/src/main/tls.c:2488: more undefined references to `tls_error_log' follow
build/objs/src/main/tls.o: In function `tls_handshake_send':
/root/freeradius-server-3.0.x/src/main/tls.c:760: undefined reference to `tls_error_io_log'
build/objs/src/main/tls.o: In function `tls_handshake_recv':
/root/freeradius-server-3.0.x/src/main/tls.c:681: undefined reference to `tls_error_io_log'
/root/freeradius-server-3.0.x/src/main/tls.c:716: undefined reference to `tls_error_log'
build/objs/src/main/tls.o: In function `tls_new_session':
/root/freeradius-server-3.0.x/src/main/tls.c:557: undefined reference to `tls_error_log'
build/objs/src/main/tls.o: In function `tls_new_client_session':
/root/freeradius-server-3.0.x/src/main/tls.c:506: undefined reference to `tls_error_io_log'
collect2: ld returned 1 exit status
make: *** [build/bin/local/radeapclient] Error 1
```

I have no clue why the linker is unable to find the tls_error_log and tls_error_io_log functions, as these are just defined in the same tls.c file. This was probably introduced by commit 8d8234b. Could someone have a look? Thanks!

alandekok:
You probably need to re-run "configure".

tdebaets:
[...]

PR #1586, opened 2016-04-20
Fix build when PSK_MAX_IDENTITY_LEN is not defined

lytboris:


alandekok:
Commit 696e40de already fixed that.  Why this patch?

lytboris:
Haven't seen your commit. Arran merged only comment to #endif effectively: e6c143af015872f3bd393f108c25cdad227bc4d3

alandekok:
You should have received an email from #1572, which you've commented on. If you look at the commit history for v3.0.x, you'll see Arran merged two commits.

lytboris:
> You should have received an email from #1572
I did receive a message from mcnewton, but not notice about your commit with reference (not even in spam).

Spent a couple of hours tracking this down in vain :(

lytboris:
Why not to merge a9d4798?

PR #1587, opened 2016-04-20
Feature: maximum FD count & idle timeout for file logging facilities can be configured

lytboris:
Cope with "Too many different filenames" logging issue properly.
64 open log files are not enough for everybody :)
Especially, if you use NAS IP(v6) address as dir name and you have many NASes.

### DIFF
--- a/raddb/mods-available/detail
+++ b/raddb/mods-available/detail
@@ -97,4 +97,14 @@ detail {
 		# User-Password
 	#}
 
+	limit {
+		# By default FreeRADIUS will allow to open upto 64 log files
+		# Change it if it is not enough (capped at 2048)
+		# max_files = 64
+
+		# A log file that is not being used to write into will be
+		# closed after idle_timeout seconds (default 30)
+		# idle_timeout = 30
+	}
+
 }
--- a/raddb/mods-available/linelog
+++ b/raddb/mods-available/linelog
@@ -93,6 +93,16 @@ linelog {
 		#  set this to "yes".
 		#
 		escape_filenames = no
+
+		limit {
+			# By default FreeRADIUS will allow to open upto 64 log files
+			# Change it if it is not enough (capped at 2048)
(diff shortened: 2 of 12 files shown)


alandekok:
It would be good to fix the travis build issues.  They don't show up in the current v3.1.x HEAD.

lytboris:
Will do for sure. Would an adopted version go to 3.0.x branch? We had a biiiig bada-boom because of 64 being hardcoded/

alandekok:
The `max_open_logfiles` and `log_idle_timeout` configurations should probably be in a separate sub-section.  e.g.

   limit {
        max_files = 20
        idle_timeout = 30
   }

That way it's a bit clearer what's going on.  We're trying to move towards putting common configuration into common sub-sections.  So the configuration isn't full of `foo_bar_baz`, and `foo_thing_bar` and `foo_egg_drop`.

alandekok:
I'm wary of putting this into v3.0.x.  I'm Ok with upping the limit of 64 there.  But we're trying to keep v3.0.x stable with minimal changes, and no new features.

lytboris:
I will add code to `close oldest file if we have array saturated with opened files and need to open a new one` to both branches and `bump 64 to 256 limit` for 3.0.x.

alandekok:
For v3.0.x, you can just close the oldest one if the server needs a new one.  And bump the limit to 256.

The exfiles API caches files for speed.  It's fine to close an old file, as if that file is needed later, it will just be re-opened.

lytboris:
That is exactly what I was going to do. Closing the oldest file when no empty slot available is needed in 3.1 too obviously.

lytboris:
Alan, should I update this pull request somehow to be merged or you're going to re-write it your way?

alandekok:
I've pushed all changes to v3.0.x, and am working on v3.1.x.

PR #1588, opened 2016-04-21
Do not fail logging when hitting open log file limit

lytboris:
Additionally do not fail when hash collides, there is nothing bad in it, just a bit slower to lookup file.


alandekok:
I've pushed commits to v3.0.x which fix the issue.  I had some comments and your commits, and it was about as easy to just write the code as to comment on it.

See 9e9af34ea0dc0, 7508435060bf87, and 20818ca1

lytboris:
You left one thing out of scope though - https://github.com/FreeRADIUS/freeradius-server/pull/1587/commits/74f11312a945ec1a3b686bd3938bf11a16362aa3

alandekok:
Fixed by 4de73a2

PR #1590, opened 2016-04-25
Fix SIGSEGV on exit when an EAP method is referenced (v3.1.x)

lytboris:
in multiple eap stansas, e.g.
===========================
eap instance1 {
	...
	md5 {
	}
	...
}
eap instance2 {
	...
	md5 {
	}
	...
}
===========================

Sponsored by: Yandex LLC

### DIFF
--- a/src/modules/rlm_eap/eap.c
+++ b/src/modules/rlm_eap/eap.c
@@ -72,6 +72,13 @@ static char const *eap_codes[] = {
 static int _eap_module_free(eap_module_t *inst)
 {
 	/*
+	 * Check if handle is still valid. If not, type is referencing freed memory
+	 */
+
+	if (!inst->handle)
+		return 0;
+
+	/*
 	 *	We have to check inst->type as it's only allocated
 	 *	if we loaded the eap method.
 	 */
@@ -83,7 +90,7 @@ static int _eap_module_free(eap_module_t *inst)
 	 *	ad it removes the symbols needed by valgrind.
 	 */
 #else
-	if (inst->handle) dlclose(inst->handle);
+	dlclose(inst->handle);
 #endif
 
 	return 0;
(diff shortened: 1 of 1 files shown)


src/modules/rlm_eap/eap.c
```diff
@@ -72,6 +72,13 @@ static char const *eap_codes[] = {
 static int _eap_module_free(eap_module_t *inst)
 {
 	/*
+	 * Check if handle is still valid. If not, type is referencing freed memory
+	 */
+
+	if (!inst->handle)
```
arr2036: I know it seems incredibly anal, but could you put it all on one line or use braces.  http://wiki.freeradius.org/contributing/coding-standards#code-style_always-collapse-bodies-where-possible

alandekok:
See the comments for #1590.  This patch isn't necessary.

Something else is smashing memory.

lytboris:
Which comments are you referring? This is #1590 and thare are no comments about smashing memory.

More, without this patch I got SIGSEGV, with it - no crashes.

alandekok:
> On Apr 25, 2016, at 1:26 PM, Boris Lytochkin <notifications@github.com> wrote:
> 
> Which comments are you referring? This is #1590 and thare are no comments about smashing memory.

  What I means is that if `inst->handle` is NULL, then something *else* is smashing the `inst` structure.  When you have 2 EAP instances, if one calls `dlclose()` on EAP-MD5 (for example), the other module will still have a *valid pointer* to rlm_eap_md5.  i.e. module 1 does `dlclose(inst->handle)`.  Module 2 has `inst->handle` which is *still a valid pointer*, and which *still points to valid memory*.

  There is *no way* for module 1 to do `dlclose(inst->handle)`, and then have module 2's `inst->handle` suddenly become `NULL`.

  If that happens, your system is broken, and module 2's `inst` structure is being smashed by something.

  As I said, the dlhandles are reference counted.  You can (and SHOULD) call `dlclose()` as many times as you call `dlopen()`

> More, without this patch I got SIGSEGV, with it - no crashes.

  Then find out why it's crashing.

  This patch does not fix any underlying bug.

lytboris:
Yep, I got the point.

lytboris:
Please reopen this pull request. See #1590 for details.

PR #1591, opened 2016-04-25
Fix SIGSEGV on exit when an EAP method is referenced (v3.0.x)

lytboris:
in multiple eap stansas, e.g.
===========================
eap instance1 {
	...
	md5 {
	}
	...
}
eap instance2 {
	...
	md5 {
	}
	...
}
===========================

Sponsored by: Yandex LLC

### DIFF
--- a/src/modules/rlm_eap/eap.c
+++ b/src/modules/rlm_eap/eap.c
@@ -74,8 +74,7 @@ static int _eap_module_free(eap_module_t *inst)
 	/*
 	 * Check if handle is still valid. If not, type is referencing freed memory
 	 */
-        if (!inst->handle)
-		return 0;
+        if (!inst->handle) return 0;
 
 	/*
 	 *	We have to check inst->type as it's only allocated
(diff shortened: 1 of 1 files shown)


alandekok:
I don't see why this is necessary.  The dlopen / dlclose API takes care of reference counting the handles.  From the "man" page for dlopen:

    A second call to dlopen() with the same path will return the same handle, but the internal reference
    count for the handle will be incremented.  Therefore all dlopen() calls should be balanced with a
    dlclose() call.

In this case, if `inst->handle == NULL`, then something *else* has gone through and smashed the `inst` structure.

lytboris:
I tracked it down. My patch IS correct. Take a look on
https://github.com/FreeRADIUS/freeradius-server/blob/v3.1.x/src/modules/rlm_eap/eap.c#L126

When module is loaded, no dlopen() is invoked and handle is left as 0x0. The only thing to fix for this patch is to fix the comment.

arr2036:
Heh yeah, I guess that was added to support static compilation of EAP methods into rlm_eap.

arr2036:
I think we still need to call the detach method though, that's probably per rlm_eap instance.  If we don't call it we leak memory on HUP

lytboris:
detach pointer may reference freed memory when real rlm_eap_... method was unloaded previously invalidating type pointer.

PR #1593, opened 2016-04-25
Misc changes

pwdng:
Extract linelog structs to separate header file so that the linelog module can be called from within other modules.
Add second & microsecond single letter expansions

### DIFF
--- a/src/main/xlat.c
+++ b/src/main/xlat.c
@@ -1505,7 +1505,7 @@ static ssize_t xlat_tokenize_literal(TALLOC_CTX *ctx, char *fmt, xlat_exp_t **he
 			ssize_t slen;
 			xlat_exp_t *next;
 
-			if (!p[1] || !strchr("%}dlmntDGHISTYv", p[1])) {
+			if (!p[1] || !strchr("%}dlmnstDGHIMSTYv", p[1])) {
 				talloc_free(node);
 				*error = "Invalid variable expansion";
 				p++;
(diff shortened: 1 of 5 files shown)


src/main/xlat.c
```diff
@@ -2216,6 +2219,11 @@ static char *xlat_aprint(TALLOC_CTX *ctx, REQUEST *request, xlat_exp_t const * c
 			snprintf(str, freespace, "%" PRIu64 , request->seq_start);
 			break;
 
+		case 'e': /* Request second */
```
arr2036: If it really is UTC request second, we already have '%l'

src/modules/rlm_sqlhpwippool/rlm_sqlhpwippool.c
```diff
@@ -280,7 +280,7 @@ static int mod_instantiate(CONF_SECTION *conf, void *instance)
 	}
 
 	/* check if the given instance is really a rlm_sql instance */
-	if (strcmp(sql_inst->module->name, "rlm_sql") != 0) {
+	if (strcmp(sql_inst->module->name, "sql") != 0) {
```
arr2036: Ah, yes.  I think the names changed at some point to drop the rlm.

arr2036:
For linelog, is there a specific reason to extract the structs? I don't think they're used outside of the module so there's not much point.

In instances where we have a module header file, it's usually because the module is split into frontend and backend.

i.e. rlm_<foo>.c calls functions in foo.c, foo_bar.c etc...

So rlm_<foo>.c implements the actual logic, and then the other source files are helpers for whatever library is being used.

src/main/xlat.c
pwdng: %e would be for the actual seconds part of the event timestamp (just like we have %H and %G for the hours and minutes), not the unix timestamp.

arr2036:
Cherry-Picked the first two commits across from your repo.

PR #1589, opened 2016-04-25
Fix seg fault on rlm_python when multiple instances are configured

gyx1000:
If not the first configured instance of a rlm_python module is called first a segmentation fault occurs.

PR #1594, opened 2016-04-27
Fix radiusd reference counting on Init #1589

gyx1000:
Because `Py_InitModule3` return a Borrowed reference, we should not DecRef radiusd module on `mod_detach` but because we can have more than one Interpreter, we should convert the Borrowed to a new Owned reference with a simple "IncRef"

https://docs.python.org/2/extending/extending.html#reference-counting-in-python

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -853,6 +853,8 @@ static int python_interpreter_init(rlm_python_t *inst, CONF_SECTION *conf)
 		 *	Initialise a new module, with our default methods
 		 */
 		inst->module = Py_InitModule3("radiusd", module_methods, "FreeRADIUS python module");
+		Py_IncRef(inst->module);
+		
 		if (!inst->module) {
 		error:
 			python_error_log();


arr2036:
reference count should start at 1, or at least it does in every other reference counting system i've encountered.

gyx1000:
From the python doc : The borrower of a reference should not call Py_DECREF()
In `rlm_python.c`, a DecRef on this reference is called:
https://github.com/FreeRADIUS/freeradius-server/blob/v3.1.x/src/modules/rlm_python/rlm_python.c#L986

A borrowed reference can be changed in a owned (ref count 1) with a simple IncRef call

arr2036:
Ah I see "The function PyImport_AddModule() also returns a borrowed reference even though it may actually create the object it returns: this is possible because an owned reference to the object is stored in sys.modules."

arr2036:
You put it before the NULL check though, which is wrong

arr2036:
And actually we call ``PyImport_ImportModule`` which returns a new reference not a borrowed one...

arr2036:
and yes... Py_InitModule3 also returns a borrowed reference, just checked.

PR #1596, opened 2016-04-28
Fix rlm python segfault detach

gyx1000:


arr2036:
Can you provide the code for your TLS test?

arr2036:
OK, I found a few issues.  I didn't realise you were only testing with the test framework (which you may not have been, can you clarify?).

The server binary (radiusd) calls functions in the server library (libfreeradius-server).  This allows us to setup test harnesses around large blocks of server code, which is what unittest.c is.

The notable differences between radiusd and unittest are that unittest does not run with threads, at all.  So the TLS tests aren't really valid there.

Because unittest doesn't run with threads, it also means the atExit() destructors never fire before we start to destroy the module instances.

I've added an explicit call to destroy the thread state tree which should fix that.  The python tests now pass.

I'm going to close this and we can continue testing in #1589

PR #1600, opened 2016-05-05
Use radiusd constants and comment rlm_python tests

gyx1000:


gyx1000:
I think that my test "auth_with_tls" is not really sure because the fact that `pmod1` and `pmod3_withmod1` have their own interpreters, any attributes are already independents.

Do you think that enable cext_compat on this two modules will "emulate" better a multi-thread context for unittest ?

arr2036:
Yes, it'll be distinct.  Even if the same interpreter is used, rlm_python will create a thread state per thread per module instance.  It's useful to have the tests just to make sure nothing crashes :)

arr2036:
Could you make a similar pull request for v3.0.x.  The patches fail to pull across cleanly, and I don't have any time to investigate why today.  Thanks!

gyx1000:
Sure, I will do. I see that when you backport #1589 you don't backport python test folder entirely (Only *pyc". I will make a PR for the full folder content.

arr2036:
ah, ok, that's what it was. Thanks.

PR #1601, opened 2016-05-05
rlm_python tests

gyx1000:


gyx1000:
v3.0.x python modules.conf differs a little from v3.1.x. I have set the cext_compat on all module because cext_compat is true by default on v3.0.x

arr2036:
Thanks!

Issue #1464, opened 2015-12-19
Pass %RAD_CONFIG and %RAD_REPLY to python modules in rlm_python


alejandro-perez:
Hi,

in rlm_pyhon, modules only have access to RAD_REQUEST. This is generally ok, but sometimes you need access to RAD_REPLY or even RAD_CONFIG (indeed, that's the modules's interface, so why would not do it?)

I have seen that rlm_perl does provide them all to sub-modules.

Thanks

herwinw:
Don't forget RAD_STATE, RAD_REQUEST_PROXY and RAD_REQUEST_PROXY_REPLY that are available in Perl as well. While the architecture of `rlm_perl` makes it pretty easy to add another list to the input (global variables), `rlm_python` works with parameters that are passed to a function. This means we'd have to change the prototype of the function (or use the `*args, **kwargs`-feature).

I'm working on a proposal to change the interface of all the interfaces of the embedded languages, which includes this feature. I hope to have it finished this weekend.

alejandro-perez:
Or you can change the "p" parameter from tuple list to map of tuple lists.
p["REQUEST"] = what we already have
p["REPLY"] = ...
p["RAD_STATE"] = ....

alejandro-perez:
Hi, is this moving forward? 
In case it is not...is there any effective way by which I can know, from within a python module, whether the request is being proxied or not?
Thanks

alandekok:
When you open an issue, there is no guarantee that someone will jump on it, and immediately issue a fix.

This is open source software.  You're welcome to submit a patch.

alejandro-perez:
Hi Alan,

sorry if my comment sounded disrespectful, that was way far from being my intention. I was asking just because he mentioned that he might do it and I wanted to know if he was on the process. I know how opensource works, and I never mean to make any exigency. Again, sorry if it seemed to be the case. It wasn't.

Regards

gyx1000:
I would like try to fix this issue. The problem is that we can't simply pass new arguments because we will introduce a break if python functions (authorize, authenticate, ...) doesn't assume it.

@arr2036 @alandekok What do you think about inspect the python functions before calling it and "inject" params according to the python functions signatures ?

arr2036:
Could be interesting.  Do you know if it's possible to do lazy martialling and unmartialling in python? i.e. intercept hash operations (get, set, remove) and call out to C functions?

gyx1000:
[...]

PR #1609, opened 2016-05-10
Give python function access to more data #1464

gyx1000:


nwkrad:

nwkrad:

src/modules/rlm_python/rlm_python.c
```diff
@@ -445,46 +442,117 @@ static rlm_rcode_t do_python_single(REQUEST *request, PyObject *pFunc, char cons
...
-		     vp;
-		     vp = fr_cursor_next(&cursor), i++) {
+		for (vp = fr_cursor_init(&cursor, vps);
+			 vp;
+			 vp = fr_cursor_next(&cursor), i++) {
 			PyObject *pp;
 
 			/* The inside tuple has two only: */
 			if ((pp = PyTuple_New(2)) == NULL) {
+				return -1;
```
alejandro-perez: pArg might not be freed if you return like that. 
Should probably call to Py_XDECREF(pArg)

PR #1615, opened 2016-05-21
raddb: move driver-specific config out from mods-available

mcnewton:
Symlinking them into mods-enabled, the location INCLUDEd in mods-available/sql,
breaks the server as it tries to load rlm_<driver>.so, which doesn't exist.
They're only part of a module config, so don't really fit that well in
mods-available.

Additionally, turn the $INCLUDE into an $-INCLUDE to pull the config in if it
exists, rather than having to be locally uncommented. This makes it easier to
configure sql for the first time (especially to get up and running with sqlite).


mcnewton:
Thanks! Wanted another set of eyes to check that one.

arr2036:
Yeah, not 100% sure it's right, but it's better than what was there :)

mcnewton:
They don't fit too well in either location but this seemed very slightly better.

PR #1618, opened 2016-06-03
Add dictionary values in rlm_rest

qnet-herwin:
Newer and better implementation of #988 

Instead of rewriting the values, add an extra key to the object with the converted values, in case any value had a dictionary value associated with it. Add JSON null in case there was no value found in the dictionary, so the keys `value` and `dict-value` are the same length and missing `dict-value` could always be resolved to a `value` by using the same index.

The improvement over #988 is that this implementation is backwards compatible with the current implementation.

Example
```json
{"Service-Type":{"type":"integer","value":[10,999,2],"dict-value":["Call-Check",null,"Framed-User"]}}
```


qnet-herwin:
A possible future work after commit 8f763a6b21c4c28667842eef1cb422196a02a8e4 is to use the extracted method too in `jsonquote_xlat`. commit 9b6ce401a27eacaa44c0a69570c467c7709df4c8 already made some nice preparations for this by placing them in the same module.

arr2036:
I'm admonishing 2012 me for attempting to write a streaming encoder instead of just using c-json to build the outbound JSON string.  Though in 2012 I didn't expect we'd have machines with 44 physical cores and eight memory channels and 9.6Gt QPI links in four years... It changes the efficiency considerations somewhat.

Honestly we should probably just use the c-json functions. If you feel like reworking rlm_rest just to build the outgoing json structure with the contents of each list in one go, then i'd have no problem merging that change into v4.0.x.  The code will be much much shorter, and far simpler... Though you'll have to redo the post encoder too.

It's looking doubtful that we'll release v3.2.0, and will go straight to v4.0.x. So please resubmit against the v4.0.x.

arr2036:
Needs resubmitting against v4.0.x is this is still an issue.

PR #1622, opened 2016-06-08
src/lib/pair.c: fr_pair_fprint: avoid segfault

lkpdn:
* prevent from calling fputs with null file pointer.
* radclient with -q option, then you can easily come across a segfault.


arr2036:
Can you put the check after the variable declarations please.

lkpdn:
Sorry about that. fixed it. https://github.com/FreeRADIUS/freeradius-server/pull/1622/commits/0a143a1e6883d0fc71f6f4dba2bd84f450ad9e6e .

PR #1625, opened 2016-06-10
Review wanted: add all lists (like reply and control) to rlm_rest

qnet-herwin:
First of all: this is very much work in progress, I mainly want an opinion to see if things in this patch are done in a sane way and if this would be accepted in the release (once finished)

At the moment, `rlm_rest` only sends the `request` list. It can be really useful to have the other lists too (compare  #1464 where this functionality is requested for `rlm_python`). This branch fixes this, but only if the body is of the json-type (as I said before: this is work in progress).
Even though json would support a structure like `{"request:{...attributes...},"reply":{...attributes...}}`, I choose to use the old hash format and prepend every attribute with the list name. This way it would also be possible to use this functionality using post-vars (or any other body type that does not support a more complex structure)

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -58,28 +58,31 @@ rest {
 	#  the settings from the authorize section here will be used.
 	#
 	#  The following config items may be listed in any of the sections:
-	#    uri          - to send the request to.
-	#    proxy        - the request via this server, supports socks/http/https uri
-	#                   and :port.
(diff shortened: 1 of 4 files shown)


qnet-herwin:
Is there any chance that someone wants to review this? It would be nice to have this included in the upcoming 3.2 release. The same goes for #1618

arr2036:
Yes, it extended the old state machine driven, manual, encoder.  Looks like you're working towards the way I suggested, with the new JSON list function, so i'll close this PR.

PR #1628, opened 2016-06-15
Fix fr_channel_drain so that have_read counter gets updated when read…

pwdng:
… stops after header

### DIFF
--- a/src/main/command.c
+++ b/src/main/command.c
@@ -3208,8 +3208,8 @@ static int command_domain_recv_co(rad_listen_t *listener, fr_cs_buffer_t *co)
 	fr_command_table_t *table;
 	uint8_t *command;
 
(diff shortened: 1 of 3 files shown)


src/main/command.c
```diff
@@ -234,8 +234,8 @@ static int fr_server_domain_socket(char const *path, gid_t gid)
 		fr_strerror_printf("Failed determining parent directory");
 	error:
 		talloc_free(dir);
-		if (dir_fd >= 0) close(dir_fd);
-		if (path_fd >= 0) close(path_fd);
+		close(dir_fd);
+		close(path_fd);
 		if (sock_fd >= 0) close(sock_fd);
```
alandekok: Why this change?

pwdng:
Oops, that's a mistake. I'll remove it.

PR #1634, opened 2016-06-27
added 2 unhandled files. rpmbuild now works again

alanbuxey:
example.so and smsotp.so are created but not handled. they are now handled so the RPMs now build again with the spec file

### DIFF
--- a/redhat/freeradius.spec
+++ b/redhat/freeradius.spec
@@ -625,6 +625,7 @@ fi
 %{_libdir}/freeradius/rlm_eap_sim.so
 %{_libdir}/freeradius/rlm_eap_tls.so
 %{_libdir}/freeradius/rlm_eap_ttls.so
+%{_libdir}/freeradius/rlm_example.so
 %{_libdir}/freeradius/rlm_exec.so
 %{_libdir}/freeradius/rlm_expiration.so
 %{_libdir}/freeradius/rlm_expr.so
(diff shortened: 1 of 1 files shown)


mcnewton:
They are already handled at https://github.com/alanbuxey/freeradius-server/blob/a0d53f596cea97bfc88052998b2111451ef61844/redhat/freeradius.spec#L658-L661, so this doesn't look right?

alanbuxey:
thats if experimental modules flag is on - which it isnt with default spec
file -   if the 'experimental' tag has been removed from those modules (as
I suspect) then the if handler for them can be removed from spec file
(those lines

%if %{?_with_experimental_modules:1}%{!?_with_experimental_modules:0}
%{_libdir}/freeradius/rlm_example.so
%{_libdir}/freeradius/rlm_smsotp.so
%endif

can be removed..... )

alanbuxey:
...no experimental by default.... the spec file even removed the
experimental.conf file...

arr2036:
What's the status on this?

alanbuxey:
the default specfile doesn't do 'experimental' - so all stable moduels are
built....example and smsotp are marked as stable..so need to be handled,
(so really SHOULDNT be in the specfile as needing to be handled if
experimental  (rlm_winbind does though!!)

arr2036:
I build both with and without experimental modules on v4.0.x and all was well... So closing this PR.

PR #1638, opened 2016-06-30
Grab request from backlog only if thread is idle

pwdng:


alandekok:
I've pushed a separate fix.

PR #1661, opened 2016-07-20
small fix in dictionary.rfc4849 string should be octets

Cereal2nd:
according to the rfc section2:

The String field is one or more octets. so this is an octets field instead of a string field


alandekok:
No.

The data type (barring the occasional zero) is printable string.

Issue #1662, opened 2016-07-21
systemd unit files should support reload


eheydrick:
# Issue type
- [x] Feature request.
# Defect/Feature description

The systemd unit files should support the reload action so that the daemon can be HUP'd when config files are updated. The SysV init scripts supports reloading and the systemd files should too.

arr2036:
The architectural changes required for HUP to work correctly are massive.  They're only likely to be able in v4.0.x even then you might be waiting until v4.2.x.  Until it can work properly, in a way that everyone expects, FreeRADIUS should be restarted not reloaded.

eheydrick:
I've been using reload with 3.0.x and it *seems* to work. At least I haven't noticed any issues with my usage. Looking forward to proper reload support.

alandekok:
I've pushed a patch for v3.  HUP there is minimal, but it should be allowed.

The larger discussion for v4 should be in a separate issue.

PR #1663, opened 2016-07-21
Add reload support to systemd unit files

eheydrick:
Currently the systemd unit files only support the restart action. This adds reload support so the daemon can be HUP'd when configs are changed. Addresses #1662. Let me know if I should target a different branch.


arr2036:
See #1662

PR #1664, opened 2016-07-22
Fixed compile error in v3.0.x

qnet-herwin:
Commit 7412786d95ac2b39a19ee2f77a29981c9adb86b3 was probably
cherry-picked from v3.1.x, but the variable has a different name in
v3.0.x.


arr2036:
thanks

PR #1666, opened 2016-07-22
Shared method to do JSON escaping of strings

herwinw:
And actually use the functionality of json-c to escape strings, instead of reinventing the wheel.

This is one of the first steps mentioned in https://github.com/FreeRADIUS/freeradius-server/pull/1618#issuecomment-231619422


arr2036:
good catch :)

PR #1678, opened 2016-08-16
Create JSON representation of lists via json-c

qnet-herwin:
This might be a first step to implement https://github.com/FreeRADIUS/freeradius-server/pull/1618#issuecomment-231619422 It contains a method to create a JSON-string from a list. It is nowhere used yet, but it's a nice start to use this in rlm_rest.

### DIFF
--- a/src/modules/rlm_json/json.c
+++ b/src/modules/rlm_json/json.c
@@ -252,3 +252,83 @@ void fr_json_version_print(void)
 #endif
 }
 
+static void json_array_add_vp(TALLOC_CTX *ctx, json_object *arr, VALUE_PAIR *vp) {
+	json_object *to_add = NULL;
+	char *stringified_value;
+
+	switch (vp->da->type) {
+		case PW_TYPE_INTEGER:
+			to_add = json_object_new_int(vp->vp_integer);
+			break;
+		case PW_TYPE_SHORT:
+			to_add = json_object_new_int(vp->vp_short);
+			break;
+		case PW_TYPE_BYTE:
+			to_add = json_object_new_int(vp->vp_byte);
+			break;
+		case PW_TYPE_SIGNED:
+			to_add = json_object_new_int(vp->vp_signed);
+			break;
+		case PW_TYPE_INTEGER64:
+			to_add = json_object_new_int64(vp->vp_integer64);
+			break;
+		case PW_TYPE_BOOLEAN:
+			to_add = json_object_new_boolean(vp->vp_byte);
+			break;
+		default:
+			MEM(stringified_value = fr_pair_value_asprint(ctx, vp, '\0'));
+			to_add = json_object_new_string(stringified_value);
+			talloc_free(stringified_value);
+			break;
+	}
+	MEM(to_add);
+	json_object_array_add(arr, to_add);
+}
+
+const char *fr_json_from_pair_list(VALUE_PAIR **vps, const char *prefix) {
+	TALLOC_CTX *ctx;
+	vp_cursor_t cursor;
+	struct json_object *obj, *vp_object, *values, *type_name;
+	VALUE_PAIR *vp;
+	const char *name_with_prefix;
+	const char *res = NULL;
+
+	MEM(ctx = talloc_pool(NULL, 1024));
+	MEM(obj = json_object_new_object());
+
+	for (vp = fr_cursor_init(&cursor, vps); vp; vp = fr_cursor_next(&cursor)) {
+		VERIFY_VP(vp);
+
+		if (prefix) {
+			MEM(name_with_prefix = talloc_asprintf(ctx, "%s:%s", prefix, vp->da->name));
+		} else {
+			name_with_prefix = vp->da->name;
+		}
+
+		if (json_object_object_get_ex(obj, name_with_prefix, &vp_object)) {
+			if (!json_object_object_get_ex(vp_object, "value", &values)) {
+				ERROR("Something is broken in the rlm_json encoder");
+				goto error;
+			}
+		} else {
+			MEM(vp_object = json_object_new_object());
+			json_object_object_add(obj, name_with_prefix, vp_object);
+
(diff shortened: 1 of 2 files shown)


src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
...
+			to_add = json_object_new_boolean(vp->vp_byte);
+			break;
+		default:
+			len = vp->data.length+1;
+			tmp = NULL;
+			do {
+				if (tmp) talloc_free(tmp);
+				tmp = talloc_array(ctx, char, len);
+				if (!tmp) return NULL;
+				outlen = fr_pair_value_snprint(tmp, len, vp, 0);
```
arr2036: Use fr_pair_value_asprint, then you don't need the talloc calls.

arr2036: Not sure what the loop is for here either?

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
...
+			break;
+	}
+
+	if (to_add)
+		json_object_array_add(arr, to_add);
+
+	return to_add;
+}
+
+const char *fr_json_vps_to_json(TALLOC_CTX *ctx, VALUE_PAIR **vps, const char *prefix) {
```
arr2036: fr_pair_list_to_json

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
...
+	VALUE_PAIR *vp, *vp_next;
+	char *name_with_prefix;
+	const char *res = NULL;
+
+	fr_pair_list_sort(vps, fr_pair_cmp_by_da_tag);
+	vp = fr_cursor_init(&cursor, vps);
+
+	obj = json_object_new_object();
+	if (!obj) return NULL;
+	while (vp) {
```
arr2036: We usually write these as

for (vp = fr_cursor_init(&cursor, vps); vp; vp = fr_cursor_next(&cursor))

At least in current code... Some of the rest stuff was from 2012 when I was first learning C

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
...
+		vp_object = json_object_new_object();
+		if (!vp_object) goto fr_json_vps_to_json_failure;
+
+		if (prefix) {
+			name_with_prefix = talloc_zero_array(ctx, char, strlen(prefix) + strlen(vp->da->name) + 2);
+			if (!name_with_prefix) {
+				json_object_put(vp_object);
+				goto fr_json_vps_to_json_failure;
+			}
+			sprintf(name_with_prefix, "%s:%s", prefix, vp->da->name);
```
arr2036: just use talloc_asprintf, much easier

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
 #endif
 }
 
+static json_object *add_vp_to_json_array(TALLOC_CTX *ctx, json_object *arr, VALUE_PAIR *vp) {
```
arr2036: verb last, file first, so

	json_array_add_vp

There is a document on the wiki.  If it's unclear in parts, let me know and i'll add more examples.

src/modules/rlm_json/json.c
arr2036: That's why we use from and to, so you can have function names that make sense without duplicating bits

src/modules/rlm_json/json.c
arr2036: Also these only fail on memory allocation errors, so wrapping with ``MEM()`` is sufficient.

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,105 @@ void fr_json_version_print(void)
...
+		} else {
+			json_object_object_add(obj, vp->da->name, vp_object);
+		}
+
+		type_name = json_object_new_string(fr_int2str(dict_attr_types, vp->da->type, "<INVALID>"));
+		if (!type_name) goto fr_json_vps_to_json_failure;
+		json_object_object_add(vp_object, "type", type_name);
+
+		values = json_object_new_array();
+		if (!values) goto fr_json_vps_to_json_failure;
```
arr2036: Same here. The general rule is, bail early on allocation errors unless you're in libfreeradius

qnet-herwin:
Pushed a new version of the patch, especially the `MEM` macro cleaned up a lot of the code. The logic for attributes that occured more than once has also been updated, which saves us from the loop in the loop.

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,83 @@ void fr_json_version_print(void)
...
+
+			MEM(values = json_object_new_array());
+			json_object_object_add(vp_object, "value", values);
+		}
+
+		json_array_add_vp(ctx, values, vp);
+	}
+	res = json_object_to_json_string_ext(obj, JSON_C_TO_STRING_PLAIN);
+
+rf_json_from_pair_list_failure:
```
arr2036: usually just use ``error:`` but not a huge issue

src/modules/rlm_json/json.c
```diff
@@ -252,3 +252,83 @@ void fr_json_version_print(void)
...
+
+const char *fr_json_from_pair_list(VALUE_PAIR **vps, const char *prefix) {
+	TALLOC_CTX *ctx;
+	vp_cursor_t cursor;
+	struct json_object *obj, *vp_object, *values, *type_name;
+	VALUE_PAIR *vp;
+	const char *name_with_prefix;
+	const char *res = NULL;
+
+	MEM(ctx = talloc_init("fr_json_from_pair_list"));
```
arr2036: Could allocate a 1k pool here  for efficiency

qnet-herwin:
New version pushed, with those two remarks processed

PR #1679, opened 2016-08-19
Request for comments: systemd startup notification and watchdog support

terryburton:
P.o.C. patches against 3.1.x since that's what I'm running but will resubmit for 4.0.x later if applicable.

I'm initially wanting to find out whether patches for systemd features such as these will be received.

The first patch adds the AC framework for detecting and linking systemd. Must be explicitly enabled during ./configure via --with-systemd.  It adds trivial support for informing systemd of successful startup by a notification emitting from the parent once the child has signalled successful initialisation.

The second patch extends this by adding [watchdog notifications](http://0pointer.de/blog/projects/watchdog.html) to the main loop. Currently all this does it to ensure that the main loop is running. To be useful we should probably suppress and reschedule watchdogs when we detect critical problems such as when all threads are hung, etc. (Accepting that problems such as waiting on external resources may be be cleared by the supervisor simply restarting the daemon.)

If the concept of systemd integration isn't in itself an anathema then I'll see if I can extend the watchdog support to cover some failure cases.


terryburton:
Ignore 8724b0ecc7da61cb8c1dfa244a9ef81352744b10.

mcnewton:
This seems reasonable to me, and should be good for recent Debian/Linux distros using systemd, assuming it doesn't affect systems that don't use systemd. There's configure checks for that, so it should be OK as far as I can tell.

Minor disclaimer; I'm sitting next to @terryburton and have been discussing this directly, but didn't want to push in case there's large argument against any internal systemd support... (he's come from never touching FreeRADIUS before two weeks ago, to building a complete working config here and writing patches :-) ).

mcnewton:
Travis failing because autoheader/autoconf need to be re-run.

terryburton:
Yes, my autotools generating lots of distracting, irrelevant changes so omitted from P.o.C. Pushed result of make reconfig.

alandekok:
I'm generally ok with this, so long as it's a configure option. There are enough distributions using systemd that we might as well go with the flow

terryburton:
@alandekok Thanks.

It might be worth listing a few realistic circumstances that we can detect where it is worth deferring the watchdog - all threads stuck, for example.

I don't yet have the experience or knowledge of the codebase to make it worth me inventing such cases as that will inevitably end up as a solution desperately looking for a problem.

alanbuxey:
wonder if any systemd things can be hooked into the event system and/or
panic action so that if bad things happen, systemd can just restart the
process - thus meaning things like 'monit' can be
taken out of caretaker duty on systemd 'enabled' systems ?

terryburton:
@alanbuxey Replacing a monit-like system is my use case. (Beyond the watchdog notification system systemd is able to relate units such that one unit is considered to have failed when some other unit reports as failed. So you could have a simple radclient-based service that continuously ensures that certain known authorisations succeed causing FR to be restarted in the event of problems. This should allow you to replace the majority of monit instances and monitoring cronjobs.)

It appears that FR's panic action is usually to quit which should already be sufficient for basic panic supervision via systemd. Specifically systemd detects when the last process of the cgroup exits and marks the unit as failed. If configured it will take action such as restarting the service. The shipped .service files include the following which already accomplishes this out of the box:

```
[Service]
Restart=on-failure
RestartSec=5
````

Watchdogs are ideal for situations where things stall (due to programming errors or blocking on offline resources) and restarting the service could be better than waiting indefinitely. For example if a fraction of your requests are hitting some offline database which is gradually tying up threads then restarting might unstick things for the remainder of the traffic (at least for a time).

Perhaps there are other "risky operations" such as reinitialising a module - I haven't looked at whether the is such a concept - during which you might set a global `watchdog_deferred` flag (or incrementing a refcount) that reschedules them for one second later until the it is indicated that the operation is complete by unsetting the flag (decrementing the refcount). That way if the operation never completes then FR will be restarted even if the event loop were still running.

Just examples of what's possible, without any detailed working knowledge...

arr2036:
Looks good, but please fix the autoconf output.  You should be using 2.69... Can @mcnewton or @alandekok run make reconfig and verify that the latest output has been committed... Don't have git repos on new laptop yet.

I'm also not sure there's much point merging this into v3.1.x.  GitHub specifies v4.0.x as the master branch so it should have been submitted against that... I guess we do need to update doc.freeradius.org and the wiki page.

Regarding liveliness, yes, checking threads are still processing their backlogs is useful.

arr2036:
For stuck threads compare top of the heap of backlog to now... If > max_request_timeout thread likely stuck

mcnewton:
On Fri, Aug 19, 2016 at 03:19:19PM -0700, Arran Cudbard-Bell wrote:
> Looks good, but please fix the autoconf output.  You should be
> using 2.69... Can @mcnewton or @alandekok run make reconfig and
> verify that the latest output has been committed... Don't have
> git repos on new laptop yet.

I'm happy to tidy up the autoconf and push this. May not be until
Monday, it's been a long week...


-- 
Matthew Newton, Ph.D. <mcn4@leicester.ac.uk>

Systems Specialist, Infrastructure Services,
I.T. Services, University of Leicester, Leicester LE1 7RH, United Kingdom

mcnewton:
I've merged this into v4.0.x with an updated configure script. @terryburton a couple of suggestions for moving this forward (if you want a longer break from creating barcodes...) may be to look at stuck threads as @arr2036 mentioned above, and also updating the Debian packaging when building with systemd on >= jessie :-).

Thanks!
Matthew

terryburton:
I wasn't expecting these to find their way in on the first round, but great.

@arr2036 Thanks for the advice RE stuck threads. I'll see what this week brings...

@mcnewton Thanks for merging. I'll propose my updated deb packaging soon. I'll update the spec file whilst I'm at it.

PR #1680, opened 2016-08-22
Build with systemd on Debian.

terryburton:
@mcnewton As requested, here are simple packaging updates to Debian to use systemd, tested on jessie (stable) and sid (unstable).

Wheezy (old stable) packaging has been broken since at least January (due to f323757a9b44d41e271c6494561f1bfbba01bc18 – samba-dev doesn't exist before Jessie – so I've not aimed for compatibility back this far.

If there were to target Wheezy then things become tricky for libsytemd-dev (and samba-dev) as Debian does not have a concept of a soft-Build-Depends, i.e. there is no supported `Build-Depends: optional_package | TRUE`

You can fudge it with something like this: `Build-Depends: libsystemd-dev | devscripts` where devscripts is an unrelated common package. Alternatively you could introduce a `debian/control.in` file that you process in `debian/rules` but such usage seems to be discouraged.


mcnewton:
I think wheezy is out anyway due to the version of gcc required for C11 now.

What's the status of systemd on Ubuntu these days? Is this safe for all recent/current releases?

Not sure if there are other common dpkg-based systems that still don't use systemd that we should be concerned about. If there are I think the "libsystemd-dev | dpkg" depends hack looks the best to me.

Thanks,
Matthew

terryburton:
The systemd headers and library exist in Wily and later [1,2]. In terms of supported Ubuntu releases that leaves Precise (12.04 LTS, EOL 04/2017) and Trusty (14.04 LTS, EOL 04/2019) lacking systemd support.

Precise also lacks samba-dev, libjson-c2, libjson-c-dev and is therefore currently broken. Trusty includes these and currently works, so this will be newly broken by this pull, as is.

Unfortunately the `Build-Depends: package | dpkg` hack doesn't appear to work. The dependency resolver in mk-build-deps treats the installed dpkg as a sufficient condition and therefore doesn't install the anticipated package, whilst the resolver in sbuild requires that it knows about the names of all alternatives otherwise it errors.

Options for including systemd in Debian-based distro packaging seem to be:

1. Pull the patch as-is breaking the standard build process for Ubuntu Trusty (and possibly other older distributions). Provide a note that it is safe to drop libsystemd-dev from debian/control in such cases. Simplest.

2. Pull the patch except for the debian/control change so that we do not require libsystemd-dev. Add a wiki instruction step to run apt-get install libsystemd-dev after mk-build-deps. Downside may be reduced use of beneficial systemd features when they are available, but users do get the choice. 

3. Keep libsystemd-dev in debian/control, but add a `debian/control` target to the debian/rules file so that running `./debian/rules debian/control` will regenerate it from debian/control.in based on whether systemd is known to the system. Run this immediately prior to mk-build-deps so that libsystemd-dev will subsequently be installed on systemd-aware distros. [Gist here](https://gist.github.com/terryburton/85512183f3a9bdfa5f708c388b291865) that can probably be extended to samba-dev, libjson, etc. Really this is just a more complex version of option 2 that automates away the user's choice. (Some folk invoke this from the clean target so that you would instead run `./debian/rules clean`, as done upfront by sbuild, et al., but this doesn't seem to be particular advantageous and is frowned upon by many DDs.)

4. Wait until 2019 when trusty will be EOL and all of your light bulbs will be running systemd ;-)

If you want to choose an approach or suggest another then I'll cook up patches.

[1] http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libsystemd-dev&searchon=names
[2] http://packages.ubuntu.com/search?suite=all&section=all&arch=any&keywords=libsystemd0&searchon=names

arr2036:
5. Use autoconf to build the spec file :)

arr2036:
We need to automate package maintenance anyway.  That's what the README.md files were the first step in... Don't particularly care if it's autoconf or shell...

The original idea of just packaging a few modules was wrong, and has caused lots of annoying issues.  Each rlm_*/proto_* needs its own .deb/.rpm with the correct dependencies.  The default config package on each distro should then have dependencies on the modules required for the default config to work.

That way you can install your own config package, specifying your own dependencies and it all just works.

mcnewton:
Thanks, looks reasonable to me now, and should still build on any deb system. There's duplication of the control file, but removing the original will cause problems for people who don't follow build instructions...

This update is probably only required for the next 2-3 years anyway until all existing Debian-based distros have at least the systemd shim libraries available (they don't actually need to use systemd).

PR #1682, opened 2016-08-22
Use a mutex for perl_sys_init3_called

herwinw:
We want to use this variable atomic.


arr2036:
mod_instantiate will always be serialised?

PR #1684, opened 2016-08-23
Fix for bug where originating CoA requests causes segfault if proxying is disabled.

paterry:
If FreeRADIUS 3.0.11 is configured to originate CoA requests in response to certain packets using the following files:

  etc/raddb/sites-enabled/default
  etc/raddb/sites-enabled/originate-coa
  etc/raddb/clients.conf

it will segfault if the following line is present in etc/raddb/radiusd.conf:

  proxy_requests = no

In src/main/process.c request_coa_originate() calls insert_into_proxy_hash(), which expects the global variable proxy_list to be initialised. It is initialised in radius_event_start(), but only if main_config.proxy_requests is true.

The patch adds the following configuration parameter:

  originate_coa_requests = yes / no

If it is set to 'yes' the proxy_list global will be initialised if the proxy code hasn't already done so. Otherwise CoA requests will not be originated. It defaults to 'yes'.

### DIFF
--- a/src/include/radiusd.h
+++ b/src/include/radiusd.h
@@ -129,6 +129,9 @@ typedef struct main_config {
 #ifdef WITH_PROXY
 	bool		proxy_requests;			//!< Toggle to enable/disable proxying globally.
 #endif
+#ifdef WITH_COA
+	bool            originate_coa_requests;         //!< Toggle to enable/disable originating COA requests globally.
+#endif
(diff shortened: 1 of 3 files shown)


alandekok:
The simpler fix is to disallow `originate_coa` unless `proxy_requests = yes`.

Sending CoA packets is proxying.  So there's no point in allowing CoA when proxying is disallowed.

There is no problem (security, performance, etc.) when allowing `proxy_requests = yes`, but never doing proxying.

PR #1685, opened 2016-08-25
Rlm python python3

Deisuan:
Changed rlm_python to build against python3(.4).
There is currently an issue where the module (probably) loses the threadstate or the heap gets corrupted. 
This manifests itself as the python functions reading random/broken data from parameters.

PR on suggestion of Arran Cudbard-Bell on freeradius-devel to allow others to work further on python3 linkage.


alandekok:
Please rebase your banch so that it's synced with upstream.

i.e. the *only* commits visible in the PR should be the ones you made to rlm_python.

As it stands now, it's very difficult to tell what's yours, and what's just pulled from upstream.

PR #1686, opened 2016-08-25
Changed rlm_python to build against python3(.4).

Deisuan:
There is currently an issue where the module (probably) loses the threadstate
or the heap gets corrupted.
This manifests itself as the python functions reading random/broken data from
parameters -> needs work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ addons:
       - lintian
       - mysql-server
       - pbuilder
-      - python-dev
+      - python3-dev
       - quilt
       - libruby
       - ruby-dev
@@ -77,7 +77,7 @@ addons:
       - libpcre3-dev
       - libperl-dev
       - libpq-dev
-      - libpython-all-dev
+      - libpython3-all-dev
       - libreadline-dev
       - libsnmp-dev
       - libssl-dev
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ addons:
       version: master
       description: The FreeRADIUS server project
     notification_email: freeradius-devel@lists.freeradius.org
-    build_command_prepend: ./configure -with-rlm-python-bin=/usr/bin/python2.7
+    build_command_prepend: ./configure -with-rlm-python-bin=/usr/bin/python3
     build_command: make
     branch_pattern: coverity_scan
   postgresql: "9.4"
--- a/debian/control.in
+++ b/debian/control.in
@@ -27,7 +27,7 @@ Build-Depends: @BUILDDEPS@ debhelper (>= 9),
  libykclient-dev,
  libmemcached-dev,
  libhiredis-dev,
- python-dev,
+ python3-dev,
  samba-dev
 Section: net
 Priority: optional
--- a/scripts/travis/build.sh
+++ b/scripts/travis/build.sh
@@ -26,7 +26,7 @@ CFLAGS="${BUILD_CFLAGS}" ./configure -C \
     --with-udpfromto=$LIBS_OPTIONAL \
     --with-openssl=$LIBS_OPTIONAL \
     --with-pcre=$LIBS_OPTIONAL \
-    --with-rlm-python-bin=/usr/bin/python2.7 \
+    --with-rlm-python-bin=/usr/bin/python3 \
 || cat ./config.log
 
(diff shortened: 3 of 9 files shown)


herwinw:
I've got a branch (https://github.com/herwinw/freeradius-server/tree/rlm_python3) where I pushed a fix to make it compile with Python 3.5 (commit 51a7eb0be4e976f7a714a20c00a6b7c2a51b507e). Feel free to include that

arr2036:
Hm, this would drop 2.7 compatibility, how do people feel about that?

Deisuan:
Alternatively a rlm_python3 module could be introduced, concurrently with the current rlm_python(2.7) module. Python 2.7 will be the last python2 version, and support will run out in 2020, while the first python3 version was released in 2008(!). Perhaps it could even be appropriate to have an rlm_python2 (and thus have rlm_python for python3), since it is basically legacy?

arr2036:
That sounds like a lot of nearly duplicated code to maintain for no real value.  I guess my question is, is there any reason to continue using Python 2.7? Are the syntax changes between 2.7 and 3.0 so great that it's huge amounts of effort to rewrite code? Are the number of eggs for Python 3.0 limited?

I'm not really involved in the Python ecosystem, I just help maintain the rlm_python module out of hatred for Perl.

Deisuan:
Allright, IMHO the porting from python2.7 to python3 is mostly trivial (for syntax alone there are scripts as "2to3" available) - as long as one does not use any libraries that are too exotic (and scaling with the size of the code base obviously). As python3 has been around for eight years, I believe it to be justified to switch exclusively to it. No new project should have a compelling reason use python2, and any older projects should consider porting anyways, if you take the ending support into account.

Deisuan:
Porting to python3 should be done. Can be merged, as far as I can tell. Tested against python3.4.

arr2036:
Well, how about you get the tests passing first ;)

Deisuan:
Yeah, well, did build for me :P I'm on it though.

arr2036:
Oh, it builds on one the compilers at least... It also crashes in the automated tests :P

arr2036:
you can ``make test`` to run them locally.

nwkrad:

nwkrad:

arr2036:
Is there any will to get this merged? Going to close the PR in two weeks if there's no activity.

alandekok:
python2 will be used until 2020.  After that it's dead.

So we can keep using rlm_python against python3.  But at some point, it should change.

TBH, we should probably do this in the v4 branch.

arr2036:
Yeah, this PR is for V4, and would have been merged, it's just it doesn't pass the tests.

spbnick:
@arr2036, @alandekok, did I read you right here and you're not interested in getting v3.0.x to support Python 3? Asking because RHEL8 is going to only have Python 3.

alandekok:
If everyone is moving to Python 3, then we should support it in FreeRADIUS.

Given that we don't want to remove functionality, I think that adding a new module would be the simplest option.  `rlm_python` is only ~1100 LoC, much of which is boilerplate.  I would be OK with adding an `rlm_python3` module, specifically for Python 3.

I suspect it's too hard to have one module support both Python 2 and Python 3.

arr2036:
It's a lot of extremely complicated and involved boilerplate which doesn't improve with python 3.  It's not actually that hard to have a module that supports both, the API is very similar.

arr2036:
I already did a bunch of the work in the existing module.

spbnick:
I see, thank you for the response and an update, @alandekok, @arr2036. Red Hat obviously needs Python 3 support, and in v3.0.x too. I have little time at the moment to help you with that (and I'm not a Python expert), but I'll either find time myself to help with that, or find someone else to take over. Can I somehow follow your efforts on this, in the meantime?

jeffersonchua:
> I've got a branch (https://github.com/herwinw/freeradius-server/tree/rlm_python3) where I pushed a fix to make it compile with Python 3.5 (commit [51a7eb0](https://github.com/FreeRADIUS/freeradius-server/commit/51a7eb0be4e976f7a714a20c00a6b7c2a51b507e)). Feel free to include that

The branch is gone? Is there a patch somewhere for python3

Thanks,
Jeff

PR #1698, opened 2016-08-29
Update rlm_chap.5

alanbuxey:
typo


alandekok:
Can we PLEASE have commit messages which have more content than "update foo"?

Imagine a commit history where 99% of the commits are "update file".  It's *retarded*.

alanbuxey:
i understand your point - will make the subject more verbose rather than filling in the comment. sorry!

PR #1704, opened 2016-08-30
initial summary documentation

alanbuxey:
..this one didn't even have the words that were incorrectly given to another proto README file

### DIFF
--- a/src/modules/proto_vmps/README.md
+++ b/src/modules/proto_vmps/README.md
@@ -1,8 +1,9 @@
 # proto_vmps
 ## Metadata
 <dl>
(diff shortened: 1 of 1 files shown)


arr2036:
We should really standardise on Markdown or Asciidoc for server docs.  It's annoying having to lookup the reference pages for three different markup formats.

I vote Markdown, because jekyll can render it, so we can publish bits of the server docs on the FreeRADIUS website.

alanbuxey:
agree. choose the format!  :)

mcnewton:
Markdown +1

Now choose the flavour ;-)

PR #1718, opened 2016-09-08
log: fix R*DEBUG* macros

gureedo:
add `L_DBG_IFNO` level
add missing `RIDEBUG3` `RIDEBUG4` `RWDEBUG3` `RWDEBUG4`
check global and request log level in all `R*DEBUG*` macros

### DIFF
--- a/src/include/log.h
+++ b/src/include/log.h
@@ -39,10 +39,11 @@ typedef enum log_type {
 	L_ACCT = 7,			//!< Accounting messages
 
 	L_DBG = 16,			//!< Only displayed when debugging is enabled.
-	L_DBG_WARN = 17,		//!< Warning only displayed when debugging is enabled.
-	L_DBG_ERR = 18,			//!< Error only displayed when debugging is enabled.
-	L_DBG_WARN_REQ = 19,		//!< Less severe warning only displayed when debugging is enabled.
-	L_DBG_ERR_REQ = 20		//!< Less severe error only displayed when debugging is enabled.
+	L_DBG_INFO = 17,		//!< Info only displayed when debugging is enabled.
+	L_DBG_WARN = 18,		//!< Warning only displayed when debugging is enabled.
+	L_DBG_ERR = 19,			//!< Error only displayed when debugging is enabled.
+	L_DBG_WARN_REQ = 20,		//!< Less severe warning only displayed when debugging is enabled.
+	L_DBG_ERR_REQ = 21		//!< Less severe error only displayed when debugging is enabled.
 } log_type_t;
 
 typedef enum log_lvl {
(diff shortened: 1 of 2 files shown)


arr2036:
Parts of this patch are incorrect... the error macros always need to call radlog_request_error in all cases, because it adds the Message-Failure-Attribute to the request.  Which you want irrespective of the log level.

gureedo:
So, it looks like inconsistent macros naming for `R*DEBUG*`.
RDEBUG*, RIDEBUG*, RWDEBUG works only when debugging is on, but REDEBUG works always.
I think it is unexpected behavior.
May be it can be renamed to something different?

arr2036:
If you read the code... that's not actually what happens.

```c
void radlog_request_error(log_type_t type, log_lvl_t lvl, REQUEST *request, char const *msg, ...)
{
	va_list ap;

	rad_assert(request);

	va_start(ap, msg);
	if (request->log.func) request->log.func(type, lvl, request, msg, ap);
	else if (!(type & L_DBG)) vradlog_request(type, lvl, request, msg, ap);
	vmodule_failure_msg(request, msg, ap);
	va_end(ap);
}
```
```c
void vradlog_request(log_type_t type, log_lvl_t lvl, REQUEST *request, char const *msg, va_list ap)
{
	char const	*filename = request->log.output->file;
	FILE		*fp = NULL;

	char		*p;
	char const	*extra = "";
	uint8_t		unlang_indent, module_indent;
	va_list		aq;

	char		*msg_prefix = NULL;
	char		*msg_module = NULL;
	char		*msg_exp = NULL;

	rad_assert(request);

	/*
	 *	Debug messages get treated specially.
	 */
	if ((type & L_DBG) != 0) {
		if (!radlog_debug_enabled(type, lvl, request)) return;
```

gureedo:
I've added second commit, that fixes `REDEBUG*` behavior.
It still confusing me, that something with DEBUG in name works while debug not enabled, i.e. set module failure message.

arr2036:
**It doesn't emit debug messages**, it adds an attribute to the request, so that when something isn't working, you have access to all the ERROR messages that **WOULD** have been emitted, in a form that can be easily serialised and written to a post-authentication record.

Every call to RERROR or REDEBUG **MUST** add a Module-Failure-Message else the facility gets far less useful, because by default, on production servers, people don't run with -xxxx, they run with no debugging, and most of the ERROR messages would not get added to the request.

So in magic, perfect, super logical, OCD land, yes, REDEBUG wouldn't do anything if the current debug level wasn't as high as the macro level, back in the real world, where people actually need to get shit done, they don't give a crap about whether a macro behaves identically to other similarly named macros, they'd prefer to know why user XYZ failed authentication.

Can you please squash your commits into a single commit.

gureedo:
Thanks for explanation! Merged PR into single commit.

arr2036:
Thanks!

...and yes the idea is that you get a stack of Module-Failure-Message attribute that give you the history of errors during request processing.  If it just kept overwriting the last instance, then i'd tend to agree with you.

PR #1723, opened 2016-09-12
Use JSON library in rlm_rest

qnet-herwin:
This saves around 150 lines of custom code to create a JSON string from an attribute list.


arr2036:
I don't think that'd actually work though? If there's more JSON than buffer, there's nothing to maintain state between calls...

arr2036:
I'm not going to merge code that will break the rlm_rest module.  So if you want this finished, you're going to have to fix it so it works with the chunking API, or remove the chunking API :)

qnet-herwin:
@arr2036: I'll try to fix it later this week. As far as I could find, json-c has no streaming output possibilities. Of course you could emulate it with copying chunks of data, but that would be pretty stupid. I think I'll have to dive into the code before finding out what the best solution is here.

qnet-herwin:
This version just copies the solution used in `HTTP_BODY_CUSTOM` (which showed a bug in that logic, see #1754). I tested it by adding just a bunch of data to request-Tmp-String-1, gdb shows it will be copied in chunks.

PR #1725, opened 2016-09-13
Initialize exp before calling radius_axlat

pwdng:


arr2036:
radius_xlat should set the pointer to NULL.  The general assumption is the function you pass the pointer to should do the initialisation.

pwdng:
ok, moved initialisation to radius_axlat & radius_axlat_struct

PR #1727, opened 2016-09-13
Feature/eap fast 3.0.x

jrouzierinverse:
Back port eap fast to 3.0.x.

Please consider this for inclusion in freeradius.


arr2036:
FWIW, I vote no, v3.0.x is feature frozen, you should be developing on the v4.0.x branch.

arr2036:
But of course you knew that... so instead of actually contributing to the development of the project, your backporting code to make your lives easier, whilst potentially reducing the stability of the v3.0.x branch, and making it a worse experience for everyone else.

alandekok:
Let's be nice here... a stand-alone feature is a reasonable thing to add, even if 3.0 is allegedly in feature freeze.

PR #1739, opened 2016-09-16
Redis patch

pwdng:


arr2036:
Out of interest, when did the key calculation code change, and will this cause issues for older versions of Redis?

pwdng:
> On 18 Sep 2016, at 04:46, Arran Cudbard-Bell <notifications@github.com> wrote:
> 
> Out of interest, when did the key calculation code change, and will this cause issues for older versions of Redis?
> 
The Redis key calculation code hasn’t been updated, the logic has always been the same since tags were introduced in keys.
See https://github.com/antirez/redis/commit/142281dc79a1428a8deae4f531e7340851e66797 <https://github.com/antirez/redis/commit/142281dc79a1428a8deae4f531e7340851e66797>
Maybe there was a typo in the Redis spec page at the time of the initial cluster implementation in rlm_redis.

arr2036:
But the cluster code worked fine for three other deployments with no remapping or memory leak issues, what did you do different that triggered this?

pwdng:
You might not have noticed the remapping messages. After all, it’s not an error.
And the memory  leak only happens when we have remaps in quick succession.
So we’re not leaking that much.
I noticed it because I’m looking for a leak, but there’s a bigger one somewhere I haven’t found yet :-(

arr2036:
But you tested it too right, and didn't notice hundreds of remaps? and i've run the code in debugging code on many many occasions, and during developing the IP pool modules.

pwdng:
In my case, the remaps only occurred frequently because the TLS session id being binary ends up containing ‘{‘ and ‘}’ characters that meant the slot calculating was wrong when we had ‘…}…{…’
With the IP pool, you wouldn’t run into that problem.

arr2036:
Ahhhh! That makes sense. OK!

PR #1742, opened 2016-09-17
Update install-sh to version 2016-01-11.22

stapelberg:
Copied (https://www.gnu.org/software/gnulib/’s recommendation) from
http://git.savannah.gnu.org/cgit/gnulib.git/plain/build-aux/install-sh

This fixes an issue with running e.g. `make -j8 install`, where make
would run e.g. the following two rules in parallel:

  1. raddb/mods-config/sql/ippool
  2. raddb/mods-config/sql/ippool/sqlite

make would run install-sh for each of these rules, and install-sh would
try to run mkdir for each path component. When the concurrently running
install-sh processes would interleave in such a way that line 182¹ was
executed in both processes before line 184² was executed, both processes
would try to run mkdir, and one of them would fail, effectively failing
the build with an error like this:
  mkdir: cannot create directory 'debian/tmp/etc/freeradius/mods-config/sql/ippool': File exists

① https://github.com/FreeRADIUS/freeradius-server/blob/d17d160f319f22143911c033d207c8909775a4da/install-sh#L182
② https://github.com/FreeRADIUS/freeradius-server/blob/d17d160f319f22143911c033d207c8909775a4da/install-sh#L184

### DIFF
--- a/install-sh
+++ b/install-sh
@@ -1,251 +1,501 @@
 #!/bin/sh
-#
 # install - install a program, script, or datafile
-# This comes from X11R5 (mit/util/scripts/install.sh).
+
+scriptversion=2016-01-11.22; # UTC
+
+# This originates from X11R5 (mit/util/scripts/install.sh), which was
+# later released in X11R6 (xc/config/util/install.sh) with the
+# following copyright and license.
 #
-# Copyright 1991 by the Massachusetts Institute of Technology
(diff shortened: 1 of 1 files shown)


alandekok:
Another fix is to update the build rules so that installation targets have an order dependency on their directories.

I'll take a look.

PR #1743, opened 2016-09-17
replace EVP_cleanup check with EVP_PKEY_new

stapelberg:
EVP_cleanup is no longer shipped in libcrypto starting with OpenSSL
1.1. Instead, EVP_cleanup is a no-op define:

  /usr/include/openssl/evp.h:#  define EVP_cleanup() while(0) continue

This resulted in rlm_eap_pwd not being compiled when OpenSSL 1.1 is
installed.


arr2036:
Nice.  @alandekok and I already did a couple of passes for 1.1, so hopefully there shouldn't be many more fixes required.

PR #1748, opened 2016-09-19
Call pthread_setspecific for every thread that calls fr_thread_local_…

pwdng:
…init, not just the first one. Otherwise, the value associated with the key is NULL and the destructor isn't called.


arr2036:
I can see the issue, but that is definitely not a good solution.  You want to call pthread_setspecific multiple times per request?

arr2036:
Never mind, sleepy :p

arr2036:
I think that'll fix Python too, yay.

PR #1751, opened 2016-09-19
Feature/eap fast 3.0.x

jrouzierinverse:
Fix pac reuse

### DIFF
--- a/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c
+++ b/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c
@@ -761,9 +761,8 @@ static rlm_rcode_t CC_HINT(nonnull) process_reply( eap_handler_t *eap_session,
 						  REQUEST *request, RADIUS_PACKET *reply)
 {
 	rlm_rcode_t			rcode = RLM_MODULE_REJECT;
-	VALUE_PAIR			*vp, *tunnel_vps = NULL;
+	VALUE_PAIR			*vp;
 	vp_cursor_t			cursor;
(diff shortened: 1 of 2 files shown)


src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c
```diff
@@ -905,18 +903,6 @@ static rlm_rcode_t CC_HINT(nonnull) process_reply( eap_handler_t *eap_session,
...
-	 * to the supplicant.
-	 */
-	if (tunnel_vps) {
-		RDEBUG("Sending tunneled reply attributes");
-		rdebug_pair_list(L_DBG_LVL_2, request, tunnel_vps, NULL);
-
-		eap_vp2fast(tls_session, tunnel_vps);
-		fr_pair_list_free(&tunnel_vps);
-	}
-
```
arr2036: Hmm, this looks pretty used to me... Why are you removing it? If it's because of a previous commit, can you reference that commit...

alandekok: Hmm... EAP-FAST still seems to work without that.  Weird.

jrouzierinverse: eap_fast2vp is still called here in the challenge.

https://github.com/jrouzierinverse/freeradius-server/blob/67f18684ac0c3d03747294d46836ede2587d72cd/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c#L887

The reason I changed that is that I was trying to keep the style similar to the other eap modules in the 3.0.x. branch.

PR #1758, opened 2016-09-26
Check the CVE ID, not the name

TheMysteriousX:
I think the ID is intended to be checked, not the name.

### DIFF
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -2483,7 +2483,7 @@ int tls_global_version_check(char const *acknowledged)
 			/*
 			 *	If the CVE is acknowledged, allow it.
 			 */
-			if (strcmp(acknowledged, defect->name) == 0) return 0;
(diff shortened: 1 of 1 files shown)


alandekok:
And pulled over to v4.0.x, too.

arr2036:
The comment in the code says newest first, but the new issues were added newest last which breaks the acknowledgement check.

alandekok:
The comments in the code also said it looped over all CVEs, but the loop had a hard-coded check against entry[0].

I'll push a fix.

arr2036:
In you read the code that's actually not true.  It did check all the defects, but it only checked to see if the first entry has been acknowledged.

	https://github.com/FreeRADIUS/freeradius-server/blob/a02b1725cba0119e31b5d4039ed3b2b42ed1ad25/src/main/tls/global.c

Which works find if all the defects are on the same branch, but it doesn't work when the defects are on multiple branches.

PR #1770, opened 2016-10-05
If sid2gid mapping fails for a group winbind reports the gid as UINT32_MAX and doesn't log anything very useful - provide a hint of where to look

gclinch:
Is checking against UINT32_MAX too much magic?  Perhaps this would be better simply as a documentation note?


qnet-herwin:
If Winbind really uses `UINT32_MAX` as a special value, I would expect that there is a constant defined somewhere within Winbind to indicate that. If that is the case, we could simply use that constant and get rid of the magic.

mcnewton:
Looks like winbind sets all GIDs to `UINT32_MAX`, then tries to fill in the blanks. So if any are still this value it does mean the ID mapping failed. It seems a bit nasty (well, it is, there not being a Samba constant for it) but does help the user debug the problem.

mcnewton:
@gclinch out of interest: when this happens, what is the `err` return value?

gclinch:
@mcnewton WBC_ERR_DOMAIN_NOT_FOUND, which does kind of make sense since I don't have an idmap range that covers UINT32_MAX.

I think it should be possible to do these lookups purely with SIDs and not GIDs and thus not need working idmaps - I'll dig further and perhaps open a further PR.

PR #1778, opened 2016-10-11
*vps being set to NULL is valid

jrouzierinverse:
Receiving lots of 
'SOFT ASSERT FAILED src/modules/rlm_perl/rlm_perl.c[703]: *vps' in the radius log.

However get_hv_content and do_perl set the value pair to NULL before pairadd_sv is called.

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_perl/rlm_perl.c#L741

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_perl/rlm_perl.c#L870


alandekok:
I've pushed an alternate fix.

PR #1787, opened 2016-10-13
Add enum mapping in JSON output

qnet-herwin:
But only if at least one value of the list is mapped. Add NULL values
for all unmapped entries. This has been attempted previously in  #1618

The JSON output now looks like this in case any value in the list is mapped:

```
"Service-Type":{"type":"integer","value":[999,1,111,10],"mapping":[null,"Login-User",null,"Call-Check"]}
```

By using an extra key `mapping` in the object, it should be completely backwards compatible.

The majority of the change has to do with testing if there is a mapped value, and fixing things up we've found one. The implementation could be really simplified by one of the following two changes:
- Do a lookup to see if there are mappings defined for the attribute. We can find in the dictionary that `Service-Type` has mappings, but I couldn't find any function to do that.
- Put the mapped value in the `value` array instead of the `mapping` array, so the example above would result in `"value":[999,"Login-User",111,"Call-Check"]`. This breaks backwards compatibility, but since we're talking about 4.0, that does not sound like a really big problem. One could argue that you'd need the dictionaries to interpret the value, but that also holds for the attribute names.

### DIFF
--- a/src/modules/rlm_json/json.c
+++ b/src/modules/rlm_json/json.c
@@ -299,6 +299,7 @@ const char *fr_json_from_pair_list(TALLOC_CTX *ctx, VALUE_PAIR **vps, const char
 	TALLOC_CTX *local_ctx;
 	vp_cursor_t cursor;
 	struct json_object *obj, *vp_object, *values, *type_name;
+	fr_dict_enum_t const *dv;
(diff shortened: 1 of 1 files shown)


arr2036:
Very nice. Looks good 👍

PR #1794, opened 2016-10-17
Fixed configure flag --without-libfreeradius-json

### LINKED ISSUE #1791: ./configure --without-libfreeradius-json has no effect
As suggested by the ./configure output:

configure: WARNING: json-c libraries not found. Use --with-jsonc-lib-dir=<path>.
configure: error: set --without-libfreeradius-json to disable it explicitly.
configure: error: ./configure failed for src/modules/rlm_json

I set ./configure --prefix=/usr/local/freeradius \
... \
--without-libfreeradius-json

but ./configure continues to error out with the same:
configure: WARNING: json-c libraries not found. Use --with-jsonc-lib-dir=<path>.
configure: error: set --without-libfreeradius-json to disable it explicitly.
configure: error: ./configure failed for src/modules/rlm_json

herwinw:
This should fix  #1791

### DIFF
--- a/src/modules/rlm_json/configure
+++ b/src/modules/rlm_json/configure
@@ -1896,7 +1896,7 @@ ac_compiler_gnu=$ac_cv_c_compiler_gnu
 ac_config_headers="$ac_config_headers config.h"
 
 
-if test x$with_libfreeradius-json != xno; then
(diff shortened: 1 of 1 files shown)


arr2036:
You need to make changes to configure.ac not just the configure script...

PR #1799, opened 2016-10-20
rlm_ldap: cleanup memory after ldap version query

gureedo:
rlm_ldap: cleanup memory after ldap version query

Signed-off-by: Aleksey Katargin gureedo@intersvyaz.net


arr2036:
Why the zero length string for inst->server?

alandekok:
`inst->server` should be allocated, so that `talloc_asprintf_append()` works.

It probably should be

    if (!inst->server) inst->server = talloc_strdup(inst, "");

gureedo:
Amended commit with `if (!inst->server) inst->server = talloc_strdup(inst, "");`

gureedo:
So, would you merge this PR?

PR #1800, opened 2016-10-20
Create RFC7542 policy

spaetow:
Deals with RFC7542-style usernames. Using User-Name attribute in control list. Concerns about this is something else using User-Name and us overwriting it...


arr2036:
My strong preference would be to modify split_username_nai, and if local_domain= is "" then not implement the fancy rfc7542 behaviour... then people might actually use it ;)

spaetow:
Ok, I'll have a look tomorrow. 👍

Issue #1790, opened 2016-10-15
EAP-Fast coding issues.


arr2036:
# Issue type
- [x] Contributor needs schooling on coding practices.
# Defect/Feature description

A certain contributor to the FreeRADIUS code base has gotten lazy writing code in the city, and needs schooling on how to write propa good code, init.

The code snippets below are taken from https://github.com/FreeRADIUS/freeradius-server/blob/v4.0.x/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c

But all code in EAP-Fast should be checked.
---

``` c
scratch = talloc_size(request, ksize + sizeof(*t->keyblock));
```

Incorrect talloc function used, should be one of the talloc array functions, so the type is set correctly -  _Yes, talloc allows type safety checks_
---

``` c
RHEXDUMP(4, "S-IMCK[0]", t->simck, EAP_FAST_SIMCK_LEN);
```

Magic numbers should be replaced with macros, in this case the macro already exists in `log.h` (guess which one).
---

``` c
static int openssl_get_keyblock_size(REQUEST *request, SSL *ssl)
```

Code copied/pasted from other projects should be converted to FreeRADIUS style, including being renamed to match the FreeRADIUS naming scheme... and that function probably should be somewhere beneath `src/main/tls`
---

``` c
eap_fast_tunnel_t       *t = (eap_fast_tunnel_t *) tls_session->opaque;
```

Casts from void should use `talloc_get_type_abort`, unless it's a super hot code path, in which case only do it for debug builds.
---

``` c
    case AUTHENTICATION:
```

Unqualified macro names, in general, are bad, but seeing them mixed with macros from another enum type makes me want to kill small furry animals.
---

All functions are lacking doxygen headers, which means maintainers need to read up on how eap-fast works, and then map that to functions in `eap_fast.c` to be able to debug behavioural/compliance issues.

Explain how the code relates to the standard, and give hints about how EAP-FAST in general is supposed to work.
---

``` c
memcpy(t->simck, t->keyblock, EAP_FAST_SKS_LEN);    /* S-IMCK[0] = session_key_seed */
```

What's a simck? Is it a s-imck? in which case it should be s_imck, cos otherwise people will think it's related to EAP-SIM or SIMs...
---

[...]

PR #1804, opened 2016-10-24
Deal with EAP-Fast coding issues

jimdigriz:
Like it says on the tin, references bug #1790


arr2036:
Looks good, thanks for making those changes.

arr2036:
Looks like @alandekok is having fun with travis, so I'll merge theses now...

PR #1812, opened 2016-10-28
Refer to radiusd.conf for more examples

Caspinol:


alandekok:
I'd prefer to not refer to the `instantiate` section for documentation.  It would be better to add examples into the "man" page.

PR #1813, opened 2016-10-28
Update freeradius.spec

spaetow:
For some reason, this SPEC doesn't seem to set the executable bit for configure or install.sh.


alandekok:
`configure` and `install.sh` are executable in the source tree.  Why aren't they marked executable in the tar file / src RPM?  I'd prefer to fix that issue.

Also, the code to do `chmod a-x *.[ch]` is likely unnecessary.  That issue has been fixed for ages.

spaetow:
@alandekok, quite frankly, I don't know why they're not set. This is what I've found when downloading the ZIP file from Github... :-/

alanbuxey:
ZIP file??  ;-)

spaetow:
Yes Alan. ZIP file. Github does ZIP. Smartypants! :-D

alanbuxey:
well, it shouldn't ;-)

..and if it does, their ZIP tool needs to support file flags and ensure its
zipping things up with them correctly.  if they do,
then its your ZIP client or choice of flags to extract with :)

spaetow:
@alanbuxey, you're welcome to prod Github over their choice of archiving tool... I'm just reporting that this is an issue and fixing it. 

@alandekok, the chmod a-x *.[ch] part I can delete, sure. It's been there since v2.0.x...

alandekok:
I'd prefer to not merge this.  The `tar.gz` format supports the correct permissions.  I'm disinclined to add code to fix problems created by other peoples tools.

arr2036:
http://stackoverflow.com/questions/8377081/github-api-download-zip-or-tarball-link

PR #1822, opened 2016-11-04
implement radiusd::radius_xlat in rlm_perl

lytboris:
Main goal is to get access to things like `%{client:group}` inside perl scripts

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -22,6 +22,8 @@ FreeRADIUS 3.1.0 Thu 26 Nov 2015 15:48:14 EDT urgency=medium
 	* TLV Attributes can now be nested to nearly any depth.
 	* radsnmp client which can listen on Net-SNMP "pass persist"
 	  and translate that into queries to the server.
+	* rlm_perl: radiusd::xlat to evaluate xlat string within
+	  perl script
 
 	Bug fixes
 	* Attribute names in unlang MUST be prefixed with &
--- a/src/modules/rlm_perl/rlm_perl.c
+++ b/src/modules/rlm_perl/rlm_perl.c
@@ -126,7 +126,7 @@ static const CONF_PARSER module_config[] = {
  */
 EXTERN_C void boot_DynaLoader(pTHX_ CV* cv);
 
-_Thread_local REQUEST *rlm_perl_request;
+static _Thread_local REQUEST *rlm_perl_request;
 
 #ifdef USE_ITHREADS
 #  define dl_librefs "DynaLoader::dl_librefs"
--- a/src/modules/rlm_perl/rlm_perl.c
+++ b/src/modules/rlm_perl/rlm_perl.c
@@ -126,6 +126,8 @@ static const CONF_PARSER module_config[] = {
  */
 EXTERN_C void boot_DynaLoader(pTHX_ CV* cv);
 
+_Thread_local REQUEST *rlm_perl_request;
+
 #ifdef USE_ITHREADS
 #  define dl_librefs "DynaLoader::dl_librefs"
 #  define dl_modules "DynaLoader::dl_modules"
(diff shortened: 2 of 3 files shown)


arr2036:
Get rid of the radius and just do radiusd::xlat

arr2036:
and use radius_axlat instead of a static buffer

doc/ChangeLog
```diff
@@ -22,6 +22,8 @@ FreeRADIUS 3.1.0 Thu 26 Nov 2015 15:48:14 EDT urgency=medium
 	* TLV Attributes can now be nested to nearly any depth.
 	* radsnmp client which can listen on Net-SNMP "pass persist"
 	  and translate that into queries to the server.
+	* rlm_perl: radiusd::radius_xlat to evaluate xlat string
```
arr2036: Just use radiusd::xlat, no need for the extra qualification

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -296,6 +296,40 @@ static XS(XS_radiusd_radlog)
...
+	dXSARGS;
+	char *in_str;
+	char expanded[1024]; /* XXX FIXME Dynamic allocation in case slen >= sizeof(expanded) ? */
+	ssize_t slen;
+	SV *rad_requestp_sv;
+	REQUEST *request;
+
+	if (items != 1) croak("Usage: radiusd::radius_xlat(string)");
+
+	rad_requestp_sv = get_sv("RAD___REQUESTP", 0);
```
arr2036: Do it like this instead:
- https://github.com/FreeRADIUS/freeradius-server/blob/v4.0.x/src/modules/rlm_lua/lua.c#L40
- https://github.com/FreeRADIUS/freeradius-server/blob/v4.0.x/src/modules/rlm_lua/lua.c#L247
- https://github.com/FreeRADIUS/freeradius-server/blob/v4.0.x/src/modules/rlm_lua/lua.c#L770

Many many many fewer cycles.

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -296,6 +296,40 @@ static XS(XS_radiusd_radlog)
...
+	if (rad_requestp_sv == NULL) croak("Can not evalue xlat, RAD___REQUESTP is not set!");
+
+	request = INT2PTR(REQUEST *, SvIV(rad_requestp_sv));
+
+	in_str = (char *) SvPV(ST(0), PL_na);
+	memset(expanded, 0, sizeof(expanded));
+	slen = radius_xlat(expanded, sizeof(expanded), request, in_str, NULL, NULL);
```
arr2036: use ``radius_axlat`` so the strings aren't truncated.

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -296,6 +296,40 @@ static XS(XS_radiusd_radlog)
...
+ *	Now users are able to get data that is accessible only via xlat
+ *	e.g. %{client:...}
+ *	Call syntax is radiusd::radius_xlat(string), string will be handled the
+ *	same way it is described in EXPANSIONS section of man unlang
+ */
+static XS(XS_radiusd_radius_xlat)
+{
+	dXSARGS;
+	char *in_str;
+	char expanded[1024]; /* XXX FIXME Dynamic allocation in case slen >= sizeof(expanded) ? */
```
arr2036: Yes..

lytboris:
Pushed updated version. It is not tested in prod though, will do it on week following (on 3.0.x with __THREAD REQUEST *rlm_perl_request).

arr2036:
Looks good, thanks!

arr2036:
v3.0.x does have thread_local macros which fallback to mutexes. Honestly any compiler written in the last five years will have support for thread local variables, even if it's not using the C11 keyword.

PR #1825, opened 2016-11-04
Use SSL_SESSION_get_master_key (OpenSSL 1.1)

stapelberg:
This unbreaks compilation with OpenSSL 1.1

### DIFF
--- a/src/modules/rlm_eap/lib/base/mppe_keys.c
+++ b/src/modules/rlm_eap/lib/base/mppe_keys.c
@@ -52,6 +52,17 @@ size_t SSL_get_server_random(const SSL *s, unsigned char *out, size_t outlen)
 	memcpy(out, s->s3->server_random, outlen);
 	return outlen;
 }
+
+static size_t SSL_SESSION_get_master_key(const SSL_SESSION *s, unsigned char *out, size_t outlen);
+static size_t SSL_SESSION_get_master_key(const SSL_SESSION *s, unsigned char *out, size_t outlen)
+{
+	if (!outlen) return s->master_key_length;
+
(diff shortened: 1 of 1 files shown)


arr2036:
But breaks all other versions :/

alandekok:
I put patches into v4.0.x which fix this in a compatible way. Please check that.

stapelberg:
> But breaks all other versions :/

Updated the pull request so that it actually is backwards compatible. The only still failing travis builds were broken before, I believe.

> I put patches into v4.0.x which fix this in a compatible way. Please check that.

…but my pull request is for v4.0.x? Where can I find your patches?

alandekok:
The other commits I have were for v4.0.x.  Look at "git log" would show you.

It's commit 9a450772305 which fixes it for the get server/client random.  Something similar is necessary for `SSL_SESSION_get_master_key()`

stapelberg:
> The other commits I have were for v4.0.x. Look at "git log" would show you.
> 
> It's commit 9a45077 which fixes it for the get server/client random. Something similar is necessary for SSL_SESSION_get_master_key()

Yes, I have seen that code, and my pull request follows exactly the same pattern that the referenced commit uses.

I don’t understand what you’re trying to tell me. Did you look at my pull request? Does it need any changes?

alandekok:
Ah... it's inside of the previously-defined `#ifdef`.  I had missed that, sorry.

The only real comment I have then is that the `SSL_SESSION_get_master_key()` should probably be made `static`.  Nothing else in the server source needs it.  And exporting a public version of that symbol could cause problems for others.

stapelberg:
Done.

PR #1826, opened 2016-11-04
Use -h to check for symlink existance

stapelberg:
-e dereferences symlinks, making the package installation fail when the target does not exist.


alandekok:
Hmm... users can create non-symlink files in mods-enabled.  It probably needs to check if the file is a symlink OR a regular file, and create the symlink only if the check fails.

alandekok:
maybe

    if [ ! ( -h /etc/freeradius/mods-enabled/$mod || -f /etc/freeradius/mods-enabled/$mod ) ]; then

stapelberg:
Updated to use `-h` and `-e` (as the Debian package now does).

PR #1827, opened 2016-11-05
Allow build with openssl 1.1.0b-2

herwinw:
**This should not be merged, but serves as an example**

Which is the current version of Debian Sid. Many things are no longer
public, but have to be accessed through functions.  It does compile, but
I haven't tried running it. The value 1024 is more or less randomly
chosen, I've got no idea if this is too small, or if there is some
constant that gives us the exact length. (And using the magic value
twice is stupid anyway)

It looks like `SSL_SESSION_get_master_key` is not available in OpenSSL before version 1.1.0. Of course the man page at https://www.openssl.org/docs/manmaster/man3/SSL_SESSION_get_master_key.html doesn't mention anything about a version, but downloading the latest 1.0.2 and grepping throught the source didn't make anything show up. `SSL_get_session` appears to be available though.

### DIFF
--- a/src/include/tls-h
+++ b/src/include/tls-h
@@ -107,7 +107,7 @@ do {\
  *	exiting to prevent memory leaks.
  */
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L
-#  define FR_TLS_REMOVE_THREAD_STATE() ERR_remove_thread_state();
+#  define FR_TLS_REMOVE_THREAD_STATE()
 #elif OPENSSL_VERSION_NUMBER >= 0x10000000L
 #  define FR_TLS_REMOVE_THREAD_STATE() ERR_remove_thread_state(NULL);
 #else
--- a/src/main/tls/global.c
+++ b/src/main/tls/global.c
@@ -376,6 +376,8 @@ int tls_global_init(void)
 		ERROR("FATAL: Failed to set up SSL mutexes");
 		return -1;
 	}
+
+	OPENSSL_config(NULL);
 #else
 	OPENSSL_init_crypto(OPENSSL_INIT_LOAD_CONFIG | OPENSSL_INIT_ENGINE_ALL_BUILTIN, NULL);
 #endif
@@ -390,7 +392,6 @@ int tls_global_init(void)
 	ENGINE_register_all_complete();
 
 	tls_done_init = true;
-	OPENSSL_config(NULL);
 
 	return 0;
 }
--- a/src/main/tls/session.c
+++ b/src/main/tls/session.c
@@ -855,7 +855,11 @@ int tls_session_pairs_from_x509_cert(vp_cursor_t *cursor, TALLOC_CTX *ctx,
 
 	int		attr_index, loc;
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+	STACK_OF(X509_EXTENSION) const *ext_list = NULL;
+#else
 	STACK_OF(X509_EXTENSION) *ext_list = NULL;
+#endif
 
 	ASN1_INTEGER	*sn = NULL;
 	ASN1_TIME	*asn_time = NULL;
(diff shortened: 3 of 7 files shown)


herwinw:
Another fun fact: the manual hints that you might want to use SSL_export_keying_material, but there are no working links to the documentation of that function.

alandekok:
The fixes from v4.0.x mppe_keys.c and eap-fast should really be back-ported.  Adding a call to `SSL_SESSION_get_master_key()` is good, but it means that the server no longer builds with older versions of OpenSSL.

herwinw:
I agree about the need to backport this, but I guess it's better to make a working 4.0 first before backporting anything. I'm bumping into a lot more compilation errors (after the first commit the compilation stalled on `rlm_perl.c`, which was not related to OpenSSL (and has been fixed via a different pull request).

On a different note: I never really understood some of the rants about the mess of OpenSSL on the mailings lists. Now I do

herwinw:
I added a bunch of extra commits. It now compiles on my machine, I've got no idea how well this works with older versions of OpenSSL (although I do have my expectations, and they aren't very hopeful). I might try to experiment with a Debian Stable VM (OpenSSL 1.0.1t) tomorrow.

src/modules/rlm_eap/lib/base/mppe_keys.c
```diff
@@ -213,7 +216,8 @@ void eap_tls_gen_mppe_keys(REQUEST *request, SSL *s, char const *prf_label)
 		(void) SSL_get_server_random(s, p, SSL3_RANDOM_SIZE);
 		prf_size += SSL3_RANDOM_SIZE;
 
-		PRF(s->session->master_key, s->session->master_key_length,
+		master_key_length = SSL_SESSION_get_master_key(session, master_key, 1024);
```
arr2036: ``sizeof(master_key)`` please

src/modules/rlm_eap/lib/base/mppe_keys.c
```diff
@@ -257,7 +264,8 @@ void eap_tls_gen_challenge(SSL *s, uint8_t *buffer, uint8_t *scratch, size_t siz
 	(void) SSL_get_server_random(s, p, SSL3_RANDOM_SIZE);
 	p += SSL3_RANDOM_SIZE;
 
-	PRF(s->session->master_key, s->session->master_key_length,
+	master_key_length = SSL_SESSION_get_master_key(session, master_key, 1024);
```
arr2036: ``sizeof(master_key)``

src/modules/rlm_eap/lib/base/mppe_keys.c
```diff
@@ -239,6 +243,9 @@ void eap_tls_gen_challenge(SSL *s, uint8_t *buffer, uint8_t *scratch, size_t siz
 	uint8_t seed[128 + 2*SSL3_RANDOM_SIZE];
 	uint8_t *p = seed;
 	size_t len;
+	SSL_SESSION *session = SSL_get_session(s);
+	unsigned char master_key[1024];
+	size_t master_key_length;
```
arr2036: Just re-use len?

src/modules/rlm_eap/lib/base/mppe_keys.c
```diff
@@ -200,6 +200,9 @@ void eap_tls_gen_mppe_keys(REQUEST *request, SSL *s, char const *prf_label)
 	{
 		uint8_t seed[64 + (2 * SSL3_RANDOM_SIZE)];
 		uint8_t buf[4 * EAP_TLS_MPPE_KEY_LEN];
+		SSL_SESSION *session = SSL_get_session(s);
+		unsigned char master_key[1024];
+		size_t master_key_length;
```
arr2036: Just re-use len?

herwinw:
@arr2036 I'm trying to get it to compile with older openssl versions too (currently abusing travis for this), cleanups can happen afterwards

arr2036:
If you're done, could you squash related commits

alandekok:
I'd really prefer to pull over just the changes, and skip the "Merge branch ..." commits.  Those just pollute the commit log with nonsense.

arr2036:
They should go away automatically when this branch is rebased off v4.0.x

herwinw:
I just pushed one last fix (there was a remaining fixme) and wait if this builds on Travis. If it does, I'll rebase it into some more logical commits

herwinw:
There you go, wrapped up nicely in two commits: The essential to compile, and one to fix the deprecation warnings.

arr2036:
Thanks for your work on this!

PR #1831, opened 2016-11-08
CA path reload interval v3.0.x

lytboris:
Created to ease review


nwkrad:

nwkrad:

alandekok:
Hmm... this should really go in.  I don't have time to do it before 3.0.16, but it will be on my short list after that.

Sorry for the delay.

noahwilliamsson:
Is there any chance this PR could be rebased and merged to both the v3.0.x and master branches?

It would be great to have this feature available so AAA didn't temporarily break during FreeRADIUS restarts because of the need to refresh CRLs.

Thanks!

lytboris:
@alandekok , do you have a milestone to merge it in?
It would be a bit painful to re-write this PR so it could be merged into today's 3.0.x and wait forever for a merge still.

alandekok:
It's been too long, sorry.

Yes, we should add this to v3.

bobmc67:
The lack of this feature has bugged me for years (using FreeRADIUS for 802.1x EAP-TLS auth) so I can't wait to see this merged and later show up in Long-Term Support releases of popular Linux distributions.

lytboris:
OpenSSL 1.1.1 support is done somewhat wrong, it makes radiusd crash upon store swap. Will update patch once I find a way to fix that.

PR #1835, opened 2016-11-09
Allow authentication retry in winbind

qnet-herwin:
A setup with the following properties:

  * Active Directory backend
  * FreeRadius with eap-inner-proxy
  * Windows client with single sign-on
  * User using different casing in username than in backend

may result in failing connections. It looks like Windows reads the
correct username from the domain server once it has logged in, and uses
that to create the MS-CHAP2-Response attribute. The User-Name attribute
is still the one with the incorrect casing, causing the authentication
to fail.

The introduced config option kicks in after a failed authentication: it
reads the correct username from the backend, tries another
authentication, and uses the found User-Name to calculate
MS-CHAP2-Response if the second authentication works.

### DIFF
--- a/raddb/mods-available/mschap
+++ b/raddb/mods-available/mschap
@@ -78,6 +78,15 @@ mschap {
 #	winbind_username = "%{mschap:User-Name}"
 #	winbind_domain = "%{mschap:NT-Domain}"
 
+	# When using single sign-on with a winbind connection and the
+	# client uses a different casing for the username than the
+	# casing is according to the backend, reauth may fail because
+	# of some Windows internals. This switch tries to find the
+	# user in the correct casing in the backend, and retry
+	# authentication with that username.
+	#
+#	winbind_retry_with_normalised_username = no
+
 	#
 	#  Information for the winbind connection pool.  The configuration
 	#  items below are the same for all modules which use the new
--- a/src/modules/rlm_mschap/auth_wbclient.c
+++ b/src/modules/rlm_mschap/auth_wbclient.c
@@ -35,6 +35,42 @@ RCSID("$Id$")
 
 #define NT_LENGTH 24
 
+/** Use Winbind to normalise a username
+ *
+ * @param[in] tctx The talloc context where the result is parented from
+ * @param[in] ctx The winbind context
+ * @param[in] dom_name The domain of the user
+ * @param[in] name The username (without the domain) to be normalised
+ * @return The username with the casing according to the Winbind remote server,
+ *         or NULL if the username could not be found.
+ */
+static char *wbclient_normalise_username(TALLOC_CTX *tctx, struct wbcContext *ctx, char const *dom_name, char const *name)
+{
+	struct wbcDomainSid sid;
+	enum wbcSidType name_type;
+	wbcErr err;
+	char *res_domain = NULL;
+	char *res_name = NULL;
+	char *res = NULL;
+
+	/* Step 1: Convert a name to a sid */
+	err = wbcCtxLookupName(ctx, dom_name, name, &sid, &name_type);
+	if (!WBC_ERROR_IS_OK(err))
+		return NULL;
+
+	/* Step 2: Convert the sid back to a name */
+	err = wbcCtxLookupSid(ctx, &sid, &res_domain, &res_name, &name_type);
+	if (!WBC_ERROR_IS_OK(err))
+		return NULL;
+
+	MEM(res = talloc_strdup(tctx, res_name));
+
+	wbcFreeMemory(res_domain);
+	wbcFreeMemory(res_name);
+
+	return res;
+}
+
 /*
  *	Check NTLM authentication direct to winbind via
  *	Samba's libwbclient library
@@ -49,7 +85,7 @@ int do_auth_wbclient(rlm_mschap_t *inst, REQUEST *request,
 		     uint8_t nthashhash[NT_DIGEST_LENGTH])
 {
 	int rcode = -1;
-	struct wbcContext *wb_ctx;
+	struct wbcContext *wb_ctx = NULL;
 	struct wbcAuthUserParams authparams;
 	wbcErr err;
 	int len;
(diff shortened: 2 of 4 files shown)


qnet-herwin:
This change has been written for 3.0.x because we had the problem here. If it is accepted, it should be merged into 4.0.x as well (but that may require some rework, since the winbind code has been moved to a separate module)

alanbuxey:
dont forget 3.1 - most people are using 3.1 because of the winbind
module/functions (and enhanced caching...)

mcnewton:
Nasty :-( (AD, not your code).

Would it not just be better to always normalise the username and not do the double retry? This seems like something that should Just Work and not need an option to configure.

I don't know what the overhead of doing the double name<->SID lookup is for every auth, versus doing a double auth and double lookup in the failure case. It might be that after the first auth someone has done then the SID being cached by winbind makes the lookup for subsequent auths negligible and therefore might as well just do it every time anyway.

Matthew

mcnewton:
@qnet-herwin when you say the auth fails, does the client retry again and get on, or is that a permanent fail until they retype their username "correctly" (as far as AD cares)?

qnet-herwin:
> Would it not just be better to always normalise the username and not do the double retry? This seems like something that should Just Work and not need an option to configure.

What fun would it be if it was that simple?
But on a more serious note: The change of the username used to calculate the hashes is only done once a use is logged on into the system. Always normalizing the username would result in a reject for the first login attempt. I guess it fetches the correct casing of the username from AD once the client has logged in and uses that for the MSCHAP-hashes for the length of the session. A more correct implementation should be to check if the client has an active session and only perform the normalization if that is the case, but that would require some kind of state tracking which is a bit outside the scope of the winbind libs.


> when you say the auth fails, does the client retry again and get on, or is that a permanent fail until they retype their username "correctly" (as far as AD cares)?

That is a permanent error. It is seen as an incorrect username or password without retry (see https://tools.ietf.org/html/rfc2759#section-6)

Sometimes it works to logout and login again after a failure, because Windows will use the normalized username in both the User-Name attribute/EAP-Identity and the hashes. But not always. Things like this really make testing your workaround code a PITA (hooray, it works, ow wait, that's not because of my changes)


> dont forget 3.1 - most people are using 3.1 because of the winbind module/functions (and enhanced caching...)

I kind of remembered that rlm_winbind was new in 4.0, guess I was wrong. Next time I'll look instead of guess ;)

qnet-herwin:
Any change that this could be merged? ("No" is a perfectly valid answer as well, then we know we have to integrate it into our own builds)

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -186,6 +265,7 @@ int do_auth_wbclient(rlm_mschap_t *inst, REQUEST *request,
 
 
 done:
+	fr_connection_release(inst->wb_pool, wb_ctx);
```
mcnewton: This can't go here like this - 'done' is sometimes called before wb_ctx is initialised.

Maybe initialise wb_ctx = NULL, then replace this with
```
if (wb_ctx) ...
```

src/modules/rlm_mschap/auth_wbclient.c
```diff
@@ -147,6 +181,51 @@ int do_auth_wbclient(rlm_mschap_t *inst, REQUEST *request,
 		REDEBUG2("Domain not found");
 		break;
 	case WBC_ERR_AUTH_ERROR:
+		if (inst->wb_retry_with_normalized_username) {
```
mcnewton: Rather than this chunk of code here, if it is put in before the switch with just a simple
```
if (err == WBC_ERR_AUTH_ERROR && inst->wb_retry_with_normalized_username) {
  ...
}
```
then the existing err handling and useful error messages will be displayed after a username rewrite.

Otherwise if the actual error results in e.g. password expired, that will be missed if the auth is retried.

You can then also leave the fr_connection_release before the switch (and after this if() ), avoiding the other issue I just wrote.

mcnewton:
Finally had some time to take a look at this. Some comments added. As well as the general annoyance at Microsoft. Why couldn't usernames always be treated as lowercase in the first case? It's not as if the system will let you have 'user' and 'User' anyway, so the hash should be calculated the same for both...

raddb/mods-available/mschap
```diff
@@ -78,6 +78,15 @@ mschap {
...
 #	winbind_domain = "%{mschap:NT-Domain}"
 
+	# When using single sign-on with a winbind connection and the
+	# client uses a different casing for the username than the
+	# casing is according to the backend, reauth may fail because
+	# of some Windows internals. This switch tries to find the
+	# user in the correct casing in the backend, and retry
+	# authentication with that username.
+	#
+# winbind_retry_with_normalized_username = no
```
arr2036: Server uses British/Canadian spelling.   Authorize is an aberration.

qnet-herwin:
```
src/modules/rlm_mschap/rlm_mschap.c:                    RWDEBUG("NT-Password has not been normalized by the 'pap' module (likely still in hex format).  "
src/modules/rlm_mschap/rlm_mschap.c:                    RWDEBUG("LM-Password has not been normalized by the 'pap' module (likely still in hex format).  "
```

I'll leave those two as they are now, that's not what this patch is aboot ;)

qnet-herwin:
Thanks for the review @mcnewton. All points looked valid, and all points have been adressed in the branch now.

I made separate commits for every review point, so it's easier to validate the changes based on the review. Of course we could alwasy squash it to a single commit before merging if desired.

mcnewton:
@qnet-herwin thanks - can you rebase to v3.0.x and squash this on top as a single commit, and I'll try and look at it again today.

qnet-herwin:
> can you rebase to v3.0.x and squash this on top as a single commit

done

alanbuxey:
..and 3.1.x  ? ;-)

mcnewton:
@alanbuxey patience, patience...

qnet-herwin:
@alanbuxey I'll look into 3.1/4.0 later this week

mcnewton:
@qnet-herwin I'm doing it now.

qnet-herwin:
@mcnewton I think I still have a working test setup here, just notify me if you have the code but aren't able to test it

mcnewton:
pushed to v4.0.x & v3.1.x. Thanks @qnet-herwin

PR #1839, opened 2016-11-10
Make Freeradius 3.1.x compile with OpenSSL 1.1

herwinw:
This only fixes the compile errors, there are still a lot of warnings
(deprecated functions, missing const quantifiers)

And once again: I'm abusing Travis to see if it still compiles with older versions of OpenSSL


herwinw:
That should do it. Two down, one to go

PR #1842, opened 2016-11-12
Use 'openssl dhparam' to create DH-params

herwinw:
'openssl gendh' is removed in newer version of OpenSSL, and using
dhparam appears to work since at least release 0.9.8


herwinw:
The build error looks pretty much unrelated to the change

PR #1844, opened 2016-11-12
Use stack based char array to generate JSON keys

herwinw:
This saves us from having to allocate memory in a loop. It simplifies
the code a little bit too, by removing an if statement that was far away
from the code that generated the requirement to call it.


alandekok:
The only minor suggestion is to use a better size for the on-stack buffer.  i.e. `FR_DICT_ATTR_MAX_NAME_LEN + 32`.  That way if the maximum name length changes, the buffer will automatically re-size.

herwinw:
The value 255 was more or less random, changed the buffersize to the suggested value.

PR #1849, opened 2016-11-17
Fix ordering or DHCP options

pwdng:


arr2036:
What's the reason for needing this ordering scheme?

pwdng:
Without it, options don't get grouped together and some clients don't like it.
In my case, a relay needs to have all options 82 packed together or it just drops the reply.

alandekok:
Makes sense.  RADIUS has similar requirements, that attributes split into fragments MUST all appear next to each other in the packet.

arr2036:
These aren't fragments, these are multiple instances of options.  The DHCP clients are arguably broken, but there's probably no harm in doing it.  The TLV sorting code doesn't work for multiple-levels of nested TLVs though, which you can have for option 82 IIRC... Um ha, yes, technically option 82 suboption 7 can have nested RADIUS attributes... but ignoring that, suboption 9 is a VSA embedded within option 82.

arr2036:
You need to build the TLV stack and then walk from the root towards the leaf doing the comparisons.

arr2036:
Look at the RADIUS encoder, it has an example... and there are already support functions to do it. Can probably create a fast path if the parent is not a TLV.

pwdng:
Yes, I realised this wouldn't work for multiple tested options but given it's not currently possible to have a depth of more than 1, I left it there.
I'll look at making it more generic.

alandekok:
travis tests should be fixed.  But other than that it looks good.

arr2036:
It doesn't look good, it looks like code i'm going to fix at some point when vendor specific attributes are reintroduced back into the DHCP parser.  I'd prefer that it be done right the first time instead of doing it the braindead hacky way and pushing the real work onto someone else.

arr2036:
Does this do what you want? 

https://github.com/FreeRADIUS/freeradius-server/commit/3ddc2d8bca823dbbf3836a7d2f0f2f4e9f59c7d2

arr2036:
I think the deeper attribute hierarchy check might be reversed, but not sure it's important.

arr2036:
ah no, need to check each of the attributes individually to see if their parent is a TLV

https://github.com/FreeRADIUS/freeradius-server/commit/64c113f1d851235f3843e4dda43552f2305e3340

pwdng:
They were some minor issues I fixed in https://github.com/FreeRADIUS/freeradius-server/pull/1852
But yes, that it a more generic implementation :-)

PR #1851, opened 2016-11-19
Fixed pair comparison.

tuyan:
Fixed pair comparison, possibly meant to compare both sides. Cookie goes to -Wtautological-compare .


alandekok:
The code is definitely wrong.  I'll push a different fix.  Your changes just make the modified checks the same as the line immediately above them.

PR #1852, opened 2016-11-21
Fix typos and option 82 handling in DHCP sort changes

pwdng:


arr2036:
Thanks! 

But blerg even that's not quite right, I hate nested attributes.  I'll push another fix for the OPTION_82 check, we need to test the heritage of the attribute :/

arr2036:
OK I think 1f821b021fa04953acbe77418cb83944d04b2bd7 should be the last of it.

PR #1861, opened 2016-11-27
Fix build issues with sd_watchdog_event

herwinw:
Probably not the best fix around, but at least it fixes the build
process


arr2036:
Looks good, thanks 👍

PR #1863, opened 2016-11-28
DHCP fix

pwdng:


src/modules/proto_dhcp/dhcp.c
```diff
@@ -1348,10 +1348,10 @@ int8_t fr_dhcp_attr_cmp(void const *a, void const *b)
...
 	 *	Check if either of the options are descended from option 82.
 	 */
-	a_82 = fr_dict_parent_common(dhcp_option_82, my_a->da, true);
-	b_82 = fr_dict_parent_common(dhcp_option_82, my_b->da, true);
-	if (a_82 && !b_82) return +1;
-	if (!a_82 && !b_82) return -1;
+	ancestor_a = fr_dict_parent_common(dhcp_option_82, my_a->da, true) ;
+	ancestor_b = fr_dict_parent_common(dhcp_option_82, my_b->da, true);
+	if (ancestor_a && ancestor_a->attr == PW_DHCP_OPTION_82 && !ancestor_b) return +1;
+	if (!ancestor_a && ancestor_b && ancestor_b->attr == PW_DHCP_OPTION_82) return -1;
```
arr2036: Ah, yes, the ancestor can be the dictionary root, doh!

src/modules/proto_dhcp/dhcp.c
```diff
@@ -1070,9 +1067,8 @@ ssize_t fr_dhcp_decode_option(TALLOC_CTX *ctx, vp_cursor_t *cursor,
 
 		for (i = 1; i < data_len; i++) {
 			if (p[i] != 0) {
-				fr_strerror_printf("%s: Non padding option follows end of options signifier",
-						   __FUNCTION__);
-				return -1;
+				WARN("DHCP: Non padding option follows end of options signifier");
+				break;
```
arr2036: As per rfc 2131:

```text

4.1 Constructing and sending DHCP messages

   DHCP clients and servers both construct DHCP messages by filling in
   fields in the fixed format section of the message and appending
   tagged data items in the variable length option area.  The options
   area includes first a four-octet 'magic cookie' (which was described
   in section 3), followed by the options.  The last option must always



Droms                       Standards Track                    [Page 22]

RFC 2131          Dynamic Host Configuration Protocol         March 1997


   be the 'end' option.
```

pwdng:
Sure, but unfortunately, there are some appliances that are buggy and include junk data after the ‘end’ marker.
This doesn’t follow the RFC but given the rest of the packet is fine, it seems silly to discard the packet entirely.
We could have an extra configuration flag to be flexible and process these requests if you think the preferred default behaviour should be to drop them.

arr2036:
Are the options after the end important?  Could we just stop parsing the packet when we see the end option?

alandekok:
When we see "end of option", we should stop processing the packet.  i.e. all subsequent data is ignored.

src/modules/proto_dhcp/dhcpclient.mk
```diff
@@ -1,5 +1,5 @@
 TARGET		:= dhcpclient
 SOURCES		:= dhcpclient.c dhcp.c
 
-TGT_PREREQS	:= libfreeradius-radius.a
-TGT_LDLIBS	:= $(LIBS)
+TGT_PREREQS	:= libfreeradius-radius.a libfreeradius-server.a
```
alandekok: why does dhcpclient now have to link to libfreeradius-server?

The idea of `dhcp.c` was that it was code which could work "as is" in either the client or the server.

I think the correct fix here is just to update the `WARN` message, and leave the rest of the headers / linking alone.

alandekok:
can you just squash the commits so that we don't have "forward / backward" commits?

i.e. `git rebase -i HEAD ^^^^^` and then pick the commits and use `s` to squash them.  Then, `git push --force` your branch, and the pull request will get updated.

Thanks.

arr2036:
Could you comment on the reason for all the other changes? Changing the arguments for the option encoders/decoders is not an option unless you change the RADIUS attribute encoder/decoder...

pwdng:
I want to be able to ignore whatever comes after the end of options marker.
I therefore need to get this information returned from  fr_dhcp_decode_option.
Currently, the function returns the length of decoded data and 0 is an error.
That is why I changed the return argument to be the status and made the length argument in out.

I agree that it’s preferable to keep the prototypes consistent.
Looking at the current code, it doesn’t seem possible to return 0 so I guess we can break on length == 0 and not consider it to be an error.
It’s that or changing the prototypes for all the encoding/decoding functions.
Having the length parameter be in out and returning the status might be a cleaner option but if there is no real use case, it’s probably not worth it.

arr2036:
But we can just return the number of junk bytes at the end of the packet?

pwdng:
Except that you do want to break the while loop in dhcp_options_xlat in case you’ve reached the end of options marker.
You don’t want to consider anything after that as a possible option.
That’s why you need a special return value.

arr2036:
```c

	for (vp = tmpl_cursor_init(NULL, &src_cursor, request, &src);
	     vp;
	     vp = tmpl_cursor_next(&src_cursor, &src)) {
		uint8_t const	*p = vp->vp_octets, *end = p + vp->vp_length;
		ssize_t		len;
		VALUE_PAIR	*vps = NULL;
		vp_cursor_t	options_cursor;

		fr_cursor_init(&options_cursor, &vps);
		/*
		 *	Loop over all the options data
		 */
		while (p < end) {
			len = fr_dhcp_decode_option(request->packet, &options_cursor,
						    fr_dict_root(fr_dict_internal), p, end - p, NULL);
			if (len <= 0) {
				RWDEBUG("DHCP option decoding failed: %s", fr_strerror());
				fr_pair_list_free(&head);
				goto error;
			}
			p += len;
		}
		fr_cursor_merge(&cursor, vps);
	}
```

Looks like it would?

pwdng:
Sorry, got the wrong function. fr_dhcp_decode is the one:

		while (p < end) {
			len = fr_dhcp_decode_option(packet, &cursor, fr_dict_root(fr_dict_internal),
						    p, ((end - p) > UINT8_MAX) ? UINT8_MAX : (end - p), NULL);
			if (len <= 0) {
				fr_pair_list_free(&head);
				return len;
			}
			p += len;
		}

We’d have a problem if (end - p) > UINT8_MAX

arr2036:
I think we can just replace that with ``end - p``, it was a safety thing.

alandekok:
Why do we care about the junk bytes at the end of the packet?

If anything, we should have a check as is done in RADIUS.  e.g. if the UDP packet is larger than the RADIUS packet, we just reset the UDP packet size.

i.e. fr_dhcp_packet_ok() should really just walk through the options, and verify that they make sense.  It doesn't do that right now.  Once it's done walking through the options, it can set the *actual* packet size.

and then we don't need to change the decoder.

arr2036:
I think the original issue has been fixed. I'll open something for v4.0.x about the verify function.

PR #1885, opened 2017-01-20
Update elasticsearch examples

mcnewton:


mcnewton:
Still hoping that maybe one day Travis will have the intelligence to work out the difference between "code" and "examples"...

alanbuxey:
can it (travis) not be told to ignore certain sub-directories?

mcnewton:
I guess it might be possible to do as part of the Travis build - check modified files early on and exit with success if they are all out of the source dirs. No idea how exactly though.

PR #1901, opened 2017-02-07
remove Reply-Message from Access-Challenge if EAP-Message

alanbuxey:
RFC 3579 doesn't allow Reply-Message if EAP-Message exists..which it may in Access-Challenge.
We already do a basic filter in the Auth-Type EAP but thats to match 2865 and gets a little tricky/messy if its not an EAP challenge-response..so best cleared up here.


arr2036:
This was merged manually

Issue #1909, opened 2017-02-13
radrelay do not start


jimdigriz:
# Issue type
- [ ] Defect - Crash or memory corruption.
- [ ] Defect - Non compliance with a standards document, or incorrect API usage.
- [X] Defect - Unexpected behaviour (obvious or verified by project member).
- [ ] Feature request.

# Defect/Feature description

Instructions say that `radrelay` should work, it does not.

## Output of ``[radiusd|freeradius] -X`` showing issue occurring

````
alex@90f259b96e75:/usr/src/networkradius/freeradius-server$ ./build/bin/local/radiusd -X -d raddb -n radrelay
Info  : FreeRADIUS Version 4.0.0
Info  : Copyright (C) 1999-2017 The FreeRADIUS server project and contributors
Info  : There is NO warranty; not even for MERCHANTABILITY or FITNESS FOR A
Info  : PARTICULAR PURPOSE
Info  : You may redistribute copies of FreeRADIUS under the terms of the
Info  : GNU General Public License
Info  : For more information about these matters, see the file named COPYRIGHT
Info  : Starting - reading configuration files ...
Error : Unable to open file "raddb/modules/always": No such file or directory
Error : Error reading or parsing raddb/radrelay.conf
````

Fixing the `rlm_always` reference (which if you remove the) you get:

````
alex@90f259b96e75:/usr/src/networkradius/freeradius-server$ ./build/bin/local/radiusd -X -d raddb -n radrelay
[snipped]
Debug :         acct_pool = radrelay
Debug :  }
Debug : radrelay: #### Loading Clients ####
Info  : Found debugger attached
Warn  : No 'thread pool {..}' found.  Server will be single threaded
Error : No virtual servers found
````

alandekok:
fixed in pull request.

PR #1910, opened 2017-02-13
fix radrelay to...erm...work

jimdigriz:
Fixes issue https://github.com/FreeRADIUS/freeradius-server/issues/1909


arr2036:
Probably should be back ported to v3

Issue #1876, opened 2017-01-12
Crash with EAP-PWD


pauldekkers:
# Issue type
- [X] Defect - Crash or memory corruption.

# Defect/Feature description

EAP-PWD crashes FreeRADIUS (tested on CentOS 6, CentOS 7 and Ubuntu 16.04)

## How to reproduce issue

Compiled FreeRADIUS 3.0.x (git, 3.0.13 at time of writing) with:
./configure --prefix=/home/centos/freeradius --enable-developer

Crash both experienced with ldap and files module for accounts.

Minimal FreeRADIUS config is 
- a user in raddb/mods-config/files/authorize
paul	Cleartext-Password := "testing123"
- eap_pwd enabled in raddb/mods-enabled/eap

eapol_test file:
```
network={
        key_mgmt=IEEE8021X
	eap=PWD
        identity="paul"
        password="testing123"
        }
```
## Full backtrace from LLDB or GDB

Backtrace:
```
(gdb) run -X
Starting program: /home/centos/freeradius/sbin/radiusd -X
[...]
(2) files: users: Matched entry paul at line 1
(2)     [files] = ok
(2)     [expiration] = noop
(2)     [logintime] = noop
(2) pap: No User-Password attribute in the request.  Cannot do PAP
(2)     [pap] = noop
(2)   } # authorize = ok
(2) eap_pwd: } # server inner-tunnel
(2) eap_pwd: Got tunneled reply code 0

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff745b79d in EVP_PKEY_CTX_ctrl () from /usr/lib64/libcrypto.so.10
Missing separate debuginfos, use: debuginfo-install glibc-2.12-1.192.el6.x86_64 keyutils-libs-1.4-5.el6.x86_64 krb5-libs-1.10.3-57.el6.x86_64 libcom_err-1.41.12-22.el6.x86_64 libpcap-1.4.0-4.20130826git2dbcaa1.el6.x86_64 libselinux-2.0.94-7.el6.x86_64 libtalloc-2.1.5-1.el6_7.x86_64 nss-softokn-freebl-3.14.3-23.3.el6_8.x86_64 openssl-1.0.1e-48.el6_8.3.x86_64 zlib-1.2.3-29.el6.x86_64
(gdb) bt
#0  0x00007ffff745b79d in EVP_PKEY_CTX_ctrl () from /usr/lib64/libcrypto.so.10
[...]

PR #1916, opened 2017-02-20
V3.0.x openssl 1.1 fix 2

spbnick:
Some more fixes for OpenSSL v1.1.0 compatibility.
Not actually tested apart from building.

### DIFF
--- a/raddb/certs/Makefile
+++ b/raddb/certs/Makefile
@@ -50,7 +50,7 @@ passwords.mk: server.cnf ca.cnf client.cnf
 #
 ######################################################################
 dh:
-	$(OPENSSL) gendh -out dh -2 $(DH_KEY_SIZE)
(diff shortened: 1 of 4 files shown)


spbnick:
@mcnewton, could you please take a look at my rlm_eap_pwd changes? They touch what you were fixing recently. Thank you.

mcnewton:
@spbnick not me? Or I've been coding in my sleep again.

spbnick:
@mcnewton, no, it's you after all, even though I failed to verify that before calling you :) I'm talking about the fix in #1876. The a28b2a49dff in particular. You added calls to HMAC_CTX_init and now I removed them.

mcnewton:
Oh, yeah, I did. My memory has gone, I don't even remember that :-(

So from the look of it you've kept initialising the context by now calling HMAC_CTX_new rather than HMAC_CTX_init? As long as #1876 doesn't come back.

PR #1919, opened 2017-02-21
Handle hostnames in fr_pton4/6

spbnick:
Make fr_pton4/6 handle hostnames longer than the longest address +
prefix.

This fixes the issue where a long hostname would be rejected by e.g. radclient as an invalid IP address. E.g.:

    $ radclient -x foofoof.barbarb.bazbazb.com auth testing123
    radclient: Invalid IPv4 address string "foofoof.barbarb.bazbazb.com"

Whereas the expected response (produced after the fix) is:

    radclient: Failed resolving "foofoof.barbarb.bazbazb.com" to IPv4 address: Name or service not known

### DIFF
--- a/src/lib/misc.c
+++ b/src/lib/misc.c
@@ -302,10 +302,12 @@ static int ip_prefix_from_str(char const *str, uint32_t *paddr)
 }
 
 
-/** Parse an IPv4 address or IPv4 prefix in presentation format (and others)
+/**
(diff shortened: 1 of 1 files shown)


arr2036:
Just increase the buffer size to 256.   This has already been changed in v4.0.x.

There's actually a 255 byte limit defined in RFC 1035 Section 2.3.4.

spbnick:
Ah, I looked but failed to find that RFC. Updated, thanks, Arran.

PR #1921, opened 2017-02-22
Clarify ipaddr resolution preference

spbnick:
Note in the raddb/sites-available/default that the "ipaddr" option
always prefers A records on hostname resolution, regardless of system
configuration (such as "/etc/gai.conf"), to clear up the possible
confusion.


alandekok:
I've pushed clarifying text in commit 99e08b85b3.  The "depending on system configuration" text was a unclear to me.

The updated text documents each option individually, and gives explanations as to how they work, with examples.

spbnick:
That's much better! Thanks a lot, Alan! There is only one little issue: you write "AAA" instead of "AAAA" in the "ipaddr" description there.

Issue #1928, opened 2017-03-06
newline missing in radclient stats


pauldekkers:
In `src/main/radclient.c`:

```
                printf("Packet summary:\n"
                       "\tAccepted      : %" PRIu64 "\n"
                       "\tRejected      : %" PRIu64 "\n"
                       "\tLost          : %" PRIu64 "\n"
                       "\tPassed filter : %" PRIu64 "\n"
                       "\tFailed filter : %" PRIu64,
```

I think Failed filter should have a newline too.

pauldekkers:
Let's not be lazy ;-) I'll create a pull request for this!

pauldekkers:
Fixed in 8962e549cc5d344263e5df3837121e69c64c6743 / #1929

Issue #1926, opened 2017-03-03
Segfault in radclient with -x -q [-s]


pauldekkers:
# Issue type

- [X] Defect - Crash or memory corruption.

# Defect/Feature description

I noticed radclient -s wasn't showing statistics anymore (since I've used it in 2.x.x), unless I combine this with -x. Also adding -q however to only show statistics and not every packet makes radclient segfault.

The problem is with -x and -q; the process segfaults before any request is done.

As for -s, I think it makes sense to show stats without -x?

## How to reproduce issue

Tested this on Ubuntu 16.04.2 LTS with FreeRADIUS 3.0.x (.13) from today.

## Full backtrace from LLDB or GDB

```
# cat test1
User-Name=test@test,User-Password=testing123,NAS-Port=1
```

...

```
(gdb) run -x -s -q -f test1 10.0.0.1 auth testing123
Starting program: /opt/freeradius/bin/radclient -x -s -q -f test1 10.0.0.1 auth testing123
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
__GI__IO_fputs (str=0x7fffffffdeb0 "\tUser-Name = \"test@test\"\n", fp=0x0) at iofputs.c:36
36	iofputs.c: No such file or directory.
(gdb) bt
#0  __GI__IO_fputs (str=0x7fffffffdeb0 "\tUser-Name = \"test@test\"\n", fp=0x0) at iofputs.c:36
#1  0x00007ffff7baaf29 in vp_print (fp=0x0, vp=0x7931b0) at src/lib/print.c:727
#2  0x00007ffff7baaf95 in vp_printlist (fp=0x0, const_vp=0x7931b0) at src/lib/print.c:744
#3  0x0000000000404942 in send_one_packet (request=0x617600) at src/main/radclient.c:981
#4  0x0000000000405e94 in main (argc=4, argv=0x7fffffffe660) at src/main/radclient.c:1545
```

pauldekkers:
This resolves the crash, but there's still no statistics shown with just -s (without -x), right? (IMHO, I think using -s also with -q is a good combination to just get a summary. This was the behaviour in 2.x.x if I'm not mistaken.)

alandekok:
I've pushed  fix.

PR #1929, opened 2017-03-06
print newline after stats.failed in summary

pauldekkers:
As described in #1928 there's a newline missing now we have printf instead of DEBUG as of the last fix in #1926


alandekok:
I've pushed  fix.

PR #1939, opened 2017-03-13
Add --disable-openssl-compat-check configure option

spbnick:
Add support for --disable-openssl-compat-check configure option, which
disables checking whether the linked OpenSSL (or related) libraries are
compatible with the ones FreeRADIUS was built with.

This is needed for distribution, which are tracking OpenSSL ABI changes
themselves and make sure shared library major and minor numbers change
if there is any breakage, thus ensuring incompatible applications and
libraries are not linked. That is also supported by the new OpenSSL
upstream policy to keep ABI compatible between "fix" and "patch"
version numbers.

See also https://bugzilla.redhat.com/show_bug.cgi?id=1299388

### DIFF
--- a/configure
+++ b/configure
@@ -757,6 +757,7 @@ with_openssl
 with_openssl_lib_dir
 with_openssl_include_dir
 enable_openssl_version_check
+enable_openssl_compat_check
 with_talloc_lib_dir
 with_talloc_include_dir
 with_pcap_lib_dir
@@ -1404,6 +1405,9 @@ Optional Features:
   --enable-werror         causes the build to fail if any warnings are generated.
   --disable-openssl-version-check
                           disable vulnerable OpenSSL version check
+
+  --disable-openssl-compat-check
+                          disable built/linked OpenSSL compatibility check
 
 
 Optional Packages:
@@ -5466,6 +5470,17 @@ else
   openssl_version_check_config=
 fi
 
+
+# Check whether --enable-openssl-compat-check was given.
+if test "${enable_openssl_compat_check+set}" = set; then :
+  enableval=$enable_openssl_compat_check;
+fi
+
+if test "x$enable_openssl_compat_check" != "xno"; then
+
+$as_echo "#define ENABLE_OPENSSL_COMPAT_CHECK 1" >>confdefs.h
+
+fi
 
 
 
--- a/configure.ac
+++ b/configure.ac
@@ -597,6 +597,18 @@ else
 fi
 AC_SUBST([openssl_version_check_config])
 
+dnl #
+dnl #  extra argument: --disable-openssl-compat-check
+dnl #
+AC_ARG_ENABLE(openssl-compat-check,
+[AS_HELP_STRING([--disable-openssl-compat-check],
+                [disable built/linked OpenSSL compatibility check])]
+)
+if test "x$enable_openssl_compat_check" != "xno"; then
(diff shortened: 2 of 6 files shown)


mcnewton:
I thought this came up before a couple of years ago and that there was already a way to do this? (Or that the recommended way was for distributions to apply their own packaging patches to remove the appropriate checks if they really needed to.)

If it's made too easy then too many people start to use it and then complain that FR is broken, when it was actually OpenSSL's ABI that got subtly changed and they had disabled the checks.

arr2036:
The current check only complains if there are major, minor, fix differences.

It's not just about ABI compatibilities, even if 1.0.0 and 1.0.1 are completely ABI compatible, it doesn't guarantee that the OpenSSL API behaves in the same way.  I can't remember the exact versions but we did see a breaking change between "fixes", where additional informational messages were passed to one of our registered callbacks, breaking EAP-TLS and friends.

I am very much against merging this, and very much against packagers doing this.  We are following the same methodology as the OpenSSH project, and other Open Source projects that have a dependency on OpenSSL.

spbnick:
Thank you for a quick response, Matthew, Arran.

I would appreciate it, if you visited the linked Red Hat bug, and explained
your concerns to Nikos and Tomas there, instead of me trying to convey your
points through the filter of my limited understanding and experience in
this matter.

Thank you.


On Mar 13, 2017 5:19 PM, "Arran Cudbard-Bell" <notifications@github.com>
wrote:

The current check only complains if there are major, minor, fix differences.

It's not just about ABI compatibilities, even if 1.0.0 and 1.0.1 are
completely ABI compatible, it doesn't guarantee that the OpenSSL API
behaves in the same way. I can't remember the exact versions but we did see
a breaking change between "fixes", where additional informational messages
were passed to one of our registered callbacks, breaking EAP-TLS and
friends.

I am very much against merging this, and very much against packagers doing
this. We are following the same methodology as the OpenSSH project, and
other Open Source projects that have a dependency on OpenSSL.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/FreeRADIUS/freeradius-server/pull/1939#issuecomment-286139533>,
or mute the thread

alandekok:
OpenSSL now has public APIs for (almost) all of it's functionality.  Which means that after OpenSSL version 1.1.0, we can likely disable this check.  I'm still in favor of keeping it for earlier revisions, due to the issues noted by Arran.

I'd also prefer to keep the check for OpenSSL > 1.1.0,  but only issue a warning message, instead of refusing to start.  That way if there is an OpenSSL compatibility issue, we can tell from the warning message.

alandekok:
See the 3.0.11 ChangeLog:

	* Fix compatiblity issues with OpenSSL 1.0.2.  Ignore
	  calls to msg_callback with 'pseudo' content types.

And commit b7b5493c61a

spbnick:
Thank you, Alan.

As you might have noticed, for now we made our FreeRADIUS packages require OpenSSL we built with, or a later version. According to Tomas Mraz, our OpenSSL maintainer, this should help. If we still see issues, I'll submit a pull request with the relaxed check and a warning for OpenSSL above 1.1.0, as you describe.

alandekok:
Thanks.  For v3.0.14, we'll probably add the warning for OpenSSL > 1.1.0.  At that point, everything should be done via the public APIs.  Which should avoid crashing issues due to OpenSSL struct changes.

PR #1940, opened 2017-03-13
Async sql

pwdng:
Initial work to make rlm_sql async

### DIFF
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -1410,9 +1410,6 @@ static int mod_instantiate(CONF_SECTION *conf, void *instance)
 	 */
 	INFO("Attempting to connect to database \"%s\"", inst->config->sql_db);
 
-	inst->pool = module_connection_pool_init(inst->cs, inst, mod_conn_create, NULL, NULL, NULL, NULL);
-	if (!inst->pool) return -1;
-
 	if (inst->config->do_clients) {
 		if (generate_sql_clients(inst) == -1){
 			ERROR("Failed to load clients from SQL");
@@ -1423,10 +1420,42 @@ static int mod_instantiate(CONF_SECTION *conf, void *instance)
 	return RLM_MODULE_OK;
 }
 
+/** Create a thread specific pool
+ *
+ * Each request uses its own connection.
+ * Each thread therefore needs a pool of connections to handle multiple requests in parallel
+ *
+ * @param[in] conf		section containing the configuration of this module instance.
+ * @param[in] instance	of rlm_sql_t.
+ * @param[in] el		The event list serviced by this thread.
+ * @param[in] thread	specific data.
+ * @return
+ *	- 0 on success.
+ *	- -1 on failure.
+ */
+static int mod_thread_instantiate(CONF_SECTION const *conf, void *instance, fr_event_list_t *el, void *thread)
 {
+	rlm_sql_t			*inst = talloc_get_type_abort(instance, rlm_sql_t);
+	rlm_sql_thread_t	*t = talloc_get_type_abort(thread, rlm_sql_thread_t);
+	CONF_SECTION		*my_conf;
+
+	t->el = el;
+	t->inst = instance;
 
 	/*
+	 *	Temporary hack to make config parsing
+	 *	thread safe.
 	 */
+	my_conf = cf_section_dup(NULL, conf, cf_section_name1(conf), cf_section_name2(conf), true);
+	t->pool = fr_connection_pool_init(NULL, my_conf, t, mod_conn_create, NULL, inst->name);
+	talloc_free(my_conf);
+
+	if (!t->pool) {
+		ERROR("Pool instantiation failed");
+		return -1;
+	}
+
+	return 0;
 }
 
 static rlm_rcode_t mod_authorize_release(REQUEST *request, void *instance, void *thread, void *ctx);
@@ -2256,6 +2285,7 @@ rad_module_t rlm_sql = {
 	.config		= module_config,
 	.bootstrap	= mod_bootstrap,
 	.instantiate	= mod_instantiate,
+	.thread_instantiate	= mod_thread_instantiate,
 	.detach		= mod_detach,
 	.methods = {
 		[MOD_AUTHORIZE]		= mod_authorize,
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -2229,13 +2229,13 @@ release:
 /*
  *	Postauth: Write a record of the authentication attempt
  */
-static rlm_rcode_t mod_post_auth(void *instance, UNUSED void *thread, REQUEST *request) CC_HINT(nonnull);
-static rlm_rcode_t mod_post_auth(void *instance, UNUSED void *thread, REQUEST *request)
+static rlm_rcode_t mod_post_auth(void *instance, void *thread, REQUEST *request) CC_HINT(nonnull);
+static rlm_rcode_t mod_post_auth(void *instance, void *thread, REQUEST *request)
 {
-	rlm_sql_t const *inst = instance;
+	rlm_sql_t *inst = talloc_get_type_abort(instance, rlm_sql_t);
 
 	if (inst->config->postauth.reference_cp) {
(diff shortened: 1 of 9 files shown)


src/modules/rlm_sql/rlm_sql.c
```diff
@@ -779,6 +779,86 @@ static int sql_get_grouplist(rlm_sql_t const *inst, rlm_sql_handle_t **handle, R
...
+	}
+	if (xlat_aeval(request, &group_ctx->sql_getvpdata_ctx.query, request, inst->config->groupmemb_query,
+			 inst->sql_escape_func, *group_ctx->sql_getvpdata_ctx.handle) < 0) {
+		group_ctx->sql_getvpdata_ctx.rows = -1;
+		return RLM_MODULE_OK;
+	}
+
+	rcode = rlm_sql_select_query(inst, request, group_ctx->sql_getvpdata_ctx.handle, group_ctx->sql_getvpdata_ctx.query);
+
+	switch(rcode) {
```
arr2036: I guess this is OK, but for general sanity and consistency, it might be better to return control back to unlang even if the result was RLM_SQL_OK, and let it call the next state in the state machine.  Might need some tweaking to add some sort of NULL I/O handler.

I guess I don't have solid reasons for this... just:
- Seems odd to me to do a state transition by adding frames onto the C stack.
- We're very likely going to want to add stats that go down to a per-state resolution.  Having states visible for some SQL drivers and not for others is going to confuse someone.

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -779,6 +779,86 @@ static int sql_get_grouplist(rlm_sql_t const *inst, rlm_sql_handle_t **handle, R
...
+
+	default:
+		talloc_free(group_ctx->sql_getvpdata_ctx.query);
+		return RLM_MODULE_FAIL;
+	}
+}
+
+static rlm_rcode_t sql_get_grouplist_resume(REQUEST *request, void *instance, UNUSED void *thread, void *ctx)
+{
+	rlm_sql_thread_group_t *group_ctx = ctx;
```
arr2036: In general it's good practice to use ``talloc_get_type_abort`` when doing these casts from void.  If your code is smashing memory or pointers on the stack you get to find out about it quicker.  If something passes the wrong pointer as ctx you also find out about it quick :)

Don't worry about the performance penalty, there's a macro for turning ``talloc_get_type_abort`` off in non-debug builds.

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -779,6 +779,86 @@ static int sql_get_grouplist(rlm_sql_t const *inst, rlm_sql_handle_t **handle, R
...
+
+	rlm_sql_row_t row;
+	rlm_sql_grouplist_t *entry;
+
+	entry = group_ctx->group_list = NULL;
+	group_ctx->sql_getvpdata_ctx.rows = 0;
+
+	talloc_free(group_ctx->sql_getvpdata_ctx.query);
+
+	while (rlm_sql_fetch_row(&row, inst, request, group_ctx->sql_getvpdata_ctx.handle) == RLM_SQL_OK) {
```
arr2036: Can any drivers return yield here or do they all pull back the complete result set?

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -935,64 +935,88 @@ static int sql_groupcmp(void *instance, REQUEST *request, UNUSED VALUE_PAIR *req
...
+	rad_assert(ctx);
 
 	rad_assert(request->packet != NULL);
 
+	/*
+	 * Because of the new async model, this function gets called multiple times
+	 * at different stages of the group processing.
+	 * Simply jump to where processing needs to resume.
+	 */
+	switch (group_ctx->next_step) {
```
arr2036: I guess this works, but it terms of understandability, breaking this out into separate state functions would be better. Otherwise it's a weird mix of state functions and state enums.

Elsewhere in the server we've moved away from switch statements and enums to just using function pointers.  It tends to make the code clearer.

src/modules/rlm_sql/rlm_sql.h
```diff
@@ -170,6 +171,25 @@ typedef struct {
...
 typedef struct sql_grouplist rlm_sql_grouplist_t;
 typedef struct sql_thread_group rlm_sql_thread_group_t;
+
+/** Context data for rlm_sql_select_query
+ *
+ */
+typedef enum {
+	SELECT_QUERY_START = 0,
+	SELECT_QUERY_RESUME,
+} sql_thread_select_query_state_t;
```
arr2036: Again functions are better than enums, and it's more consistent with the two_step function you added.

src/modules/rlm_sql/sql.c
```diff
@@ -576,31 +576,46 @@ rlm_rcode_t rlm_sql_select_query_async(REQUEST *request, void *instance, UNUSED
...
+	TALLOC_CTX *talloc_ctx = db_ctx->talloc_ctx;
+	rlm_sql_t const *inst = instance;
+
+	rad_assert(talloc_ctx);
 
-	while (rlm_sql_fetch_row(&row, inst, request, handle) == RLM_SQL_OK) {
-		if (sql_fr_pair_list_afrom_str(ctx, request, pair, row) != 0) {
+	db_ctx->rows = 0;
+
+	while (rlm_sql_fetch_row(&row, inst, request, db_ctx->handle) == RLM_SQL_OK) {
```
arr2036: As before, there are no drivers that perform I/O when retrieving rows?

src/modules/rlm_sql/sql.c
```diff
@@ -576,31 +576,46 @@ rlm_rcode_t rlm_sql_select_query_async(REQUEST *request, void *instance, UNUSED
...
+			(inst->driver->sql_finish_select_query)(*db_ctx->handle, inst->config);
 
-			return -1;
+			return RLM_MODULE_FAIL;
 		}
-		rows++;
+		db_ctx->rows++;
 	}
-	(inst->driver->sql_finish_select_query)(*handle, inst->config);
+	(inst->driver->sql_finish_select_query)(*db_ctx->handle, inst->config);
```
arr2036: Or finalizing queries?

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -779,6 +779,86 @@ static int sql_get_grouplist(rlm_sql_t const *inst, rlm_sql_handle_t **handle, R
...
 
+static rlm_rcode_t sql_get_grouplist_resume(REQUEST *request, void *instance, void *thread, void *ctx);
+
+/*
+ * Get group list
+ * Yield if SQL driver supports async queries.
+ */
+static rlm_rcode_t sql_get_grouplist_async(REQUEST *request, void *instance, void *thread, void *ctx)
+{
+	rlm_sql_thread_group_t *group_ctx = ctx;
```
pwdng: The sql_thread_group structure is there. Why do you say it's missing?
It holds the context for the group related queries.

I used the _async suffix because in some cases, I kept the original implementation to make the transition work easier. The goal is to remove it at some point.

arr2036: It's not in the headers at this point in the commit series.

src/modules/rlm_sql/rlm_sql.c
pwdng: The problem with that is that it makes the code impossible to read.
That's what I started doing, but it was impossible to follow the overall logic.
It's mainly because we need to exit some functions from within a loop and therefore resume from that same point.

src/modules/rlm_sql/rlm_sql.c
pwdng: It's easier if they all retrieve the complete result set. It makes managing memory during the initial fetch a bit more complex but it's more consistent across drivers and it avoids having to yield during rlm_sql_fetch_row.

src/modules/rlm_sql/rlm_sql.c
arr2036: Could you just store the state of the loop in the iterator? Doesn't seem too too awful.

src/modules/rlm_sql/rlm_sql.c
arr2036: Yeah that's fair.

nwkrad:

nwkrad:

PR #1942, opened 2017-03-14
Static analyzer issue fixes

spbnick:
This is a bunch of fixes for some new issues found by Coverity and Clang in v3.0.13, compared to v3.0.4.

Please feel free to pick and choose as you see fit.

### DIFF
--- a/src/main/evaluate.c
+++ b/src/main/evaluate.c
@@ -363,8 +363,6 @@ static size_t regex_escape(UNUSED REQUEST *request, char *out, size_t outlen, ch
 		case '[':	/* we don't list close braces */
 		case '{':
 		case '(':
-			if (outlen < 3) goto done;
-
 			*(p++) = '\\';
 			outlen--;
 			/* FALL-THROUGH */
@@ -376,7 +374,6 @@ static size_t regex_escape(UNUSED REQUEST *request, char *out, size_t outlen, ch
 		}
 	}
 
-done:
 	*(p++) = '\0';
 	return p - out;
 }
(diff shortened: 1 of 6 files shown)


src/main/evaluate.c
```diff
@@ -167,6 +167,10 @@ static int cond_do_regex(REQUEST *request, fr_cond_t const *c,
 	rad_assert(lhs_type == PW_TYPE_STRING);
 	rad_assert(lhs != NULL);
 
+	if (lhs_type != PW_TYPE_STRING || lhs == NULL) {
+		return -1;
+	}
+
```
alandekok: The assertions a few lines up should also check for this.  Either Coverity is missing the assertions, OR we should have only one check.  i.e. there's not much point in having assertions *and* run-time checks.

alandekok:
I'll push some simpler changes, and then manually pull over some of the patches.  I would rather not have 3-4 patches that all touch the same code, and I'd rather not have both asserts and run-time checks.

spbnick:
Thanks, Alan. Yeah, we run Coverity and clang-scan on non-debug build, because that's what we package, so assertions are ignored by Coverity.

spbnick:
Alan, are you done with these? Asking because I'd like to make sure I picked and backported all the fixes you took from here.

alandekok:
I've pushed a bunch of fixes over.  I'll be doing the rest in a few days

nwkrad:

nwkrad:

arr2036:
@alandekok did you merge all the fixes, can this be closed?

alandekok:
There are some fixes remaining, but I don't recall which ones.  I'll find time to look at it again.

alandekok:
These have all been merged.

spbnick:
Thanks, Alan!

PR #1945, opened 2017-03-18
Ignore non http version headers in init state

kck325:
Version: FreeRadius 3.0.x
Issue: Some servers return server name and date headers along with http version header. It is throwing out freeradius rest server. While in init state, ignore other headers.

### DIFF
--- a/src/modules/rlm_rest/rest.c
+++ b/src/modules/rlm_rest/rest.c
@@ -1556,6 +1556,12 @@ static size_t rest_response_header(void *in, size_t size, size_t nmemb, void *us
 	case WRITE_STATE_INIT:
 		RDEBUG2("Processing response header");
 
+		/**
+		* Ignore other headers since we are looking for only http version header
+		* in this state.
+		*/
+		if (strncasecmp("HTTP", p, 4) != 0) return t;
+
 		/*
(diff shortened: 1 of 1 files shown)


arr2036:
Hm, weird.  Are they allowed to do that? :)

kck325:
yeah rfc says so.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.2

nwkrad:

nwkrad:

arr2036:
You've not really cited anything that supports your claims.  

The format of the response is:
```text
After receiving and interpreting a request message, a server responds with an HTTP response message.

       Response      = Status-Line               ; Section 6.1
                       *(( general-header        ; Section 4.5
                        | response-header        ; Section 6.2
                        | entity-header ) CRLF)  ; Section 7.1
                       CRLF
                       [ message-body ]          ; Section 7.2
```

The status line is:

```text
Status-Line = HTTP-Version SP Status-Code SP Reason-Phrase CRLF
```

So the first component of any HTTP response should be the status line.  If your server is responding without the Status line, your server is not in compliance with RFC2616 and is unsupported.

PR #1963, opened 2017-04-20
stopped rotation of the session database files

alanbuxey:
you really don't want to be rotating these under the server - they are not normal log files but are stateful session files (used by various utilities). these were removed from the logrotate some time ago but appear to have crept back in.

### DIFF
--- a/redhat/freeradius-logrotate
+++ b/redhat/freeradius-logrotate
@@ -25,13 +25,6 @@ delaycompress
 }
 
 #
-#  Session database modules
-#
-/var/log/radius/radutmp /var/log/radius/radwtmp {
-	nocreate
-}
-
-#
 #  SQL log files
 #
 /var/log/radius/sqllog.sql {


arr2036:
Include changes for all relevant copies of the logrotate files and they won't creep back in

alanbuxey:
eh?

alanbuxey:
commit 2ee2f5c743cbcbc68733b0bb67086406fcaf81f4   stopped utmp previous, 
commit db805b29292bcc6e31dc4f98be8657ef87dc7f09  put it back  ;-)


alan

mcnewton:
```
redhat/freeradius-logrotate
debian/freeradius.logrotate
suse/radiusd-logrotate
scripts/logrotate/freeradius
```

all the same.

arr2036:
So the fix should be applied to all the logrotate scripts, not just the one that benefits the OP 👍

alanbuxey:
doing it...strange but I thought those distros had maintainers - i dont USE
any of those other scripts so wasn't aware of them ;-)

PR #1964, opened 2017-04-20
stop rotation of the session database files

alanbuxey:
you really don't want to be rotating these under the server - they are
not normal log files but are stateful session files (used by various
utilities). these were removed from the logrotate some time ago but
appear to have crept back in.

### DIFF
--- a/debian/freeradius.logrotate
+++ b/debian/freeradius.logrotate
@@ -26,13 +26,6 @@ notifempty
 }
 
 #
-#  Session database modules
-#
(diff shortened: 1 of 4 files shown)


arr2036:
I guess the default location should really be ``/var/lib/`` too... Oh well.

alanbuxey:
I was thinking the same.....those files really are special... but then
theres lots of utils etc that expect the current path...  maybe a big bang
change for 4.x ???

alan

arr2036:
I think the module is being removed in v4

PR #1966, opened 2017-04-21
Use Framed-MTU value correctly

lytboris:
Framed-MTU holds full packet length rather than EAP-Message payload size.
Calculate maximum EAP-Message payload based on Framed-MTU value correctly
to ensure IP fragment-free authentication.

This pull request is a proof of concept that is tested in my lab. Will do patches for 3.1 and 4.0 once this one is discussed and merged.

### DIFF
--- a/src/include/radiusd.h
+++ b/src/include/radiusd.h
@@ -432,6 +432,7 @@ void		rad_suid_set_down_uid(uid_t uid);
 void		rad_suid_down(void);
 void		rad_suid_up(void);
 void		rad_suid_down_permanent(void);
+size_t		rad_eap_calculate_mtu(size_t fragment_size, REQUEST *request);
 /* regex.c */
 
 #ifdef HAVE_REGEX
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -566,7 +566,6 @@ tls_session_t *tls_new_session(TALLOC_CTX *ctx, fr_tls_server_conf_t *conf, REQU
 	tls_session_t	*state = NULL;
 	SSL		*new_tls = NULL;
 	int		verify_mode = 0;
-	VALUE_PAIR	*vp;
 
 	rad_assert(request != NULL);
 
@@ -668,11 +667,7 @@ tls_session_t *tls_new_session(TALLOC_CTX *ctx, fr_tls_server_conf_t *conf, REQU
 	 *	of EAP-TLS in order to calculate fragment sizes is
 	 *	just too much.
 	 */
-	state->mtu = conf->fragment_size;
-	vp = fr_pair_find_by_num(request->packet->vps, PW_FRAMED_MTU, 0, TAG_ANY);
-	if (vp && (vp->vp_integer > 100) && (vp->vp_integer < state->mtu)) {
-		state->mtu = vp->vp_integer;
-	}
+	state->mtu = rad_eap_calculate_mtu(conf->fragment_size, request);
 
 	if (conf->session_cache_enable) state->allow_session_resumption = true; /* otherwise it's false */
 
--- a/src/main/util.c
+++ b/src/main/util.c
@@ -1700,3 +1700,46 @@ int rad_segid(gid_t gid)
 	}
 	return 0;
 }
+
+/** Calculate fragment-free EAP-Message payload
+ * Calculate payload length based on fragment_size and Framed-MTU received from NAS.
+ *
+ * @param fragment_size fragment size configured for instance
+ * @param request The current #REQUEST.
+ * @return correct mtu value to use
+ */
+size_t rad_eap_calculate_mtu(size_t fragment_size, REQUEST *request)
+{
+	size_t		mtu = fragment_size;
+	VALUE_PAIR	*vp;
+
+	vp = fr_pair_find_by_num(request->packet->vps, PW_FRAMED_MTU, 0, TAG_ANY);
+	if (vp == NULL || vp->vp_integer < 100 + 96 /* min possible offset */) return mtu;
+
+	/*
+	 *	Framed-MTU means total size of a packet so we need to
+	 *	take into account so actual EAP message will be smaller than
+	 *	Framed-MTU value.
+	 *	This offset includes:
+	 *	    - IP/IPv6 header, 20/40 bytes
+	 *	    - UDP/TCP header, 8/20 bytes
+	 *	    - RADIUS:
+	 *		* RADIUS header, 4 bytes
+	 *		* Authenticator, 16 bytes
+	 *		* Message-Authenticator, 2 + 16 bytes
+	 *		* State, 2 + 16 bytes
+	 *		* Header for each EAP-Message chunk, 2 bytes
+	 *		* EAP header, 4 bytes + upto 6 bytes for:
+	 *		    o EAP-TLS header (type + flags + tls message length), 6 bytes
+	 *		    o EAP-PWD header, 4 bytes
+	 */
+	mtu =	vp->vp_integer - \
+		(request->client->ipaddr.af == AF_INET ? 20 : 40) - \
+		(((listen_socket_t *)request->listener->data)->proto == IPPROTO_UDP ? 8 : 20) - \
+		(4 + 16 + 18 + 18 + 4 + 6);
+	/* calculate number of EAP-Message chunks and reserve space for header of each chunk */
+	mtu -= 2 * (1 + (mtu/255));
+	/* fallback to value in configuration */
+	if (mtu < 100 || mtu > fragment_size) return fragment_size;
+	return mtu;
+}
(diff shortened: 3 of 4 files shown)


arr2036:
You've misunderstood the use of Framed-MTU.  It either hints (Access-Accept) or describes (Access-Request) the MTU of the connection between the Supplicant and the NAS.  It does not describe the maximum size of IP/RADIUS packets between the NAS and the RADIUS server.

The MTU in this case is used to fragment EAP-Messages so that when reassembled, and are wrapped in an EAPOL frame*, they do not exceed the MTU between the NAS and the supplicant.

In cases where they fragments do exceed the the MTU between the NAS and the supplicant, the NAS will usually drop the packet, which is why it's important to calculate the fragment size dynamically if the Framed-MTU is available.

Your patch would artificially/incorrectly limit the size of the fragments between the NAS and the supplicant, which may make authentication less efficient by introducing additional rounds.

If you want to avoid IP fragmentation, either increase the MTU between the NAS and the RADIUS server, or artificially configure a lower MTU.

* Assuming the transport is EAPOL isn't great.  We might be able to use Service-Type in some cases, but it's really not a reliable indicator, so we're sort of stuck.

arr2036:
```
5.12.  Framed-MTU

   Description

      This Attribute indicates the Maximum Transmission Unit to be
      configured for the user, when it is not negotiated by some other
      means (such as PPP).  It MAY be used in Access-Accept packets.  It
      MAY be used in an Access-Request packet as a hint by the NAS to
      the server that it would prefer that value, but the server is not
      required to honor the hint.

   A summary of the Framed-MTU Attribute format is shown below.  The
   fields are transmitted from left to right.

    0                   1                   2                   3
    0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
   |     Type      |    Length     |             Value
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
              Value (cont)         |
   +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+

   Type

      12 for Framed-MTU.

   Length

      6

   Value

      The Value field is four octets.  Despite the size of the field,
      values range from 64 to 65535.
```

arr2036:
So your effort doesn't go to waste, you could submit an unlang policy with similar calculations for other people who wanted to avoid IP fragmentation

arr2036:
If you wanted to prevent IP fragmentation, you could encapsulate similar logic as an unlang policy and alter Framed-MTU.  Then at least the effort wouldn't go to waste :)

lytboris:
Thanks for clarifying that, I was confused by
`
It MAY be used in an Access-Request packet as a hint by the NAS to the server that it would prefer that value
`

Meanwhile the main limitation factor that ` artificially/incorrectly limit the size of the fragments between the NAS and the supplicant` is default value of fragment_size to 1024 by default. And you need it being low because you don't know for sure the size of EAP payload that will not produce IP fragments. What do you think if I will introduce, let's say `mtu` parameter to listen block and it will be used by EAP modules the same way Framed-MTU value is used in currrent pull request?

alandekok:
I don't know why you want to avoid IP fragments.  If your network doesn't deal with fragmented IP packets, then your network is broken.

So why do you need to set MTU between the client and the server?

lytboris:
Alan, I would quote your RFC as an answer for your question:
>  Fragmentation of RADIUS/UDP packets is generally inadvisable due to lack of fragmentation support within intermediate devices such as filtering routers, firewalls, and NATs.  However, since RADIUS/UDP implementations typically do not support MTU discovery, fragmentation can occur even when the maximum RADIUS/UDP packet size is restricted  to 1500 octets.

My guess default value of fragment_size set to 1024 has the same reasoning. Am I wrong?

Additionally, using 1.5k packets instead of 1.1k packets saves one round-trip.

alandekok:
I'm well aware of that.  However, quoting an RFC doesn't explain why *you* need to set the MTU.

NFS fragments UDP packets.  People use it all of the time on *local* networks.  Which is a use-case for RADIUS, too.

i.e. if your *local* network equipment doesn't support UDP fragmentation, it's broken.

If you're trying to proxy, the wider internet doesn't support UDP fragmentation that well.  In which case you should do proxying via RadSec (TLS / TCP), or via IPSec.

So again, why you *do* need UDP MTU settings on the server?

How can setting MTU on the server `listen` socket help anything?

* if you're sending small amounts of data in a packet, it will be under the MTU, and MTU settings won't help

* if you're sending large amounts of attributes in an Access-Accept, you *need* to send that data, no matter what the MTU says

* EAP packets which fit into EAPoL frames will *likely* overflow the MTU on the IP section, because of the overhead of IP / UDP / RADIUS.  But... the local equipment *should* support UDP fragmentation

* Where the local equipment doesn't support UDP fragmentation, you can just set Framed-MTU manually:

    if (from socket on link with low MTU) {
        update request {
            Framed-MTU = 800
        }
    }

The EAP module will then ensure that the EAP packets are no larger than this.

Again, what problem are you trying to solve?  Please be specific.

lytboris:
General idea behind this PR is to minimize packet flow when performing EAP-TLS authentication in a most efficient way. If your RADIUS server is running on a host with Jumbo frames enabled (MTU 9k) and your NAS is IPv6 then your server is responsible for proper fragmentation (since routers do not fragment IPv6) and just using PMTU means that negotiation will take  longer than really needed. Of course you can tell me that my network/routing is broken and what I need to do is to make a bunch of static routes for each NAS with altered MTU so fragmentation will be done here and now. That's an option but I'd prefer not to use static routes for this.

My original confusion was that Framed-MTU value is set by NAS indicating maximum packet length it can handle. E.g. Cisco 2960 switches are capable of Jumbo frames on all gig interfaces to forward frames but CPU can not handle packets more than 1998 bytes. When you try to authenticate a client behind such port, Framed-MTU will be set to 9k (jumbo mtu value you configure by yourself). If RADIUS is configured with fragment_size = 4k and jumbo frames are enabled on the interface - 4k long packet will be sent to NAS and that NAS is incapable of handling such big packets. So my thoughts are to make a way to hint radiusd when it is safe to use jumbos and when it is not. E.g. I can write something like
```
client generic_switches {
ipv6addr = 2001:DB8::/32
mtu = 1950
}
```
to hint radiusd that even though fragment_size = 4k packets to this NAS should not be greater than 1950 bytes long.

As a summary I tend to think that if there's a good way not to produce fragments - it is the best way to do that.

Like @arr2036 mentioned, the same approach can be programmed in unlang but I still think this could be a nice core feature too.

alandekok:
> My original confusion was that Framed-MTU value is set by NAS indicating maximum packet length it can handle

Not entirely.  As arran mentioned, it's the max packet length to the *user*. 

> When you try to authenticate a client behind such port, Framed-MTU will be set to 9k 

Which looks like a bug in the switch.

> If RADIUS is configured with fragment_size = 4k

That's not allowed.  The maximum RADIUS packet size is 4K.  For the RADIUS header + all attributes.

I'll see if I can update the EAP module so that the EAP packets don't overflow RADIUS packets.  Or if they do, issue a large warning.

> ...to hint radiusd that even though fragment_size = 4k packets to this NAS should not be greater than 1950 bytes long.

You can do that in unlang.

    update request {
      Framed-MTU = "%{client:mtu}"
    }

> the same approach can be programmed in unlang but I still think this could be a nice core feature too.

I think this situation is rare enough that it's better to just have unlang rules for it.

lytboris:
> I'll see if I can update the EAP module so that the EAP packets don't overflow RADIUS packets. Or if they do, issue a large warning.

That is exactly what my patch is doing - it lowers down EAP payload so it fits into a value (mistakenly derived from Framed-MTU). Source of "MTU" can be changed in a easy way.

arr2036:
Whilst it's true that IPv6 does not support fragmentation by intermediate nodes, and PMTU discovery is used to figure out the path, the originating host is able to fragment packets.

One of the major advantages of having a layered networking stack is that the complexity of the lower layers are hidden from the application.  Forcing this logic into the application as you're suggesting is IMHO wrong.  The kernel in this case should fragment the UDP frames using the MTU it discovered.

The reason why we support controlling the EAP fragment sizes, is because there's no MTU discovery in EAP or any of the EAP methods, so we have no choice but to limit the EAP fragments to a size that are likely to make it to supplicant.

If you really want to do this, unlink is the place, as it is for all other hacky workarounds.

alanbuxey:
so...if you have

client generic_switches {
ipv6addr = 2001:DB8::/32
mtu = 1950
}


then, in your virtual server, before call to EAP, do


if(%{client:mtu}) {

   update request {

      Framed-MTU = %{client:mtu}

   }

}

arr2036:
Yeah it's easily doable in unlang, OP is just being stubborn :)

lytboris:
> OP is just being stubborn :)
LOL. You all a way too good example to follow :)

I'll make a policy as a separate PR. Thanks for conversation.

PR #1968, opened 2017-04-22
updated ldap default config

alanbuxey:
net_timeout has a default setting of 10 in the code but the enabled value in the default config is 1 - therefore
the 'default' would be , for users, 1 - so commented out. users can enable it to change.
likewise, same for ldap_debug - 'default' is 0x0000 but it had a preconfigured value of 0x0028 - so commented
out so default is in play and admins can choose to change it.

### DIFF
--- a/raddb/mods-available/ldap
+++ b/raddb/mods-available/ldap
@@ -451,7 +451,7 @@ ldap {
 		#  failures) default: 10
 		#
 		#  LDAP_OPT_NETWORK_TIMEOUT is set to this value.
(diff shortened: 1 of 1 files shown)


mcnewton:
In the spirit of not changing the stable version, I'd be more inclined to also change the defaults in the code to match the config that was currently used. Which means less surprises all round if someone has commented them out already and then upgrades.

arr2036:
@mcnewton agreed

PR #1972, opened 2017-04-25
Implement client:mtu as eap policy

lytboris:
Framed-MTU value is tuned to insure fragment-free IP conversation between NAS and radiusd.

This is implementation based on talks in #1966.

### DIFF
--- a/raddb/clients.conf
+++ b/raddb/clients.conf
@@ -182,6 +182,14 @@ client localhost {
 #	response_window = 10.0
 
 	#
+	#  MTU for packets sent to NAS, currently works only for EAP
+	#  and ensures IP fragment-free packet exchange. Enable
+	#  apply_client_mtu_as_framed_mtu before calling eap in
+	#  authorize section.
+	#
+#	mtu = 1500
+
+	#
 	#  Connection limiting for clients using "proto = tcp".
 	#
 	#  This section is ignored for clients sending UDP traffic
--- a/raddb/policy.d/eap
+++ b/raddb/policy.d/eap
@@ -83,3 +83,53 @@ remove_reply_message_if_eap {
 	}
 }
 
+
+#	Override Framed-MTU value based on mtu settings for NAS
+
+#	Framed-MTU specifies desired length of EAP payload to send.
+#	For some NAS this could lead to IP packet fragmentation which can
+#	be undesirable. If so, enable this policy in authorize section just
+#	before eap call and set nas attribute for a NAS you need to avoid
+#	IP fragmentation for.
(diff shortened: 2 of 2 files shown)


alandekok:
Again, the Framed-MTU attribute is for the link between the NAS and the end user.  It has *no* relation to the MTU between the NAS and the RADIUS server.

arr2036:
This is about artificially lowering the Framed-MTU so the EAP record fragments don't cause the RADIUS packet to exceed the network MTU.  It's not perfect but I don't see any harm in including it?

arr2036:
After some out of band discussions adding a separate attribute to control fragmentation instead of overloading Framed-MTU would be less confusing... I'll take a look.

alandekok:
some more notes... using "mtu" in this way is confusing. i.e. mtu is being used both for local MTU, and for Framed-MTU.  That will cause confusion.

The change should instead set the maximum RADIUS packet size.  The limit on Framed-MTU can be set from that.

The code should also *not* deal with IPv4 / IPv6 issues.  That information isn't really available in unlang (unless it checks for the existence of Packet-Src-IP-Address versus Packet-Src-IPv6-Address).

If there's a "max radius packet size" attribute, then the EAP module can auto-calculate it's maximum fragment size.  And that kind of policy should arguably go into the EAP module.

arr2036:
In all instances it seems the Maximum RADIUS packet size should be associated with the client? For IPv6 PMTU is client specific, in other future cases having jumbo RADIUS packets would likely be a negotiated feature?

alandekok:
Jumbo radius packets can be a negotiated feature. See RFC 7930

alanbuxey:
aye..dont forget that in the management layer, the AP (for example) may
only have access to the RADIUS server over <1500 byte ethernet..... but in
the data plane, the
clients of the AP (who need Framed-MTU) might be quite happily doing 2000
MTU etc

arr2036:
You'd want the target fragment size to be the smallest of the two MTUs after taking account of headers.

alanbuxey:
well, you'd want the NAS to be treated with small MTU to deal with the EAP
auth stuff through the 802.1X and the wireless client to be told the right
MTU for the service network its dropped onto after the auth.

arr2036:
reply:Framed-MTU isn't really relevant for wireless clients, it's more for VPNs.

lytboris:
> reply:Framed-MTU isn't really relevant for wireless clients, it's more for VPNs.

Well, in my case it is quite conversely: it has nothing to do with VPNs but all wireless & wired clients are on the scene.

PR #1973, opened 2017-04-27
initial mention of one_shot option in config

alanbuxey:
default is no - nothing explains what it is yet, what it does, how it works etc

### DIFF
--- a/raddb/sites-available/buffered-sql
+++ b/raddb/sites-available/buffered-sql
@@ -90,6 +90,12 @@ server buffered-sql {
 		retry_interval = 30
 
 		#
+		#  this option is not yet correctly documented.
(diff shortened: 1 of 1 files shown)


mcnewton:
A patch that did happen to explain what it is and what it does might possibly be better...?

alanbuxey:
first step mention it - then it can have correct (I dont know its operational modes) documentation  ;-)

mcnewton:
For any detail reader (so not just buffered-sql - e.g. copy-acct-to-home-server as well), it will mean that once the end of the file has been reached the server will quit. Which makes it possible to do a "one-shot" configuration - "read this file, process it, and exit".

I've done another patch.

PR #1988, opened 2017-05-11
Fix typo in previous commit

alejandro-perez:
I made a mistake in the call, that should be "tls_server_conf_alloc" and not "fr_tls_server_conf_alloc".

### DIFF
--- a/src/modules/rlm_realm/trustrouter.c
+++ b/src/modules/rlm_realm/trustrouter.c
@@ -70,7 +70,7 @@ static fr_tls_server_conf_t *construct_tls(TIDC_INSTANCE *inst,
 	char *hexbuf = NULL;
 	DH *aaa_server_dh;
 
-	tls = fr_tls_server_conf_alloc(hs);
+	tls = tls_server_conf_alloc(hs);
 	if (!tls) return NULL;
 
 	aaa_server_dh = tid_srvr_get_dh(server);


nwkrad:

alanbuxey:
it [original PR] shouldn't have passed travis-ci with that error anyway...

alan

alejandro-perez:
Yes, that's what is more weird. If you perform "make" with that typo, it actually works. It is not until you try to execute it when you get the error (dlload error, but not related to that particular call, but another one (debug_something).

mattrose:
I think the travis tests don't link against openssl.  I've been digging through the travis tests and saw this 
```
Executing: /bin/sh -c ./build/bin/radiusd -xxxv -n test  
Wed May 10 09:57:34 2017 : Info: test: FreeRADIUS Version 3.0.14 (git #56aa3b7), for host x86_64-unknown-linux-gnu, built on May 10 2017 at 09:55:13
Wed May 10 09:57:34 2017 : Debug: Server was built with: 
[...]
Wed May 10 09:57:34 2017 : Debug:   threads                  : no
Wed May 10 09:57:34 2017 : Debug:   tls                      : no
[...]
Wed May 10 09:57:34 2017 : Debug: Server core libs:
Wed May 10 09:57:34 2017 : Debug:   freeradius-server        : 3.0.14
Wed May 10 09:57:34 2017 : Debug:   talloc                   : 2.0.*
Wed May 10 09:57:34 2017 : Debug:   ssl                      : not linked
```
and this
```
[1m[33mrlm_eap (EAP): Ignoring EAP method tls because we don't have OpenSSL support[0m
[1m[33mrlm_eap (EAP): Ignoring EAP method ttls because we don't have OpenSSL support[0m
[1m[33mrlm_eap (EAP): Ignoring EAP method peap because we don't have OpenSSL support[0m
```

alandekok:
There is a build without OpenSSL.  But there should be builds with OpenSSL

PR #1991, opened 2017-05-18
Add SQL backing to Moonshot-*-TargetedId generation

spaetow:
This gives some SQL backing to the Moonshot TargetedId generation policies.


spaetow:
@arr2036, @alandekok, can we merge this into 3.0.14? :-)

Issue #990, opened 2015-05-04
Switch to MRUBY because normal ruby was not designed for embedding and is very broken when used this way


alanbuxey:
3.0.8 doesnt compile when Ruby is used:
# ruby -v

ruby 1.8.7 (2013-06-27 patchlevel 374) [x86_64-linux]

```
./configure --prefix=/usr --enable-static --with-large-files --with-raddbdir=/etc/raddb --with-snmp --with-experimental-modules  | grep ruby
=== configuring in src/modules/rlm_ruby (/usr/src/freeradius-server-3.0.8/src/modules/rlm_ruby)
checking whether ruby executable path has been provided... no
checking for ruby... /usr/bin/ruby
checking for ruby version... 1.8.7
checking for Ruby include path...  -I/usr/lib64/ruby/1.8/x86_64-linux
checking for Ruby linking flags... -lruby
checking for Ruby site-packages path... /usr/lib64/ruby/site_ruby/1.8/x86_64-linux
checking consistency of all components of ruby development environment... yes
configure: Sanitized cflags are "-O2 -g -pipe -Wall -Wp,-D_FORTIFY_SOURCE=2 -fexceptions -fstack-protector --param=ssp-buffer-size=4 -m64 -mtune=generic -fno-strict-aliasing  -fPIC    -isystem /usr/lib64/ruby/1.8/x86_64-linux"
```

make

```
CC src/modules/rlm_ruby/rlm_ruby.c
In file included from /usr/src/freeradius-server-3.0.8/src/freeradius-devel/libradius.h:85,
                 from /usr/src/freeradius-server-3.0.8/src/freeradius-devel/radiusd.h:29,
                 from src/modules/rlm_ruby/rlm_ruby.c:35:
/usr/src/freeradius-server-3.0.8/src/freeradius-devel/regex.h:71: error: expected declaration specifiers or ‘...’ before ‘regex_t’
/usr/src/freeradius-server-3.0.8/src/freeradius-devel/regex.h:73: error: expected ‘)’ before ‘*’ token
In file included from /usr/src/freeradius-server-3.0.8/src/freeradius-devel/radiusd.h:64,
                 from src/modules/rlm_ruby/rlm_ruby.c:35:
/usr/src/freeradius-server-3.0.8/src/freeradius-devel/tmpl.h:223: error: expected specifier-qualifier-list before ‘regex_t’
In file included from src/modules/rlm_ruby/rlm_ruby.c:35:
[...]

PR #1992, opened 2017-05-20
Add rlm_mruby

herwinw:
Had this only lurking around for quite some time now. See #990


src/modules/rlm_mruby/rlm_mruby.c
```diff
@@ -0,0 +1,468 @@
...
+
+	return 0;
+}
+
+#define BUF_SIZE 1024
+static mrb_value mruby_vps_to_ary(mrb_state *mrb, VALUE_PAIR **vps)
+{
+	mrb_value res;
+	VALUE_PAIR *vp;
+	vp_cursor_t cursor;
```
arr2036: That should be fr_cursor_t, yes I know it's confusing hopefully we'll hopefully unify them again soon.

arr2036: Nm, i'll fix it after merging.

PR #1993, opened 2017-05-21
Remove incorrect UNUSED marker in rlm_mruby

herwinw:


herwinw:
Travis will still fail with this change, but with 3 errors instead of 4

PR #1994, opened 2017-05-21
Removed rlm_mruby from travis build

herwinw:
The mruby version used on travis is very old (I can't say how old, the
ubuntu package used version 0.0.0, `mruby --version` doesn't show
anything other than a copyright notice.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ addons:
       - libldap2-dev
       - libluajit-5.1-dev
       - libmemcached-dev
+      - libmruby-dev
       - libmysqlclient-dev
       - libnl-genl-3-dev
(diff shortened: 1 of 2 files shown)


herwinw:
I couldn't find any way to install a more recent version of mruby in Travis, so the next best thing is just to ignore the problem and skip the build of rlm_mruby on travis

arr2036:
Maybe just check for the failing function in the autoconf script?

arr2036:
...and no, I checked too.  The only way would be to build our own packages, or transition to running docker images in Travis.  I'm actually in favour of the second, as it means we could run the same tests on travis and our local jenkins cluster, but it'll require some development work.

PR #1996, opened 2017-05-26
Fix deadlock

alejandro-perez:
There seem to be a regression introduced in 8391d0ba8ebd2599212317259d26a17cfebb5b2a


alejandro-perez:
Oh, if you accept it you might want to edit the comment and remove the # symbol so the commit hash appears nicely in github.

PR #2004, opened 2017-06-08
If proxy mode is enable use proxy_reply timestamp if available on xlat_print function

geaaru:
If proxy mode is enable use proxy_reply timestamp if available on xlat_print function

### DIFF
--- a/src/main/xlat.c
+++ b/src/main/xlat.c
@@ -2143,8 +2143,18 @@ static char *xlat_aprint(TALLOC_CTX *ctx, REQUEST *request, xlat_exp_t const * c
 		when = request->timestamp;
 		usec = 0;
 		if (request->packet) {
+#ifdef WITH_PROXY
+			if (request->proxy && request->proxy_reply) {
+				when = request->proxy_reply->timestamp.tv_sec;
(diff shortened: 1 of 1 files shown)


alandekok:
We can't change the behavior of the xlat expansion in a stable release.

geaaru:
Hi, but is not a change of the behavior, but a bug fix. Currently on logging proxy response it is used request timestamp instead of response timestamp.
Is not an issue this or is there some reason ?
Unlucky is not possible use request->reply->timestamp on logging response for not proxy mode configuration because when it is used linelog module probably timestamp is not initialized (i think that this could be correct) until packet terminate unlang processing. WDYT about this ? could be correct set timestamp of the reply packet before post_auth section to trace outgoing packet timestamp?
I think that could be interesting trace this timestamp to understand quality and effort of the authentication /accounting process and fix eventually too slow queries or other.

alandekok:
The point is that %t is the timestamp of the original request.  It's OK to add a new expansion for proxy timestamp.  It's not OK to change the meaning of an existing expansion.

geaaru:
Ok, thank you very much for clarification. I will open an issue to 4.0.x tree for this if is ok for you.
Thanks again.

PR #2006, opened 2017-06-12
Fix duplicate atrribute name.

skids:
Depending on where you look, some of these other names on the recently
added entries are incorrect.

Two Cisco-VPN3000-WebVPN-Outlook-Exchange-Proxy-Enable defs cause
a server startup failure when this dictionary is included, so fix that pronto.


skids:
This seems to be a mistake promulgating from some other data source, e.g.
https://www.ccexpert.us/posture-validation/cisco-vpn-3000-concentratorasapix-7x-dictionary-of-radius-vsas.html
has obvious errors near this attribute.

PR #2007, opened 2017-06-16
Add Trust Router rekeying functionality

alejandro-perez:
Hi,

This set of focused patches provide rekeying capabilities for the Trust Router client in rlm_realm.
This rekeying functionality basically consists of a dedicated thread which negotiates, in the background, new TLS keys for existing dynamic REALMs before they expire. The achieved aim is to prevent the performance issues that appear when this re-negotiation happens in the foreground as a consequence of a end user authentication, which can then take up to 10 seconds (due to the TR protocol).

Performing this task in the background with the enough margin of time ensures that new keys are used before expiration happens.


alejandro-perez:
Hi guys,

is there anything I can do to help with the review/acceptance process of this PR? I decided to split it into 4 focused commits, but I could split it even more if required.

alandekok:
It mostly looks good.  I'll have to pull the changes over manually, as all of the trust router things should be in a config sub-section.  I'll likely do that Monday.

arr2036:
@alejandro-perez you guys should also start thinking about how you want this implemented in v4.0.x.  rlm_realm has been removed completely, so we need to determine what you need exactly for trust router, and whether it can be implemented in policies instead of modules.

alejandro-perez:
@alandekok Thanks. Let me know if I can help somehow.
@arr2036 Speaking as an individual (I'm far from being a leader on the trust router matters), the requirements are basically being able to dynamically create home_server_t/REALM structs (TLS-PSK secrets and all) and insert them into the realm tree in a dynamic way, when the destination realm does not exists. In other words, the TR libraries are able to establish a PSK between two RADIUS entities (off-line of the RADIUS protocol), but then we need a way to enforce that REALM with the PSK into FR.
 I don't know whether policies can control that. 

On the IDP side there is a dedicated daemon (tids) which inserts the negotiated PSK keys into a sqlite database, which is then queried by a FR policy to get the proper key. A similar approach could be followed on the client.

I can create a dedicated Issue to discuss that, in order to avoid mixing things into this PR and to call the rest of the TR community into the conversation.

alanbuxey:
I did raise this over a year ago (at moonshot meetings)  when initial notes of rlm_realm going in v4 were raised.   the more interesting part would be of new policy requirements to ensure that IF its not a moonshot capable construct, then do NOT use it and instead use the normal RADIUS path :)

arr2036:
@alejandro-perez yes, a separate issue would be good.  Dynamic realms are definitely on the roadmap, and will be implemented by the new rlm_radius module.

alandekok:
`rlm_realm` was always horrible in the server.  It was the only module which looked at `proxy.conf`, and the two were welded together.  Having "core" functionality in a module made everything hard.

I'm OK with these patches going into v3, as they're bug / stability fixes.

For v4, the framework should allow for dynamic home servers.  At which point the trust router could likely be a module all to itself.  It would just have to return IP / port / cert information to the server core, which would then track the connection independently of anything else.

And all of the lifetime / cache / renew management for dynamic home servers has to be done anyways.  So removing that from the trust router code is a good thing.

arr2036:
>  It would just have to return IP / port / cert information to the server core, which would then track the connection independently of anything else.

Sqlite3 does appear to deal ok with concurrent accesses from different applications, with the caveat that queries might return 'busy' and need to be retried.

For remote access a REST interface would definitely be preferable to a dedicated C module just query the tids.

alejandro-perez:
> I'm OK with these patches going into v3, as they're bug / stability fixes.

Indeed, and given that you mentioned that renew management functionality will be in the core of v4, they can be seen as a backport in some sense (since v3 does not have that).

> For v4, the framework should allow for dynamic home servers. At which point the trust router could likely be a module all to itself. It would just have to return IP / port / cert information to the server core, which would then track the connection independently of anything else.

I guess PSK would still be possible, wouldn't it?
Having support for dynamic servers in the core would actually make the integration with the TR more natural. Having a dedicated module seems like a good idea to me. And it would be much more simpler than the current one as much functionality would be in the core.

> And all of the lifetime / cache / renew management for dynamic home servers has to be done anyways. So removing that from the trust router code is a good thing.

Yes.

alejandro-perez:
Please, use #2019 for discussion about how to deal with v4.0 and keep this thread to discuss about this specific PR.

arr2036:
There's only one new config item, and I don't think moving the config items to a subsection is a good idea for v3.0.x, so i'm going to merge this.

PR #2009, opened 2017-06-20
adding docker scripts for building

mattrose:
as requested.  These docker scripts set up docker containers able to build freeradius easily

### DIFF
--- a/src/modules/rlm_dict/README.md
+++ b/src/modules/rlm_dict/README.md
@@ -5,8 +5,4 @@
 </dl>
 
 ## Summary
-<<<<<<< Updated upstream
-Registers xlats to access dictionary data
-=======
 Registers xlats and maps to access dictionary data
->>>>>>> Stashed changes


alandekok:
The repository shouldn't have binary RPMs in it.  Is there any way to load these packages from elsewhere, or build them from source.. e.g. like we do already with eapol_test.

alanbuxey:
Was going to say the same thing. RPMs = no. Just pull them directly from
required repo using standard docker/yum scripts. In that way they also stay
up to date and you're not using stale code.

alan

mattrose:
It's not ideal, but these packages I found are definitely not in any standard repository.  I ended up finding the source rpms  on pkgs.org and rebuilding them on a CentOS server to get these binaries.  

The problem with building them on the box is that the freeradius RPMs themselves depend on the RPM packages, not just the libraries, so to build free radius we have to get the source code, then get the RPM manifest from somewhere, and then installing the packages necessary for building RPMs, before we can even get to building the RPMs.

Not impossible, I'll see what I can do.

arr2036:
libkqueue one is already solved.  Shallow clones of git repositories are fine.

mattrose:
retest this please

arr2036:
We needed that? :P

arr2036:
We do actually send systemd hearbeats from the main event loop

PR #2022, opened 2017-07-19
Update rlm_python log level declarations

alejandro-perez:
LOG level declarations in radiusd.py were very outdated.
L_WARN was duplicated in rlm_python.c

Would need fixing on v4.0.x too.


qnet-herwin:
Does anyone actually use this file? The content is pretty stupid (a bunch of constants repeated from the source) and a specification of a `radlog` method that includes a `level = level` statement. "This should only be used when testing modules."

The other scripting modules (perl, (m)ruby and lua) don't contain anything comparable.

if we really want to keep this file, I guess it would be better to generate it from the C constants instead of keeping a file with this much duplication

alejandro-perez:
I'm using it since for testing, you don't have access to the embedded radiusd module.
It makes sense to me, since otherwise I would need to create it myself.

How are other modules tested without access to the logging function or to the constant declarations?

Having it generated from the C constants automatically would be the optimal choice, sure.

PR #2029, opened 2017-07-26
Convert freeradius LDAP schema to Samba4/AD format

candlerb:
Added conversions of `doc/schemas/ldap/openldap/freeradius.schema` and `doc/schemas/ldap/openldap/freeradius-clients.schema` into a form which can be loaded into Samba4 (and presumably also into MS Active Directory)


arr2036:
You know OpenLDAP can do this right? :)  But sure it's useful to have the ldif versions.

candlerb:
> You know OpenLDAP can do this right?

Do you mean, it can convert an OpenLDAP schema into an AD schema?  If so, please can you point to it -  I did google quite extensively for exactly that, and couldn't find anything.

There is already an LDIF in [doc/schemas/ldap/openldap/freeradius.ldif](https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/doc/schemas/ldap/openldap/freeradius.ldif), but that's using OpenLDAP-specific attributes and cannot be loaded into Samba4.

AD describes its schema in a completely different way to OpenLDAP.

arr2036:
Ahh, ok, then no it cannot.

PR #2041, opened 2017-08-17
fix segfault when setting pythonpath in rlm_python extension.

tremolo:
The bug might only be triggered when specifying "python_path" in the config file for the module and results in trying to memcpy to an uninitialized pointer (char *path). Copying the string is not really necessary, so the code is just removed.

The bug appears to be present in 3.0.x and also 4.0.x

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -926,14 +926,11 @@ static int python_interpreter_init(rlm_python_t *inst, CONF_SECTION *conf)
 #if PY_VERSION_HEX > 0x03050000
 			{
 				inst->wide_path = Py_DecodeLocale(inst->python_path, strlen(inst->python_path));
-				PySys_SetPath(path);
+				PySys_SetPath(inst->wide_path);
 			}
 #else
 			{
-				char *path;
-
-				memcpy(&path, inst->python_path, sizeof(path));
-				PySys_SetPath(path);
+				PySys_SetPath(inst->python_path);
 			}
 #endif
 		}


arr2036:
Wasn't uninitialised, was necessary, hence travis failing ;)

tremolo:
There is still a bug in there, you can reproduce it with python (2.7) and by specifying the python_path in the config.

What you are doing is copying a string (char *) to a pointer of a string (char **)
then 
I think this should fix it and please the compilers (tested with gcc and -Werror )

`memcpy(&path, &(inst->python_path), sizeof(path));`

tremolo:
stack trace:
```
  # Instantiating module "python" from file /opt/freeradius/etc/raddb/mods-enabled/python
Python version: 2.7.12 (default, Nov 19 2016, 06:48:10)  [GCC 5.4.0 20160609]

Program received signal SIGSEGV, Segmentation fault.
__strchr_sse2 () at ../sysdeps/x86_64/multiarch/../strchr.S:32
32	../sysdeps/x86_64/multiarch/../strchr.S: No such file or directory.
(gdb) bt
#0  __strchr_sse2 () at ../sysdeps/x86_64/multiarch/../strchr.S:32
#1  0x00007ffff25507c1 in PySys_SetPath () from /usr/lib/x86_64-linux-gnu/libpython2.7.so.1.0
#2  0x00007ffff29c4a3f in python_interpreter_init (inst=0x92c530, conf=0x882f60) at src/modules/rlm_python/rlm_python.c:923
#3  0x00007ffff29c4cc7 in mod_instantiate (conf=0x882f60, instance=0x92c530) at src/modules/rlm_python/rlm_python.c:1006
#4  0x0000000000427d25 in module_instantiate (modules=0x8538b0, askedname=0x883060 "python") at src/main/modules.c:758
#5  0x000000000042a92e in modules_init (config=0x6a65f0) at src/main/modules.c:2145
#6  0x0000000000433a28 in main (argc=2, argv=0x7fffffffe718) at src/main/radiusd.c:472
(gdb) up
#1  0x00007ffff25507c1 in PySys_SetPath () from /usr/lib/x86_64-linux-gnu/libpython2.7.so.1.0
(gdb) up
#2  0x00007ffff29c4a3f in python_interpreter_init (inst=0x92c530, conf=0x882f60) at src/modules/rlm_python/rlm_python.c:923
923					PySys_SetPath(path);
(gdb) print path
$1 = 0x7479702f74706f2f <error: Cannot access memory at address 0x7479702f74706f2f>
(gdb) print inst->python_path
$3 = 0x92cc90 "/opt/python"
```

arr2036:
The bug was fixed in v4.0.x, which was the branch you opened this pull-request against.  Merging to v3.0.x was pending you confirming that the issue was fixed.

As per your other email - ``PY_VERSION_HEX`` is defined on my system, which version of python are you using?

tremolo:
ok, thanks for fixing this. the code seems to be fine now on v4.0.x and v3.0.x

PR #2058, opened 2017-09-07
Updated Huawei dictionary

herwinw:
Source: http://support.huawei.com/onlinetoolsweb/aaa/en/export.jsp?domain=0&lang=EN&product=&version=&key=&flag=111111

I'm not sure how the previous version was assembled, but it was completely different from the upstream version

### DIFF
--- a/share/dictionary.huawei
+++ b/share/dictionary.huawei
@@ -2,7 +2,7 @@
 # Copyright (C) 2015 The FreeRADIUS Server project and contributors
 ##############################################################################
 #
-#	Dictionary for Huawei.  See also dictionary.h3c
+#	Dictionary for Huawei.
 #
 #	$Id$
 #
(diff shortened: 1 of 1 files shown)


herwinw:
(For my own reference: Q136980)

alandekok:
I'm wary of changing vendor dictionaries.  But doing that in v4 is probably OK, so long as it's documented.

herwinw:
After searching in the mailing list archives: [this post](http://lists.freeradius.org/pipermail/freeradius-users/2007-August/021479.html) implies the exact dictionary used might differ among Huawei devices.

I didn't do a real check, but if I just look at the first 24 attributes it seems the types of attributes do not change (attribute 17 is a string in both the old and the new dictionary, all the others that occur in both old and new dict are integers). Maybe we could merge the two and use double names for certain attributes? This would at least make it possible to add the attributes in a regular way, but displaying them in debugging could still create some confusion.

arr2036:
Hm, this seems to remove a lot of attributes... Maybe copy the old dictionary to dictionary.huawei.legacy unless you can find model/version mappings?

PR #2059, opened 2017-09-07
Use proper key expiration time from the trust router response

alejandro-perez:
v3.x of the trust router lib has added an accessor to the key
expiration information from the Trust Router server reply.
Hence, that information must be used to expire the realm,
instead of the hardcoded 600 seconds that were included as a
workaround until this accessor was implemented.


alandekok:
If the function doesn't exist in earlier versions of the library, then calls to it should be protected with a pre-processor flag.

alejandro-perez:
It seems that the trust router header files do not have versioning information (although there is a tr_versioning.h file). So, I'm trying to modify src/modules/rlm_realm/configure.ac and re-generate the configure script, but I'm failing. The generated script is not valid. It says:
```
./configure: line 2701: syntax error near unexpected token `tr_tid,'
./configure: line 2701: `	FR_SMART_CHECK_LIB(tr_tid, tidc_create)'
```
even if I don't modify the configure.ac file, just try to generate either with "autoreconf" or "autoconf configure.ac > configure"

Do I need some other additional parameter to properly generate the configure script?

alejandro-perez:
Nevermind, I found out I was missing the -I parameter to autoconf.

alejandro-perez:
It works, but seems like ./configure script grew a bit too much for just adding an additional line to configure.ac. If you want to build configure out of configure.ac yourself in a more controlled environment, that might be even better.

alandekok:
Just submit a patch without the `configure` script updated.  That can be done as a separate step.

alejandro-perez:
> Just submit a patch without the configure script updated. That can be done as a separate step.

Said and done!

alejandro-perez:
There is still a missing line to be merged. That's why there is a conflict.

alejandro-perez:
```

alandekok:
Should be fixed now.

PR #2063, opened 2017-09-13
Fixes for docker build for debian9

herwinw:
It still doesn't work, but fails further down the line.

### DIFF
--- a/scripts/docker/build-debian9/Dockerfile
+++ b/scripts/docker/build-debian9/Dockerfile
@@ -8,7 +8,7 @@ RUN apt-get update && apt-get install -y devscripts equivs git quilt
 #
 #  Install eapol_test dependencies
 #
-RUN apt-get install libnl-3-dev libnl-genl-3-dev
+RUN apt-get install -y libnl-3-dev libnl-genl-3-dev
 
 #
 #  Setup a src dir in /usr/local


arr2036:
Stretch does have a new enough version of cake, we need 3.8.2 or later for the packaging for libkqueue to function correctly.

arr2036:
Where does it current fail?

herwinw:
> Stretch does have a new enough version of cmake, we need 3.8.2 or later for the packaging for libkqueue

It looked like it worked with cmake from debian when trying it earlier today on my workstation at work, but it looks differently on my workstation af home. Reverted that part


> Where does it current fail?

It looks like the build of libkqueue only makes a package `libkqueue-2.1.1-Linux.deb`, but no `-dev` package. The `mk-build-deps` of `v4.0.x` errors on this

```
Starting pkgProblemResolver with broken count: 1
Starting 2 pkgProblemResolver with broken count: 1
Investigating (0) freeradius-build-deps:amd64 < 4.0.0+git @iU mK Nb Ib >
Broken freeradius-build-deps:amd64 Depends on libkqueue-dev:amd64 < none @un H >
  Removing freeradius-build-deps:amd64 because I can't find libkqueue-dev:amd64
Done
```

mcnewton:
This is fixed, if there are still issues with the Dockerfiles please do a new PR.

PR #2088, opened 2017-10-09
Misc updates

jsynack:
Here are some misc updates.

- mostly docs, minor typos/corrections and minor adjustments (http -> https) to gauge interest of these changes
- one dictionary update for Foundry/Brocade for an overlapping VSA (both are strings). A vendor case has been brought up about this - given the company split that is taking place for Brocade (Oct-2017) this will probably see some areas of churn until they get their PEN / VSA tracking sorted.
- 

Questions/thoughts/directions - just let me know!


arr2036:
Thanks! Polishing the config/docs is always appreciated.  The only thing in future would be splitting unrelated commits into separate pull requests.

PR #2091, opened 2017-10-11
add new Juniper/Unisphere/ERX VSAs

bmork:
Importing new attributes from
https://www.juniper.net/documentation/software/junos/junos161/radius-dictionary/unisphereDictionary_for_JUNOS_v16-1.dct
using the exact attribute names Juniper use, following the precedence set
by the original conversion from dictionary.erx to dictionary.unisphere

Yes, Juniper *do* use four different attributes prefixes here:
 Unisphere, Pcp (sic), Jnpr and ERX(!)

Quite impressive.  And quite a mess.  But at least it's not *our* mess
if we just got with their names...

Signed-off-by: Bjørn Mork <bjorn@mork.no>


bmork:
Can you please cherry-pick this for v3.1.x too?  Or should I do that and create a separate pull request?

alandekok:
I can pull the changes in, but the v3.1 branch is dead and no longer supported.  It has been removed from github, and exists only as a tag with the last commit.

bmork:
Ah, right.  I guess I should have seen that before asking.  Thanks

PR #2093, opened 2017-10-17
logrotate: don't accidentally define global options

zmousm:
> The logrotate config file that ships with freeradius sets global options 
> in its file (/etc/logrotate.d/freeradius). These options do not just 
> apply to the freeradius log files, but to all log files that logrotate 
> encounters after processing /etc/logrotate.d/freeradius (e.g. other logs 
> defined in /etc/logrotate.d with a filename that come later 
> alphabetically). [...]

As reported to/fixed for Debian:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=872158
https://anonscm.debian.org/cgit/pkg-freeradius/freeradius.git/commit/?id=42fb7f66d29771f658c97d1812dc9594f76f7083

### DIFF
--- a/debian/freeradius.logrotate
+++ b/debian/freeradius.logrotate
@@ -1,34 +1,30 @@
-# You can use this to rotate the /var/log/freeradius/* files, simply copy
-# it to /etc/logrotate.d/radiusd
-
-#
-#    Global options for all logfiles
-#
-daily
-rotate 52
-missingok
-compress
-delaycompress
-notifempty
-
-#
 #  The main server log
-#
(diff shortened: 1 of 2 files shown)


jsynack:
Random thought: 
This also bumped the rotate count to 52 days/files. 

Was the intent to make this a weekly rotate in order to keep a year of logfiles around?  52 otherwise just sounds like an oddball number for days, especially when it was previously 14/2-weeks worth :)

zmousm:
Sorry about that, it was inadvertent. I will submit a new PR to revert the value for `daily` from 52 to 14.

PR #2094, opened 2017-10-18
add suport for IA_PD in sqlippool

maximumG:
The *sqlippool* module currently handles the IPv4 and IPv6 address (IA_NA) dynamic assignation.

This pull request gives the ability for the *sqlippool* module to handle IPv4 / IPv6 address as well as IPv6 delegated prefix assignation (linked to the mailing list discussion http://freeradius.1045715.n5.nabble.com/Delegated-IPv6-Prefix-amp-sqlippool-td5746143.html).

The idea is being able to handle IPv6 IA_PD the same way as the IPv6 IA_NA and IPv4 address.

### DIFF
--- a/src/main/client.c
+++ b/src/main/client.c
@@ -1358,7 +1358,8 @@ RADCLIENT *client_afrom_request(RADCLIENT_LIST *clients, REQUEST *request)
 			for (parse = client_config; parse->name; parse++) {
 				if (parse->offset == dynamic_config[i].offset) break;
 			}
-			rad_assert(parse);
+
+			if (!parse) break;
 
 			cp = cf_pair_alloc(c->cs, parse->name, strvalue, T_OP_SET, T_BARE_WORD, T_SINGLE_QUOTED_STRING);
 		}
--- a/src/modules/proto_dhcp/dhcpd.c
+++ b/src/modules/proto_dhcp/dhcpd.c
@@ -423,6 +423,9 @@ static int dhcp_process(REQUEST *request)
 
 	/* BOOTREPLY received on port 67 (i.e. from a server) */
 	if (vp->vp_byte == 2) {
+		if (request->reply->code == 0) {
+			return 1;
+		}
 		return dhcprelay_process_server_reply(request);
 	}
 


alandekok:
Hmm... simple enough patch, I guess.

Tho at that point, I'm inclined to remove all of the booleans, and just let the admin define an attribute which is used for IP address assignment.

We could probably also update the queries to reference that attribute, which means less editing of queries when you change from Framed-IP-Address to Framed-IPv6-Address, etc.

maximumG:
> Hmm... simple enough patch, I guess.

Exactly simply enough, but it will definitely help to offload our current BSS which is currently handling the dynamic IPv6 PD assignation.

> Tho at that point, I'm inclined to remove all of the booleans, and just let the admin define an attribute which is used for IP address assignment.

I think it will be better than all of the boolean indeed. As I am new to the freeradius development, is there any available function to convert a string to an actual freeradius type ?

> We could probably also update the queries to reference that attribute, which means less editing of queries when you change from Framed-IP-Address to Framed-IPv6-Address, etc.

I don't really get what you mean. On my side, I am currently using the same column 'framedipaddress' to host both IPv4 and IPv6 prefix without changing anything to the queries. Do you have any example in mind ?

maximumG:
Based on your previous remarks, I dig into the freeradius code and found an awser to all of my questions. 

I updated the feature request with the ability for the admin to configure the RADIUS attribute he wants to assign the IP ressources to (*ip_attribute_name* in sqlippool). This dynamic attributes is then referenced in all related config file (sqlippool, queries.conf).

Hope this feature request fullfill what you had in mind previously ?

maximumG:
@alandekok : Are you still interessted in this pull request ? I guess that everything you mentionned in your first response is now completed after my updated pull request ?

alandekok:
It looks pretty good.  I'll have to do some tests on my end, and maybe some minor tweaks.

I've been busy the last few weeks, so it might take a bit.  But I'll try to pull it in before 3.0.16 is released.

alandekok:
I've pushed some patches base on your pull request.  Change include:

- checking that the attribute is IPv4 / IPv6 / prefix
- using `%{${attribute_name}}` instead of `${attribute_name}`.  The first one will do dynamic expansion of the attribute.  The second one is just a fixed string.
- allowing NAS-IP-Address to be v4 or v6 by default.  i.e. it's possible for IPv6 assignment to be done via RADIUS over IPv6.

Please double-check and test.

maximumG:
Thanks for the work. It looks pretty good, excepts for those 2 points below.

- There might be some uses cases where the RADIUS needs to send to the NAS mutliple *Delegated-IPv6-Prefix* VP. With the current implementation it is not possible as we check for the VP presence before doing any assignation (line 514 in rlm_sqlippool.c).

- Regarding the code the use of *framed-ip-address* variable name is a bit confusing as we do not deal with only framed-ip-addresss but rather different attribute_name. It could be interessting to rename this variable for all developper that might deal with this code, don't you think ?

Anyway I tested the implementation with Framed-IP-Address, Framed-IPv6-Adress, Delegated-IPv6-Prefix and some other vendor attributes and it works as expected. Thanks for the work and your quick involvment.

maximumG:
Just tested the last commit you made. Working as expected. From my side you can close the pull request.

Hope it will be added in the next official release 3.0.16 so that we can start using it in production.

Again thanks for the work you made :smile:

maximumG:
I took the opportunity of this request to also use the "attribute_name" in the DHCP with a value of _DHCP-Your-Client-IP-Address_.

However as of the code we wrote, we cannot use vendor attribute as VSA. Would it be acceptable to remove this lock in order to use a vendor VSA (e.g : DHCP dictionary) ?

PR #2096, opened 2017-10-19
Coverity scan

jsynack:
I was looking at some of the Coverity results and several are dated due to branch freshness of the coverity_scan branch (I presume it is set to scan this branch only based on the age / dates).

 Would like to get the coverity_branch caught up with HEAD/v4.0.x for freshness of scan/results (approx 1k commits behind) so that things can be reviewed.

This is a large merge -all from HEAD/v4.0.x with no other local changes. 

If the Coverity scan/freshness is triggered elsewhere / someplace else - more than happy to help to get it caught up for more accurate scanning results.  Once caught up - it should be more manageable for regular syncs.

After the refresh of the scanning results, any fixes would then go back to HEAD/v4.0.x and potentially v3.0.x  .It may make sense to have a dedicated coverity branch for v3.0.x - but I dunno how the back-end is organized/works for Coverity scanning setup.


alandekok:
We just do a force push of the branch instead of merging.  We don't really care about the history of the coverity_scan branch, just that the scan runs.

I've done a force push, and Coverity should run later today.

PR #2098, opened 2017-10-20
A few adjustments for radsniff

jsynack:
1. minor grammar (were / we're, a/an)
2. update the usage text to show proper use of -C <checksum_type>
3. Make use of FR_AUTH_UDP_PORT and FR_COA_UDP_PORT where it makes sense

### DIFF
--- a/src/main/radsniff.c
+++ b/src/main/radsniff.c
@@ -972,7 +972,7 @@ static int _request_free(rs_request_t *request)
 	bool ret;
 
 	/*
-	 *	If were attempting to cleanup the request, and it's no longer in the request_tree
+	 *	If we're attempting to cleanup the request, and it's no longer in the request_tree
 	 *	something has gone very badly wrong.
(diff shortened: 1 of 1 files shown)


arr2036:
Does radsniff actually work for you on the v4.0.x branch? I tried it little while back and it seemed non-functional, but maybe that's just linux...

jsynack:
Oddly enough - radsniff isnt working for me on linux while reading from interfaces. I assumed it was something I was breaking on my host as I have lots of other libpcap / PF_RING items going on.

I did compile/run on MacOS X without problems while reading from interfaces - so it would appear there is as a linux-specific item preventing interface-reading. Files seemed to work out.

I figured it was all the code churn taking place in some of the event work.

arr2036:
It's more libkqueue is not a perfect mirror of kqueue and we're still working out the areas where the behaviour differs.

PR #2100, opened 2017-10-23
Virtual server fixup

jsynack:
Update virtual-server example to be a mostly working example with the new requirements for namespace to be set on a server {} block.

There are other example config files that need this same sort of update - just making sure this is useful before updating the remaining ones.

While here - also fix a small typo and expand the error message when a virtual server lacks the namespace directive.


arr2036:
Yes, it'd definitely be helpful to have the other demo virtual servers fixed up.  All the changes look fine.

PR #2102, opened 2017-10-25
Fixed rlm_rest

herwinw:
Add an explicit type to the value assigned to REST-HTTP-Status-Code.
Without this, the following error occurs:

CONSISTENCY CHECK FAILED src/lib/util/pair_cursor.c[478]: VALUE_PAIR
attribute 0xb6c639a8 "REST-HTTP-Status-Code" data type (int8) does not
match da type (uint32)


arr2036:
Good catch :)

PR #2106, opened 2017-11-03
fix: correct receiving of multiple RADIUS packets through RadSec

semik:
On heavy loaded FR servers occurs a situation when multiple RADIUS packets are received at one time. Existing code inside function ``tls_socket_recv()`` isn't able to handle this situation.  Function ``tls_application_data()`` always reads only one RADIUS packet from OpenSSL buffer. Rest of received packets remains in OpenSSL buffer and waiting for more data coming through RadSec to trigger reading of next RADIUS packet.

On small deployments is ``idle_timeout = 30`` good enough to prevent any serious impact of this problem. On large deployments isn't this enough even with combination ``lifetime = 300`` which forces RadSec connection restart every 5minutes.

To easily reproduce the problem enable ``sleep(1)`` at ``src/main/tls_listen.c:208``. This will slow down PEAP-MSCHAPv2 authentication to more than 10s but the problem will show quickly. When running ``/usr/sbin/freeradius -fxxx -l stdout`` you can see following lines in output in the situation when the problem is fixed:
```
Thu Nov  2 22:55:56 2017 : Debug: more TLS records after dual_tls_recv
Thu Nov  2 22:55:56 2017 : Debug: (0) Reading pending buffered data
```

I work for CESNET as Czech eduroam operator. During summer 2017 we had campaigned for upgrading to freeRADIUS v3 and in the result, upgraded institutions stated to have troubles authenticating their traveling users. We believe that this patch correctly fixes receiving multiple RADIUS packets at once.

### DIFF
--- a/src/main/tls_listen.c
+++ b/src/main/tls_listen.c
@@ -186,8 +186,29 @@ static int tls_socket_recv(rad_listen_t *listener)
 
 	request = sock->request;
 
+	/*
+	 * PEAK: The previous invocation of tls_socket_recv might have
+	 * already received enough data and another RADIUS packet is ready
+	 * in the record-layer queue. (See dual_tls_recv.)
+	 */
+	if (SSL_pending(sock->ssn->ssl)) {
+		RDEBUG3("Reading pending buffered data");
+		sock->ssn->dirty_in.used = 0;
+		goto skip_socket_io;
+	}
+
 	RDEBUG3("Reading from socket %d", request->packet->sockfd);
 	PTHREAD_MUTEX_LOCK(&sock->mutex);
+#if 0
+	/*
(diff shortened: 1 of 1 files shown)


alandekok:
I'll push a reformatted patch based on this one.

* the change in `tls_socket_recv()` does not lock the mutex, but jumps to code which tries to unlock the mutex
* there's no need for testing code in the server, so the `#if 0` section is not necessary
* adding "PEAK" comments everywhere is unnecessary.  If every contribution had similar comments, the code would quickly become unreadable.

arr2036:
Ugh yes, it's equivalent to scrawling "I woz 'ere" over random parts of the code.

semik:
Thanks for reformatting patch, for your feedback and quick response. :+1:

PR #2109, opened 2017-11-07
fix rpm build errors

mattrose:
As much as I don't like wildcards in spec files, there's just too much churn with modules right now in 4.0.x

Also, removed the file attributes from /etc/raddb/sites-enabled as all of the files in there are symlinks.

### DIFF
--- a/redhat/freeradius.spec
+++ b/redhat/freeradius.spec
@@ -648,75 +648,7 @@ fi
 %attr(755,root,root) %{_libdir}/freeradius/lib*.so*
 # RADIUS Loadable Modules
 %dir %attr(755,root,root) %{_libdir}/freeradius
-%{_libdir}/freeradius/proto_arp.so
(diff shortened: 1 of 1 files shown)


arr2036:
is it smart enough not to package the same files twice? How does this work when specific .so files are meant to go in other packages?

mattrose:
yes and we'll have to split them up at that point.

alandekok:
It should be that smart.  The Suse packages have been doing that for 10 years, and no one's complained.

If not, we'll have to manually build the list.  TBH, the best thing would be to write a script to generate the list for the spec file.

mattrose:
SuSE and Redhat RPMs use the exact same format, so what works for suse should work for redhat.

arr2036:
Yeah I was more wondering why this was done manually in the first place, and how the rpmbuild figures out which package an so belongs to if it matches the wildcard and is listed explicitly.

PR #2111, opened 2017-11-08
Remove debian patch to disable dhcp

mattrose:
The way this patch interacts with quilt, and rules/all.mk seems to be very brittle, and I was thinking that it would be better to leave dhcp enabled for the development period.


alandekok:
fix pushed.

PR #2115, opened 2017-11-09
remove __DATE__ and __TIME__

mcnewton:
"reproducible builds" is apparently a thing, and gcc
has started complaining

### DIFF
--- a/src/main/radclient.c
+++ b/src/main/radclient.c
@@ -77,7 +77,7 @@ static char const *radclient_version = "radclient version " RADIUSD_VERSION_STRI
 #ifdef RADIUSD_VERSION_COMMIT
 " (git #" STRINGIFY(RADIUSD_VERSION_COMMIT) ")"
 #endif
-", built on " __DATE__ " at " __TIME__;
+;
 
 static void NEVER_RETURNS usage(void)
 {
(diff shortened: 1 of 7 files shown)


mattrose:
This has been a kind of holy grail for debian for a while now.

mcnewton:
Yeah, for whatever reason. But TBH saying when it was built doesn't tell you much (apart from that Debian are recently building really old software... maybe that's the real reason ;-) )

arr2036:
yeah now we have commits in the version number date/time not needed so much.

mcnewton:
I sort of like the build time, but I think that's because I'm used to seeing it there, rather than it actually being useful. occasionally if you build with uncommitted changes and can't remember which is which? maybe dev builds is the right compromise

arr2036:
It's useful when you're trying to figure out if the new packages you just built were actually deployed.

arr2036:
v3.0.x doesn't pass through the release number... not even sure if it passes through the commit hash to the various packaging systems.  I'd much rather this be an option to the configure script ``--enable-reproducible-builds``, than to always strip data/time by default.

mcnewton:
Yeah, OK - configure option is a good idea. Then the packaging scripts can turn them off, but they're on for everyone else. Though without that set probably need to start setting -Wno-date-time to stop the silly warnings coming up. I'll have a look.

PR #2118, opened 2017-11-10
Fix status-server retransmits

pwdng:


alandekok:
There's a lot of whitespace reformatting in the patch, which makes it harder to tell what's changed.

Can do update the patch without the whitespace changes, and do a force-push to the branch?

arr2036:
The underlying issue is that the LTB and RedHat packages both use the same soname, but do not provide exactly the same set of exported symbols.

alandekok:
FIxes pulled into v4.0.x

PR #2125, opened 2017-11-16
set OpenSSL < v1.1.0 threads callbacks in a proper way

lkpdn:
Noteworthy is that this ensures threads pool robustness e.g., when embedded interpreter-based rlm_* modules utilises external modules, which expects openssl < v1.1.0 threads callbacks being set up properly on main thread setup_ssl_mutexes().

### DIFF
--- a/src/main/threads.c
+++ b/src/main/threads.c
@@ -223,7 +223,13 @@ static const CONF_PARSER thread_config[] = {
 static pthread_mutex_t *ssl_mutexes = NULL;
 
 #ifdef HAVE_CRYPTO_SET_ID_CALLBACK
-static unsigned long get_ssl_id(void)
+
+/*
+ *	Use preprocessor magic to get the right function and argument
+ *	to use.  This avoids ifdef's through the rest of the code.
+ */
+#if OPENSSL_VERSION_NUMBER < 0x10000000L
+static unsigned long ssl_id_function(void)
 {
 	unsigned long ret;
 	pthread_t thread = pthread_self();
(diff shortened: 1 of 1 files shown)


src/main/threads.c
```diff
@@ -236,6 +237,21 @@ static unsigned long ssl_id_function(void)
...
+	unsigned long ret;
+	pthread_t thread = pthread_self();
+
+	if (sizeof(ret) >= sizeof(thread)) {
+		memcpy(&ret, &thread, sizeof(thread));
+	} else {
+		memcpy(&ret, &thread, sizeof(ret));
+	}
+
+	CRYPTO_THREADID_set_numeric(id, ret);
```
alandekok: https://www.openssl.org/docs/man1.0.2/crypto/threads.html says:

*use CRYPTO_THREADID_set_numeric() if thread IDs are numeric, or CRYPTO_THREADID_set_pointer() if they are pointer-based*

So it's probably best to do that instead.

alandekok: And https://linux.die.net/man/3/pthread_self says:

*POSIX.1 allows an implementation wide freedom in choosing the type used to represent a thread ID; for example, representation using either an arithmetic type or a structure is permitted. Therefore, variables of type pthread_t can't portably be compared using the C equality operator (==); use pthread_equal(3) instead.*

So using `pthread_id` as an ID might work on some platforms, but it's not portable.

It's probably better to use thread-local storage, and to put an integer ID there... though that's a bit more work.

alandekok: Hmm... that's all original code.  I'll go fix that.

lkpdn: Ok thanks. And I also agree if you would change it to *_pointer to delegate the original logic to CRYPTO implementation.

lkpdn:
The original intention has been directly reflected in v3.0.x commit 509c7e971955 ("set thread callback for newer OpenSSL. Fixes #2125") with 'Fixes' which just reduces ifdefs, but in the way that does not reflect what you pointed out in the previous comments. If it is okay for you now, I will close this PR, since not using *_pointer interface is not harmful in almost all cases.

PR #2127, opened 2017-11-17
Fast crypto binding fix

frenche:
PR is against v3.

### DIFF
--- a/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c
+++ b/src/modules/rlm_eap/types/rlm_eap_fast/eap_fast.c
@@ -1086,10 +1086,11 @@ static PW_CODE eap_fast_process_tlvs(REQUEST *request, eap_handler_t *eap_sessio
 				break;
 			case EAP_FAST_TLV_CRYPTO_BINDING:
 				if (!binding) {
+					int offset = 2 * sizeof(uint16_t);
 					binding = talloc_zero(request->packet, eap_tlv_crypto_binding_tlv_t);
-					memcpy(binding, vp->vp_octets, sizeof(*binding));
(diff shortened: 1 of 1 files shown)


alandekok:
Hmm... there's a lot wrong with the EAP-FAST code.  The simplest thing is that the tlv_type and length should *not* be part of the crypto_binding structure.

I've pushed fixes:

* return `code` from crypto_binding() as per your patch
* checks that the VP containing the crypto_binding is the right size
* since the contents of the VP are just the data, and the crypto_binding check takes just the data...
* pass the `vp->vp_octets` directly to the  crypto_binding()  function

Please test.  I'd like to release 3.0.16 next week.

frenche:
now eapol_test ins't happy with the binding that we send, trying to figure what changed.

EAP-FAST: Received Compound MAC - hexdump(len=20): d8 da fc 86 a4 c2 db 41 52 d2 ae ee 20 78 80 5f d1 8b 16 8e
EAP-FAST: Calculated Compound MAC - hexdump(len=20): f9 d9 a4 1e 56 78 cb d6 ba 34 68 d9 89 4e d6 02 9b 9e ee 87
EAP-FAST: Compound MAC did not match
EAP-FAST: Add Result TLV(status=2)

alandekok:
:( Sorry... I suspected something like that might happen.

alandekok:
Arg... the TLV header *and* length are part of the signature:

`fr_hmac_sha1(binding.compound_mac, (uint8_t *)&binding, ..`

so that part has to be re-done.  I'll take a look.

alandekok:
OK, pushed a fix.  That should help.

frenche:
Works perfect now, thanks!

PR #2129, opened 2017-11-20
Safer openssl threads setup

lkpdn:
This prevents needless crashes or hard-to-debug behaviors, e.g., when using rlm_perl with a custom script which "requires" Net::SSLeay that would bootstrap XS code setting up openssl callbacks if not ready.
Furthermore, 1d9a6d6 ("do setup_ssl_mutexes() before modules_init()") will make all fair in the sense that we do not have to care any more about which module instantiation (sometimes unintentionally) will win the race to set up the callbacks in the first place.

### DIFF
--- a/src/include/radiusd.h
+++ b/src/include/radiusd.h
@@ -537,6 +537,10 @@ void	thread_pool_lock(void);
 void	thread_pool_unlock(void);
 void	thread_pool_queue_stats(int array[RAD_LISTEN_MAX], int pps[2]);
 
+#ifdef HAVE_OPENSSL_CRYPTO_H
+int	setup_ssl_mutexes(void);
+#endif
+
 #ifndef HAVE_PTHREAD_H
 #  define rad_fork(n) fork()
 #  define rad_waitpid(a,b) waitpid(a,b, 0)
--- a/src/main/radiusd.c
+++ b/src/main/radiusd.c
@@ -338,6 +338,15 @@ int main(int argc, char *argv[])
 	 */
 #ifdef HAVE_OPENSSL_CRYPTO_H
 	tls_global_init();
+	/*
+	 *  Set up the mutexes and enable the thread callbacks before individual modules
+	 *  .instantiate() would happen to do it. We need them ONLY if we're spawning,
+	 *  AND we're running normally.
+	 */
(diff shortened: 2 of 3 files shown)


alandekok:
I'm not sure why the first patch is necessary.  The code in `thread_pool_init()` only calls `setup_ssl_mutexes()` if `spawn_flag == true`.  And doing `radiusd -XC` sets `spawn_flag = false`

Moving that code to `radiusd.c` doesn't seem to add any benefit.

alandekok:
Hmm... OK, I've pushed a change which sets up the mutexes in `tls_global_init()`.  That has fewer visible code changes than the alternative.

That also ensures that the mutexes are set up before `modules_init()` is called.

I think it should work now.  The patch "set openssl callbacks only if needed" should no longer be necessary.

i.e. the mutexes and callbacks are set up before the modules.  So the code:

    if (!CRYPTO_get_locking_callback())
        CRYPTO_set_locking_callback(ssl_locking_function);

Should *always* have the `CRYPTO_get_locking_callback()` function return `NULL`.  Because nothing else will set before that call.

lkpdn:
Ah you're right, we don't need the former one as we just call CRYPTO apis, never bundling it with any extra setups.

About the latter one, what I thought was that:
To prevent external library from doing any such extra setups, which is invoked only when they recognise that they are the first one to initialize mutexes/callbacks, clears off hard-to-debug issues. That is because:
  - If such module requires mutexes/callbacks initialisation being successfully called once, the system stability depends on whether or not the initialisation of the module occurs in the first place, in main thread. If it's just one, we may set `instantiate { } ` section appropriately, but seems a bit tricky and we cannot have two or more modules like that anyway.
  - Conversely, if external library can do without it but rather should not call mutexes/callbacks initilisation on .instantiate().
  - So, having freeradius initialise it in the first place seemed harmless and make things simple and deterministic.

but now I notice that second one is obviously the problem of that module itself and the way we load it, so it's totally pointless. So, let me close this one. I have been debugging around this in the last few days, so maybe I will ask you some help. Thanks.

lkpdn:
And thanks for 09759f7 ! I noticed just now.

PR #2131, opened 2017-11-22
Fix missing id column in radusergroup

dulin:
Some ORM drops support for table without primary key. I see missing id is only in MySQL schema.


arr2036:
Can you squash the commits please, then we'd be happy to merge this.

dulin:
Done

arr2036:
thanks

sayajin101:
could we also merge this into freeRadius v3, since all database schema's other than MySQL has an auto_incrementing primary key id column, this is causing issues with table syncing for me.

alandekok:
Done.

PR #2137, opened 2017-12-11
Added new Patton Attribute

BriceSchaffner:
Added new Patton Attribute called Patton-Disconnect-Reason (37). This attribute contains usually the same information as the Patton-Disconnect-Cause but in string format instead. This attribute is available on Patton device from version 3.13.0.


alandekok:
fixed in commit 45d018d7.  And pulled into v3.0.x with commit f14227dbf17c62ee

PR #2140, opened 2017-12-12
Removed unused attributes

BriceSchaffner:
The Patton-Calling-Station-Id is not used by Patton device, the regular Radius Callilng-Station-Id is used instead. Same for Patton-Called-Stations-Id


alandekok:
Some people may be running older firmware which still uses the old attributes.  As such, we can't remove attributes from the dictionaries.

i.e. the dictionaries contain *known* attributes, even for vendors who have been out of business for 20 years.  The dictionaries don't contain only *used* attributes.

If the dictionaries contained only used attributes, then everyone would remove all of the dictionaries that they don't use.  And then get hit when they bought new equipment.

PR #2143, opened 2017-12-17
V3.0.x

jusjose24:
Fix 'radiusd' crash when OCSP certificate response error is unknown
    
    UNKNOWN and REVOKED errors are treated same in 'radiusd'. Sometimes
    'rev' is NULL for the UNKNOWN error case. This makes 'radiusd' crash
    while printing the revoke timestamp.

diff --git a/src/main/tls.c b/src/main/tls.c
index 2690ebca0..fca2ddcac 100644
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -1939,7 +1939,7 @@ static ocsp_status_t ocsp_check(REQUEST *request, X509_STORE *store, X509 *issue
                REDEBUG("ocsp: Cert status: %s", OCSP_cert_status_str(status));
                if (reason != -1) REDEBUG("ocsp: Reason: %s", OCSP_crl_reason_str(reason));
 
-               if (bio_out) {
+               if (bio_out && rev) {
                        BIO_puts(bio_out, "\tRevocation Time: ");
                        ASN1_GENERALIZEDTIME_print(bio_out, rev);
                        BIO_puts(bio_out, "\n");


alandekok:
I've pushed a simple patch.

We strongly prefer to NOT merge patches which have hundreds of changesets in them.

It should be simple to rebase your changes to the latest v3.0.x HEAD, and submit the patch from there.

PR #2144, opened 2017-12-18
V3.0.x

jusjose24:
Author: Justin Jose <justin.jose@alliedtelesis.co.nz>
Date:   Tue Dec 19 09:27:54 2017 +1300

    Fix 'radiusd' crash if 'bresp' is NULL
    
    'radiusd' is crashed in my test setup due to accessing NULL pointer.
    If 'brsep' is NULL, radiusd' should not proceed further processing the response.

diff --git a/src/main/tls.c b/src/main/tls.c
index 2f586c057..9bfe5cbf5 100644
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -1892,6 +1892,10 @@ static ocsp_status_t ocsp_check(REQUEST *request, X509_STORE *store, X509 *issue
                goto ocsp_end;
        }
        bresp = OCSP_response_get1_basic(resp);
+    if (!bresp) {
+       REDEBUG("ocsp: Error parsing response\n");
+        goto ocsp_end;
+    }
        if (conf->ocsp_use_nonce && OCSP_check_nonce(req, bresp)!=1) {
                REDEBUG("ocsp: Response has wrong nonce value");
                goto ocsp_end;

*************************
Sorry, the merge request from v3.0.x into v4.0.x was wrong. Please ignore.
*************************
Would this patch be pushed to V3.0.x if it make sense? Thanks


arr2036:
Please pull --rebase from origin v3.0.x and force push to yout git pull branch.

alandekok:
> jusjose24  wants to merge 7,152 commits into v4.0.x from v3.0.x

That is NOT going to happen.

Please pay attention to the branches you're using.  v4.0.x and v3.0.x are COMPLETELY different at this point.

Any merge request with 7000+ commits means you're doing something wrong.

jusjose24:
Would you please push #2144 into v3.0.x? Thanks

alandekok:
Don't put patches into the comments in github.  It's pointless.  You can use a tool called `git` to submit patches.

Don't submit PRs which merge 1000s of commits.

Don't submit PRs which try to merge v4.0.x branch into the v3.0.x branch.

Don't ignore comments which ask you to fix all this.

Don't post a PR, and immediately follow up with a comment of "Please merge this".   We DO get email about each PR.  We DO read each email.  We suggest that you do the same.

PR #2148, opened 2017-12-30
Cordel patch1

cordelster:
Fix in Makefile to remove variable OPENSSL that is no longer set.
Several patches to make ca.csr for external CA to create a subordinate CA which can not create/sign any further CA. 
Make file tested for vanilla test certs called from bootstrap and for subca in FreeIPA.
Makefile now looks for ca.key.sub to properly test and run each stage. Running the following begins the process:
$ make ca.csr

DogTag/FreeIPA profile added.


raddb/certs/Makefile
```diff
@@ -44,7 +44,7 @@ verify: server.vrfy client.vrfy
 
 passwords.mk: server.cnf ca.cnf client.cnf inner-server.cnf ocsp.cnf
 	@echo "PASSWORD_SERVER	= '$(shell grep output_password server.cnf | sed 's/.*=//;s/^ *//')'"		> $@
-	@echo "PASSWORD_INNER	= '$(shell grep output_password inner-server.cnf | sed 's/.*=//;s/^ *//')'"	>> $@
+	@echo "PASSWORD_INNER	= '$(shell grep output_password inner-server.cnf | sed 's/.*=//;s/^ *//')'">> $@
```
alandekok: We'd prefer to not pull in local whitespace patches.

alandekok:
Is this patch for v4.0.x?  It looks like it's for v3.0.x.

And please fix the commits to remove unnecessary changes.  i.e. "add dogtag / delete dogtag".  We prefer to have a clean history.

PR #2156, opened 2018-01-10
Remove undef var

cordelster:
Remove $(OPENSSL) from previous merge where the variable was already dropped.


alandekok:
thanks.  I'll just change the variable and push a commit.  I'd prefer to avoid multiple merge commits.

PR #2167, opened 2018-01-24
3.0.x - rest - curl option CURLOPT_CAINFO

nchaigne:
As proposed on the ML, this patch adds support for curl option
CURLOPT_CAINFO (through a new parameter "ca_info_file").
This new parameter can be used instead of "ca_file" (which sets curl
option CURLOPT_ISSUERCERT).
(They can also be used both at the same time.)

CURLOPT_CAINFO is useful to validate a chain of certificate authorities.
If this option is not set, curl will try to validate the CA chain using
a default location, and it may fail.

CURLOPT_ISSUERCERT is useful to check the issuer of the server
certificate.

For reference, see:
https://curl.haxx.se/libcurl/c/CURLOPT_ISSUERCERT.html
https://curl.haxx.se/libcurl/c/CURLOPT_CAINFO.html

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -5,8 +5,18 @@ rest {
 	#  server.
 	#
 	tls {
-#		ca_file	= ${certdir}/cacert.pem
-#		ca_path	= ${certdir}
+		#  Certificate Authorities:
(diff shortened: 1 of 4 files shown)


alandekok:
Looks good.  Documentation, too!

Once the tests pass, I'll pull it in.

PR #2193, opened 2018-03-12
radclient "-i" didn't work, so just remove it to save confusion

mcnewton:
There's no current way to tell fr_packet_list_id_alloc to use a particular ID, and fr_packet_socket_t isn't available to us to change the ID after it's been allocated.

If anyone really needs to do this for testing it can be temporarily hacked into send_one_packet().

Might be what #2179 meant, though the issue wasn't clear.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -31,6 +31,7 @@ FreeRADIUS 3.0.17 Thu 11 Jan 2018 12:00:00 EST urgency=low
 	  not Stripped-User-Name.
 	* RedHat Systemd updates.  Fixes #2184
 	* Use correct API for State variable in rlm_securid.
+	* Remove broken radclient option "-i".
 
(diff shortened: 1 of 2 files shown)


arr2036:
Or fix it because it's a really useful diagnostic feature? :(

alandekok:
For v4, we should allow setting ID, and Request Authenticator.  That's very useful for testing.

mcnewton:
Yeah, I know. But at the moment it's in the help as a feature but it is just silently ignored, which is even worse as people think it does something :(

The easy hack for v3 is to just override the packet id after fr_packet_list_id_alloc has run, which seems to work enough for radclient (hence the commit comment), but isn't great as it will leave fr_packet_socket_t->id inconsistent. The proper fix is probably to be able to pass a requested ID to fr_packet_list_id_alloc, but that means touching src/packet.c, which I didn't want to do in the stable version for something so minor.

Looking at it for v4 at some point is my intention (I want to use it...)

PR #2200, opened 2018-03-26
A few fixes in fr_dhcpv4_packet_decode

nchaigne:
1) Options 26 and 57 are short => use vp_uint16

2) Field "opcode" is byte => use vp_uint8

3) Value 3 does not exist for field "Opcode". Client request is 1.
Option 63 should have been 60 (Vendor Class Identifier).
That said, I don't think this "MSFT 98" code ever worked - unless MS does some really weird stuff with DHCP. So maybe it's worth considering removing this altogether ? I'll let you decide. :)


alandekok:
Yeah, some of those were fixed in v2 and v3, but we didn't pull them over to v4.

PR #2160, opened 2018-01-17
Some OCSP fixes

frenche:
Two patches for OCSP and intermediate CA scenario.

PR #2202, opened 2018-03-27
OCSP fixes v4

frenche:
Fixes from PR #2160 for v4 branch (somewhat simplified).

Some logs from the different stages:
Initial, ocsp skipped although softfail set to false
https://pastebin.com/L80vNmzc
After first patch, ocsp not skipped:
https://pastebin.com/L7zCnEwQ
After second patch, we succeed to get issuer cert (fails just because it can't connect):
https://pastebin.com/tUStSWM9

### DIFF
--- a/src/lib/tls/ocsp.c
+++ b/src/lib/tls/ocsp.c
@@ -331,6 +331,11 @@ int tls_ocsp_check(REQUEST *request, SSL *ssl,
 		break;
 	}
 
+	if (issuer_cert == NULL) {
+		RWDEBUG("Could not get issuer certificate");
+		goto skipped;
+	}
+
 	/*
 	 *	Setup logging for this OCSP operation
 	 */
--- a/src/lib/tls/validate.c
+++ b/src/lib/tls/validate.c
@@ -324,11 +324,12 @@ int tls_validate_cert_cb(int ok, X509_STORE_CTX *x509_ctx)
 	 */
 	if (my_ok && conf->ocsp.enable){
 		RDEBUG2("Starting OCSP Request");
-		if (X509_STORE_CTX_get1_issuer(&issuer_cert, x509_ctx, cert) != 1) {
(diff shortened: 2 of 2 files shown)


frenche:
The "issuer_cert = NULL" was only done in the first patch, and then removed in second.

The reason is, before the first patch we don't call tls_ocsp_check() if X509_STORE_CTX_get1_issuer() doesn't return true, but we do need to call it in order to be able to apply softfail logic.
So momentarily, if X509_STORE_CTX_get1_issuer() fails (return value !=1), I set "issuer_cert = NULL" just to make sure.

After the second patch it is not necessary, as X509_STORE_CTX_get0_current_issuer() returns a pointer (so I just check for NULL to issue log with "common_name").

alandekok:
> The "issuer_cert = NULL" was only done in the first patch, and then removed in second.

I would strongly prefer to have patches which just work, instead of doing / undoing changes.

It may be better to just merge them into one patch which is clearer, and less confusing.

frenche:
Ok, I've squashed the two commits.

frenche:
Hmm, the CI errors seem unrelated.

arr2036:
@frenche can you rebase.  Thanks.

frenche:
Yea, that worked.
Thanks.

src/lib/tls/ocsp.c
```diff
@@ -331,6 +331,11 @@ int tls_ocsp_check(REQUEST *request, SSL *ssl,
 		break;
 	}
 
+	if (issuer_cert == NULL) {
+		REDEBUG("ocsp: couldn't get issuer certificate");
+		goto skipped;
```
arr2036: this will print "tls - oscsp - oscsp:", and it's not really an error in that it doesn't signal a fatal condition, so it's more RWDEBUG.

frenche: Switched to RWDEBUG, without prefix.

PR #2204, opened 2018-04-01
Decode additional DHCP options according to option Overload

nchaigne:
If option Overload is present in the 'options' field, then fields 'file' and/or 'sname' are used to hold more options. They are partitioned and must be interpreted in sequence (cf. RFC 2131 and 2132).


nchaigne:
Mmh... clang is not happy. I don't think it is related to this commit though.

PR #2205, opened 2018-04-04
EKU: Extract raw EKU OIDs into attributes

frenche:
This helps with matching a single OID regardless of its name.

Currently, the EKU attribute is a comma separated list of mixed OIDs and names (as known to ossl), e.g. 
TLS-Client-Cert-X509v3-Extended-Key-Usage += 1.2.250.1.4.1.311.21.8.16553884.13926527.9456536.14411893.1032733.83.1698795.44, 1.3.6.1.4.1.311.10.3.8, 1.3.6.1.4.1.311.10.3.6, TLS Web Client Authentication, E-mail Protection, Microsoft Encrypted File System

While with this patch, it adds and attribute for each OID, like:
TLS-Client-Cert-X509v3-Extended-Key-Usage-OID += 1.3.6.1.4.1.311.10.3.8
TLS-Client-Cert-X509v3-Extended-Key-Usage-OID += 1.3.6.1.4.1.311.10.3.6
TLS-Client-Cert-X509v3-Extended-Key-Usage-OID += 1.3.6.1.5.5.7.3.2
...

Thanks!

### DIFF
--- a/share/dictionary.freeradius.internal
+++ b/share/dictionary.freeradius.internal
@@ -533,8 +533,9 @@ ATTRIBUTE	TLS-Client-Cert-X509v3-Basic-Constraints 1930	string
 ATTRIBUTE	TLS-Client-Cert-Subject-Alt-Name-Dns	1931	string
 ATTRIBUTE	TLS-Client-Cert-Subject-Alt-Name-Upn	1932	string
 ATTRIBUTE	TLS-PSK-Identity			1933	string
+ATTRIBUTE	TLS-Client-Cert-X509v3-Extended-Key-Usage-OID 1936	string
 
-# 1934 - 1939: reserved for future cert attributes
+# 1937 - 1939: reserved for future cert attributes
 
 # 1940 - 1949: reserved for TLS session caching, mostly in 3.1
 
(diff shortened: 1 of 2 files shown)


alandekok:
That's a good idea.  Unless there are other objections, I'll put it in later today.

I'll also do minor cleanups to fix coding style consistency.  I suspect also that travis will fail because `eku_oid_attr` isn't declared `const`.

The solution there is to just delete that variable, and pass the raw string name directly to `fr_pair_make`.  Which is what all of the other code does.

Plus, definint variables like `for (int j = 0;...` isn't done elsewhere in the server.  For consistency, that variable should be declared earlier.

This is less about what's allowed in C (2018) than consistency in styling.  When we look at the source, we should see one consistent coding style.

I've worked with code bases where you could tell who wrote what based on style.  It was a nightmare. :(

arr2036:
Yeah looks useful.  Can you pull it over to v4 when you merge it

frenche:
Yea, was a bit hasty, also missed to EXTENDED_KEY_USAGE_free(eku).
I think I'll also try to take it out of the X509 extensions loop, to make it simpler.

frenche:
I updated with a slightly different version, hope it looks better.
Previous commit at: https://github.com/FreeRADIUS/freeradius-server/pull/2205/commits/73e18056cfd0966a3157c508c974d2cf90957e44

PR #2206, opened 2018-04-04
fix postinst script when installing freeradius with own config (no freeradius-config package)

it-eddy:
The postinst script from the freeradius package expects files that are installed by the freeradius-config package. When using my own config package they do not exist.
Quick fix: create link only if destination exists, insert additional check.


arr2036:
You can provide your own freeradius-config package that satisfies the dependency.  You also seem to be submitting fixes for unrelated things in the same PR.

PR #2211, opened 2018-04-10
SASL: Let configure SASL_SECPROPS per LDAP instance

frenche:
Active-Directory does not allow to use GSS-API signing / sealing when
over TLS (to avoid double encryption?).
While SASL_SECPROPS (or maxssf) coul be set globally via system's
ldap.conf, it is not enough if you want one instead to work with
TLS and the other without TLS but with signing and sealing of GSS-API.

Add a directive sasl_secprops, to allow setting it per instance.

See reference to MS bug at:
https://buildfarm.opencsw.org/source/xref/samba/source3/librpc/crypto/gse.c#303

### DIFF
--- a/raddb/mods-available/ldap
+++ b/raddb/mods-available/ldap
@@ -438,6 +438,12 @@ ldap {
 		chase_referrals = yes
 		rebind = yes
 
+		# SASL Security Properties (see SASL_SECPROPS in ldap.conf man page).
+		# Note - uncomment when using GSS-API sasl mechanism along with TLS
+		# encryption against Active-Directory LDAP servers (this disables
+		# sealing and signing at the GSS level as required by AD).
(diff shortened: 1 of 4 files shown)


arr2036:
This should be set inside the subsection and also configurable for user binds.

frenche:
For my use-case, setting globally per instance would suffice since TLS is set per instance.
Not sure the it is worth the efforts therefore, thoughts?

arr2036:
Ah ok, yeah I guess it effects both. Probably belongs in the option_config section then, that's where the rest of those per instance configurables live.

arr2036:
Are the security properties well defined? If so how many are there? Could we build the string internally, or would that be too much effort?

frenche:
Updated, and also changed suggested sasl_secprops to start with 'noanonymous,noplain' which is the defaults according to the man page.

PR #2232, opened 2018-05-22
fr_dhcpv4_packet_get_option - fix overload handling

nchaigne:
This fixes two errors in fr_dhcpv4_packet_get_option:

- value of "overload" option is not retrieved where it should be.
- it is possible to overload in 'sname' without overloading in 'file'.


nchaigne:
OK thanks, I merged your changes. Build is ok, and dhcpclient still works 👍

PR #2237, opened 2018-05-25
Convert MD5 xlat to new API

pwdng:


src/main/xlat_func.c
```diff
@@ -1024,34 +1024,63 @@ static int fr_value_box_to_bin(TALLOC_CTX *ctx, REQUEST *request, uint8_t **out,
...
  * Example: "%{md5:foo}" == "acbd18db4cc2f85cedef654fccc4a4d8"
  */
-static ssize_t md5_xlat(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
-			UNUSED void const *mod_inst, UNUSED void const *xlat_inst,
-			REQUEST *request, char const *fmt)
+static xlat_action_t xlat_md5(TALLOC_CTX *ctx, fr_cursor_t *out,
+			      REQUEST *request, UNUSED void const *xlat_inst, UNUSED void *xlat_thread_inst,
+			      fr_value_box_t **in)
 {
 	uint8_t		digest[16];
```
arr2036: That can be MD5_DIGEST_LENGTH, I know you didn't change it, but may as well fix it up here.

src/main/xlat_func.c
```diff
@@ -1024,34 +1024,63 @@ static int fr_value_box_to_bin(TALLOC_CTX *ctx, REQUEST *request, uint8_t **out,
...
+		vb_in = *in;
+	} else {
+		MEM(vb_in = fr_value_box_alloc_null(ctx));
+	}
+
+	MEM(vb_out = fr_value_box_alloc_null(ctx));
 
 	fr_md5_init(&md5_ctx);
-	fr_md5_update(&md5_ctx, p, inlen);
+	fr_md5_update(&md5_ctx, vb_in->vb_octets, vb_in->vb_length);
```
arr2036: Probably simpler just to call fr_md5_update with NULL and 0, then you save the alloc.

PR #2239, opened 2018-05-29
Convert rand xlat to new API

pwdng:


src/main/xlat_func.c
```diff
@@ -632,25 +632,33 @@ static ssize_t xlat_debug(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
...
+	/* Make sure it isn't too big */
+	if (result > (1 << 30)) result = (1 << 30);
 
 	result *= fr_rand();	/* 0..2^32-1 */
 	result >>= 32;
 
-	snprintf(*out, outlen, "%ld", (long int) result);
-	return strlen(*out);
+	MEM(vb = fr_value_box_alloc(ctx, FR_TYPE_UINT64, NULL, false));
+	vb->datum.uint64 = result;
```
arr2036: Use vb_uint64, as you did above with vb_uint32, but otherwise this looks great.

src/tests/keywords/rand
```diff
@@ -0,0 +1,29 @@
...
+}
+
+#
+#  Make sure random number is whithin limit
+#
+if (&Tmp-Integer-2 < 0 || &Tmp-Integer-2 > 123) {
+	test_fail
+}
+
+success
```
arr2036: Yay tests!

PR #2240, opened 2018-05-31
Convert randstr xlat to new API

pwdng:


src/main/xlat_func.c
```diff
@@ -658,20 +658,58 @@ static ssize_t rand_xlat(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
...
+		RPEDEBUG("Failed concatenating input");
+		return XLAT_ACTION_FAIL;
+	}
+
+	p = (*in)->vb_strvalue;
+
+	/*
+	 * Calculate size of output
+	 */
+	while (*p) {
```
arr2036: You can't assume the input string is binary safe anymore (though the original code did).

General pattern is declare `char const *end`

```
end = p + (*in)->vb_length;

while (p < end) {
<stuff>
}
```

src/tests/keywords/randstr
```diff
@@ -0,0 +1,41 @@
+
+update request {
+        &Tmp-String-0           := "%{randstr:%{Tmp-String-0}}"
+        &Tmp-String-1           := "%{randstr:nnn}"
+        &Tmp-String-2           := "%{randstr:24a}"
+        &Tmp-String-3           := "%{randstr:123aa}"
```
arr2036: Be good to have a repetition test for the double digit outputs like hex

src/main/xlat_func.c
pwdng: The fact we pass FR_TYPE_STRING when calling fr_value_box_list_concat is misleading then. I was expecting the resulting valuebox to contain a NULL terminated string.

arr2036: The string is actually '\0' terminated for safety, but you should use the length anyway so we can error out correctly if there's a '\0' somewhere in the format string... Which yes I know sounds pretty stupid, but there are all kinds of fun attacks you can do with functions which don't deal with embedded '\0s' correctly.  I remember about 10 years ago a subset of PHP regex functions were not binary safe, so if you didn't want the remainder of your input string to be validated against a regex, you just stuck a '\0' in there.

PR #2241, opened 2018-06-01
Convert urlquote and urlenquote xlat

pwdng:


src/main/xlat_func.c
```diff
@@ -841,20 +841,57 @@ static xlat_action_t xlat_randstr(TALLOC_CTX *ctx, fr_cursor_t *out,
...
+		    *p == '.' ||
+		    *p == '~') {
+			outlen++;
+		} else {
+			outlen += 3;
+		}
+		p++;
+	}
+
+	buff = buff_p = talloc_zero_size(NULL, sizeof(char) * (outlen + 1));
```
arr2036: you should really use talloc_array here specifying a char type (likewise with the other xlats), just so that the type of the buffer is set to char.

src/main/xlat_func.c
```diff
@@ -925,38 +925,76 @@ static xlat_action_t xlat_urlquote(TALLOC_CTX *ctx, fr_cursor_t *out,
...
+	while (p < end) {
+		if (*p == '%') {
+			p += 3;
+		} else {
+			p++;
+		}
+		outlen++;
+	}
+
+	buff = buff_p = talloc_zero_array(NULL, char, outlen + 1);
```
arr2036: If it's going to be overwritten straight after allocation it's better just to use talloc_array.

src/main/xlat_func.c
```diff
@@ -925,38 +925,76 @@ static xlat_action_t xlat_urlquote(TALLOC_CTX *ctx, fr_cursor_t *out,
...
+			talloc_free(buff);
+
+			return XLAT_ACTION_FAIL;
 		}
 		p++;
-		*out_p++ = ((c1 - hextab) << 4) + (c2 - hextab);
+		*buff_p++ = ((c1 - hextab) << 4) + (c2 - hextab);
 	}
 
-	*out_p = '\0';
```
arr2036: and still explicitly '\0' terminate.

PR #2242, opened 2018-06-06
Fix to fr_pair_list_afrom_file - file done

nchaigne:
After fr_cursor_merge, local vp is NULL, so caller gets "*pfiledone = true" and will not read more than one value pair list.

Here is a fix.


arr2036:
This was already fixed

nchaigne:
It wasn't when I started... you're faster than I am :)
Thanks!

arr2036:
np :)

PR #2257, opened 2018-07-03
Allow for sql acct session to be created in post-auth section

pwdng:
Currently, it's possible to start multiple sessions in parallel  and have them all successfully authenticate, even with simultaneous session check in place.
That's because the check looks for an existing accounting session that only gets created with the first accounting packet. Until it's been received and processed, other auth requests will be accepted.
As a workaround, we can choose to create the acct session in the post-auth section.


alandekok:
I've pulled in the patch and re-worded some of the docs.  Please check for sanity. :)  If it's OK, just close this PR.

The simple solution to large blocks of commented text is to use `-query = ...`.  Since `rlm_sql` only looks for `query = ...`, the `-query` lines will be ignored.

This config sort of matches what we already do with `-ldap` in the virtual servers.  So it's at least not surprising.

PR #2259, opened 2018-07-05
Adding new Nomadix VSA’s and use official names

cnelissen:
Just adding some new VSA's and renaming one based on official documentation: http://business.nomadix.com/aghelp/common/ag_userguide/QuickReference/Nomadix_Vendor-Specific_RADIUS_Attributes.htm

### DIFF
--- a/share/dictionary.nomadix
+++ b/share/dictionary.nomadix
@@ -21,6 +21,7 @@ ATTRIBUTE	Nomadix-Subnet				6	string
 ATTRIBUTE	Nomadix-MaxBytesUp			7	integer
 ATTRIBUTE	Nomadix-MaxBytesDown			8	integer
 ATTRIBUTE	Nomadix-EndofSession			9	integer
+ATTRIBUTE	Nomadix-Session-Terminate-End-Of-Day		9	integer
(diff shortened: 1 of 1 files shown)


cnelissen:
I see that you are requiring commits to be signed. I added a GPG key after I created this PR, but I think I probably need to re-do the request for the commit to be signed... Unless you can do something from your end.

arr2036:
Force push into your PR branch

alandekok:
The existing names shouldn’t be deleted. That would break people’s configurations, which is bad. 

  Just add the new name, and leave the old names there.

cnelissen:
@alandekok So you are saying to leave duplicate attributes for the same index? I.e.

```
ATTRIBUTE	Nomadix-EndofSession			        9	integer
ATTRIBUTE	Nomadix-Session-Terminate-End-Of-Day		9	integer
```

alandekok:
Yes.  It's OK to have duplicate attributes like that.

arr2036:
You still need to sign your original commit.  If you ``git rebase -i HEAD~2`` and squash your second commit, then ``git push --force`` to this branch, it should fix things.

cnelissen:
@arr2036 Thank you sir. Done and done.

PR #2262, opened 2018-07-13
Removed expiry_time IS NULL from allocate_update query

michael-neumann:
The allocate_find query fetches rows with expiry_time < NOW() OR expiry_time IS NULL or expiry_time = 0

If we fetch a row where the expiry_time is not NULL this ip was not marked as used and the next allocate_find query is fatching the same ip again .... the result was duplicated ips


alandekok:
I also fixed the sqlite query.  And ported it to the master branch.

PR #2280, opened 2018-08-17
Added support for IPv6 RADIUS accounting in SQL - 3.0.x

mjducharme:
Added new fields and queries in SQL to store Framed-IPv6-Address, Framed-IPv6-Prefix, Framed-Interface-Id and Delegated-IPv6-Prefix


alandekok:
That looks good, and consistently formatted, too.  Very nice!

PR #2281, opened 2018-08-19
Two small dictionary fixes

lemenkov:
I'm trying to improve one particular small dictionary "compiler", which is very fragile and very format-demanding right now. On a good side it allows me to find some issues within dictionaries. Here are two small fixes for issues I've just found.

Btw do we have any formal specification for dictionaries?


arr2036:
There's a basic formatting script written in perl, but unfortunately no formal specification.

alandekok:
The changes look good to me.  We should also pull them over to v3.0.x.

Issue #2282, opened 2018-08-20
sqlippool Pool-Name not defined


maximumG:
# Issue type
- sqlippool module requires that a Pool-Name is defined
- Feature request.

# Feature description

I have a use case for the sqlippool module where we defined our own Pool-Name attribute (specific dictinary) with the following meaning:

- Pool-Name is used for IP reservation when a request is a pure RADIUS one (PPP user) which sends back a Framed-IP-Address attribute.
- DHCPv4-Pool-Name is used for IP reservation when a request is a pure DHCPv4 message (DHCPv4 user) which sends back a DHCP-Your-IP-Address attribute.
- DHCPv6-Pool-Name-Pd ise used for IPv6 delegated prefix when a request is a pure RADIUS one which sends a IPv6-Delegated-prefix attribute.

For one of our group in radgroupcheck we only have DHCPv4-Pool-Name & DHCPv6-Pool-Name-Pd attributes set but NOT Pool-Name. When the module is running it is not possible to allocate any IP address as the Pool-Name attribute is 'hard coded'.

https://github.com/FreeRADIUS/freeradius-server/blob/a9c75f03dfda610453b92753d1e5b225965373cf/src/modules/rlm_sqlippool/rlm_sqlippool.c#L487

https://github.com/FreeRADIUS/freeradius-server/blob/a9c75f03dfda610453b92753d1e5b225965373cf/src/modules/rlm_sqlippool/rlm_sqlippool.c#L189

My suggestion would be to have this specific check based on the inst->pool_name variable instead of a hardcoded pool-name. By doig this way we would be able to defined our own Pool-Name attribute inside the module configuration.

```
if (fr_pair_find_by_da(request->control, inst->pool_name, TAG_ANY) == NULL) {
		RDEBUG("No %s defined",  inst->pool_nam);

		return do_logging(inst, request, inst->log_nopool, RLM_MODULE_NOOP);
}
```

I can work on this modification and send a pull request if you want.

alandekok:
I think that makes sense.

maximumG:
@alandekok : Thanks for the reply. I will work on this ASAP. Do you need this pull request on both master (v4) and v3.0.x branches ?

alandekok:
I think v3 already has a pool name in it's configuration.  So only v4 would be best.

maximumG:
[...]

PR #2286, opened 2018-08-24
Add support for dynamic Pool-Name attribute

maximumG:
Regarding issue #2282, I added the support to specify a custom Pool Name instead of the traditional *Pool-Name*.

the *pool_name* attribute in the sqlippool file (default: 'Pool-Name') can now be used to specify any other specific attribute to be used as Pool Name.


maximumG:
If you need some explanation regarding the modification I made, please let me know.

alandekok:
We're reviewing it and will get back to you. It just takes time.

PR #2288, opened 2018-08-24
Better "un-xlating" of value pairs where xlat is not supported

nchaigne:

For example, we will have:

### DIFF
--- a/src/lib/util/value.c
+++ b/src/lib/util/value.c
@@ -37,8 +37,8 @@
  *
  * - PRESENTATION format is what we print to the screen, and what we get from the user, databases
  *   and configuration files.
- *   - #fr_value_box_asprint is used to convert INTERNAL format PRESENTATION format.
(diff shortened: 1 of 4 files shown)


nchaigne:
Dunno what the conflicts are... the changes are rather simple :/

arr2036:
maybe your branch head isn't up to the same point as the master branch? The interface won't let me merge with conflicts, so if you could fix it up on your side that'd be appreciated.

nchaigne:
I just merged right before creating the branch, it was one commit behind Alan's latest.
GitHub doesn't tell me what's wrong, only "the conflicts are too complex to resolve in the web editor".

Don't know what to do besides dropping the PR and making a new one :/

nchaigne:
Can you please cancel / close this PR ? really sorry for the mess.

nchaigne:
Just noticed I could close these myself. :)

PR #2289, opened 2018-08-24
Better "un-xlating" of value pairs where xlat is not supported (2)

nchaigne:

For example, we will have:


nchaigne:
Trying again.

No idea why all those commits show up... I've had these issues since 4.0.x was renamed to master.

Any idea how to fix this ?

nchaigne:
Can you please cancel / close this PR ? really sorry for the mess.

PR #2290, opened 2018-08-24
Better "un-xlating" of value pairs where xlat is not supported

nchaigne:

For example, we will have:


nchaigne:
A hard reset on master solved my issues. :)
Sorry for the two previous pull requests.

PR #2294, opened 2018-08-30
fix typo about ifdef for TLS 1.3 Version

alagoutte:


alagoutte:
There is the same issue on 3.0.x branch

PR #2305, opened 2018-09-17
Typo on doc reference links

ratan12rk:


alandekok:
the fix was already in my local tree.  I've pushed it back, thanks.

PR #2316, opened 2018-09-28
HMAC -> OpenSSL

cipherboy:
I was debating making a mailing list post but decided that opening a PR was just as easy.

For compliance reasons, I'm interested in replacing the in-tree implementation of HMAC with the version used in OpenSSL. I need to look at the implementations of MD5 and SHA1 to see if I need to switch those as well.

I'm not quite sure the historical significance of these functions, so perhaps there is some reason you'd like to keep them as is. Also, while I'm comfortable replacing the bodies, I think keeping the functions proper should preserve ABI/API compatibility, which shouldn't be broken in the `v3.0.x` branch. I'm happy to target the `master` branch and then backport changes, but I think the one targeting master should be more invasive and remove them completely (as iirc, 4.x.x isn't released yet?).

This PR comes in two commits:
 - One to replace `fr_hmac_md5`'s implementation with OpenSSL calls
 - One to replace `fr_hmac_sha1`'s implementation with OpenSSL calls

Feedback requested! Marking WIP as I'd like to see if CI passes (tests pass locally). :)

### DIFF
--- a/src/lib/hmacmd5.c
+++ b/src/lib/hmacmd5.c
@@ -27,10 +27,41 @@
 
 RCSID("$Id$")
 
+#ifdef HAVE_OPENSSL_EVP_H
+#include <openssl/hmac.h>
+#include <openssl/evp.h>
+#endif
+
 #include <freeradius-devel/libradius.h>
 #include <freeradius-devel/md5.h>
 
-/** Calculate HMAC using MD5
+#ifdef HAVE_OPENSSL_EVP_H
+/** Calculate HMAC using OpenSSL's MD5 implementation
+ *
+ * @param digest Caller digest to be filled in.
+ * @param text Pointer to data stream.
+ * @param text_len length of data stream.
+ * @param key Pointer to authentication key.
+ * @param key_len Length of authentication key.
+ *
+ */
+void fr_hmac_md5(uint8_t digest[MD5_DIGEST_LENGTH], uint8_t const *text, size_t text_len,
+		 uint8_t const *key, size_t key_len)
+{
+	HMAC_CTX *ctx  = HMAC_CTX_new();
+
+#ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
+	/* Since MD5 is not allowed by FIPS, explicitly allow it. */
+	HMAC_CTX_set_flags(ctx, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW);
+#endif /* EVP_MD_CTX_FLAG_NON_FIPS_ALLOW */
+
+	HMAC_Init_ex(ctx, key, key_len, EVP_md5(), NULL);
+	HMAC_Update(ctx, text, text_len);
+	HMAC_Final(ctx, digest, NULL);
+	HMAC_CTX_free(ctx);
+}
+#else
+/** Calculate HMAC using internal MD5 implementation
  *
  * @param digest Caller digest to be filled in.
  * @param text Pointer to data stream.
@@ -101,6 +132,7 @@ void fr_hmac_md5(uint8_t digest[MD5_DIGEST_LENGTH], uint8_t const *text, size_t 
 					      * hash */
 	fr_md5_final(digest, &context);	  /* finish up 2nd pass */
 }
+#endif /* HAVE_OPENSSL_EVP_H */
 
 /*
 Test Vectors (Trailing '\0' of a character string not included in test):
--- a/src/lib/hmacsha1.c
+++ b/src/lib/hmacsha1.c
@@ -10,13 +10,19 @@
 
 RCSID("$Id$")
 
+#ifdef HAVE_OPENSSL_EVP_H
+#include <openssl/hmac.h>
+#include <openssl/evp.h>
+#endif
+
 #include <freeradius-devel/libradius.h>
 
 #ifdef HMAC_SHA1_DATA_PROBLEMS
 unsigned int sha1_data_problems = 0;
 #endif
 
-/** Calculate HMAC using SHA1
+#ifdef HAVE_OPENSSL_EVP_H
+/** Calculate HMAC using OpenSSL's SHA1 implementation
  *
  * @param digest Caller digest to be filled in.
  * @param text Pointer to data stream.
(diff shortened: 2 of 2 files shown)


src/lib/hmacsha1.c
```diff
@@ -10,12 +10,10 @@
 
 RCSID("$Id$")
 
+#include <openssl/hmac.h>
+#include <openssl/evp.h>
 #include <freeradius-devel/libradius.h>
```
cipherboy: I notice this lacks a `#include <freeradius-devel/sha1.h>` -- would you like me to include it? I think it would be good to include.

src/lib/hmacsha1.c
```diff
@@ -28,119 +26,15 @@ unsigned int sha1_data_problems = 0;
...
-				printf("_");
-				j=0;
-			}
-			j++;
-
-			printf("%02x", digest[i]);
-		}
-		printf("\n");
-	}
+#ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
```
cipherboy: I believe SHA-1 to be allowed by FIPS 140-2 so perhaps I can drop these lines (I mirrored the MD5 code closely), but I'll take an AI to investigate this.

arr2036:
You did see this... right?

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/include/md5.h#L71

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/include/sha1.h#L34

arr2036:
Assuming that we still want a functional server without OpenSSL, we can't just replace them (which we do).  We could switch to using them if OpenSSL is available, as we do currently with the MD4 and MD5  hashing functions.

SHA1 is special because we need to get at one of the intermediary states... I forget which protocol that's for I'm sure @alandekok knows.

arr2036:
That said... what's your reason for wanting to switch them to OpenSSL? IIRC the EVP functions need an EVP_CTX malloced on each run, the current HMAC functions require no heap allocations.

cipherboy:
@arr2036 said:
> Assuming that we still want a functional server without OpenSSL, we can't just replace them (which we do).

Ahh is that why the first two builds are failing on Travis? I'm happy to make them conditional then. 

> SHA1 is special because we need to get at one of the intermediary states...

Nifty--I'd appreciate a reference on this to see what I can do. :) I'll admit I haven't read (most) of the RADIUS RFCs.

arr2036:
> @arr2036 said:
> 
> > Assuming that we still want a functional server without OpenSSL, we can't just replace them (which we do).
> 
> Ahh is that why the first two builds are failing on Travis? I'm happy to make them conditional then.
> 

The two builds that build without OpenSSL to test our ability to build without OpenSSL, yes :)

> > SHA1 is special because we need to get at one of the intermediary states...
> 
> Nifty--I'd appreciate a reference on this to see what I can do. :) I'll admit I haven't read (most) of the RADIUS RFCs.

I'm not sure it was even RADIUS... Oh jees, it's the EAP-SIM fips186prf.c code.

Ok, if there's a way of using the EVP interface for that code, and you can write a patch, then we could switch SHA1 hashing to OpenSSL too.  There's some a test vector as well as a cutest harness at the bottom of the file in the master branch you could use to double check the PRF outputs.

I'd do this work against master initially and then backport it.

arr2036:
Currently, although it looks like the SHA1 functions get switched out for EVP functions too, I don't think ``WITH_OPENSSL_SHA1`` is ever defined anywhere.

arr2036:
If you can't figure out the PRF code I can take a look tomorrow.  I'm familiar with the EAP-SIM, EAP-AKA, EAP-AKA' PRFs as I wrote implementations of most of them a year or so ago... The fips one was left over from the original implementation of EAP-SIM.

cipherboy:
@arr2036 said:
> That said... what's your reason for wanting to switch them to OpenSSL? IIRC the EVP functions need an EVP_CTX malloced on each run, the current HMAC functions require no heap allocations.

Compliance and security. I wish I could give a better answer than that, but we'd like to (and have) removed as much "custom"/hand written crypto as we could and replaced it with OpenSSL. Fedora is building this with OpenSSL, so we can carry it there if need be. OpenSSL should be faster on nearly every platform due to the optimized assembly.

I'll take a look at those changes tonight, else I'll get to it on Monday. :)

cipherboy:
@arr2036 I had a chat with our crypto team. We've been granted an exception to keep using the PRF+SHA1 implementations, so this PR is fine for compliance. OpenSSL doesn't expose the internal hash state so there's no way currently to replace the functionality without relying on unstable/potentially changing APIs.

This PR is ready for review then. :)

src/lib/hmacmd5.c
```diff
@@ -27,6 +27,11 @@
 
 RCSID("$Id$")
 
+#ifdef HAVE_OPENSSL_EVP_H
```
cipherboy: I guarded around `HAVE_OPENSSL_EVP_H`; would you prefer I guard around something else?

arr2036:
Guard macro is fine.  Small style point, could you do these as two separate functions, it just makes the code easier to read.

cipherboy:
Sure, I'll rework to do that. Thanks!

src/lib/hmacsha1.c
cipherboy: To answer the question, `./src/include/libradius.h` includes `sha1.h` so it'd be technically extraneous to do so. Actually, it includes `md4.h` and `sha1.h` but not `md5.h`. Go figure.

cipherboy:
Okay, this one looks good to go. The one against master (#2317) seems to be having issues though.

arr2036:
You need to add a call to ``HMAC_CTX_Free`` at the end else we'd leak contexts.

cipherboy:
@arr2036 Updated. Could you kick the Travis gcc build that failed? I'm 90% sure I didn't break `apt-get install` :)

alandekok:
Travis has random networking failures.

cipherboy:
@alandekok Thanks! 

In one project, I became convinced Travis also has disk problems (and/or they're using network storage when building inside a container)...  but they've gotten better lately I think.

PR #2317, opened 2018-10-01
HMAC -> OpenSSL against master

cipherboy:
Mirror of #2316 against master branch.

See #2316 for more information.

### DIFF
--- a/src/lib/util/hmacmd5.c
+++ b/src/lib/util/hmacmd5.c
@@ -30,9 +30,40 @@
  */
 RCSID("$Id$")
 
+#ifdef HAVE_OPENSSL_EVP_H
+#include <freeradius-devel/tls/base.h>
+#endif
+
 #include <freeradius-devel/util/md5.h>
 
-/** Calculate HMAC using MD5
+#ifdef HAVE_OPENSSL_EVP_H
+/** Calculate HMAC using OpenSSL's MD5 implementation
+ *
+ * @param digest Caller digest to be filled in.
+ * @param text Pointer to data stream.
+ * @param text_len length of data stream.
+ * @param key Pointer to authentication key.
+ * @param key_len Length of authentication key.
+ *
+ */
+void fr_hmac_md5(uint8_t digest[MD5_DIGEST_LENGTH], uint8_t const *text, size_t text_len,
+		 uint8_t const *key, size_t key_len)
+{
+	HMAC_CTX *ctx  = HMAC_CTX_new();
+
+#ifdef EVP_MD_CTX_FLAG_NON_FIPS_ALLOW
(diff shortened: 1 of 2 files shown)


alandekok:
It's awkward to have layers of functions, e.g. `fr_hmac_md5_internal`.  It would be simple to just build one function or the other.

e.g.

```
#ifdef HAVE_OPENSSL_EVP_H
fr_hmac_md5(...)
 ... using OpenSSL
#else
fr_hmac_md5(...)
 ... not using OpenSSL
```

cipherboy:
@alandekok Okay, I'll switch to that. Apologies I don't usually CI spam but, `master` doesn't build on Fedora as it lacks `libkqueue`. :/

alandekok:
> @alandekok Okay, I'll switch to that. Apologies I don't usually CI spam but, `master` doesn't build on Fedora as it lacks `libkqueue`. :/

Hmm... we do need libkqueue to work.  Moving to another even library would be enormous amounts of work, unfortunately.  The kqueue API is much simpler (and saner) than most other eventing APIs.

cipherboy:
Regarding `libkqueue`, it looks like there were two main packaging attempts:

https://bugzilla.redhat.com/show_bug.cgi?id=684446
https://bugzilla.redhat.com/show_bug.cgi?id=889505

At this point you're probably too heavily invested in `libkqueue` to switch, but [libverto](https://github.com/latchset/libverto) is what [gssproxy](https://pagure.io/gssproxy/blob/master/f/src/gssproxy.c#_284) and [krb5](https://github.com/cipherboy/krb5/blob/master/src/kdc/main.c#L1050) use, and all of these are packaged in RHEL-like distros.

---

Out of curiosity, what is your time scale looking like for releasing v4.x.x? Just wondering if it is measured in months or over a year. 

--- 

This PR and the other are both updated. They should be good to go, but I'll watch CI and make sure it passes.

alandekok:
libverto seems a little odd.  It's a wrapper around wrapper event libraries?  i.e. it doesn't do anything itself, but it abstracts libev, livevent, etc.  That's not something I'm comfortable with.  The extra layers of abstraction cause layers of bugs, make debugging more difficult, etc.

As for v4, "when it's ready".  We don't have funding to work in it exclusively.  So the work is done whenever we can find time.  We hope to have an alpha release early next year.

cipherboy:
Weird, it looks like you're building against OpenSSL 1.0-series versus 1.1-series on master, but 1.1-series on `v3.0.x`.

Thoughts @alandekok? Should I support both or switch to 1.1.x only?

alandekok:
We should support the commonly used OpenSSL versions.

arr2036:
I'll fix it up locally.  There's an ongoing effort to not use monolithic header files so that this doesn't happen.

PR #2322, opened 2018-10-04
Change Extended-Location-Policy-Rules to string

herwinw:
The description in the RFC implies the text is printable and not a
binary value:

   The Ruleset Reference field of this attribute is of variable length.
   It contains a URI that indicates where the richer ruleset can be
   found.  This URI SHOULD use the HTTPS URI scheme.  As a deviation
   from [RFC4119], this field only contains a reference and does not
   carry an attached, extended ruleset.  This modification is motivated
   by the size limitations imposed by RADIUS.


alandekok:
I've pushed fixes to both v3.0.x and master.  Travis is having problems, so this is the fasted way to do it.

PR #2333, opened 2018-10-18
Clarify shebangs to be python2

cipherboy:
To align with [Fedora packaging standards](https://fedoraproject.org/wiki/Changes/Make_ambiguous_python_shebangs_error), replace all occurrences of `#!/usr/bin/env python` with `#!/usr/bin/env python2` as that is the correct interpreter.

The only file which could arguably be Python3-compatible is `src/modules/rlm_python/radiusd.py`, but I kept `python2` for consistency. 

Note that Python 2 support will be dropped [in the near future](https://fedoraproject.org/wiki/FinalizingFedoraSwitchtoPython3), with "near future" looking like 30 removing most all `python2` packages, 32 making `/usr/bin/python` point to `python3` and fully removing `python2` by 33. 

(29 will be shipped soon, 30 is the current `rawhide` release). 

### DIFF
--- a/scripts/radtee
+++ b/scripts/radtee
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
 from __future__ import with_statement 
 
 # RADIUS comparison tee v1.0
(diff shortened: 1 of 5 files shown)


cipherboy:
(enabled signed commits again)

cipherboy:
Thanks @arr2036! The v3.0.17 release has been pushed to [rawhide](https://koji.fedoraproject.org/koji/buildinfo?buildID=1154471). :)

PR #2334, opened 2018-10-19
Implement dict-based input/output for rlm_python

alejandro-perez:
Current rlm_python interface only allows writing to the "reply" and "config" VPS lists. While this is enough for most use cases, there are some scenarios where you might also want to be able to modify the request (for instance, if your module modifies the "User-Password" or User-Name" attribute). The Perl module allows modifying all the VPS lists so I thought this should be allowed also from Python.

This PR does not modify current behaviour. If a module provides a 5-tuple, the PR will take care of those extra 2 tuple items and process them as "request" and "state". But if an existing module is using the 3-tuple, it will see no difference.

The code updating the cached copies for username and password has been borrowed from the Perl version.

### DIFF
--- a/raddb/mods-available/python
+++ b/raddb/mods-available/python
@@ -17,6 +17,10 @@ python {
 
 	module = example
 
+	# Pass all VPS lists as a 6-tuple to the callback
+	# (request, reply, config, state, proxy_req, proxy_reply)
+	# pass_all_vps = no
+
 	mod_instantiate = ${.module}
 #	func_instantiate = instantiate
 
--- a/raddb/mods-available/python
+++ b/raddb/mods-available/python
@@ -21,6 +21,12 @@ python {
 	# (request, reply, config, state, proxy_req, proxy_reply)
 	# pass_all_vps = no
 
+	# Pass all VPS lists as a dictionary to the callback
+	# Keys: "request", "reply", "config", "session-state", "proxy-request",
+	# 	    "proxy-reply"
+	# This option prevales over "pass_all_vps"
+	# pass_all_vps_dict = no
+
 	mod_instantiate = ${.module}
 #	func_instantiate = instantiate
 
--- a/raddb/mods-available/python
+++ b/raddb/mods-available/python
@@ -17,13 +17,13 @@ python {
 
 	module = example
 
-	# Pass all VPS lists as a 6-tuple to the callback
+	# Pass all VPS lists as a 6-tuple to the callbacks
 	# (request, reply, config, state, proxy_req, proxy_reply)
 	# pass_all_vps = no
 
-	# Pass all VPS lists as a dictionary to the callback
+	# Pass all VPS lists as a dictionary to the callbacks
 	# Keys: "request", "reply", "config", "session-state", "proxy-request",
-	# 	    "proxy-reply"
+	#       "proxy-reply"
 	# This option prevales over "pass_all_vps"
 	# pass_all_vps_dict = no
 
--- a/src/modules/rlm_python/example.py
+++ b/src/modules/rlm_python/example.py
@@ -7,26 +7,30 @@
 
 import radiusd
 
+# Check post_auth for the most complete example using different
+# input and output formats
+
 def instantiate(p):
   print "*** instantiate ***"
   print p
   # return 0 for success or -1 for failure
 
+
 def authorize(p):
   print "*** authorize ***"
-  print
   radiusd.radlog(radiusd.L_INFO, '*** radlog call in authorize ***')
   print
   print p
   print
   print radiusd.config
-  print
   return radiusd.RLM_MODULE_OK
+
 
 def preacct(p):
   print "*** preacct ***"
   print p
   return radiusd.RLM_MODULE_OK
+
 
 def accounting(p):
   print "*** accounting ***"
@@ -35,25 +39,53 @@ def accounting(p):
   print p
   return radiusd.RLM_MODULE_OK
 
+
 def pre_proxy(p):
   print "*** pre_proxy ***"
   print p
   return radiusd.RLM_MODULE_OK
+
 
 def post_proxy(p):
   print "*** post_proxy ***"
   print p
   return radiusd.RLM_MODULE_OK
 
+
 def post_auth(p):
   print "*** post_auth ***"
-  print p
-  return radiusd.RLM_MODULE_OK
+
+  # This is true when using pass_all_vps_dict
+  if type(p) is dict:
+    print "Request:", p["request"]
+    print "Reply:", p["reply"]
+    print "Config:", p["config"]
+    print "State:", p["session-state"]
+    print "Proxy-Request:", p["proxy-request"]
+    print "Proxy-Reply:", p["proxy-reply"]
+
+  else:
+    print p
+
+  # Dictionary representing changes we want to make to the different VPS
+  update_dict = {
+        "request": (("User-Password", ":=", "A new password"),),
+        "reply": (("Reply-Message", "The module is doing its job"),
+                  ("User-Name", "NewUserName")),
+        "config": (("Cleartext-Password", "A new password"),),
+  }
+
+  return radiusd.RLM_MODULE_OK, update_dict
+  # Alternatively, you could use the legacy 3-tuple output
+  # (only reply and config can be updated)
+  # return radiusd.RLM_MODULE_OK, update_dict["reply"], update_dict["config"]
+
 
 def recv_coa(p):
   print "*** recv_coa ***"
   print p
   return radiusd.RLM_MODULE_OK
+
 
 def send_coa(p):
   print "*** send_coa ***"
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -550,7 +550,7 @@ static rlm_rcode_t do_python_single(REQUEST *request, PyObject *pFunc, char cons
 		int tuple_size = PyTuple_GET_SIZE(pRet);
 
 		if (tuple_size < 2 || tuple_size > 3) {
-			ERROR("%s - Tuple must be (return, replyTuple, configTuple) or (return, updateDict)", funcname);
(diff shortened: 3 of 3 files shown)


herwinw:
```c
if (tuple_size < 3 || tuple_size > 7) {
```

Some nitpicking here: The maximum value could be 5 instead of 7, in case `WITH_PROXY` is not defined. It could be fixed by something like:

```c
#ifdef WITH_PROXY
int max_tuple_size = 7;
#else
int max_tuple_size = 5;
#end
if (tuple_size < 3 || tuple_size > max_tuple_size) {
```

(And the error string would need something comparable)

herwinw:
And it's too bad that the input of the methods uses the ordering request, reply, config, while the output uses reply, config, request, but we'll have to break backwards compatibility to fix that.

alejandro-perez:
> And it's too bad that the input of the methods uses the ordering request, reply, config, while the output uses reply, config, request, but we'll have to break backwards compatibility to fix that.

Yes. TBH, if I implemented `pass_all_vps` again, I would use a dictionary. The same goes for the output.

alejandro-perez:
> Some nitpicking here: The maximum value could be 5 instead of 7, in case `WITH_PROXY` is not defined. It could be fixed by something like:

True, let me fix it

herwinw:
> Yes. TBH, if I implemented `pass_all_vps` again, I would use a dictionary. The same goes for the output.

Might still be an option for the output, the C code can switch codepaths if the value is a dictionary or a list/tuple/array.

alejandro-perez:
> Might still be an option for the output, the C code can switch codepaths if the value is a dictionary or a list/tuple/array.

Yes, but there is already a comment indicating `xxx This code is messy!`. Doing just another codepath seemed to make things even more complicated (we already have 3 codepaths!)

alejandro-perez:
> Yes, but there is already a comment indicating `xxx This code is messy!`. Doing just another codepath seemed to make things even more complicated (we already have 3 codepaths!)

If we decided to go down this road, would you think including the return code into the dictionary would be sensible, rather than having (return, vpsDict)? I think so, since it would make checks simpler (tuple vs dict) rather than having to check the type of the second element of the return tuple.

herwinw:
> If we decided to go down this road, would you think including the return code into the dictionary would be sensible, rather than having (return, vpsDict)? I think so, since it would make checks simpler (tuple vs dict) rather than having to check the type of the second element of the return tuple.

Adding the status to the dict would result in structures like this:

```python
{
  'status': radiusd.RLM_MODULE_UPDATED,
  'request': (('Tmp-String-0', 'foo'),),
  'reply': (('Tmp-String-0', 'bar'),)
}
```

This feels to me like the status code is at the same level as the request and reply lists, which feels wrong. Returning a tuple `(status-code, update-dict)` would be my preferred choice.
This can still be uniquely identified: a 2-tuple is interpreted as `(status-code, update-dict)`, a 3-tuple is interpreted as `(status-code, reply, config)`, other return-values remain as they are now.

alejandro-perez:
> This feels to me like the status code is at the same level as the request and reply lists, which feels wrong. Returning a tuple `(status-code, update-dict)` would be my preferred choice.
> This can still be uniquely identified: a 2-tuple is interpreted as `(status-code, update-dict)`, a 3-tuple is interpreted as `(status-code, reply, config)`, other return-values remain as they are now.

Fair enough. So, you more inclined to have the dict-based output interface, right? I have no problem rewriting my PR for that. Should we wait for Alan's opinion here?

herwinw:
Yes, and Yes. And maybe wait for a reaction of Arran as well, he's been more active in the development of the modules that embed languages.

arr2036:
Just use a Python method to get and set the status code.  It's not an attribute.  The existing mechanism of marshalling and unmarshalling was extremely ill conceived.  It's very inefficient and doesn't scale.  rlm_lua in v4.0.x does it right, using a metatable to intercept gets and sets, if you could so something like that for the extra lists it would be much better, and something we could use going forward.

alejandro-perez:
I'll take a look to rlm_lua and see whether I can use similar approach. not so convinced that if it implies such a big change it would be backwards compatible. I'll analyse it and will come back with my impression.

alejandro-perez:
I've been looking at rlm_lua but I'm not sure I fully understand what you meant :(. I can implement the approach of having a 2-tuple `(status-code, update-dict)` if that's still wanted.

alejandro-perez:
Indeed, I have rebased so now the proposal is dict based. Please, let me know what you think about it.

alejandro-perez:
This is how a working python module would look like:
```
  updateDict = {
        "reply": (("User-Name", "VALUE"), ("User-Password", "Another")),
        "config": (("Cleartext-Password", "NONO"),),
  }
  return radiusd.RLM_MODULE_OK, updateDict
```
I think that's definitely clearer than returning a 3-tuple of N-tuples of 2-tuples

alejandro-perez:
I forgot to show server log output for it:
```
*** post_auth ***
(('User-Name', 'alex'), ('User-Password', 'alex'), ('NAS-IP-Address', '172.17.0.2'), ('NAS-Port', '2222'), ('Message-Authenticator', '0xc7f832c1594eae884f5353e639a0254d'), ('Event-Timestamp', 'Oct 31 2018 11:21:47 UTC'))
post_auth - 'reply:User-Name' = 'VALUE'
post_auth - 'reply:User-Password' = 'Another'
post_auth - 'config:Cleartext-Password' = 'NONO'
```

alandekok:
It looks good.

herwinw:
It looks good, but that doesn't stop me from nitpicking ;)

* I would switch the blocks of `if (tuple_size == 3) {` and `else if (tuple_size == 2) {`, I think it's more logical to put the lowest one first, and IMHO the 2-tuple should be the preferred way to work with this module (any other opinions on this?)
* Should we check the dict for other unused keys? Using `dict['relpy']` now break implicitely. Adding something to the proxy keys on an install without proxy support simply skips them. A simple log warning might help, an error results in users not being able to add additional information in the dict (do we want to support that?)
* There is no documentation to show the new return format. Using it in one of the example Python files might be nice. Once again: if we want to use the dict as the recommended way of working, all example files should use this syntax and the old return type can be added as a legacy supported item.

alejandro-perez:
I've added two commits. One is for documenting the "pass_all_vps" which was missing. The other one adds a new option "pass_all_vps_dict" which effectively makes that the module calls the callback passing a dict rather than a 6-tuple.

herwinw:
> An extra note: Even though I implemented pass_all_vps not that long ago, I think that under the light of what we've discussed for this PR, an additional option called dict_input (or vps_as_dict, or something similar) might be useful to get p as a dict[] rather than a 6-tuple.

That sounds like another nice improvement as well.

Once this change is merged, I'll try to have another look at `rlm_mruby` to implement this new interface as well.

alejandro-perez:
Ok, so implemented, rebased and tided up (squased a couple of small commits into larger-meaningful ones)

alejandro-perez:
Let me know whether anything else is required to have this PR accepted/reviewed.

alejandro-perez:
All commits signed :+1:

alejandro-perez:
I'm sorry to insist. Let me know if there is anything I can do to help speeding up the integration of this PR.

alandekok:
I'm OK with it.  I think it should be fine.

arr2036:
This whole method of bridging the data domains is pretty brain-dead, I do not want to see it extended to mruby.  Say you have a 20 AVP list, to manipulate one value in that list, the entire list has to be converted into python tuples, and the original pairs deleted.  Then, the one value is changed in the list, and after the python module finishes, all the python tuples need to be converted back to AVPs again.

arr2036:
If there's a real requirement for this in v3.0.x, then merge it, and we'll come up with an entirely new interface in v4.0.x

alejandro-perez:
> This whole method of bridging the data domains is pretty brain-dead, I do not want to see it extended to mruby. Say you have a 20 AVP list, to manipulate one value in that list, the entire list has to be converted into python tuples, and the original pairs deleted. Then, the one value is changed in the list, and after the python module finishes, all the python tuples need to be converted back to AVPs again.

Right, but that was the code that was on the module already. My extension just gives more information to the module and provides a better-organised interface (dicts rather than tuples).
I think I get now what you mentioned in a previous comment. I think I might be able to create some Python Class called `VPList` or similar that provides `get` and `set` methods, operating directly on the actual C lists. The interface would then consist of a dictionary containing 6 keys and their respective VPLists objects. That would definitely be better as you won't need to use Python structs other than Strings (I don't think you can use the actual char*).

> If there's a real requirement for this in v3.0.x, then we'll merge it, and we'll come up with an entirely new interface in v4.0.x

My current requirement is that I've written a Python module that allows to perform 2FA using TOTP (Google Authenticator and the likes), and that current interface did not allow me to overwrite User-Password. I've written this for Moonshot users, but I would be willing to push it upstream if you think it's interesting. It consists of less than 200 lines that could serve as a more elaborate example of Python module.

Since I'm also the maintainer of the Moonshot FreeRadius fork/backport, it is very easy to me to have this extension included into my copy while is not included upstream. I don't want to push code that you feel is not good enough or does not follow your coding standards or expectations.

I can see if I can implement what I've mentioned above if you feel that would be something you'd like to see in v4.

herwinw:
Mainly for @alejandro-perez (not sure if you are subscribed to the mailing list): I've done a proposal for a new interface [2 years ago](http://lists.freeradius.org/pipermail/freeradius-users/2016-November/085817.html), and then never implemented it. I think we should continue the discussion on the mailing list (although I don't have any objection to continuing it on github, but in that case we should create a new issue for it)

alejandro-perez:
Interesting. I don't know if much code could be shared between languages, though, but the idea is basically the same so the wireframes could be the same as well. I like the idea of instantiating a `ServerStatus` object (or whichever name we want to give to it) as the sole parameter for all the callbacks, so we can do:
```
def authorize(server_status):
  server_status.request.get_attribute('User-Name')
  server_status.reply.set_attribute('User-Name', 'Alex')
  
  return radiusd.RLM_MODULE_OK
```
We can discuss this on the ML, that probably would have a broader impact than discussing on an already closed PR.

PR #2337, opened 2018-10-23
Make Rlm_rest available in pre-proxy and post-proxy

fdurand:
Added rlm_rest call available in pre-proxy and post-proxy section.


alandekok:
I've put the common code into a common function, and manually pulled things over.

PR #2340, opened 2018-10-23
Fix a couple of v3 build warnings

mcnewton:


arr2036:
Why do you need to trim the commit in the configure script?

arr2036:
ah for the magic number stuff

mcnewton:
`src/include/libradius.h` builds `RADIUSD_MAGIC_NUMBER` from "f", `RADIUSD_VERSION` and `RADIUSD_VERSION_COMMIT`. The version is 3 bytes, the commit is 5 bytes, which is 64bit. The "f" adds an extra nibble which pushes it over 64 bits and the compiler spews warnings everywhere.

PR #2342, opened 2018-10-26
V3.0.x radclient.c change in support of --without-tcp

doug-fsu:
### Suggested Patch for src/main/radclient.c
Up until some unknown time ago, the freeradius source could not be compiled using --without-tcp.  Thus, in radclient.c, "#ifdef WITH_TCP" was always true and it was the case that "request->packet->dst_port = server_port;" was always included.

Now, if you compile using --without-tcp, that instruction will not be executed and "request->packet->dst_port" will not get set if we pass a port as part of the command-line argument.  Please be skeptical about this but it looks like it might be as simple as this change:

```
--- radclient.orig	2018-10-26 12:54:15.476727635 -0400
+++ radclient.c	2018-10-26 12:44:46.147461712 -0400
@@ -355,10 +355,10 @@
 		request->packet->src_ipaddr = client_ipaddr;
 		request->packet->src_port = client_port;
 		request->packet->dst_ipaddr = server_ipaddr;
-		request->packet->dst_port = server_port;
 		request->packet->proto = ipproto;
 #endif
 
+		request->packet->dst_port = server_port;
 		request->files = files;
 		request->packet->id = -1; /* allocate when sending */
 		request->num = num++;
```


alandekok:
> 	doug-fsu wants to merge 7,495 commits into master from v3.0.x

No, that's not going to happen.

Please re-do the pull request from v3.0.x plus your patches to v3.0.x.

PR #2344, opened 2018-10-26
radclient.c support for compiling using --without-tcp

doug-fsu:
### Suggested Patch for src/main/radclient.c
Up until some unknown time ago, the freeradius source could not be compiled using --without-tcp.  Thus, in radclient.c, "#ifdef WITH_TCP" was always true and it was the case that "request->packet->dst_port = server_port;" was always included.

Today, if you compile using --without-tcp, that instruction will not be executed and "request->packet->dst_port" will not get set when we pass a port as part of the command-line argument.  Please be skeptical about this but it looks like it might be as simple as this change.
 (I apologize for my earlier trouble in trying to create this pull request)


alandekok:
I've pushed a different patch which achieves the same goal

PR #2346, opened 2018-11-02
Allow driver specific escape methods in rlm_sql

herwinw:
After a discussion on the mailing list these last two days: this patch backports the driver specific escape modules for SQL from master to v3.0.x.

This change will break backwards compatibility, since the escape mechanism will be changed. It will need an extra config option, turn the new behaviour on in the default config, turn it off in the defaults in the code, so new installations will use the improved escape methods, while current installations keep using the old version.

### DIFF
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -378,7 +378,7 @@ static int generate_sql_clients(rlm_sql_t *inst)
 static size_t sql_escape_func(UNUSED REQUEST *request, char *out, size_t outlen,
 			      char const *in, void *arg)
 {
-	rlm_sql_handle_t *handle = arg;
+	rlm_sql_handle_t *handle = talloc_get_type_abort(arg, rlm_sql_handle_t);
 	rlm_sql_t *inst = handle->inst;
 	size_t len = 0;
 
@@ -490,7 +490,7 @@ static size_t sql_escape_func(UNUSED REQUEST *request, char *out, size_t outlen,
 static size_t sql_escape_for_xlat_func(REQUEST *request, char *out, size_t outlen, char const *in, void *arg)
 {
 	size_t			ret;
-	rlm_sql_t		*inst = arg;
+	rlm_sql_t		*inst = talloc_get_type_abort(arg, rlm_sql_t);
 	rlm_sql_handle_t	*handle;
 
 	handle = fr_connection_get(inst->pool);
--- a/src/modules/rlm_sql/rlm_sql.c
+++ b/src/modules/rlm_sql/rlm_sql.c
@@ -122,6 +122,8 @@ static const CONF_PARSER module_config[] = {
 	CONF_PARSER_TERMINATOR
 };
 
+static size_t sql_escape_for_xlat_func(REQUEST *request, char *out, size_t outlen, char const *in, void *arg);
+
 /*
  *	Fall-Through checking function from rlm_files.c
  */
@@ -375,7 +377,8 @@ static int generate_sql_clients(rlm_sql_t *inst)
 static size_t sql_escape_func(UNUSED REQUEST *request, char *out, size_t outlen,
 			      char const *in, void *arg)
 {
-	rlm_sql_t *inst = arg;
+	rlm_sql_handle_t *handle = arg;
+	rlm_sql_t *inst = handle->inst;
 	size_t len = 0;
 
 	while (in[0]) {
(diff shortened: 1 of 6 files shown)


herwinw:
By the way: I have not tested this in any way, it's just a very selective copy-paste from master, and it compiles.

alandekok:
This patch is probably worth merging, as it doesn't do much.

The next step would be to *selectively* enable the new behavior.  e.g. if `safe_characters = auto`, or something like that.

We don't want to change the existing behavior in v3, of course.  But selectively allowing new behavior is OK.

herwinw:
I've fixed a bug, and seen it working as well:

Database:
```
mysql> select * from radcheck;
+----+----------+--------------------+----+---------+
| id | username | attribute          | op | value   |
+----+----------+--------------------+----+---------+
|  1 | bob      | Cleartext-Password | := | hello   |
|  2 | bob'     | Cleartext-Password | := | goodbye |
+----+----------+--------------------+----+---------+
```

Request:
```
bin/radtest "bob'" goodbye 127.0.0.1 0 testing123
```

Some loglines:

```
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: EXPAND SELECT id, username, attribute, value, op FROM radcheck WHERE username = '%{SQL-User-Name}' ORDER BY id
Fri Nov  2 20:49:52 2018 : Debug: (0) sql:    --> SELECT id, username, attribute, value, op FROM radcheck WHERE username = 'bob\'' ORDER BY id
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: Executing select query: SELECT id, username, attribute, value, op FROM radcheck WHERE username = 'bob\'' ORDER BY id
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: User found in radcheck table
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: Conditional check items matched, merging assignment check items
Fri Nov  2 20:49:52 2018 : Debug: (0) sql:   Cleartext-Password := "goodbye"
...
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: EXPAND INSERT INTO radpostauth (username, pass, reply, authdate) VALUES ( '%{SQL-User-Name}', '%{%{User-Password}:-%{Chap-Password}}', '%{reply:Packet-Type}', '%S')
Fri Nov  2 20:49:52 2018 : Debug: (0) sql:    --> INSERT INTO radpostauth (username, pass, reply, authdate) VALUES ( 'bob\'', 'goodbye', 'Access-Accept', '2018-11-02 20:49:52')
Fri Nov  2 20:49:52 2018 : Debug: (0) sql: Executing query: INSERT INTO radpostauth (username, pass, reply, authdate) VALUES ( 'bob\'', 'goodbye', 'Access-Accept', '2018-11-02 20:49:52')
```

I can see `bob'` in the `radpostauth`  table as well.

arr2036:
Agreed.  This needs to be selectively enabled, else you'll probably break any installs that have consumers of the data from the MySQL database.

herwinw:
Added the config option as well

arr2036:
Can you sign your commits and force push them to the branch

herwinw:
Rebased and signed

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -1595,6 +1622,11 @@ static rlm_rcode_t mod_checksimul(void *instance, REQUEST * request)
 		return RLM_MODULE_FAIL;
 	}
 
+	if (radius_axlat(&expanded, request, inst->config->simul_count_query, inst->sql_escape_func, handle) < 0) {
```
arr2036: Why did this call move?

src/modules/rlm_sql/rlm_sql.c
```diff
@@ -478,6 +481,28 @@ static size_t sql_escape_func(UNUSED REQUEST *request, char *out, size_t outlen,
...
+/** Passed as the escape function to map_proc and sql xlat methods
+ *
+ * The variant reserves a connection for the escape functions to use, and releases it after
+ * escaping is complete.
+ */
+static size_t sql_escape_for_xlat_func(REQUEST *request, char *out, size_t outlen, char const *in, void *arg)
+{
+	size_t			ret;
+	rlm_sql_t		*inst = arg;
+	rlm_sql_handle_t	*handle;
```
arr2036: Can you switch this out to ``handle = talloc_get_type_abort(arg, rlm_sql_handle_t)`` to make extra sure that argument is correct.

arr2036:
Looks fine, just need that abort statement to ensure the correct argument is always passed in.

PR #2365, opened 2018-12-04
Convert hex, sha & hmac xlats

pwdng:


src/tests/keywords/sha2
```diff
@@ -9,46 +9,63 @@ update {
...
 
 #
 #  SHA512 and SHA256 share common code paths, so the tests don't need to be
 #  as exhaustive.
 #
-if ("%{sha512:This is a string\n}" != '56b57df5cce42d4e35c644649798ea23ec16f4f4626e78faf4d2d8f430ea349bcc28cd5532457c82f0aa66bf68988346039fe75b900a92ff94fd53993d45990f') {
+update {
+	request:Tmp-Octets-1 := "%{sha512:This is a string\n}"
+	request:Tmp-Octets-2 := "%{sha512:%{&Tmp-String-0}}"
+	request:Tmp-Octets-3 := "%{sha512:%{&request:Tmp-Octets-0}}"
```
arr2036: Mmm, those don't look right, shouldn't have attribute references with expansions like that.  I guess they probably still work because the tmpl code ignores the '&' but it should be removed.

PR #2372, opened 2018-12-14
Add support for HP-user-role attribute

horsitis:
Added support to send user-role for newer software versions of Aruba Switches (previously HP ProCurve), for example 2930F.


alandekok:
I've added the fix to the v3.0.x branch, and will pull it over to "master".

In v3, there is a script `share/format.pl` which formats the dictionaries, so that you don't need multiple commits to get them fixed.

PR #2373, opened 2018-12-14
Update systemd units to depend on network-online

cipherboy:
Due to the use of `SO_BINDTODEVICE`, FreeRADIUS requires the address to be
up when binding to a specific IP address. In systemd, depending on
`network.target` is not sufficient to require the network to be up, so
FreeRADIUS can still fail to bind; instead, `network-online.target` is
required if the network is required to be online.

This manifests in errors such as:

    Error: Failed binding to auth address x.x.x.x port 1812 bound to
           server default: Cannot assign requested address

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1637275

I'm slightly hesitant to add the `network-online.target` as I personally feel it should be configuration-specific (IMO, same with updating the service file to depend on MySQL, PostgreSQL, or any other backend targets -- that's up to the system administrator to depend on and correctly set). Alternatively, as discussed [here in the last footnote](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/), `IP_FREEBIND` can be used, but that seems to be the exact opposite of what [listen wants to do](https://github.com/freeradius/freeradius-server/blob/v3.0.x/src/main/listen.c#L2411-L2430).

Also, if for whatever reason, the network fails to come online, FreeRADIUS will not be started. I'm not sure how much of a problem this is in practice. (e.g., binding to `0.0.0.0` or `127.0.0.1` won't strictly require `network-online.target` and might result in the service being unavailable for some users because systemd didn't attempt to start it because the network wasn't online for some reason).

Thoughts?

### DIFF
--- a/debian/freeradius.service
+++ b/debian/freeradius.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=FreeRADIUS multi-protocol policy server
-After=network.target
+After=network-online.target
 Documentation=man:radiusd(8) man:radiusd.conf(5) http://wiki.freeradius.org/ http://networkradius.com/doc/
 
 [Service]
--- a/redhat/radiusd.service
+++ b/redhat/radiusd.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=FreeRADIUS multi-protocol policy server
-After=network.target
+After=network-online.target
 Documentation=man:radiusd(8) man:radiusd.conf(5) http://wiki.freeradius.org/ http://networkradius.com/doc/
 
 [Service]


alandekok:
The issue seems to be that the user has set `interface = vlan0` in the configuration.  This tells the server to bind to that interface, which has to exist.   So changing the systemd config seems to be the right thing here.

I don't think that the `IP_FREEBIND` will help, as that's used for IPs, and not interfaces.

The `interface = ...` configuration in the server is generally intended to work around routing issues.  e.g. it lets you listen on multiple interfaces, which may duplicate the same IP.  Or, for things like DHCP, it lets you listen for broadcast packets, but only ones which are received on that interface.

In this case, the user could likely drop the `interface = vlan0` configuration in FreeRADIUS, as the IPs it's listening on are all unique.  But that would still require listening on a particular IP.  So either `IP_FREEBIND` would have to be used, or the systemd configuration would have to be updated as is done here.

I think the best solution here is to just fix the systemd config.  The cases where people want a RADIUS server without any networking are vanishingly small.

cipherboy:
@alandekok Okay, thanks for the sanity check. :) I'll push this out to Fedora.

PR #2374, opened 2018-12-14
Fixed mikrotik name

natanmedeiros:
Fixed mikrotik name at line 40.


alandekok:
I've added the fix to the v3.0.x branch, and will pull it over to "master".

PR #2376, opened 2018-12-17
There is no corresponding attr for these two vals

lemenkov:
It looks like a large part of the FreeDHCP dictionary was created with
find+replace magic from share/dictionary.dhcp (later renamed into
share/dictionary/dhcpv4/dictionary.rfc2131). These two values either
slipped through or their corresponding attribute wasn't added. In any
case there is no way to use them now. So let's just remove them.

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>

### DIFF
--- a/share/dictionary/radius/dictionary.freedhcp
+++ b/share/dictionary/radius/dictionary.freedhcp
@@ -233,9 +233,6 @@ ATTRIBUTE	FreeDHCP-Cisco-TFTP-Server-IP-Addresses	150	ipaddr array
 
 ATTRIBUTE	FreeDHCP-End-Of-Options			255	byte
 
-VALUE	FreeDHCP-Opcode			Client-Message		1
-VALUE	FreeDHCP-Opcode			Server-Message		2
-
 VALUE	FreeDHCP-Message-Type		FreeDHCP-Discover	1
 VALUE	FreeDHCP-Message-Type		FreeDHCP-Offer		2
 VALUE	FreeDHCP-Message-Type		FreeDHCP-Request	3


arr2036:
It can probably be removed.  Part of reorganising the dictionaries was that we can now reference the DHCP dictionary from RADIUS, and not need to duplicate all the attribute definitions.  I'll wait for @alandekok to comment though.

alandekok:
TBH, I think we should just delete dictionary.freedhcp.  It's not used for anything.

arr2036:
> TBH, I think we should just delete dictionary.freedhcp. It's not used for anything.

Sorry, yes, I meant delete the whole thing, not just the values.

herwinw:
A bit offtopic, but maybe we should add a check in the unit tests to see if all VALUEs in the dictionaries have corresponding ATTRIBUTEs. Sounds like something one could hack together in Perl in a few minutes.

alandekok:
In v3, it already complains about a `VALUE` no `ATTRIBUTE`:

`Errors reading dictionary: dict_init: No ATTRIBUTE "Foo-Bar" defined for VALUE "maybe"`

lemenkov:
> In v3, it already complains about a `VALUE` no `ATTRIBUTE`:
> 
> `Errors reading dictionary: dict_init: No ATTRIBUTE "Foo-Bar" defined for VALUE "maybe"`

I think it should be checked in Travis as (I believe) @herwinw suggests. Anyway this is likely falls into "patches are welcome" category :)

alandekok:
@lemenkov The travis builds already run `radiusd -XC`, which checks for `VALUE`s without `ATTRIBUTE`.

PR #2379, opened 2018-12-21
More xlat conversions to use to async API.

pwdng:


src/lib/server/xlat_func.c
```diff
@@ -1439,32 +1439,48 @@ static xlat_action_t base64_xlat(TALLOC_CTX *ctx, fr_cursor_t *out,
...
 	}
 
-	if ((size_t)((declen * 2) + 1) > outlen) {
-		REDEBUG("Base64 conversion failed, output buffer exhausted, needed %zd bytes, have %zd bytes",
-			(declen * 2) + 1, outlen);
-		return -1;
-	}
+	MEM(vb = fr_value_box_alloc_null(ctx));
+
+	fr_value_box_memdup(vb, vb, NULL, decbuf, declen, false);
```
arr2036: Use fr_value_box_memsteal instead to avoid the extra alloc and memcpy

src/lib/util/value.c
```diff
@@ -3073,7 +3073,7 @@ int fr_value_box_bstrsnteal(TALLOC_CTX *ctx, fr_value_box_t *dst, fr_dict_attr_t
 	dst->type = FR_TYPE_STRING;
 	dst->tainted = tainted;
 	dst->vb_strvalue = str;
-	dst->datum.length = len - 1;
+	dst->datum.length = inlen;
```
arr2036: Ah good catch :)

arr2036:
Looks good! I like returning a list of pairs from %{pairs:} so that it can be fed into concat.

src/lib/server/xlat_func.c
pwdng: No, that's not possible as the buffer for decoding the base64 data can be longer that the actual decoded data. In some cases, we would end up with too much data. We need to only copy `declen` bytes.

PR #2383, opened 2019-01-11
Include SKIP LOCKED queries as options for sqlippool, some notes, and some tidy up.

nward:
sqlippool uses `SELECT .. FOR UPDATE` queries, which in many configurations cannot run concurrently with other queries.
 - MySQL (even with InnoDB) locks the indexes used in the WHERE and ORDER BY parameters
 - The "assign the same IP address as before" query will wait for the lock to be released on all DBs.
-  The "random" queries will block on some DBs, and fail on others.

The `SKIP LOCKED` option on `FOR UPDATE` ignores locked rows in a `SELECT`. This means that queries can run concurrently. This significantly improves performance of the sqlippool module.

I have added additional query examples including `SKIP LOCKED`, after each of the relevant current examples. I considered replacing the current examples, however, a very recent version of MySQL is required (8.0.1+), and distros such as CentOS/RHEL don't ship a recent enough PostgreSQL (they ship 9.2.x, and 9.5 is required).

I have added some notes to the "random" queries. These run very slowly if you have large pools of free IPs. The `SELECT` runs, copies the results in to a temporary table, then sorts them randomly, then selects the first row. The copy and random sort are very costly. I intend to publish some more query options soon but I think a note here about this is helpful. My first instinct when reading these configs was that the random option would be faster than the "assign the same IP address as before" option - but it most definitely is not, so I think it's important to note.

I have also commented out the "random" query in the PostgreSQL config. It's commented out for the other DBs, not sure why it's not commented out in PostgreSQL. As this was the second in the file, does this mean it changes the default? Happy to scrub this if we don't want the default to change at this time.

### DIFF
--- a/raddb/mods-config/sql/ippool/mysql/queries.conf
+++ b/raddb/mods-config/sql/ippool/mysql/queries.conf
@@ -66,6 +66,7 @@ allocate_find = "\
 
 #
 #  If you prefer to allocate a random IP address every time, use this query instead.
+#  Note: This is very slow if you have a lot of free IPs.
 #
 #allocate_find = "\
 #	SELECT framedipaddress FROM ${ippool_table} \
--- a/raddb/mods-config/sql/ippool/mysql/queries.conf
+++ b/raddb/mods-config/sql/ippool/mysql/queries.conf
@@ -50,6 +50,21 @@ allocate_find = "\
 	FOR UPDATE"
 
 #
+#  The above query again, but with SKIP LOCKED. This requires MySQL >= 8.0.1,
+#  and InnoDB.
+#
+#allocate_find = "\
+#	SELECT framedipaddress FROM ${ippool_table} \
+#	WHERE pool_name = '%{control:${pool_name}}' \
+#	AND (expiry_time < NOW() OR expiry_time IS NULL) \
+#	ORDER BY \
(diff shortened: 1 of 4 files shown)


maximumG:
Thanks for this merge request. We are also considering using a more recent version of MySQL in order to have this 'SKIP LOCKED' feature.  

Without it the overall performance of our DHCP virtual server was about 45-50 DHCP messages/reservations per second whereas our recent test with the SKIP LOCKED allows for more than 180 messages/reservations per second.

nward:
Do you use several IP pools, or do you use the query which tries to assign a previous address?

MySQL also has a performance penalty for this sort of work where the index record is locked, which effectively prevents similar queries from running in parallel if they are locking the row. Splitting pools and I think using the “prefer previous address” query would go some way towards lessening the pain there.

I don’t entirely understand how, but this doesn’t appear to impact Postgres. I would encourage you to test the performance on postgres and see. I saw, from memory (I’m on my mobile right now), at least one order of magnitude increase - I think I was doing thousands per second on Postgres with skip locked vs. 10s per second on MySQL without it.

Accounting is also impacted as it updates the expiry time, especially when bringing a lot of subscribers online at once. I’d recommend doing the buffered accounting for that if you are likely to have many customers coming at once, to reduce retransmissions.

PR #2388, opened 2019-01-15
dhcpclient: Clean up

jpereira:
Removing redundant macro just to keep clean. :)


alandekok:
TBH, I'm OK with leaving the macro in.  Less typing means fewer possibilities for mistakes.

i.e. typing `fr_perror("dhcpclient: ...` is not as nice as `ERROR(...)`

Issue #2385, opened 2019-01-14
ttls fails with TLS 1.3 (openssl 1.1.1)


hartmans:
# Issue type
- **Questions about the server or its usage MUST be posted to the [users mailing list](http://freeradius.org/list/users.html).  If you post those issues here, they will be closed and locked. Repeat offenders will be BANNED**.
- **Remote security exploits MUST be sent to security@freeradius.org.**

***

**REMOVE THOSE WHICH DO NOT APPLY**
- Defect - Non compliance with a standards document, or incorrect API usage.
- Defect - Unexpected behaviour (obvious or verified by project member).

See [here](https://github.com/FreeRADIUS/freeradius-server/blob/master/doc/bugs.md) for debugging instructions and how to obtain backtraces.

# Defect
## How to reproduce the issue
I was putting together automated tests for moonshot-gss-eap in Debian accidentally ended up running with OpenSSL 1.1.1  and without specifying either a minimum or maximum tls version in mods-available/eap.
TLS 1.3 ended up getting selected, and freeradius produced a tls session ticket that moonshot was unable to decrypt at the TLS layer.
Using TLS 1.2 worked fine.
I was several layers deep in trying to get things done, and didn't capture  very good logs, but since it was writing automated tests, reproducing from the test suite is easy to do.

My suspicion is that ttls is just broken with TLS 1.3, but if you think it is more complex than that I'm happy to provide logs.
If it is broken, it would be really nice to get a patch that caps out tls at 1.2 at least for ttls.  It would be best not to cap it out for other things because for example it seems like tls 1.3 would be quite helpful for radsec.

alandekok:
It's possible to set up multiple EAP modules and/or multiple `tls` configurations, which lets you set `tls_min_version` and `tls_max_version`.

I'd prefer to use those as work-arounds instead of coming up with fixes specific for TTLS.

That gives a bit of breathing room, which lets us figure out why TTLS session tickets have issues with TLS 1.3.

From what I can tell, there's nothing specific in TTLS that interacts with TLS 1.3.  The session tickets *should* just be part of the TLS exchange, and therefore completely invisible to TTLS.

hartmans:
So, the problem with using config changes is that administrators tend to
get a radius config that works and not take changes to it from their
packaging.
So, for someone like Debian:

1) If we include it in the config, then people will get stuck at TLS 1.2
well after it's no longer necessary.

2) Conversly, other people will upgrade and it will break because they
[...]

PR #2390, opened 2019-01-16
Set TLSv1.2 as the maximum version for ABFAB (for now)

alejandro-perez:
It seems that a combined TLS configuration where both certificates and PSK
are acceptable does not work properly with OSSL 1.1 and TLSv1.3.

Hence, for now, we establish TLSv1.2 as our default.

### DIFF
--- a/raddb/sites-available/abfab-tls
+++ b/raddb/sites-available/abfab-tls
@@ -21,6 +21,11 @@ listen {
 		ca_path = ${cadir}
 		cipher_list = "DEFAULT"
 
+		# Having a TLS configuration allowing both PSK and certificates at the
+		# same time does not work with OpenSSL 1.1 and TLS 1.3 at this time.
+		# We restrict ABFAB to TLS 1.2 until this is resolved
+		tls_max_version = 1.2
+
 		cache {
 			enable = no
 			lifetime = 24 # hours


arr2036:
Eh, EAP-TTLS-PAP doesn’t work either... do you have specific info on why this doesn’t work, or is this just a temporary hack?

alejandro-perez:
There is a mail thread where I discussed with Alan some of my findings.
http://lists.freeradius.org/pipermail/freeradius-devel/2019-January/013344.html

The problem seems to be related somehow when you want to use PSK *and* certificates at the same time. With TLS 1.2, if PSK is used, the rest of parameters (including whether to request a certificate from the client, etc.) are just ignored. With TLS 1.3 there seem to be some problem with that. If you use TLS 1.3 and PSK without any certificate-related parameter, it works. If you use TLS 1.3 and certificates without PSK, it works as well. It's the mixture of both what precludes PSK to work when they are both enabled at the same time.

I've made some tests with `openssl s_client` and `openssl s_server` and I think that it also happens there. From the little I've been able to learn about TLS 1.3, it seems that PSK-specific ciphersuites are no longer needed. PSK becomes a sort of TLS session resumption. 

Some tests one can easily do with openssl to see this (just use any freeradius installation on a OSSL 1.1 system - eg. debian buster):

Test 1: TLS 1.3 - PSK only (works and you end up with a Session)
```
openssl s_client -connect localhost:2083 -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cipher aPSK

openssl s_server -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cert /etc/freeradius/certs/server.pem -key /etc/freeradius/certs/server.key -port 2083 -nocert
```

Test 2: TLS 1.3 - PSK and Certificates (It does not fail completely, but you don't get a session)
```
openssl s_client -connect localhost:2083 -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cipher aPSK

openssl s_server -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cert /etc/freeradius/certs/server.pem -key /etc/freeradius/certs/server.key -port 2083
```

Test 3: TLS 1.2 - PSK and Certificates (works, you get your session)
```
openssl s_client -connect localhost:2083 -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cipher aPSK -no_tls1_3

openssl s_server -psk 03592CFAD91DB9E6DBDC022A80E14D15B62493C7D6D114272FD70374EA920E75DEB5DFC760487D96F8E1D4310FC617167DD539CB4245BEFE083BE71CA28F937A13EE3ECA0EBC4FC6687A948CE0F32FBF99AAF2B9024EAEF5688B1CA054F3DDA54D65277223ED25FB3C2D7F39FDE1FE1D2EF1D8B6F04F97121095A077A33784CD2754E0E6A8511F5DC334A0394E41B7300ED7ACCC62F8903B9092A0CD40FA12F8DC825852BD080292519450F37C80D96033704E86E36874635D800F0AD4230E9C9343060E25343BF9D33A12B979EF6318A6F32FCE9791423C82DE8DCF17634247592D86F4F5029CFF258D6DDC8ADDFB6223E02C8B7843C991686966775B9044D3 -psk_identity key-2378d9 -cert /etc/freeradius/certs/server.pem -key /etc/freeradius/certs/server.key -port 2083
```

alandekok:
Ah yes. In TLS 1.3, the PSK cipher suites have been removed. And PSK is only for session resumption. 

  I’d double check the RFC, but I’m out of the office right now. 

  If that is true, when when PSK is enabled, we should warn that tls 1.3 won’t work. Or, disable tls 1.3 when PSK is enable. 

  I think this issue is baked into the standard. So we can’t work around it. We can only forbid the problematic configuration. Or, warm the admin.

alejandro-perez:
> If that is true, when when PSK is enabled, we should warn that tls 1.3 won’t work. Or, disable tls 1.3 when PSK is enable.

It does work when no certificate configuration is given (see the examples above). If the listen.tls{} configuration section just contain psk-related parameters, it does work. It's only that you cannot create a generic configuration that works with both, PSK and Certificates depending on what the client wants.

alandekok:
Then if there’s both PSK and certs configured, it should disable tls 1.3. And print a warning message saying so

alejandro-perez:
My first though was trying to disable any cert-related parameter of the SSL connection when we knew it was a PSK-based attempt. This is known in the PSK callback. But it seems you cannot unset a certificate once you've set it.

alandekok:
I pushed changes to `tls.c` which should do the same thing.  The check is a little more narrow, and should automatically work everywhere.

Please check...

alejandro-perez:
> Please check...

It does not work for me. It seems that since it is being checked *before* max/min TLS versions are processed, it gets overwritten.

I've created a gits with an alternative, that has been moved to the place where versions are processed, and it also sets the maximum to 1.2 if it was 1.3 (to avoid raising it from 1.1).
https://gist.github.com/alejandro-perez/05f7f9560d77af7a5a6453c4fbffe14a

alejandro-perez:
> I pushed changes to `tls.c` which should do the same thing. The check is a little more narrow, and should automatically work everywhere.

Also, I cannot assure this is a bug of OSSL rather than just me not understanding how to properly make this work (eg. missing some initialisation or anything else, or not understanding how TLS 1.3 works).

alandekok:
Until we can get it to work, I'd rather forbid the bad configuration than allow it, and confuse the end user.

alejandro-perez:
> I've created a gist with an alternative, that has been moved to the place where versions are processed, and it also sets the maximum to 1.2 only if it was 1.3 (to avoid raising it from 1.1).
> https://gist.github.com/alejandro-perez/05f7f9560d77af7a5a6453c4fbffe14a

Did you have the chance to look at this? The original patch you committed did not work. This one does, but you might not like where I put it. I did not create a new PR since discussion is happening here.

alejandro-perez:
I think this is related to #2385.

alandekok:
I'd like to know why the patch doesn't work.  Likely because you already set min/max version in your config.

The patch you propose only does the fix if the admin *already* sets min/max TLS version.  It should work no matter what the config.  I'll see if I can push a fix.

alejandro-perez:
> I'd like to know why the patch doesn't work. Likely because you already set min/max version in your config.

No. I think it was because the logic that handles the config came later and overwrote the value anyway.

> The patch you propose only does the fix if the admin _already_ sets min/max TLS version. It should work no matter what the config. I'll see if I can push a fix.

Not according to my tests. I did not set any min/max TLS version and yet it worked.

alejandro-perez:
> I think commit [52a2133](https://github.com/FreeRADIUS/freeradius-server/commit/52a21336103ea421aefe0e3d2060dcf4a54ee71b) will fix it.
> 
> The logic there is a bit better. You can set: `min_version = "1.3"`, in which case it will _allow_ PSK and certs to be used. Or, you can set `max_version = "1.2"`, in which case it won't complain.
> 
> Or, if you set `min_version = "1.0"` and `max_version = "1.3"`, it will complain, and reset `max_version` to 1.2.
> 
> This should work with all possible configs. It should also have the minimum number of complaints.

I'll test it out and let you know. Won't take more than 1h

alejandro-perez:
> I'll test it out and let you know. Won't take more than 1h

Done. It works fine.
So, I guess we can close this PR without merging, as it is no longer required.
Thanks!

arr2036:
I tracked down where in RFC 8446 it expressly prohibits both the certificate and PSK extensions being sent.

Section 4.1.1 states

```
   -  When authenticating via a certificate, the server will send the
      Certificate (Section 4.4.2) and CertificateVerify (Section 4.4.3)
      messages.  In TLS 1.3 as defined by this document, either a PSK or
      a certificate is always used, but not both.  Future documents may
      define how to use them together.
```

This isn't really an OpenSSL bug, TLS 1.3 simply won't let you use both together, so only one should be configured for the ``SSL *``.  IMHO it should be a fatal error if the maximum TLS version is unset or >= 1.3 and a user configured both PSKs and certs.  In the vast majority of cases a user **won't** do this, but it's better to explicitl

Another minor note PSK isn't only for session resumption, it's still a valid key derivation method, it's just that session resumption also leverages PSK authentication.

alejandro-perez:
> This isn't really an OpenSSL bug, TLS 1.3 simply won't let you use both together, so only one should be configured for the `SSL *`. IMHO it should be a fatal error if the maximum TLS version is unset or >= 1.3 and a user configured both PSKs and certs. In the vast majority of cases a user **won't** do this, but it's better to explicit. Thoughts on this?

As I understand it, you cannot use both, but OpenSSL could do a better job and just ignore the configured certificates when the client indicates it wants to use PSK. (I think this is done for TLSv1.2).

In any case, Moonshot users can live with the current behaviour. Either they use TLS1.2 (which is not insecure at all) or they can split the configuration into two `listen` sections, one using PSK exposed to the federation, and one using certificates exposed to their RPs/APs.

arr2036:
Ah, true, OK.

PR #2391, opened 2019-01-19
Tidying for next release

mcnewton:
Debian packaging lintian fixes and some other typos and formatting consistency


mcnewton:
I did some v4 updates already, but most likely some of it will apply. I want to take a bigger look at the v4 packaging at some point though so I'm going to look then.
-- 
Matthew

PR #2398, opened 2019-01-23
Add new target 'asciidoc-clean'

jpereira:


alandekok:
My comments:

* other rules are `clean.FOO`, so using that would be best
* it shouldn't print anything when running a `clean` target
* it could just do `rm -rf asciidoc`, which is simpler.  We don't need to check for the directory, and we don't need to have an `exit 0`.  Small scripts are best
* we should add a rule so that `make clean` also gets rid of the asciidoc files.

I'll push a separate patch to show what I mean.

jpereira:
Got it. thank you @alandekok for the feedback.

PR #2399, opened 2019-01-23
Update raddb/mods-available/delay

jpereira:


raddb/mods-available/delay
```diff
@@ -1,40 +1,50 @@
...
 #	relative = no
 }
 
+#
+# ## Delay example
+#
+
+#
+#  How long to delay request processing for.
+#
```
arr2036: Huh, why did you move this here? It doesn't make sense at the module level?

PR #2416, opened 2019-01-25
Add link to the rfc5080

jpereira:


alandekok:
this is now done automatically as of commit 8acca20f43

PR #2417, opened 2019-01-25
Enable hiperlinks for RFC's mentions.

jpereira:


arr2036:
Yes. @alandekok already pushed changes to the conversion script.

PR #2419, opened 2019-01-25
Complete and improve realm module documentation

alejandro-perez:
The `rekey_enabled` and `rekey_lifetime` options were undocumented.

### DIFF
--- a/raddb/mods-available/realm
+++ b/raddb/mods-available/realm
@@ -29,13 +29,14 @@ realm suffix {
 	format = suffix
 	delimiter = "@"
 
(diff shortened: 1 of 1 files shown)


alandekok:
Can you sign the commits && do a force push?

alejandro-perez:
Done. Is that something you want contributors to do? I could do it hereafter if that's required. Previously I've not been asked to do it.

alandekok:
We should have commits signed moving forward.  It helps with identification, and ongoing security of the git history.

PR #2421, opened 2019-01-25
Some fixes

jpereira:


jpereira:
@alanbuxey the "RFC XXXX, Section YYYY" should be on the same line to be parsed as well.

jpereira:
@alandekok The "RFC XXXX, Section YYYY" should be on the same line to be parsed as well.

PR #2427, opened 2019-01-25
Update raddb/mods-available/counter

jpereira:


raddb/mods-available/counter
```diff
@@ -1,81 +1,133 @@
...
+	#  |`h` | hours
+	#  |`d` | days
+	#  |`w` | weeks
+	#  |`m` | months
+	#  |===
+	#  
+	#  NOTE: If the letter is omitted days will be assumed.
+	#  
+	#  .In example:
+	#
```
arr2036: "In example" doesn't really make sense here.  It'd be "For example", but not even sure that's needed.

raddb/mods-available/counter
```diff
@@ -1,81 +1,133 @@
...
+
+	#
+	#  check_name::
+	#
+	#  The `check_name` attribute defines an attribute which will be
+	#  registered by the counter module and can be used to set the
+	#  maximum allowed value for the counter after which the user
+	#  is rejected.
+	#
+	#  .Something like:
```
arr2036: "Something like" does make sense, but isn't really something that you'd use for a caption.

PR #2429, opened 2019-01-28
issue 2395 sqlippool allocate clear v3

maximumG:
Thanks to this changes, it is now possible to specify the rate (in second) at which the allocate_clear statement should be run.

The default rate is 1 second (set in the config file) in order to be retro-compatible with the old configuration.

### DIFF
--- a/raddb/mods-available/sqlippool
+++ b/raddb/mods-available/sqlippool
@@ -28,6 +28,12 @@ sqlippool {
 	lease_duration = 3600
 
 	#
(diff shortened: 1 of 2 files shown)


alandekok:
Can you squash all of the commits together?  We'd rather not have it go "add patch, fix it, fix it. ..."

I'll then pull it over for 3.0.18

maximumG:
I did something wrong during the rebase. I will republish the code by tomorrow.

maximumG:
The pull request now contains only one commit.

PR #2437, opened 2019-01-28
Add capability to resolve the 'man XXX'

jpereira:


arr2036:
I vote no.  Man pages are system and distribution specific, the only ones we should link to are our internal FreeRADIUS man pages.  Linking to Linux man pages makes it seem like the project is more Linux centric, which is definitely not the case.

alandekok:
That's fine.  There's no pressing need to point to "man" pages, as we don't use many third-party tools.

PR #2439, opened 2019-01-29
clarify which CA certs to include when using certificate_file without ca_file

restena-sw:
Sending the root serves no useful purpose; the client has this root CA installed and verifies against that trust base, or it doesn't. No amount of sending it as part of the untrusted cert data will make the root CA any more or less trusted.
The previous wording was confusing for eduroam admins and consequently we see many who send a superfluous *root* certificate with their server cert. This typically costs another 1-2 roundtrips for no real purpose.
The new wording is more explicit that *intermediate* CAs are a good idea, but no root.

### DIFF
--- a/raddb/mods-available/eap
+++ b/raddb/mods-available/eap
@@ -190,10 +190,14 @@ eap {
 		#  name.
 		#
 		#  If ca_file (below) is not used, then the
-		#  certificate_file below MUST include not
-		#  only the server certificate, but ALSO all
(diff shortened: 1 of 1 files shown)


arr2036:
Don’t tls 1.0 and 1.1 require the complete chain? I thought the ability to omit the root CA only came in with 1.2?

-Arran

restena-sw:
No, that was never the case.

PR #2443, opened 2019-01-29
Update raddb/mods-available/digest

jpereira:


raddb/mods-available/digest
```diff
@@ -2,12 +2,20 @@
 #
 #  $Id$
 
+#######################################################################
 #
-#  The 'digest' module currently has no configuration.
+#  ## Digest Module
 #
-#  "Digest" authentication against a Cisco SIP server.
-#  See 'doc/rfc/draft-sterman-aaa-sip-00.txt' for details
```
alandekok: It turns out that the draft is what people implemented.  There's an RFC, but pretty much no one uses it.

The module implements the draft, not the RFC.  The comments should be updated to note this. 

I'll do it after the merge.

PR #2449, opened 2019-01-30
Backport some TLS1.3 fixes

mcnewton:


src/main/tls.c
```diff
@@ -3377,7 +3377,19 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client)
...
 	} else {
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+		/*
+		 *	This controls the number of stateful or stateless tickets
+		 *	generated with TLS 1.3.  In OpenSSL 1.1.1 it's also
+		 *	required to disable sending session tickets,
+		 *	SSL_SESS_CACHE_OFF is not good enough.
+		 */
+		SSL_CTX_set_num_tickets(ctx, 0);
+#endif
```
alandekok: I've pulled this into v3.0.x.  It's needed for OpenSSL 1.1.1, even without TLS 1.3.

src/main/tls.c
```diff
@@ -3897,7 +3897,7 @@ fr_tls_status_t tls_ack_handler(tls_session_t *ssn, REQUEST *request)
 		return FR_TLS_FAIL;
 
 	case handshake:
-		if ((ssn->info.handshake_type == handshake_finished) && (ssn->dirty_out.used == 0)) {
+		if (SSL_is_init_finished(ssn->ssl) && (ssn->dirty_out.used == 0)) {
```
alandekok: That should likely be `tls_session->is_init_finished`, which was added for 3.0.18

mcnewton: OK - if so should d90c4bf807c66 in master also be changed?

alandekok: Yes, likely

arr2036: Why? It’s better to get the information directly from the SSL *?

alandekok: Because the change to add `tls_session->is_init_finished` wasn't done on a whim.  See:

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/main/tls.c#L792

While the change is perhaps not needed here, it's best to have one way of doing things.  If we randomly do API calls vs checking local variables, then we have a higher potential of doing the wrong thing.

arr2036: is_init_finished is only used to implement an edge based trigger to add the TLS-Session-Version attribute to the session list. The master branch code works differently and doesn't require a hack to implement the edge based trigger.

alandekok: `is_init_finished` is used for more than that in v3.  And this patch set is for v3.

arr2036: My main concern is code from v3.0.x being applied to the master branch without careful consideration.

arr2036: My issue was with @mcnewton asking if the master branch should be changed and your response of:

> Yes, likely

That's why I was stating that it wasn't required in master.  I agree in v3.0.x that we should use the shadowed value.

mcnewton: So agreed `ssn->is_init_finished` in v3, and leave master as-is (using `SSL_is_init_finished()`)?

alandekok: That's fine.

alandekok:
The IETF is still making changes to the EAP-TLS 1.3 standard.  Hopefully we'll have finalization soon.

alandekok:
I'd be open to pulling most of these over for 3.0.20.  That way even if the IETF isn't done, the remaining patches are minimal.  We'll know more in the next few days.

alandekok:
We should also disable SSL_OP_ENABLE_MIDDLEBOX_COMPAT:

https://w1.fi/cgit/hostap/commit/?id=8b57a378081bbc33387c86e994c7d282d413395d

alandekok:
Hostap / wpa_supplicant should support the current understanding of EAP-TLS as of

https://w1.fi/cgit/hostap/commit/?id=7ad9e36d4a769f72b4b45c447a006d997787675f

It would be worth double-checking that.

arr2036:
Can this be rebased, or closed?

mcnewton:
Rebased this, some of the patches were already in. @alandekok may know better whether this final commit should actually be pulled in or not, though?

alandekok:
we should probably pull it in.  wpa_supplicant now supports TLS 1.3, and follows the preliminary spec.  We should probably ensure that we work with wpa_supplicant before the final spec is published.  Otherwise there will be a standard that doesn't work.

jimdigriz:
This probably can be closed now https://github.com/FreeRADIUS/freeradius-server/pull/3516 is merged.

PR #2460, opened 2019-02-02
Fix the pdf target

jpereira:


jpereira:
It will fix the `make pdf`

```
[jpereira@sugarloaf freeradius-server.git]$ make pdf
PDF asciidoc/templates.conf.adoc
--latex-engine has been removed.  Use --pdf-engine instead.
Try pandoc --help for more information.
make: *** [Makefile:374: asciidoc/templates.conf.pdf] Error 2
```

mcnewton:
Depends on the version of pandoc... the one in Ubuntu 18.04 doesn't have that option :-(

PR #2463, opened 2019-02-05
Some updates

jpereira:


raddb/mods-available/linelog
```diff
@@ -1,29 +1,44 @@
...
+#  = Linelog Module
 #
-#  The "linelog" module will log one line of text to a file.
+#  The `linelog` module will log one line of text to a file.
 #  Both the filename and the line of text are dynamically expanded.
 #
-#  We STRONGLY suggest that you do not use data from the
-#  packet as part of the filename.
+#  WARNING: We STRONGLY suggest that you do not use data from the packet
+#  as part of the filename.
```
arr2036: Add "unless `escape = yes`"

raddb/mods-available/linelog
```diff
@@ -34,22 +49,30 @@ linelog {
...
 
 	#
-	#  Reference the Packet-Type (Access-Request, etc.)  If it doesn't
-	#  exist, reference the "format" entry, above.
+	#  reference::
+	#
+	#  Reference the `Packet-Type` (`Access-Request`, etc.)  If it doesn't
+	#  exist, reference the `format` entry, above.
+	#
+	#  May be an attribute ref, `xlat`, `literal` or `exec`.
```
arr2036: I guess we don't have these docs merged in yet.  Maybe talk to @alandekok  and see if we can get the NR docs included in the doc source tree.

arr2036: That'd also be `attribute ref` in backticks ;)

raddb/mods-available/linelog
```diff
@@ -34,22 +49,30 @@ linelog {
...
 	#
-	#  The messages defined here are taken from the "reference"
+	#  messages { ... }::
+	#
+	#  The messages defined here are taken from the `reference`
 	#  expansion, above.
 	#
-	#  Pairs may be attributes refs, xlats, literals or execs.
+	#  Pairs may be attributes refs, `xlat`, `literal` or `exec`.
+	#
```
arr2036: Attribute ref in backticks here too.

raddb/mods-available/linelog
```diff
@@ -266,7 +387,7 @@ linelog log_accounting {
 		Accounting-On = "NAS %{%{Packet-Src-IP-Address}:-%{Packet-Src-IPv6-Address}} (%{%{NAS-IP-Address}:-%{NAS-IPv6-Address}}) just came online"
 		Accounting-Off = "NAS %{%{Packet-Src-IP-Address}:-%{Packet-Src-IPv6-Address}} (%{%{NAS-IP-Address}:-%{NAS-IPv6-Address}}) just went offline"
 
-		# don't log anything for other Acct-Status-Types.
+		# Don't log anything for other `Acct-Status-Type`.
```
arr2036: add 's' back after the backticks.

raddb/mods-available/logintime
```diff
@@ -1,23 +1,43 @@
...
+#
+# = Login time Module
+#
+# The `logintime` module handles the `Login-Time`, `Current-Time`,
+# and `Time-Of-Day` attributes.
+#
+# It should be included in the *end* of the `authorize` section in order
+# to handle `Login-Time` checks. It should also be included in the
+# `instantiate { ... }` section in order to register the `Current-Time`
+# and `Time-Of-Day` comparison functions.
```
arr2036: Remove

```
It should also be included in the
	# `instantiate { ... }` section in order to register the `Current-Time`
	# and `Time-Of-Day` comparison functions.
```

It's not really relevant anymore

raddb/mods-available/logtee
```diff
@@ -1,40 +1,59 @@
...
 	#  Additional expansions are available:
-	#  - %{Log-Message}	- The log message
-	#  - %{Log-Level}	- Debugging level
-	#  - %{Log-Type}	- Error, Warn, Info, Debug etc...
 	#
-	#  Format could contain markup (JSON for example), but it's up
+	#  [options="header,autowidth"]
+	#  |===
+	#  | Option         | Description
+	#  | %{Log-Message}	| The log message
```
arr2036: Expansions should be in backticks

jpereira:
@arr2036 I did the changes. only waiting for @alandekok to say something about the point about the https://github.com/FreeRADIUS/freeradius-server/pull/2463#pullrequestreview-199900332

raddb/mods-available/linelog
alandekok: How about:

```
May be an attribute reference, e.g. `Usert-Name`, or `xlat`, `literal` or `exec`.
```

jpereira:
@alandekok @arr2036 I did the changes. we are good to go.

raddb/mods-available/linelog
arr2036: An attribute reference is at the same level as xlat, literal or exec.  We could shorten it to `attref`? It'd be `&User-Name` if you wanted to give an example.

arr2036: For @jpereira 's benefit.

- attribute reference is a description of an attribute and where to find it. e.g. `&reply:Reply-Message` it all supports all the weird and wonderful accessor syntaxes like `&reply:Reply-Message[5]` to get the sixth instance of the reply-message in the reply list.
- xlat is a string expansion, it can call a function like `%{sql:<query>}` or be used to interpolate attribute values, in which case the syntax is the same as attribute references but without the leading & e.g. %{reply:Reply-Message[5]}.
- Literal is a string literal e.g. 'I am a string literal'.  Depending on the quotes used, different escape sequences are supported. 
- Exec is a call out to a script. The script is places in backticks `/usr/bin/myscript.sh -user %{User-Name}`

alandekok: It's probably worth adding another page that explains all of this, and then referencing that page from this (and other) locations.

The alternative is to have the same text repeated everywhere, which makes me unhappy.

mcnewton: I'm not convinced all those backticks above are correct. Backticks are generally to show code blocks or examples of something that's being typed, or will appear on screen. So "The `xlat` is set here" would be wrong ("xlat" is just what it's called), but "set the `reference` here" would be correct (as above) _if_ there is an actual option called "reference" that is being set (the option is something that appears on the screen).

So I would say that the last line should lose all its backticks - it's simply

    May be an attribute reference, e.g. `&User-Name`, an xlat, a string literal or an exec.

Better (as that sentence confused me too for a while) might be to make them all links to other pages that explain what they are, or have examples.

    May be
      - an attribute reference, e.g. `&User-Name`;
      - an xlat, e.g. `"%{sql:...}"`;
      - a string literal, such as `"a literal string"`;
      - or an exec, e.g. `` `/bin/echo hello` ``

As @alandekok said, duplication is bad. Small examples might be OK, but the main explanation  of what these are should be on another page and be linked to.

jpereira: @mcnewton The backticks here is following the asciidoc https://asciidoctor.org/docs/asciidoc-syntax-quick-reference/ manual and here in special has a different behavior because the script https://github.com/FreeRADIUS/freeradius-server/blob/master/scripts/asciidoc/conf2adoc does some extra parsing.

jpereira: About the duplication, I agree that we should avoid to duplicate. An option could be to create a kind of glossary with the usual options/sections e.g: pool { ... }

alandekok: A common glossary is useful.  Maybe `doc/raddb/common/` ?

We can put manually generated files there, and refer to them from the config items.

Though in this case, it's really more common *concepts*, and could go into a "reference" documentation

mcnewton: The asciidoc manual explains what the markup does, but I don't see any more than it says that backticks make something monospace.

So my question is, why would you write "This can be a `string literal`" rather than "This can be a string literal"?

I don't see a reason for "string literal" to be monospaced, and equivalently no need for "xlat" or "exec" to be either. The _contents_ of them, yes, such as "`"this text"` is a string literal" or "`%{sql:...}` is an xlat", but not the name itself.

The other ones are all good, though :)

I'm not sure what you mean by the conf2adoc script. It would convert e.g. `` `exec` `` to a link to the module of the same name, but in this case "exec" is nothing to do with that module.

mcnewton: > A common glossary is useful. Maybe `doc/raddb/common/` ?

That sounds good - tidying up `doc/` would be nice, and then linking to more full documentation there where it's not directly part of the config makes sense.

In theory it doesn't even need to be `doc/raddb/...`, just `doc/glossary.md` or `doc/guide/eap.md` for example, as by the fact they're being moved out of `raddb/` means they're more general documentation.

PR #2464, opened 2019-02-05
Include some feedback when trust router server is not responding

alejandro-perez:
So far, if FreeRadius failed to contact the Trust Router server, it just returned Access-Reject.
This made this error indistinguible from an authentication failure.
Just adding something meaningful in the Reply-Message

### DIFF
--- a/src/modules/rlm_realm/trustrouter.c
+++ b/src/modules/rlm_realm/trustrouter.c
@@ -625,10 +625,13 @@ REALM *tr_query_realm(REQUEST *request, char const *realm,
 	pthread_mutex_unlock(&tidc_mutex);
 
 	/* If we weren't able to get a response from the trust router server, goto cleanup (hence return NULL realm) */
-	if (!rv) goto cleanup;
+	if (!rv) {
+		DEBUG2("Could not connect with TR server, rv = %d\n", rv);
+		pair_make_reply("Reply-Message", "Could not connect with Trust Router server", T_OP_SET);
+	}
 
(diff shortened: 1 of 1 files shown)


alandekok:
I've pushed a different fix.

I'm wary of adding a `Reply-Message` to *every* related `Access-Reject`.  This is a behavior change, and it's not clear that it's necessary.

Instead, the code now adds `Module-Failure-Message`.  An administrator can then copy that to `Reply-Message` if necessary.

alejandro-perez:
I guess that could be also applied to the error below that one then, where Reply-Message is being used.

alandekok:
In v3:

```
Post-Auth-Type Reject {
  if (Module-Failure-Message) {
    update reply {
      Reply-Message += &request:Module-Failure-Messge
    }
  }
}
```

alandekok:
> I guess that could be also applied to the error below that one then, where Reply-Message is being used.

Sure, but I think it's fine to leave the existing behavior, too.  Adding a `module_failure_message()` call would be useful.  I'll do that.

alejandro-perez:
> In v3:
> 
> ```
> Post-Auth-Type Reject {
>   if (Module-Failure-Message) {
>     update reply {
>       Reply-Message += &request:Module-Failure-Message
>     }
>   }
> }
> ```

Thanks. That will probably end up into our abfab-tr-idp "site" anyway...



> Sure, but I think it's fine to leave the existing behavior, too. Adding a `module_failure_message()` call would be useful. I'll do that.

Thanks

PR #2467, opened 2019-02-06
Add *.adoc to .gitignore

jpereira:


alandekok:
I'll add that to the `doc/.gitignore` file, tho we will likely be having "adoc" files in that directory too.  I'll see if there's a more reasonable approach.

Maybe use Markdown for the manually written files in that directory, and asciidoc for the auto-generated ones?

jpereira:
@alandekok well, we can keep there a simple README.md saying the purpose of the dir or something like that. what do you think?

PR #2469, opened 2019-02-06
rlm_rediswho: Fix some missing Acct-Status-Type fields

jpereira:
Based on the rfc2866, we are missing those Acct-Status-Type fields.


jpereira:
ps: I will update the raddb/mods-available/rediswho soon.

arr2036:
Hm why did you close this, looks ok to me?

jpereira:
Mmmm, Because we don't receive the &User-Name during those events. But, we can do the query using another field.

That is it?

arr2036:
Have to push all the session IDs into another list for NAS-Ip-Address

alandekok:
I would prefer to *not* add configuration / code that does nothing.   Once the module is updated to track sessions, that's the time to add config / code for `on` and `off`.

jpereira:
Got it. I will do that today.
--
Jorge Pereira

PR #2473, opened 2019-02-08
rlm_rediswho: Add 'stub' sections for missing fields. ref #2469

jpereira:


jpereira:
Those stubs will allow the user *to do something* with those fields.

PR #2476, opened 2019-02-11
conf2adoc: Fix - treat everything as literal

jpereira:
It will treat everything in between as literal when the input string has some special character. e.g: ()[]{}

for example the URL https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/WinArchive/[MS-SOH].pdf

`\Q - Quote (disable) pattern metacharacters`


jpereira:
`\Q  - Quote (disable) pattern metacharacters`

PR #2477, opened 2019-02-11
conf2adoc: Fix - Match anything between backtick

jpereira:


jpereira:
It fixes the matches with anything between backticks, even when the input has 'regex` commands. eg: ()[]

It will fix the match with some inputs `..... (`anything`)....`, e.g:

```
#  Rather than maintaining separate (`GDBM`) databases of accounting info
```

PR #2481, opened 2019-02-12
V3.0.x

2128506:
Hi! This is backport of recent commit for TLS support in MySQL driver to 3.0.x branch along with necessary fixes.


alandekok:
Please update this so that there is *one* commit which adds the functionality.  There is no need to have multiple commits to fix broken things.  There is no need to open multiple PRs for the same issue.

arr2036:
As @alandekok states, just merge the commits into one, I don't need credit for every commit.  You'll also need to sign your commits.  There are multiple tutorials online.

2128506:
Merged and signed.

PR #2482, opened 2019-02-12
Fixed errors in MySQL TSL support

2128506:
This patch fixes some errors in recently added TLS support in MySQL driver:
1. ifdef should not be used for testing option availability as at least in MySQL 5.7 option codes are enum values, not defines. Same applies to reconect option.
2. Some option and field names were incorrect


alandekok:
It is MUCH preferred to just fix the errors in your local repo, and then do a "force push" back.  That way the previous PR gets updated, and we don't have broken code in the repository.

alandekok:
Please do:

    git rebase -i HEAD^^^^

or however many `^` is necessary to get all of the commits.  Then, edit the resulting file to "squash" all of the commits into one.  Then:

   git push --force ...

to update #2481 so that it only has one patch to add the new feature.

arr2036:
@alandekok I wrote the original enhancement, but couldn't test it because I couldn't get a high enough mysql client version locally. @2128506 is submitting a fix to my original commit as a PR.  I merged some elements manually, because @2128506 didn't offer an explanation for the contents of the PR, and it appeared the majority of the changes were personal style preference.  It now appears that they are required, so i've merged the changes from the original PR.

@2128506 also has another PR open for backporting the changes to v3.0.x.  @2128506 should merge all commits in that PR and sign the result if they want the feature backported.

PR #2483, opened 2019-02-12
Use *two* spaces through the entire raddb/mods-available/*

jpereira:


alandekok:
fixed via commit 7bf85d9caf

PR #2485, opened 2019-02-13
Use cache-ttl < 0 even if entry doesn't exist

### LINKED ISSUE #2484: Cache doesn't use Cache-TTL of < 0 if there is no existing entry
# Issue type

- Defect - Unexpected behaviour (obvious or verified by project member).

# Defect

The cache module configuration file states

> Setting a Cache-TTL of < 0 means expire the existing entry (without merging) and create a new one with TTL set to value * -1.

However, in the 3.0.x branch, if there is no existing entry then the TTL is set to the default TTL value of the cache, instead of `Cache-TTL * -1`

## How to reproduce the issue

Set the control:Cache-TTL attribute to a value < 0, then call the cache module when there is no existing entry matching the key.
 
## Output of ``radiusd -X`` showing issue occurring

```text
(2)     update config {
(2)       &Cache-TTL := -20
(2)     } # update config = noop
(2) cache: EXPAND %{User-Name}
(2) cache:    --> bob
(2) cache: No cache entry found for "bob"
(2) cache: Creating new cache entry
(2) cache:   &reply:Reply-Message += &reply:Reply-Message[*] -> 'Hello, bob'
(2) cache: Merging cache entry into request
(2) cache:   &reply:Reply-Message += "Hello, bob"
(2) cache: Committed entry, TTL 10 seconds
(2)     [cache] = updated
(2)   } # post-auth = updated
```

eeverett6:
Fixes #2484

### DIFF
--- a/src/modules/rlm_cache/rlm_cache.c
+++ b/src/modules/rlm_cache/rlm_cache.c
@@ -477,7 +477,8 @@ static rlm_rcode_t CC_HINT(nonnull) mod_cache_it(void *instance, REQUEST *reques
 	 *	If there's no existing cache entry, go and create a new one.
 	 */
 	if (!c) {
-		if (ttl <= 0) ttl = inst->ttl;
+		if (ttl < 0) ttl *= -1;
+		if (ttl == 0) ttl = inst->ttl;
 		goto insert;
 	}
 


alandekok:
I've pushed a fix.

PR #2488, opened 2019-02-14
docker: Fix debconf warnings upon build

jpereira:


jpereira:
It will fix the debconf warnings messages.

e.g:

```
Setting up ca-certificates (20180409) ...
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.26.1 /usr/local/share/perl/5.26.1 /usr/lib/x86_64-linux-gnu/perl5/5.26 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.26 /usr/share/perl/5.26 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7.)
debconf: falling back to frontend: Teletype
Updating certificates in /etc/ssl/certs...
133 added, 0 removed; done.
Setting up cron (3.0pl1-128.1ubuntu1) ...
Adding group `crontab' (GID 101) ...
Done.
update
```

scripts/docker/build-debian7/Dockerfile.deps
```diff
@@ -1,5 +1,8 @@
 FROM debian:wheezy
 
+# Fix debconf warnings upon build
```
mcnewton: No need to comment what this is, just add the ARG (next to the other ARGs there if there are some).

It's obvious what it does (and should have been there before :-) thanks.)

jpereira: OK, let me update the patch.

jpereira: Done, just refresh the commit @mcnewton

mcnewton: Ta

PR #2489, opened 2019-02-15
Fix reconnect correctly in rlm_sql_mysql

jpereira:


jpereira:
The expected behavior to set the reconnect never will be reached because `MYSQL_OPT_RECONNECT` is an enum. 

p.s: That option was added >= 5.0.13 as can be seen here https://sourceforge.net/p/mysql-python/feature-requests/9

PR #2490, opened 2019-02-15
Expose more TLS options in MySQL driver ref #2481

jpereira:
* Availability of TLS and reconnect options in MySQL driver
checked by using MySQL or MariaDB version.


jpereira:
I did several tests checking the build with MariaDB vs MySQL libraries and I fixed it based on #2481 and that verifies involving mariadb/mysql as discussed in https://github.com/perl5-dbi/DBD-mysql/issues/232

Also, I did some tests about the functionality that have worked fine.

PR #2496, opened 2019-02-16
tls: Fix memory leak

jpereira:


alandekok:
The `ssn` is talloc'd, and `request` is parented from that.  So when the `ssn` is freed, so is the `request`.

This patch shouldn't be necessary.

PR #2498, opened 2019-02-17
rlm_isc_dhcp: Fix the rcode in mod_instantiate

jpereira:


alandekok:
I've just added some more functionality, so I'll fix `mod_instantiate()` in another commit.

PR #2503, opened 2019-02-20
docker: Install the package 'ssl-cert'

jpereira:


mcnewton:
Wondering what the Dockerfiles need ssl-cert for?

jpereira:
@mcnewton I got some dependency errors during *.deb installation. 

```
# dpkg -i *.deb
......
make: Leaving directory '/etc/freeradius/certs'
dpkg: dependency problems prevent configuration of freeradius:
 freeradius depends on ssl-cert; however:
  Package ssl-cert is not installed.

dpkg: error processing package freeradius (--install):
 dependency problems - leaving unconfigured
Setting up libfreeradius3 (3.0.18+git) ...

```

mcnewton:
I thought that should get picked up automatically from the package dependencies? There shouldn't be a need to install them manually.

If not, as dpkg doesn't pull from the repos, you often have to `apt-get -f install` after a `dpkg -i ...` to get additional dependencies installed. Recent systems you can just do `apt install ./package.deb` and it'll install the dependencies for you from the repos.

jpereira:
The point is just to avoid the `apt-get -f install` to fix the missing packages.

mcnewton:
But installing this one package manually means that the next time a package is added to the FR dependencies in `debian/control`, then that needs to be added to the Dockerfiles as well.

What happens e.g. if you `dpkg -i freeradius-mysql` or some of the other library packages? I'm guessing that at least some of them try and pull in external dependencies - and this is unpredictable and they shouldn't all be added to the Dockerfile.

I'm not fussed about installing `ssl-cert` in the Dockerfile particularly (it's small), just that it feels the wrong solution generally.

FYI the solution in the v3 Dockerfiles is to just do

    apt-get install -y /location/to/whatever.deb

and that installs dependencies for you. That's supported since Debian 8, and we've pretty much dropped Debian 7, so that feels like the cleanest way and saves trying to pre-guess the required dependencies.

mcnewton:
Actually since Debian 9 - see https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/scripts/docker/debian9/Dockerfile#L48

arr2036:
Agree with @mcnewton

PR #2504, opened 2019-02-20
raddb/sites-available/default: Suppres warning message

jpereira:


arr2036:
What to the rest of the policies? :) We now enforce this in v4.0.x so it'd be good if it was done everywhere in the v3.0.x config.

PR #2505, opened 2019-02-20
Fix warn related to limited range of data type

jpereira:


jpereira:
It will fix the such warn.

```
src/lib/tls/session.c: In function 'session_msg_log':
src/lib/tls/session.c:671:50: warning: comparison is always false due to limited range of data type [-Wtype-limits]
    if (((size_t)tls_session->info.handshake_type >= NUM_ELEMENTS(tls_handshake_type_str)) ||
                                                  ^~
```

PR #2508, opened 2019-02-21
configure.ac: Fix warn literal message

jpereira:
It will fix the message.

`configure.ac:26: warning: AC_INIT: not a literal: $Id$`


arr2036:
VERSION isn't equivalent to $Id$?

PR #2510, opened 2019-02-21
rlm_isc_dhcp: Add 'pedantic' option

jpereira:


jpereira:
@alandekok It was updated.

alandekok:
I've reworked this a bit, and pushed a different patch.

- pass `rlm_isc_dhcp_t *inst` in `state`, so that we can just check it directly
- move the checks for `pedantic` to where the command is found.  That way it can complain about the command by name, and not about any arguments it's already parsed
- ERROR if `pedantic = true`
- WARN if `rad_debug_lvl && pedantic == true`

we don't want to pollute the main FR logs with tons of warnings for things the module is ignoring.

PR #2512, opened 2019-02-22
isc_dhcp: Add all ISC supported options as stub

jpereira:


jpereira:
The approach was creating that code here https://gist.github.com/jpereira/ab05d733afc519ae929286d6f270e5f8 with all fields extracted from the dhcpd isc source code and convert to our current format.

PR #2516, opened 2019-02-23
rlm_isc_dhcp: Update 'dhcpd.conf' with all ISC options

jpereira:


alandekok:
I think it's best to close this, and instead to put the list of allowed commands into the `doc` directory.  Maybe `doc/howto/modules/isc_dhcp`

PR #2520, opened 2019-02-25
rlm_isc_dhcp: Less code

jpereira:


alandekok:
I'd prefer to leave it the way it is... it's not laid out *quite* the same as the rest of the server.  But it's useful to keep the functions operating in a consistent manner.  All of the `parse_option_foo()` functions should operate in the same way.

PR #2527, opened 2019-02-26
xlat: move '%{integer: ...} to async

jpereira:


arr2036:
I'm not really sure what you were going for here, but the code doesn't make sense and the many of the integer tests now fail.

src/lib/server/xlat_func.c
```diff
@@ -227,100 +227,34 @@ static ssize_t xlat_length(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
...
-	case FR_TYPE_IPV6_ADDR:
-	case FR_TYPE_IPV6_PREFIX:
-		return fr_snprint_uint128(*out, outlen, ntohlll(*(uint128_t const *) &vp->vp_ipv6addr));
+	var = (*in)->vb_strvalue;
+	if ((xlat_fmt_get_vp(&vp, request, var) < 0) || !vp) return XLAT_ACTION_FAIL;
 
-	default:
-		break;
+	MEM(vb = fr_value_box_alloc(ctx, vp->vp_type, NULL, false));
+	if (fr_value_box_cast(ctx, vb, vp->vp_type, NULL, &vp->data) < 0) {
```
arr2036: You appear to be casting the data to the same type as it was previously, which is just a copy operation, and then inserting a copy of the VP's data into the output list.  That's not what this function does... It's meant to return the integer representation of the data you passed in.

PR #2528, opened 2019-02-26
xlat: move '%{length: ...} to async

jpereira:


src/lib/server/xlat_func.c
```diff
@@ -210,18 +210,41 @@ static ssize_t xlat_strlen(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
...
+		if ((xlat_fmt_get_vp(&vp, request, var) < 0) || !vp) {
+			talloc_free(vb);
+			return XLAT_ACTION_FAIL;
+		}
+
+		vb->vb_size = fr_value_box_network_length(&vp->data);
+	}
+
+	fr_cursor_insert(out, vb);
+	fr_cursor_append(out, vb);
```
alandekok: not sure we need both "insert" and "append" here.

jpereira: @alandekok It was a typo. I pushed the `fixup`

jpereira: @alandekok if everything is ok, let me know to rebase/update the pr.

PR #2530, opened 2019-02-27
Add fr_skip_spaces() helper to strip whitespace

jpereira:


jpereira:
It will be helpful to avoid several: `while (isspace((int) *p)) p++`

PR #2534, opened 2019-02-27
fr_skip_spaces() Check the value

jpereira:


alandekok:
we don't need this.  `isspace()` returns `false` for zero bytes.

PR #2537, opened 2019-02-27
xlat: move '%{explode: ...}' to async

jpereira:


arr2036:
As per slack messages the signature for this xlat should now be %{explode:<delim> <input>[< input>...]}, same with lpad and rpad.

PR #2539, opened 2019-02-28
xlat: Move lpad/rpad to async

jpereira:


jpereira:
It should be '%{foo: input1 inputN ...}

src/tests/keywords/pad
```diff
@@ -9,35 +9,35 @@ update request {
 #
 #	rpad tests
 #
-if ("%{rpad:&Tmp-String-0 7}" != "test   ") {
+if ("%{rpad: 7 %{Tmp-String-0}}" != "test   ") {
```
mcnewton: Why change the argument order? Padding with spaces like this looks really weird (and to be technically correct, it should have two space at the start - the pad and the separator - which is even weirder).

arr2036:
Just ignore it for now.  We need grouped xlat arguments first and probably an alternative syntax

arr2036:
Will be addressed in https://trello.com/c/wHHX6MsF

PR #2542, opened 2019-03-01
value.c: Add missing 'vb_ptr'

jpereira:


arr2036:
vb_ptr isn't meant to be a public field which is why it doesn't have an accessor function

PR #2543, opened 2019-03-01
Fix the xlat_exp_alloc()

jpereira:


jpereira:
It is necessary to fix the `master`

arr2036:
Got them already, thanks 👍

PR #2546, opened 2019-03-04
Fix radwho dict calls

jpereira:


alandekok:
already merged in another commit...

PR #2550, opened 2019-03-05
Avoid empty 'dir_name'

jpereira:


alandekok:
I'll add some `CC_HINT(nonnull)` which will catch this at the compilation stage.

PR #2555, opened 2019-03-05
binary unit tests

jpereira:


src/lib/util/trie.c
```diff
@@ -2870,7 +2870,7 @@ int main(int argc, char **argv)
 
 	if (argc < 2) {
 		fprintf(stderr, "Please specify filename\n");
-		exit(1);
+		exit(EXIT_SUCCESS);
```
alandekok: That's an exit failure.

jpereira: let me update it @alandekok

PR #2556, opened 2019-03-05
The 'make test' should call 'tests.bin' first

jpereira:


jpereira:
first, check the binaries. therefore continue the tests.

PR #2559, opened 2019-03-08
Add option '-r' in all unit_test_*:

jpereira:


src/lib/util/misc.c
```diff
@@ -1185,3 +1185,23 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
...
+
+/** Create an empty file. (It will overwrite an existing file).
+ *
+ * @param filename path to file.
+ * @param mode Specifies the file mode bits be applied.
+ */
+int fr_file_touch(char const *filename, mode_t mode) {
+	int fd;
+
+	fd = open(filename, O_WRONLY | O_CREAT | O_TRUNC, mode);
```
arr2036: Touch doesn't truncate files

src/lib/util/misc.h
```diff
@@ -171,6 +171,14 @@ int8_t		fr_pointer_cmp(void const *a, void const *b);
 void		fr_quick_sort(void const *to_sort[], int min_idx, int max_idx, fr_cmp_t cmp);
 int		fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length) CC_HINT(nonnull);
 
+int 		fr_file_touch(char const *filename, mode_t mode);
+
+/**  Remove a name from the filesystem.
+ */
+static inline int fr_file_remove(char const *filename) {
+	return remove(filename);
```
arr2036: We generally use unlink not remove, and this should set fr_strerror otherwise it's pretty pointless wrapping the function

arr2036: Should probably also check that it's a file and not a directory

src/lib/util/misc.c
```diff
@@ -1185,3 +1187,40 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
 
 	return result;		/* 0 is OK, !0 is !OK, just like memcmp */
 }
+
+/** Create an empty file. (It will overwrite an existing file).
```
arr2036: Doesn't actually overwrite the existing file

src/lib/util/misc.c
```diff
@@ -1185,3 +1187,40 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
...
+ *
+ * @param filename path to file.
+ * @param mode Specifies the file mode bits be applied.
+ */
+int fr_file_touch(char const *filename, mode_t mode) {
+	int fd;
+
+	fd = open(filename, O_WRONLY | O_CREAT, mode);
+	if (fd < 0) {
+		fr_strerror_printf("fr_file_touch() Failed creating file \"%s\": %s", filename, fr_syserror(errno));
```
arr2036: Don't really need the function name here, it tends to look messy when they're used in common utility functions

src/lib/util/misc.c
```diff
@@ -1185,3 +1187,40 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
 
 	return result;		/* 0 is OK, !0 is !OK, just like memcmp */
 }
+
+/** Create an empty file. (It will overwrite an existing file).
+ *
+ * @param filename path to file.
+ * @param mode Specifies the file mode bits be applied.
+ */
```
arr2036: Document return values

src/lib/util/misc.c
```diff
@@ -1185,3 +1187,40 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
...
+	}
+
+	close(fd);
+
+	return 0;
+}
+
+/**  Remove a name from the filesystem.
+ */
+int fr_file_unlink(char const *filename) {
```
arr2036: return

-1 on error
0 on unlink
1 if the file doesn't exist

src/lib/util/misc.c
```diff
@@ -1185,3 +1187,40 @@ int fr_digest_cmp(uint8_t const *a, uint8_t const *b, size_t length)
...
+		fr_strerror_printf("fr_file_unlink() Problems with stat(\"%s\")='%s'", filename, fr_syserror(errno));
+		return -1;		
+	}
+
+	if (!S_ISREG(path_stat.st_mode)) {
+		fr_strerror_printf("fr_file_unlink() only remove regular file: \"%s\": %s", filename, fr_syserror(errno));
+		return -1;
+	}
+
+	return remove(filename);
```
arr2036: Need to check the return code here and call fr_strerro_printf with fr_syserror.  The reason why we use fr_syserror is it's thread safe, and it prints the name of the error e.g. EINVAL or ENOENT in addition to the syserror text.

arr2036: This should also be ``unlink`` which is the POSIX function.

src/tests/map/all.mk
```diff
@@ -19,11 +19,17 @@ $(MAP_OUTPUT): $(MAP_UNIT_BIN) | $(BUILD_DIR)/tests/map/
 #
 $(BUILD_DIR)/tests/map/%.out: $(top_srcdir)/src/tests/map/%
 	${Q}echo MAP_TEST $(notdir $<)
-	${Q}if ! $(MAP_UNIT) -d $(top_srcdir)/raddb -D $(top_srcdir)/share/dictionary $< > $@ 2>&1; then \
+	${Q}rm -f "$@.ok"
```
arr2036: You don't need to remove the .ok files as this will only be invoked in the .ok file doesn't exist

src/tests/map/all.mk
```diff
@@ -19,11 +19,17 @@ $(MAP_OUTPUT): $(MAP_UNIT_BIN) | $(BUILD_DIR)/tests/map/
 #
 $(BUILD_DIR)/tests/map/%.out: $(top_srcdir)/src/tests/map/%
```
arr2036: The target should be changed to ``$(BUILD_DIR)/tests/map/%.ok``.  The build product is the 'ok' files, not the log output.

src/tests/map/all.mk
```diff
@@ -19,11 +19,17 @@ $(MAP_OUTPUT): $(MAP_UNIT_BIN) | $(BUILD_DIR)/tests/map/
...
 	${Q}echo MAP_TEST $(notdir $<)
-	${Q}if ! $(MAP_UNIT) -d $(top_srcdir)/raddb -D $(top_srcdir)/share/dictionary $< > $@ 2>&1; then \
+	${Q}rm -f "$@.ok"
+	${Q}if ! $(MAP_UNIT) -d $(top_srcdir)/raddb -D $(top_srcdir)/share/dictionary -r "$@.ok" $< > $@ 2>&1; then \
+		if [ ! -f "$@.ok" ]; then \
```
arr2036: This pretty much entirely negates the point of writing out the 'ok' files.  You need to add this to the test at the top i.e.

```
${Q}if ! $(MAP_UNIT) -d $(top_srcdir)/raddb -D $(top_srcdir)/share/dictionary -r "$@.ok" $< > $@ 2>&1 || [ ! -f "$@.ok" ]
```

src/bin/unit_test_map.c
```diff
@@ -69,6 +69,7 @@ static void NEVER_RETURNS usage(void)
 	fprintf(stderr, "  -O <output_dir>	  Set output directory\n");
 	fprintf(stderr, "  -x                     Debugging mode.\n");
 	fprintf(stderr, "  -M                     Show program version information.\n");
+	fprintf(stderr, "  -r <exit_flag>         Create the <exit_flag> as a 'success' exit.\n");
```
arr2036: Exit flag isn't really descriptive. Maybe ``receipt_file`` (for 'r')

src/bin/unit_test_map.c
```diff
@@ -165,7 +167,7 @@ int main(int argc, char *argv[])
 	}
 #endif
```
arr2036: If we're going to be producing an 'ok' file, you need to unlink it at the start of the utility's execution.  If the file doesn't exist, the resulting error should be ignored.

Removing the file at the start, means if the utility is manually executed and then fails, make test will run the test again.

src/tests/map/all.mk
arr2036: Make clean should be changed to remove the .ok files in addition to the log files.

src/tests/map/all.mk
arr2036: The example command that's written out on test failure also needs to be updated with the -r flag

src/lib/util/misc.c
```diff
@@ -36,6 +36,8 @@ RCSID("$Id$")
 #include <string.h>
 #include <sys/file.h>
 #include <sys/uio.h>
+#include <sys/types.h>
```
arr2036: Generally the include files are listed in alphabetical order.  So it should go sys/file.h sys/types.h and sys/uio.h

src/bin/unit_test_map.c
```diff
@@ -252,5 +258,7 @@ int main(int argc, char *argv[])
 
 	fr_strerror_free();
 
+	if (exit_flag) fr_file_touch(exit_flag, 0644);
```
arr2036: Need to check ``rcode == EXIT_SUCCESS`` as well

src/tests/map/all.mk
jpereira: If I don't remove the *.ok and the $(MAP_UNIT) fails, then it will rise a false positive.

alandekok: The solution here is to update `unit_test_map` and `unit_test_attribute` to write those files before they exit.  I'll take a look at it.

arr2036: See below for the comment on having those utilities remove the ok files when they start

jpereira:
@arr2036 I updated the pr following your reviews.

p.s.: Due to some tests expecting "errors" to do some checks. e.g: `src/tests/map/all.mk`, then I added an extra option `-R` used for force the `<receipt_file>` even `if (rcode != EXIT_SUCCESS)`

arr2036:
Manually merged with edits

PR #2565, opened 2019-03-09
rlm_lua: Print the version

jpereira:


jpereira:
Add simple plumbing showing the detected Lua version/lib/headers.

PR #2569, opened 2019-03-12
rlm_lua: Add capability to return

jpereira:


jpereira:
It is necessary to perform the new tests for rlm_lua. The constants are following the same idea as implemented in rlm_python.

src/modules/rlm_lua/lua.c
```diff
@@ -39,6 +39,29 @@ RCSID("$Id$")
...
+	A(RLM_MODULE_REJECT)
+	A(RLM_MODULE_FAIL)
+	A(RLM_MODULE_OK)
+	A(RLM_MODULE_HANDLED)
+	A(RLM_MODULE_INVALID)
+	A(RLM_MODULE_USERLOCK)
+	A(RLM_MODULE_NOTFOUND)
+	A(RLM_MODULE_NOOP)
+	A(RLM_MODULE_UPDATED)
+	A(RLM_MODULE_NUMCODES)
```
arr2036: - NUMCODES is not a valid constant. 
- The table they should be added to is 'fr', like the other helper functions.
- There's no point in using the full constant name.  Lua developers will have no internal knowledge of how FreeRADIUS works internally.  You should use the ``rcode_table``, and fr_str2int/fr_int2str.

The end result is that they should be able to return ``fr.ok``, ``fr.fail`` etc.. which is significantly less typing.

src/modules/rlm_lua/lua.c
```diff
@@ -471,6 +494,32 @@ static int _lua_list_iterator_init(lua_State *L)
  *
  *
  */
+
+static void rlm_lua_set_radiusd_constants(lua_State *L, char const *name)
+{
+	const radiusd_constants_t *p = radiusd_constants;
+
+	lua_newtable(L);
```
arr2036: Should take the name of the existing API table we use for iterators and other things, and insert into that instead. For robustness it should first check if the table exists, and then if it doesn't create it.

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +838,32 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
...
 		REDEBUG("Call to %s failed: %s", funcname, msg ? msg : "unknown error");
 		goto error;
 	}
 
-	return 0;
+	
+	/*
+	 *	the functions should return
+	 */
+	if (lua_isnil(L, -1)) goto error;
```
arr2036: Nil return is fine, it should default to RLM_MODULE_OK.

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +838,32 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
...
+
+	ret = lua_tointeger(L, -1);
```
arr2036: We don't know it's an integer yet.  We need to check the type first, and error out if it's not what we expect.  If you wanted to be nice and found the return value was a string, you could just look it up in the rcode_table and use that if it matched an entry.

e.g. return "ok"

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +838,32 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
...
+	lua_pop(L, 1);
+
+	/*
+	 *	... and should be a valid "radiusd.*" constant
+	 */
+	ret_str = rlm_lua_constants2str(ret);
```
arr2036: fr_int2str

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +838,32 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
...
+	if (!ret_str) {
+		RDEBUG2("The called %s() returned invalid value", funcname);
+		goto error;
+	}
+
+	RDEBUG2("The called %s() returned \"%s\"", funcname, ret_str);
```
arr2036: Not needed.  We already print the return code of each module call.

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +838,32 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
...
+
+	ret = lua_tointeger(L, -1);
+	lua_pop(L, 1);
+
+	/*
+	 *	... and should be a valid "radiusd.*" constant
+	 */
+	ret_str = rlm_lua_constants2str(ret);
+	if (!ret_str) {
+		RDEBUG2("The called %s() returned invalid value", funcname);
```
arr2036: Wrong logging macro.  It should be ``REDEBUG``.  Maybe something like ``Lua function \"%s\" returned invalid rcode \"%s\"`` and then print what was returned.

src/modules/rlm_lua/lua.c
```diff
@@ -471,6 +494,32 @@ static int _lua_list_iterator_init(lua_State *L)
...
+		lua_pushstring(L, p->name);
+		lua_pushnumber(L, p->value);
+		lua_rawset(L, -3);
+	}
+
+	lua_setglobal (L, name);
+}
+
+
+static char const *rlm_lua_constants2str(int value)
```
arr2036: Not needed, use fr_int2str.

jpereira:
@arr2036 done.

About the sigsegv. if you try a broken *.lua code like

```
function authorize()
	if true then
end
--- Missing the "end"
```

The module crash as can be seen below.
```
Tue Mar 12 21:16:26 2019 : Error : rlm_lua (lmod1): Failed loading file: src/tests/modules/lua/mod1.lua:4: 'end' expected (to close 'function' at line 1) near '<eof>'
Tue Mar 12 21:16:26 2019 : Error : src/tests/modules/lua//module.conf[1]: Instantiation failed for module "lmod1"
Backtrace of last 14 frames:
./build/bin/local/unit_test_module[0x49ef31]
./build/lib/local/.libs/libfreeradius-util.so(fr_fault+0x246)[0x7efe57284926]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7efe5586b890]
/usr/lib/x86_64-linux-gnu/libluajit-5.1.so.2(lua_close+0xa)[0x7efe4f431eda]
./build/lib/local/.libs/rlm_lua.so(+0xa4f5)[0x7efe4f6644f5]
./build/lib/local/.libs/libfreeradius-server.so(+0x122f44)[0x7efe57e83f44]
/usr/lib/x86_64-linux-gnu/libtalloc.so.2(talloc_free_children+0x99b)[0x7efe569479db]
./build/lib/local/.libs/libfreeradius-server.so(+0x15da54)[0x7efe57ebea54]
/usr/lib/x86_64-linux-gnu/libtalloc.so.2(+0xb1b4)[0x7efe569451b4]
/usr/lib/x86_64-linux-gnu/libtalloc.so.2(_talloc_free+0xd8)[0x7efe5693d8c8]
./build/lib/local/.libs/libfreeradius-server.so(modules_free+0x2e)[0x7efe57eba71e]
./build/bin/local/unit_test_module(main+0x330f)[0x51803f]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7efe54a52b97]
./build/bin/local/unit_test_module(_start+0x2a)[0x41c96a]
No panic action set
_EXIT(139) CALLED src/lib/util/debug.c[921]0
```

src/modules/rlm_lua/lua.c
```diff
@@ -39,6 +39,19 @@ RCSID("$Id$")
...
+	{ "ok",       RLM_MODULE_OK },
+	{ "handled",  RLM_MODULE_HANDLED },
+	{ "invalid",  RLM_MODULE_INVALID },
+	{ "userlock", RLM_MODULE_USERLOCK },
+	{ "notfound", RLM_MODULE_NOTFOUND },
+	{ "noop",     RLM_MODULE_NOOP },
+	{ "updated",  RLM_MODULE_UPDATED },
+	{ NULL, -1 },
+};
+
```
arr2036: You don't need to duplicate the table, use the existing one...

src/modules/rlm_lua/lua.c
```diff
@@ -466,6 +479,28 @@ static int _lua_list_iterator_init(lua_State *L)
...
+	for (; this->name != NULL; this++)
+		if (!strcmp(this->name, key)) {
+			lua_pushstring(L, this->name);
+			break;
+		}
+
+	return 1;
+}
+
+static void rlm_lua_register_rcode_table(lua_State *L, char const *name)
```
arr2036: Just create a table, don't have it call a c function every time it wants to return a value, that's horribly slow. If you want to make the table read-only use a proxy table. http://lua-users.org/wiki/ReadOnlyTables

jpereira: the pr was updated.

src/modules/rlm_lua/lua.c
```diff
@@ -466,6 +466,36 @@ static int _lua_list_iterator_init(lua_State *L)
 	return 1;
 }
 
+/*
+ * 	Initialise the table "fr." with all valid return codes.
+ */
+static int _lua_rcode_table_newindex(lua_State *L)
+{
+	lua_pushstring(L, "You can't modify the table 'fr.*' (read-only)");
```
arr2036: Why is this pushing a string, it can just call the logging functions directly and produce an error??

src/modules/rlm_lua/lua.c
```diff
@@ -782,11 +817,36 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
 		goto error;
 	}
 
-	if (lua_pcall(L, 0, 0, 0) != 0) {
+	if (lua_pcall(L, 0, LUA_MULTRET, 0) != 0) {
```
arr2036: Why LUA_MULTRET now?

src/modules/rlm_lua/lua.c
```diff
@@ -768,12 +809,17 @@ int do_lua(rlm_lua_thread_t *thread, REQUEST *request, char const *funcname)
 	lua_setmetatable(L, -2);
 	lua_setglobal(L, "request");
 
+	/*
+	 *	Setup the "fr" global table. with all RLM_MODULE_* values. e.g: "fr.reject", "fr.ok", ...
+	 */
+	rlm_lua_rcode_table_init(L, "fr");
```
arr2036: This only needs to be done when the interpreter is initialised

PR #2576, opened 2019-03-17
Remove call to redisReplyReaderGetError

herwinw:
This was a backwards compitibility macro that has been replaced by
redisReaderGetError. Release v0.14.0 removed these macros.

Source: https://github.com/redis/hiredis/blob/master/CHANGELOG.md


herwinw:
And of course it's not as widespread as I thought it would be, so we need a configure file here

PR #2583, opened 2019-03-26
Add additional attributes for Meraki dictionary

eeverett6:


jpereira:
You're missing to include that new dictionary in `$src/share/dictionary`

PR #2588, opened 2019-03-28
add rpm target to Makefile

mattrose:
Add a target to automatically create RPM, similar to how "make deb" works

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,10 +27,12 @@ endif
 #  the debian packages.
 #
 ifneq "$(MAKECMDGOALS)" "deb"
+ifneq "$(MAKECMDGOALS)" "rpm"
 ifeq "$(findstring crossbuild,$(MAKECMDGOALS))" ""
 $(if $(wildcard Make.inc),,$(error Missing 'Make.inc' Run './configure [options]' and retry))
 
 include Make.inc
+endif
 endif
 endif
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -350,6 +352,17 @@ deb:
 	fakeroot debian/rules debian/control #clean
 	fakeroot dpkg-buildpackage -b -uc
 
+.PHONY: rpm
+
+rpmbuild/SOURCES/freeradius-server-$(RADIUSD_VERSION_STRING).tar.bz2: freeradius-server-$(RADIUSD_VERSION_STRING).tar.bz2
+	@for file in `awk '/^Source...:/ {print $$2}' redhat/freeradius.spec` ; do cp redhat/$$file rpmbuild/SOURCES/$$file ; done
+	@mkdir -p rpmbuild/{SOURCES,SPECS,BUILD,RPMS,SRPMS,BUILDROOT}
```
jpereira: you have to call it first then `cp ....`

jpereira: you can't call `mkdir -p {a,b,c}` in Makefile, because it never will be expanded. if you want to do a single shot command, it could be something like:

`	bash -c "mkdir -pv rpmbuild/{SOURCES,SPECS,BUILD,RPMS,SRPMS,BUILDROOT}"`

mattrose: That works fine on my mac, and on centos (which is where this is targetting), are you trying this on ubuntu, which uses dash as /bin/sh?

alandekok: There's no need to do `bash -c` in a makefile.  The above should work.  If it doesn't, you can always do:

`mkdir -p $(addprefix rpmbuild/,SOURCES SPECS BUILD RPMS SRPMS BUILDROOT)`

which is gnu make magic. :(

mattrose: @jpereira hasn't confirmed, but I suspect he's running this on ubuntu, in which case it won't work because they have a /bin/sh that is POSIX-compliant, but lacks the feature that makes this work.  I'll try the GNU make magic.

jpereira: I am running on Ubuntu.

My default shell is bash as can be seen below.
```
[root@docker-freeradius a]# echo $SHELL
/bin/bash
[root@docker-freeradius a]# 
```

But, by default the GNU make use the /bin/sh as can be seen here.

```
[root@docker-freeradius a]# make --version
GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

[root@docker-freeradius a]# cat Makefile 
all:
	@echo $(SHELL)
[root@docker-freeradius a]# make
/bin/sh
[root@docker-freeradius a]#
```

jpereira: but, the @alandekok suggestions is very good. probably it is the best choice.

jpereira: btw, @alandekok what do you think to add new tests validating the *.rpm/*.deb generation? Maybe running on Jenkins.

PR #2597, opened 2019-04-02
Fix the make 'make rpm'

jpereira:
The 'make rpm' should be friendly and resolve all dependencies first.


mattrose:
I don't think the Makefile should alter the system it's building.  I left this out on purpose.

arr2036:
I think doing it interactively would be ok, but definitely not implicitly

alandekok:
Installing extra packages shouldn't be part of the `make` process.  It's better to check if the packages have been installed.  And maybe tell the admin what to do.

jpereira:
when we call `make deb` it calls `fakeroot debian/rules debian/control` through the `Makefile` that basically does the same as `yum-builddep -y redhat/freeradius.spec`. or not? :confused:

mcnewton:
No, `debian/rules debian/control` makes `debian/control` from `debian/control.in`. `yum-builddep` actually installs dependent pacakges

PR #2607, opened 2019-04-10
Add IPv6 listen and clients for ABFAB

alejandro-perez:


raddb/sites-available/abfab-tls
```diff
@@ -40,6 +40,43 @@ listen {
 	clients = radsec-abfab
 }
 
+# There needs to be a separated "listen" section for IPv6.
+# Typically it will be identical to the IPv4 one above, but there might be
+# some differences (e.g. if a different certificate or port is desired)
+listen {
+	ipaddr = ::
```
jpereira: it should be: `ipv6addr`

raddb/sites-available/abfab-tls
```diff
@@ -52,6 +89,11 @@ clients radsec-abfab {
 		proto = tls
 	}
 
+        client default_ip6 {
+		ipaddr = ::/0
```
jpereira: `ipv6addr` instead of `ipaddr`

right @alandekok ?

raddb/sites-available/abfab-tls
alejandro-perez: Well, according to the documentation, ipaddr works (indeed, I've tested this configuration). A different thing is whether using `ipv6addr` is preferred. In that case I guess a similar case for `ipv4addr` is recommended

raddb/sites-available/abfab-tls
arr2036: No, ipaddr works.  It figures out the type of IP address from the formatting.

The only time you really need to use ipv4addr or ipv6addr is when you want to control whether the A or AAAA is used as the result of DNS resolution.

Issue #2194, opened 2018-03-12
radclient: Add the -i option to set RADIUS IDs again


mcnewton:
# Issue type
- [x] Feature request.

# Defect/Feature description
Removed the "-i" option from radclient in ea5b99abaab because it did nothing.

We should add the the ability to specify radius IDs back in before v4 is released as it's a useful testing and debugging feature.

PR #2608, opened 2019-04-10
Fix radclient i option

bmork:
This is an attempt to fix issue #2194 

Lightly tested with radclient, but not yet with the server.

### DIFF
--- a/src/bin/radclient.c
+++ b/src/bin/radclient.c
@@ -68,6 +68,7 @@ static fr_ipaddr_t client_ipaddr;
 static uint16_t client_port = 0;
 
 static int sockfd;
+static int last_used_id = -1;
 
 static char const *proto = NULL;
 static int ipproto = IPPROTO_UDP;
@@ -169,6 +170,7 @@ static void NEVER_RETURNS usage(void)
 	fprintf(stderr, "                         If a second file is provided, it will be used to verify responses\n");
 	fprintf(stderr, "  -F                     Print the file name, packet number and reply code.\n");
 	fprintf(stderr, "  -h                     Print usage help information.\n");
+	fprintf(stderr, "  -i <id>                Set request id to 'id'.  Values may be 0..255\n");
 	fprintf(stderr, "  -n <num>               Send N requests/s\n");
 	fprintf(stderr, "  -p <num>               Send 'num' packets from a file in parallel.\n");
 	fprintf(stderr, "  -P <proto>             Use proto (tcp or udp) for transport.\n");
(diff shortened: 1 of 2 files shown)


src/protocols/radius/list.c
```diff
@@ -590,6 +593,18 @@ bool fr_packet_list_id_alloc(fr_packet_list_t *pl, int proto,
 		 *	Otherwise, this socket is OK to use.
 		 */
 
+		/*
+		 *	An explicit ID was requested
+		 */
+
+		if (id != -1) {
+			if  ((ps->id[(id >> 3) & 0x1f] & (1 << (id & 0x07))) != 0) continue;
```
arr2036: Hmm what's the idea here?

alandekok: The IDs are stored as a bitmask, so this is checking if the ID is allocated.

src/bin/radclient.c
```diff
@@ -855,7 +855,7 @@ static int send_one_packet(rc_request_t *request)
 	/*
 	 *	Haven't sent the packet yet.  Initialize it.
 	 */
-	if (request->packet->id == -1) {
+	if (fr_packet_list_find(packet_list, request->packet) == NULL) {
```
alandekok: We should really avoid expensive lookups when a simple flag would do.  It's better to add a flag, or use a flag in `request->packet`.

src/protocols/radius/list.c
```diff
@@ -508,7 +508,10 @@ bool fr_packet_list_id_alloc(fr_packet_list_t *pl, int proto,
 	 *	and near-zero cost on free.
 	 */
 
-	id = fd = -1;
```
alandekok: There's no need to move this..

src/bin/radclient.c
bmork: OK.  I thought it didn't matter that much in radclient since the packet_list is short anyway. I'll try to make this cheaper in the next version.

src/protocols/radius/list.c
bmork: Will fix

src/bin/radclient.c
bmork: I started out adding a new "sent" boolean to the request, but noticed that "tries" already has this info.  Or is there something I'm missing now?  I've updated the branch with a new version

bmork:
ping?  I believe I addressed all the comments on this request.  Please close it, with or wthout merging as you see best - that's your privilege to decide :-)

Unless there was something I missed?

alandekok:
I think it should be OK.  I'll have to take a deeper look

If this was v3, I'd be worried about changing the server functionality, just to fix radclient.  But the proxying in "master" uses an entirely different API to allocate IDs.

bmork:
> I think it should be OK. I'll have to take a deeper look
> 
> If this was v3, I'd be worried about changing the server functionality, just to fix radclient. But the proxying in "master" uses an entirely different API to allocate IDs.

Fair enough.  I believe it should have no measurale impact on the server, but verifying this would be good.

arr2036:
@bmork can you rebase and I'll merge this.

bmork:
Hope I got the rebasing right.  Didn't really remember writing this code ;-)

PR #2614, opened 2019-04-17
Some fixes to use ./autogen.sh again 

jpereira:


alandekok:
Please don't run autogen.sh.  It does all kinds of terrible things.

i.e. it adds `ltmain.sh` and `m4/libtool.m4`, none of which are needed.

alandekok:
You should just do `make reconfig`, which will do the right thing.

PR #2615, opened 2019-04-17
detail: Print the 'fd'

jpereira:


alandekok:
What problem does this solve?  None of the other sockets print the FD.

I'd rather have a `DEBUG3` message to print this.  And then it can be in the network code, where it prints out the FD for *all* sockets.

PR #2616, opened 2019-04-17
doc: Check the dependencies

jpereira:


doc/all.mk
```diff
@@ -49,15 +49,27 @@ install.doc: $(DOCINSTALL)
 clean.doc:
 	${Q}rm -f *~ rfc/*~ examples/*~ $(ADOC_FILES) $(HTML_FILES) $(PDF_FILES)
 
+
+#
+#  Check if the pandoc exist
+#
+PANDOC_EXISTS := $(shell which pandoc 2>/dev/null)
+ifeq "$(PANDOC_EXISTS)" ""
```
mcnewton: Can't you just `ifeq "$(shell which pandoc 2>/dev/null)" ""`?

doc/all.mk
```diff
@@ -49,15 +49,27 @@ install.doc: $(DOCINSTALL)
...
+ASCIIDOCTOR_EXISTS := $(shell which asciidoctor 2>/dev/null)
+ifeq "$(ASCIIDOCTOR_EXISTS)" ""
+$(error You need to install asciidoctor)
+endif
+
 #
 #  Pandoc v2 onwards renamed --latex-engine to --pdf-engine
 #
 PANDOC_ENGINE=pdf
-ifneq ($(shell which pandoc 2> /dev/null),)
```
mcnewton: In fact, like this... same for ASCIIDOCTOR_EXISTS as well?

doc/all.mk
jpereira: but, yes. pretty good, I updated the pr.

I just copied/paste the `AUTOCONF_EXISTS` from Makefile. :)

PR #2618, opened 2019-04-18
two spaces after '#'. No spaces at the end of a line.

jpereira:
via:

perl -p -i -e 's/^# ([^ \t])/#  $1/;s/^([ \t]+)# ([^ \t])/$1#  $2/;s/[ \t]+$//' raddb/site-available/*


alandekok:
Thanks.  We should probably add that Perl script to the makefile, too.

jpereira:
@alandekok for sure, please check #2623

PR #2619, opened 2019-04-18
Add Jenkins continuous integration to FreeRADIUS

mattrose:


arr2036:
Ah and sign your commits

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,41 @@
...
+                stage("build-${platform}") {
+                    docker.build("${platform}-master-test","-f ./scripts/docker/build-${platform}/Dockerfile.deps ./scripts/docker/build-${platform}").inside {
+                        checkout([$class: 'GitSCM', branches: [[name: '*/master']], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}/build"]], submoduleCfg: [], userRemoteConfigs: [[url: 'https://github.com/mattrose/freeradius-server']]])
+                        sh "cat /etc/os-release || cat /etc/redhat-release"
+                        dir("${platform}/build") {
+                            if (platform.contains("centos")) {
+                                sh 'QA_RPATHS=0x0003 RADIUSD_VERSION_STRING=$(cat VERSION) make rpm'
+                                sh "mv rpmbuild/RPMS/x86_64/*.rpm .."
+                            } else {
+                                sh "apt-get install -y unixodbc-dev"
```
mcnewton: Is there a reason this is done manually? It should be a normal .deb dependency.

arr2036: Yes, good point.  It might be there's multiple providers I remember there being something weird with odbc and debian packaging.

mattrose: It's kind of a side-effect of how the Dockerfiles were built.  Because the dockerfiles in master try to be version-agnostic, they end up installing some other odbc package that conflicts with unixodbc, so I have to over-ride the package install from the dockerfile.

mcnewton: Maybe older versions depended on iodbc instead? But the solution there is possibly just to not install dependencies from the older versions? Well, unless v3 had it I guess.

arr2036: Prior to v4.0.x there were two database drivers, one for iodbc and one of unixodbc.  In v4 we removed iodbc.  We can likely just adjust the dependencies in the packages to not build the iodbc driver in v3 and v2?

mattrose:
> Could you add some inline documentation about what this is for and how it's integrated into the build system. Nothing huge, just some notes for people who come across it in future

Sure.  I'll add a short README.md file that just talks about how to run the file.  It's pretty simple, as long as you have jenkins with the docker plugin, you can just create a job that points to that file.

PR #2620, opened 2019-04-18
Add Jenkins continuous integration to FreeRADIUS for v3.0.x

mattrose:


scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,35 @@
...
+                stage("build-${platform}") {
+                    docker.build("${platform}-test","-f ./scripts/jenkins/${platform}/Dockerfile ./scripts/jenkins/${platform}").inside {
+                        checkout([$class: 'GitSCM', branches: [[name: '*/v3.0.x']], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}"]], submoduleCfg: [], userRemoteConfigs: [[url: 'https://github.com/mattrose/freeradius-server']]])
+                        sh "cat /etc/os-release || cat /etc/redhat-release"
+                        dir(platform) {
+                            if (platform.contains("centos")) {
+                                sh 'mkdir -p rpmbuild/{SOURCES,SPECS,BUILD,BUILDROOT,RPMS,SRPMS}'
+                                sh 'tar -j -c --transform "s/./freeradius-server-$(cat VERSION)/" --exclude ".git" --exclude "rpmbuild" -f rpmbuild/SOURCES/freeradius-server-$(cat VERSION).tar.bz2 .'
+                                sh 'for file in $(grep "^Source...:" redhat/freeradius.spec | cut -d" " -f2) ; do cp redhat/$file rpmbuild/SOURCES/$file ; done'
```
jpereira: What do you think about: `awk '/^Source...:/{ print $2 }' redhat/freeradius.spec`

e.g:

```
[jpereira@sugarloaf freeradius-server.git]$ awk '/^Source...:/{ print $2 }' redhat/freeradius.spec 
radiusd.service
freeradius-tmpfiles-conf
freeradius-radiusd-init
freeradius-logrotate
freeradius-pam-conf
```

scripts/jenkins/Jenkinsfile
```diff
@@ -0,0 +1,35 @@
...
+                                sh 'rpmbuild --define "_topdir $(pwd)/rpmbuild" -bb redhat/freeradius.spec'
+                            } else {
+                                sh "make deb"
```
jpereira: It could be good to include an extra check for Debian base then an error message for `unknown` OS.

```
...
if (platform.contains("centos") || platform.contains("redhat")) {
   sh "yum-builddep -C --assumeyes redhat/freeradius.spec"
   sh "make rpm"
else if (platform.contains("debian") || platform.contains("ubuntu") ) {
   sh "make deb"
} else {
   sh "echo 'Unknown OS'; exit 1;"
}
...
```

scripts/jenkins/centos6/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ARG release=v3.0.x
+
+RUN git clone --depth 1 --branch ${release} ${source}
+WORKDIR freeradius-server
+#
+#  Other requirements
+#
+
+# Use LTB's openldap packages intead of the distribution version to avoid linking against NSS
+RUN echo $'[ltb-project]\n\
```
jpereira: That file creation could be changed to provide that file into the filesystem and just use the command `COPY`

e.g:

```
COPY ltb-project.repo /etc/yum.repos.d/ltb-project.repo
```

ps: you should create that file into the same `Dockerfile` path.

scripts/jenkins/centos7/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+ARG release=v3.0.x
+
+RUN git clone --depth 1 --single-branch --branch ${release} ${source}
+WORKDIR freeradius-server
+#
+#  Other requirements
+#
+
+# Use LTB's openldap packages intead of the distribution version to avoid linking against NSS
+RUN echo $'[ltb-project]\n\
```
jpereira: Can just reuse the previous idea about provide/reuse the file from the filesystem.

scripts/jenkins/Jenkinsfile
mattrose: That is not a bad idea.  Maybe I'll do another PR with that.

PR #2622, opened 2019-04-20
doc: Move pandoc/asciidoctor checks to configure.ac

jpereira:


alandekok:
I've pushed some changes in commit  3149f8.  I re-ran "configure" locally, so there are fewer changes to it.

jpereira:
got it. thanks @alandekok

PR #2629, opened 2019-04-22
doc: Fix sql table

jpereira:


jpereira:
Fixing the missing "==="

image

Now is fine.

image

PR #2632, opened 2019-04-22
Fix listen.ipaddr to reference an IPv6-only host #2627

jpereira:
Backport from v3.0.x https://github.com/FreeRADIUS/freeradius-server/pull/2627


cipherboy:
Thanks for this! I had looked briefly and found `misc.c` completely refactored and I didn't see a `fr_pton` function.

PR #2636, opened 2019-04-26
ERX-Acct-Request-Reason is integer. Closes #2635

### LINKED ISSUE #2635: ERX-Acct-Request-Reason field type?
I'm writing a parser for the RADIUS dictionary files, and observed that the `ERX-Acct-Request-Reason` attribute (4874.210) is defined as `octets` but also appears to have defined values.

This seems inconsistent - surely the field should have one of the integer field types?  Without knowing the field width there'd be no way to match a value.

jpereira:


alandekok:
already done.

PR #2638, opened 2019-05-01
exfile.c: Fix missing talloc_free()

jpereira:


arr2036:
Ah good catch, can you do a similar one for v4?

jpereira:
I am doing that,

PR #2642, opened 2019-05-02
exfile: Fix the file permissions

jpereira:
Sometimes the file permissions are changed externally,
just be sure to reset the permission if necessary.


jpereira:
Just backport from v3.

PR #2645, opened 2019-05-02
Show the current 'systemd' support

jpereira:


arr2036:
Can you fixup the PR, you're calling the wrong function

jpereira:
done

PR #2647, opened 2019-05-02
Warn if is running as a systemd service without proper support.

jpereira:
It could help to avoid http://lists.freeradius.org/pipermail/freeradius-users/2019-May/095311.html

### DIFF
--- a/src/main/radiusd.c
+++ b/src/main/radiusd.c
@@ -401,6 +401,18 @@ int main(int argc, char *argv[])
 			INFO("systemd watchdog is disabled");
 		}
 	}
+#else
+	/*
+	 *	Some users get frustrated due to can't handle the service using "systemctl start radiusd"
+	 *	even when the SO supports systemd. The reason is because the FreeRADIUS version was built
+	 *	without the proper support.
+	 *
+	 *	Then, as can be seen in https://www.systutorials.com/docs/linux/man/3-sd_notify/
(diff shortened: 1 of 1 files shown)


src/main/radiusd.c
```diff
@@ -401,6 +401,18 @@ int main(int argc, char *argv[])
...
+	 *	Some users get frustrated due to can't handle the service using "systemctl start radiusd"
+	 *	even when the SO supports systemd. The reason is because the FreeRADIUS version was built
+	 *	without the proper support.
+	 *
+	 *	Then, as can be seen in https://www.systutorials.com/docs/linux/man/3-sd_notify/
+	 *	We could assume that if find the NOTIFY_SOCKET, it's because we are under systemd.
+	 *
+	 */
+	if (getenv("NOTIFY_SOCKET"))
+		WARN("!!! Running as a systemd service without the proper support.");
```
arr2036: I really don't want to get this warning every time I run ``radiusd -lstdout -f -xx`` which is extremely common when debugging threading issues.

jpereira: Hi @arr2036,

But, it should appear only if the user executes "radiusd.service" without be built with the *systemd * support.

arr2036: Ah, ok.  Then sure, but maybe change the message too "Built without support for systemd watchdog, but running under systemd"

jpereira: @arr2036 done

PR #2648, opened 2019-05-02
Detect if it runs as a systemd service without the proper support

jpereira:


jpereira:
Backport from v3 as can be seen in https://github.com/FreeRADIUS/freeradius-server/pull/2647

src/bin/radiusd.c
```diff
@@ -553,6 +553,18 @@ int main(int argc, char *argv[])
...
+	 *	Some users get frustrated due to can't handle the service using "systemctl start radiusd"
+	 *	even when the SO supports systemd. The reason is because the FreeRADIUS version was built
+	 *	without the proper support.
+	 *
+	 *	Then, as can be seen in https://www.systutorials.com/docs/linux/man/3-sd_notify/
+	 *	We could assume that if find the NOTIFY_SOCKET, it's because we are under systemd.
+	 *
+	 */
+	if (getenv("NOTIFY_SOCKET"))
+		WARN("!!! Built without support for systemd watchdog, but running under systemd.");
```
arr2036: Warn makes the log output yellow, and sticks a big WARN in front of the message.  We don't need extra decoration.

jpereira: What do you prefer @arr2036 ? e.g: INFO ?

PR #2652, opened 2019-05-03
conf2adoc: Fix the '-*- text -*-' regex

jpereira:
The commit c96eda99fd1f941d7a02e8c933fc3cad202f23c7 broken that regex.


jpereira:
Now is fine again generating the appendices.

image

PR #2654, opened 2019-05-03
Document updates

jpereira:


arr2036:
Can you remove the pre-proxy and post-proxy sections from the virtual servers, they're no longer used

arr2036:
same with ``session`` if you find it anywhere

jpereira:
@arr2036 done

PR #2655, opened 2019-05-06
Support for non RFC3580 compliant Aruba AP's in rewrite_called_station_id

langemar:
This change add support for non RFC3580 compliant Aruba AP's in the rewrite_called_station_id

### DIFF
--- a/raddb/policy.d/canonicalization
+++ b/raddb/policy.d/canonicalization
@@ -85,6 +85,12 @@ rewrite_called_station_id {
 				&Called-Station-SSID := "%{8}"
 			}
 		}
+		# Support non RFC3580 compliant Aruba AP's
+		elsif (&Aruba-Essid-Name) {
+			update request {
+				&Called-Station-SSID := &Aruba-Essid-Name
+			}
+		}
 		updated
 	}
 	else {


arr2036:
I don't really like this.  We don't include vendor-specific hacks in any of the other default policies, and this is very easy for end users to implement themselves.

alandekok:
I'd be OK with updating some documentation to include this.  Even comments in the standard config would be OK.

But it's hard to mangle the default config so that *everyones* system checks for broken things from *every* vendor.  That's hard.

langemar:
I’m using FreeRadius for eduroam, the Aruba AP is not under my control, so I need to implement every vendor specific workaround any vendor might think of. Without ever having access to devices of these vendors.

So basically I agree, but the canonicalization is already a workaround for vendors who are unable to probably read, understand and implement the RFC.

And using the same argument you should not include vendor specific dictionary’s in the source code.

alandekok:
> I’m using FreeRadius for eduroam, the Aruba AP is not under my control, so I need to implement every vendor specific workaround any vendor might think of. Without ever having access to devices of these vendors.

That's a good point.  But not *everyone* is in that scenario.  So we should document these workarounds, but not *enable* them for everyone.

> And using the same argument you should not include vendor specific dictionary’s in the source code.

That comment is ridiculous.

The dictionaries are available, but aren't *used* unless they're needed.  Your patch changes the configuration for EVERYONE USING FREERADIUS.  It doesn't change the configuration for people who need it.

langemar:
Yes it's ridiculous, but the dictionaries are vendor specific hacks and Arron said: 
> I don't really like this. **We don't include vendor-specific hacks** in any of the other default policies, and this is very easy for end users to implement themselves.

As the dict's are used if needed, the code is dead until the branch is entered because the calling station did not provide the SSID in an 'nearly compliant way' and the policy is not active by default (at least not in any preconfigured sites in the repo). So yes this PR change the default behavior of this function, not the configuration for every FR installation.
But yes it changes the configuration for many's if not most installations.

May this is more what people expect when calling this function as Called-Station-SSID is not RFC compliant. To be compliant it should be:
```
if ("%{8}") {
    &Called-Station-Id := "%{toupper:%{1}-%{2}-%{3}-%{4}-%{5}-%{6}}:%{8}"
} else {
    &Called-Station-Id := "%{toupper:%{1}-%{2}-%{3}-%{4}-%{5}-%{6}}"
}
```

It might be good to evaluate if the NAS-Port-Type is 802.11 to not change the behavior for non-wifi requests. Could this be a solution you could accept?

Adding such hacks would make it easier to use FR in any environment whether in an roaming situation or not. SA's could more easily replace hardware or use hardware from multiple vendors.

alandekok:
> Yes it's ridiculous, but the dictionaries are vendor specific hacks 

No, they're not.  If you misunderstand this point, you have major misconceptions about how RADIUS works.

> As the dict's are used if needed, 

No, that's not how the dictionaries work.

> So yes this PR change the default behavior of this function,

Which is in every default install.  And which used by every default installation.

>  not the configuration for every FR installation.

Yes, a change to the default configuration does change every system that uses the default configuration.

> Adding such hacks would make it easier to use FR in any environment whether in an roaming situation or not. SA's could more easily replace hardware or use hardware from multiple vendors.

i.e. you entirely missed my comments saying "yes, it should be documented, but not enabled in the default config". 

So, no, we're not going to add more vendor-specific hacks to the default configuration.

As for roaming consortiums, if you receive a packet from a known upstream RADIUS proxy, then the packet is valid.  You don't really care what the SSID is.  There *might* be an SSID in the packet, but there doesn't have to be.  It could be missing, and stripped out by an intermediary proxy.

You've offered no reason why this is beneficial for many people.  You don't see that hacks like these multiply.  We are NOT going to put 10,000 "helpful" hacks into the server.  That way lies madness.

You can add this to your configuration.  There is no valid reason to have it in the default configuration.

arr2036:
> I’m using FreeRadius for eduroam, the Aruba AP is not under my control, so I need to implement every vendor specific workaround any vendor might think of. Without ever having access to devices of these vendors.

- If you're running an IdP you don't care what the SSID is.
- If you're running an SP you have control over your WLC/APs.

So your assertion that you need to accommodate all vendors is incorrect.

> And using the same argument you should not include vendor specific dictionary’s in the source code.

I disagree.

arr2036:
If you really want to help people, add a vendor page for aruba in the wiki, and document this there.

PR #2658, opened 2019-05-06
Fixes in client_read()

jpereira:


alandekok:
Fixed in master, thanks.

PR #2663, opened 2019-05-07
rlm_couchbase: remove 'read client' functionality.

jpereira:
It can now be better done via dynamic clients


arr2036:
I'd rather leave this in until there's a map added.  If you feel like adding map functionality to couchbase then that'd be useful.

jpereira:
Ok @arr2036 , I will add it.

PR #2665, opened 2019-05-07
rlm_couchbase: Add missing 'username' option

jpereira:


jpereira:
As can be seen in https://github.com/couchbaselabs/devguide-examples/blob/master/c/retrieving.cc#L37

PR #2666, opened 2019-05-07
su to radiusd user/group when rotating logs

cipherboy:
The `su` directive to `logrotate` ensures that log rotation happens under the
owner of the logs. Otherwise, `logrotate` runs as `root:root`, potentially
enabling privilege escalation if a RCE is discovered against the
FreeRADIUS daemon.

This attack avenue seems quite unlikely to me. The other alternative is making the `/var/log/radiusd` directory owned by root. It seems like the `logrotate` method is better hygiene though. Thoughts?

### DIFF
--- a/debian/freeradius.logrotate
+++ b/debian/freeradius.logrotate
@@ -9,6 +9,7 @@
 	notifempty
 
 	copytruncate
+	su freerad freerad
 }
 
 # (in order)
@@ -26,6 +27,7 @@
 	notifempty
 
 	nocreate
+	su freerad freerad
 }
 
 # There are different detail-rotating strategies you can use.  One is
@@ -45,4 +47,5 @@
 	notifempty
 
 	nocreate
+	su freerad freerad
 }
--- a/redhat/freeradius-logrotate
+++ b/redhat/freeradius-logrotate
@@ -9,6 +9,7 @@ rotate 4
 missingok
 compress
 delaycompress
+su radiusd radiusd
 
 #
 #  The main server log
--- a/scripts/logrotate/freeradius
+++ b/scripts/logrotate/freeradius
@@ -17,6 +17,7 @@
 	notifempty
 
 	copytruncate
+	su radiusd radiusd
 }
 
 # (in order)
@@ -34,6 +35,7 @@
 	notifempty
 
 	nocreate
+	su radiusd radiusd
 }
 
 # There are different detail-rotating strategies you can use.  One is
@@ -53,4 +55,5 @@
 	notifempty
 
 	nocreate
+	su radiusd radiusd
 }
--- a/suse/radiusd-logrotate
+++ b/suse/radiusd-logrotate
@@ -11,6 +11,7 @@ missingok
 compress
 delaycompress
 notifempty
+su radiusd radiusd
 
 #
 #  The main server log


arr2036:
Could you do a similar PR for master

cipherboy:
Happy to, will do. Thanks!

ret2libc:
CVE-2019-10143 has been assigned to this issue.

alandekok:
@ret2libc What security issue mandates a CVE?  I'm not sure there's a need for a CVE which has no current way to exploit it.  Are we issuing CVEs now for potential issues?

ret2libc:
@alandekok we are aware of a way to exploit this, though it requires the attacker to already have "high privileges" (that is, he needs to have access to the radiusd user)

alandekok:
@ret2libc " we are aware of a way to exploit this"

Uh... any thought that you might *tell* us about this?  The email address `security@freeradius.org` has existed for ~20 years, and is available on the public web page.  Why did you not look there?

Is it common RedHat security practice to file for a CVE, and then never tell the authors about it?

On top of that, I'm skeptical of "security" issues which require already privileged access.  If the "attacker" has access to the radiusd user, then he can run the RADIUS server, and authenticate anyone he wants.  That's a huge security problem, too. 

So maybe you need to issue another CVE saying "someone running as radiusd can run the RADIUS server"!

I'm more than a bit surprised at this process.  It's utterly bizarre to not inform the authors of an issue, and file for a CVE without giving the authors any prior notification.

We will be issuing a PGP signed vendor statement about this process, along with our opinion of the validity of the CVE.

Do you have any feedback which mitigates the disastrous process you've used here?

ret2libc:
> Uh... any thought that you might tell us about this? The email address
> security@freeradius.org has existed for ~20 years, and is available on the
> public web page. Why did you not look there?

Sorry for not using your security email first, but the issue was already public
in our bugzilla since May 1st
(https://bugzilla.redhat.com/show_bug.cgi?id=1705143), which, I guess, led
@cipherboy to submit this PR upstream. It was not discovered by Red Hat as far
as I know. Anyway, the comment of the commit clearly explain that the impact is
privilege escalation from radius user to root and given the CVE was assigned
weeks later, after the bug was public and the patch was already acknowledged and
accepted upstream, I did not think an extra email was required.

> @ret2libc " we are aware of a way to exploit this"

For "a way to exploit this", you can see the reproducer in the first comment of
https://bugzilla.redhat.com/show_bug.cgi?id=1705143, which confirms the problem
is real.

> On top of that, I'm skeptical of "security" issues which require already
> privileged access. If the "attacker" has access to the radiusd user, then he
> can run the RADIUS server, and authenticate anyone he wants. That's a huge
> security problem, too.

As you said, the flaw has limited impact, since it requires the attacker to
already have control of the radiusd user, but it let him escalate his privileges
to root after having compromised the freeradius server. It is true that radiusd
user compromise is already not good, but we still believe that privilege
escalation from radius user to root has a security impact, which deserves a CVE.

> Is it common RedHat security practice to file for a CVE, and then never tell
> the authors about it?

We do have very careful security response processes in place, but I agree in
this particular instance we made a mistake and we could have handled the issue
better. We will try to work more closely with you from now on.

PR #2670, opened 2019-05-08
Fix dynamic submodule load

jpereira:
It fixes the use of 'dynamic_clients = true' with/without the properly 'new client{...}' section


jpereira:
Enable `dynamic_clients = true` in `protocol { ... }`

PR #2671, opened 2019-05-08
Don't generate certificates in a reproducible build

cipherboy:
This is something I've been meaning to do for a while.

[Fedora standards](https://fedoraproject.org/wiki/Packaging:Initial_Service_Setup#First-time_Service_Setup) have changed since FreeRADIUS was initially packaged. Now they want us to generate certificates [at first service start](https://bugzilla.redhat.com/show_bug.cgi?id=1672284) instead of at installation time. The default build process generates its own certificates (as part of `make install`), only for us to blow them away (to be created at RPM install time or at service start time).  

This changes the build process to not generate certificates only in reproducible environments (`--enable-reproducible-build`, as certificates generated at build time are unique and hence not reproducible).

If you'd like me to change this to gate behind a flag specifically for certificate generation (`--disable-cert-generation` or similar), I'm happy to update this. 

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,6 +84,7 @@ before_script:
 # Run CLANG analyzer if we're building with CLANG
   - if [ "${DO_BUILD}" = 'yes' -a "${COVERITY_SCAN_BRANCH}" != 1 -a ${CC} = 'clang' ]; then make -j8 scan && [ "$(find build/plist/ -name *.html)" = '' ]; fi
 script:
+  - if [ "${DO_BUILD}" = 'yes' -a "${COVERITY_SCAN_BRANCH}" != 1 -a "%{REPRODUCIBLE}" = 'yes' ]; then sh $HOME/freeradius/etc/raddb/certs/bootstrap; fi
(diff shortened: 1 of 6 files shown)


cipherboy:
Looks like this breaks CI; I'll probably keep it as a patch then.

alandekok:
The travis tests fail because they run tests with SSL and the auto-created certificates.

You could likely fix the build errors by updating the build to check for the `REPRODUCIBLE_BUILDS` flag, and then not run the relevant tests.

cipherboy:
Sure, we could also go and generate the certificates afterwards (during the test suite if given that flag) to make sure that the build artifacts are enough to run the test suite (i.e., that we have enough to generate all the relevant certs).

Is the general PR something you're interested in?

alandekok:
Yes, that works.

cipherboy:
\o/ CI passes (I swapped the default from yes -> no on ENABLE_REPRODUCIBLE_BUILDS and that seems to have done the trick).

You don't actually test reproducible builds in CI. Is this something you'd like to see in a separate PR?

alandekok:
Sure, testing reproducible builds is fine.

cipherboy:
I'll go ahead and rebase this once #2674 is merged so we can test this change in CI.

PR #2674, opened 2019-05-10
[v3.0.x] Travis reproducible

cipherboy:
Enable reproducible builds in the Travis testing matrix. Complements #2671. 

I lean towards testing reproducible builds with optional libraries (openssl); but if you think it should be tested without as well (or only without), I can do that too. IMO, once is probably enough. 


cipherboy:
That's a lot better, sorry about the noise. I'll open a single PR for v4.0.x branch once these two are merged.

PR #2678, opened 2019-05-10
rlm_couchbase: Replace 'read_clients' by 'map{...}'

jpereira:


alandekok:
Can you fix up the doxygen comments so that the doxygen build passes?

arr2036:
@jpereira the issue with these commits is that you didn't ask anyone on the team for advice on how to do this work, you just went ahead and did it the way you thought was best.  This might work for small changes, it does not work for large ones.  Between us, Alan and I have 30+ years working on FreeRADIUS, we have a good handle on the direction and what work needs to be done, and how it should be done.  This isn't an issue of competence, it's one of knowledge transfer, and please don't feel you're being singled out for this, I've raised complaints like this against a couple of other contributors too.

As I said on slack, I'm going to close this commit.  Using older deprecated APIs makes more work for anyone attempting to remove them.  The only way I'd be happy with this code being merged is if it was left in as an example, but ifdefd out so its not part of the active build.

Please work with Alan and I to create an async version of this map function.  The async map does not require couchbase be async, but it does require that the evaluation of the map is async, i.e. any xlat expansions and tmpl expansions are done on the unlang stack.

arr2036:
With the async conversion, I absolutely practice what I preach.  I do not add new code that performs synchronous evaluation, and any code I do add is fully asycnhronous, see the async_eap_aka branch.

I know the async API isn't fully complete in all places, but the only way it will become complete is if people do the hard work of understanding the requirements and implementing the necessary functions.

PR #2679, opened 2019-05-13
tmpl: Add some missing talloc_free()

jpereira:


alandekok:
nice catch.  I'll push a slightly different patch, mostly for consistency.

PR #2681, opened 2019-05-13
map_to_vp() Fix memory leak

jpereira:


arr2036:
It's just talloc_free as you're not freeing multiple VPs in a list

arr2036:
No &n, just n :)

jpereira:
done

PR #2682, opened 2019-05-13
map_to_vp() Another slow memory leak

jpereira:


arr2036:
NB: Not actually a slow memory leak.  Yes it's good to fix them, no they won't cause the server to leak memory at runtime.

PR #2685, opened 2019-05-15
Fix fr_json_object_to_value_box()

jpereira:


arr2036:
Replace
```
		if (!enumv) goto no_enumv;

		if (fr_dict_valid_name(value, len) < 0) goto no_enumv;

		/*
		 *	If an alias exists, use that value instead
		 */
		found = fr_dict_enum_by_alias(enumv, value, len);
		if (found) {
			if (fr_value_box_copy(ctx, out, found->value) < 0) return -1;
			return 0;
		}

	no_enumv:
		/*
		 *	Just copy the string to the box.
		 */
		fr_value_box_bstrndup(ctx, out, NULL, value, len, tainted);

		/*
		 * e.g: the LHS 'FreeRADIUS-Client-IP-Address' != 'string', then verify the type as the dict.
		 */
		if (enumv) {
			out->type = enumv->type;
		}
```

With a call to ``fr_value_box_from_str`` with quote char '\0'

jpereira:
@arr2036 done

jpereira:
@arr2036 done, you mean like this?

src/lib/json/json.c
```diff
@@ -50,29 +50,16 @@ int fr_json_object_to_value_box(TALLOC_CTX *ctx, fr_value_box_t *out, json_objec
 	{
 		char const	*value;
 		size_t		len;
-		fr_dict_enum_t	*found;
+		fr_type_t dst_type = enumv->type;
```
arr2036: enumv may be null.  You need to set dst_type to FR_TYPE_STRING if it is

src/lib/json/json.c
```diff
@@ -45,6 +45,9 @@
 int fr_json_object_to_value_box(TALLOC_CTX *ctx, fr_value_box_t *out, json_object *object,
 				fr_dict_attr_t const *enumv, bool tainted)
 {
+	rad_assert(out != NULL);
+	rad_assert(object != NULL);
+
```
arr2036: The preferred way is to use CC_HINT(nonnull) in the header file.  Look for existing examples ``CC_HINT(nonnull(2,3))`` would be the markup for this case

PR #2686, opened 2019-05-16
fix travis script to start services

mattrose:
Not sure why this was removed but it's necessary for tests to run


arr2036:
Not removed, never there in the first place :P

PR #2688, opened 2019-05-16
V4/couchbase add map

jpereira:


jpereira:
As suggested by @alandekok ,

I just create that PR as we have today. even we are facing issues with the `map_to_list_mod()` API.

arr2036:
This will need to be redone when nested maps are complete

PR #2698, opened 2019-05-18
Send systemd event READY=1 before daemonize

geaaru:


arr2036:
Why would we do it before daemonizing?

mcnewton:
From comments on -users, systemd doesn't always seem to be noticing the ready from processes with a different pid. I was pretty sure that there was a option set somewhere so that it should, but haven't had a look yet. So maybe either a bug or change of functionality in some particular versions of systemd, or a misunderstanding (or omission) of something on the FR side.

I'm not convinced this is the best solution until there's a reason to explain why it won't/can't work after daemonisation. Though I guess the likelihood of the fork failing is fairly minimal.

For instance, the error could be occurring because someone's messed up the unit file to tell freeradius not to daemonise, which will mean READY isn't ever sent. That's the unit file messed up, not the code that needs fixing to work around it...

arr2036:
Hm, but we send all the watchdogs from the child, so if systemd is confused about mapping between of PID to service then it would fail there...

You can specify the PID of the process you want to notify for with ``sd_pid_notify()``.  I would try that as a first step, seeing as this fix would tell systemd the process was ready way before it's actually ready, i.e. before its even bootstrapped the modules or virtual servers.

arr2036:
```
MAINPID=…
The main process ID (PID) of the service, in case the service manager did not fork off the process itself. Example: "MAINPID=4711"
```

arr2036:
Probably need to call that or something...

arr2036:
*pass that

arr2036:
Actually, looking at the code in master, this is the right idea, but the wrong location. 

The notification should be sent https://github.com/FreeRADIUS/freeradius-server/blob/1c9074f0f6c4fb887c50d6779005fb9aebac495a/src/main/radiusd.c#L480 <- there, from the parent thread, after the child has notified it that all was well.

arr2036:
So the code in master was correct, I guess this was yet another defect in the backported code *sigh*.

arr2036:
...and you probably want to use 

```
#ifdef HAVE_SYSTEMD
			/*
			 *	Tell systemd we're ready, and pass it the PID
			 *	of our child, which is the new MAINPID as
			 *	this process is about to exit...
			 */
			sd_notify(0, "READY=1\nMAINPID=%lu", (unsigned long)pid);
#endif
```

To pass it the PID of the child.

arr2036:
Pushed fixes to master, leaving this for @geaaru to fix in v3.0.x

geaaru:
Hi, just to understand, why you want to maintain sd_notify under daemonize? Systemd doesn't require it.

From `systemd.service` documentation:

```
Note that systemd assumes here that the process started by systemd will continue running until the service terminates. If the program daemonizes itself (i.e. forks), please use Type=forking instead.
```
I admit that I have few experiences with services of Type=Notify but I think that is equal to Type=simple and normally you run radiusd with `-f` option (at least currently I run it with this option) that avoids daemonizing and this means that sd_notify is never executed.

arr2036:
Your patch sends the notify message too early before the server is actually ready.  I gave a pointer to where you should call sd_notify if it needs to be called in the main process, after initialisation is complete.

I think you're correct in that we shouldn't fork, but your patch didn't actually fix the unit file to do that.

Could you test what I just committed to master branch and see if it works OK.  
Ideally you'd test starting the process with the ``-f`` flag which makes it run in the foreground (i.e. non-forking), and without the ``-f`` flag.  Theoretically the code in master should now work for both cases, as we inform systemd of the PID change.

If all that works correctly, maybe submit a similar patch for v3.0.x, i.e. something that:
- Informs systemd of the MAINPID change.
- Sends a ready notification once the main event loop starts, instead of in radiusd.c.
- Sends an exiting notification once the main event loop has exited.
- (optionally) calls radiusd with the ``-f`` flag to prevent it forking in the relevant unit files (there's multiple).

We mostly develop on Macs and use Docker (which lacks systemd) for testing Linux stuff, so it's annoying to test systemd interoperability.

arr2036:
You also need to sign your commits https://git-scm.com/book/en/v2/Git-Tools-Signing-Your-Work.  We make exceptions for small changes, but code contributions need to be signed, especially if you're going to be sending more pull requests in the future :)

geaaru:
> Your patch sends the notify message too early before the server is actually ready. I gave a pointer to where you should call sd_notify if it needs to be called in the main process, after initialisation is complete.
> 
> I think you're correct in that we shouldn't fork, but your patch didn't actually fix the unit file to do that.

Ok.

> 
> Could you test what I just committed to master branch and see if it works OK.
> Ideally you'd test starting the process with the `-f` flag which makes it run in the foreground (i.e. non-forking), and without the `-f` flag. Theoretically the code in master should now work for both cases, as we inform systemd of the PID change.

Ok. Personally, I think that could be cleaner automatically apply foreground option if NOTIFY_SOCKET is present and systemd built-in module is enabled. Just, because I don't see a reason to permit the use of daemonize with systemd. IMHO. However, I will follow master commit.

> 
> If all that works correctly, maybe submit a similar patch for v3.0.x, i.e. something that:
> 
> * Informs systemd of the MAINPID change.
> * Sends a ready notification once the main event loop starts, instead of in radiusd.c.
> * Sends an exiting notification once the main event loop has exited.
> * (optionally) calls radiusd with the `-f` flag to prevent it forking in the relevant unit files (there's multiple).

Copy that.

In addition, it could be added also RELOADING event. WDYT?

> 
> We mostly develop on macOS and use Docker (which lacks systemd) for testing Linux stuff, so it's annoying to test systemd interoperability locally.

A good reason to integrate MottainaiCI cluster and configure commit webhook for executing automatic tests. :)

Thanks for the support.

arr2036:
@geaaru any progress on this?

geaaru:
@arr2036 sorry, very few time in this period. I will try to look at this weekend.

arr2036:
Thanks :)

src/main/radiusd.c
```diff
@@ -621,6 +625,12 @@ int main(int argc, char *argv[])
 	 */
 	state = fr_state_init(NULL);
 
+#ifdef HAVE_SYSTEMD
+	if (sd_notify(0, "READY=1") <=0) {
+		WARN("Something goes wrong on send READY event to systemd");
```
arr2036: Could you change that to:

```
ret = sd_notify(0, "READY=1");
if (ret < 0) WARN("Failed notifying systemd that process is READY: %s", fr_syserror(ret)));
```

You'll have to add an ```int ret``` somewhere too.  A return code of 0 means that $NOTIFY_SOCKET wasn't set, so that'd happen any time we were running in debug mode, so we don't want to emit warning messages there...

jpereira: @arr2036 what do you think to change the STATUS also?

e.g:

```
sd_notifyf(0, "READY=1\n"
        "STATUS=Processing requests…\n"
        "MAINPID=%lu",
        (unsigned long) getpid());
```

arr2036: Sure...

jpereira: therefore, it could be something like:

```
ret = sd_notifyf(0, "READY=1\n"
            "STATUS=Processing requests…\n"
            "MAINPID=%lu",
            (unsigned long) getpid());
if (ret < 0) {
    WARN("Failed notifying systemd that process is READY: %s", fr_syserror(ret)));
}
```

geaaru: Hi, just a summary of the last changes:
* avoid WARN if NOTIFY_SOCKET is not available
* add STATUS event
* avoid daemonizing in existing .service (missing in the last changes)

@jpereira I sent MAINPID only if I daemonized the process and not with READY event. I think that could be more correct. Let me know WDYT.

arr2036:
Looks fine to me. Maybe @mcnewton or @terryburton can have a quick look too.

geaaru:
Maybe it's better fix this:
```
src/main/radiusd.c:460:12: error: declaration of ‘ret’ shadows a previous local [-Werror=shadow]
    uint8_t ret = 0;
            ^
src/main/radiusd.c:99:14: error: shadowed declaration is here [-Werror=shadow]
  int status, ret;
```

geaaru:
travis said OK now :)

arr2036:
Unfortunately Travis doesn't run Systemd, so you also have to test it works locally.  Did the mainpid stuff work as expected without -f?

geaaru:
Yes, tested already this night. :)

OT: Just to do a bit of advertisement if you have an infra you could register an additional webhook and push test to Mottainai with LXD container or Virtualbox VM and test SystemD bootstrap. :)

geaaru:
Sorry, it's needed to fix a typo with last changes. It's better to add a task in Travis with systemd enable :'(

geaaru:
Tested again locally both modes. It seems ok.

arr2036:
One final thing, can you sign your commits.

geaaru:
Yes, sorry, I drop signature in the last rebase. Fixed.

geaaru:
Just last note:

https://github.com/geaaru/geaaru_overlay/blob/master/net-dialup/freeradius/files/freeradius-3.0.19-systemd-service.patch#L27

Currently, in debian and redhat .service is missing ExecReload and this means that is not possible execute `systemctl reload freeradius` command. Is there some reason for this?

arr2036:
If we're doing -f, then no, the pidfile is not needed.  ExecReload isn't there because the server doesn't support HUP properly.

src/main/radiusd.c
```diff
@@ -621,6 +625,14 @@ int main(int argc, char *argv[])
 	 */
 	state = fr_state_init(NULL);
 
+#ifdef HAVE_SYSTEMD
+	int ret_notif;
```
arr2036: That'll cause some compilers to throw warnings about declarations only being allowed at the top of the function.  Wrap the whole lot in a pair of curly braces.

geaaru:
> ExecReload isn't there because the server doesn't support HUP properly.

Interesting, I currently use it in production for logrotate 8-| with something like this:
```
        postrotate
                systemctl reload freeradius > /dev/null
                radmin -n myinstance -e "hup log_accounting" > /dev/null
                radmin -n myinstance -e "hup log_auth" > /dev/null
        endscript
```

Are the main problems related to missing clean of [structure](https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/main/mainconfig.c#L1236)  or related to the fact that is needed to execute hup for specific modules?

(sorry for OT question)

arr2036:
It's specific modules.  It'd be more confusing for users to attempt a reload and then find the majority of the config wasn't in fact reloaded.

geaaru:
Yeah, I imagined. Thanks for the clarification.

geaaru:
So, could be merged or are there changes to do?

arr2036:
Waiting for you to remove PIDFile from the units.

geaaru:
Ah ok. Sorry. I lost a piece in the route.

geaaru:
It seems that Travis returns errors not related to my changes. It seems that Travis returns errors not related to my changes. Have you changed something on Travis stuff that must be fixed? Could be related to a new default version of clang package?

geaaru:
I found the previous task... it seems that in the previous Travis job it was used clang 5.0.0 and now with clang 7.0.0 there are a lot of things to fix.

alandekok:
I've mostly fixed Travis.  Let's see if there's anything else left to do.

alandekok:
I think that this can be pulled over. @arr2036 any objections?

arr2036:
@alandekok, nope looks good to me.  @geaaru has removed the references to the PID file (which is no longer created) from the various unit files, which was the last thing I was waiting on.

geaaru:
I spoke with my friend in Suse and also from Suse side now it's used systemd, so probably there is something to fix under `suse` directory but could be done in a separate fix.

PR #2702, opened 2019-05-18
Fix module methods

jpereira:


jpereira:
Some pieces of evidence.

1. In my rlm_cache (new feature)

```
static const module_method_names_t method_names[] = {
	{ "delete",	CF_IDENT_ANY,	mod_method_delete },
	{ "add",	CF_IDENT_ANY,	mod_method_add },
	{ "lookup",	CF_IDENT_ANY,	mod_method_lookup },
	MODULE_METHOD_NAME_TERMINATOR
};
```

2. Once in my `send Access-Accept `

```
	cache.lookup
	cache.delete
	cache.add
```

3. Debug

```
Sat May 18 18:58:11 2019: Debug : /opt/freeradius4/etc/raddb/mods-enabled/cache[35]: Calling cache.lookup
Sat May 18 18:58:11 2019: Debug : /opt/freeradius4/etc/raddb/mods-enabled/cache[35]: Calling cache.delete
Sat May 18 18:58:11 2019: Debug : /opt/freeradius4/etc/raddb/mods-enabled/cache[35]: Calling cache.add

```

src/lib/server/module.h
```diff
@@ -160,8 +160,7 @@ struct rad_module_method_names_s {
 	module_method_t	method;
 };
 
-#define MODULE_NAME_TERMINATOR { .name1 = NULL }
-
+#define MODULE_METHOD_NAME_TERMINATOR { .name1 = NULL, .name2 = NULL, .method = NULL }
```
alandekok: I think the old one is fine.  Any field not explicitly initialized is set to NULL  / 0.

src/lib/server/module.c
```diff
@@ -966,7 +966,7 @@ module_instance_t *module_by_name_and_method(module_method_t *method, rlm_compon
 	 *
 	 *	Loop over the method names, seeing if we have a match.
 	 */
-	len = q - p;
+	len = strlen(p);
```
alandekok: I don't think this is correct.  The name can be `module.method1.method2`.  So doing `strlen(p)` here means it's comparing `method1.method2`, and not just `method1`.

PR #2703, opened 2019-05-18
Fix the header conflict

jpereira:


jpereira:
```
src/lib/server/main_loop.c:58:25: error: redefinition of 'sd_watchdog_interval' with a different type: 'fr_time_delta_t' (aka 'long') vs 'struct timeval'
static fr_time_delta_t          sd_watchdog_interval;
                                ^
src/lib/server/main_loop.h:46:23: note: previous declaration is here
```

Now the build with systemd is fine.

PR #2712, opened 2019-05-23
note recent changes

jpereira:


arr2036:
That's completely nondescriptive.  **WHAT** issue was fixed with log rotation?

jpereira:
@arr2036 Now is better! :)

arr2036:
thanks!

PR #2713, opened 2019-05-24
misc.c: Fix redefinition of 'NSEC'

jpereira:


arr2036:
Yeah there's loads of them. Already fixed a few

jpereira:
It was the last one.

```
[jpereira@sugarloaf freeradius-server.git]$ grep NSEC -r src/ | grep define
src/lib/util/time.h:#define NSEC	(1000000000)
```

PR #2720, opened 2019-05-25
radclient: Update to use fr_time_delta_t

jpereira:


src/bin/radclient.c
```diff
@@ -49,7 +50,7 @@ typedef struct REQUEST REQUEST;	/* to shut up warnings about mschap.h */
 #define pair_update_request(_attr, _da) fr_pair_update_by_da(request->packet, _attr, &request->packet->vps, _da)
 
 static int retries = 3;
-static float timeout = 5;
+static fr_time_t timeout = 5;
```
arr2036: 5 nanoseconds? And it's not a wallclock time, should be fr_time_delta_t

src/bin/radclient.c
```diff
@@ -1013,10 +1014,10 @@ static int send_one_packet(rc_request_t *request)
 /*
  *	Receive one packet, maybe.
  */
-static int recv_one_packet(int wait_time)
+static int recv_one_packet(fr_time_t wtime)
 {
 	fd_set		set;
-	struct timeval	tv;
+	fr_time_t 	wait_time;
```
arr2036: the general convention is prefixing 'our_' i.e. wait_time and our_wait_time.

arr2036: and again should be an fr_time_delta_t

src/bin/radclient.c
```diff
@@ -1027,13 +1028,12 @@ static int recv_one_packet(int wait_time)
 	max_fd = fr_packet_list_fd_set(packet_list, &set);
 	if (max_fd < 0) exit(1); /* no sockets to listen on! */
 
-	tv.tv_sec = (wait_time <= 0) ? 0 : wait_time;
-	tv.tv_usec = 0;
+	wait_time = (wtime <= 0) ? 0 : fr_time_delta_from_sec(wtime);
```
arr2036: The type of wtime is a type that holds the time in nanoseconds, but you're converting it to seconds here, but wait_time is also a time delta in nanoseconds, and you convert the delta to a timeval later... so the conversion from delta to seconds isn't needed, and will result in huge waiting periods.

src/bin/radclient.c
```diff
@@ -1337,9 +1337,10 @@ int main(int argc, char **argv)
 		       break;
 
 		case 't':
-			if (!isdigit((int) *optarg))
-				usage();
-			timeout = atof(optarg);
+			if (fr_time_delta_from_str(&timeout, optarg, FR_TIME_RES_SEC) < 0) {
```
arr2036: That's fine.  Just to clear up the confusion.  Passing FR_TIME_RES_SEC, just makes the parsing function assume that unqualified values (i.e. those without s, us, ms, ns suffixes) are values in seconds.  The actual value is still provided in nanoseconds.

src/bin/radclient.c
```diff
@@ -1608,7 +1604,7 @@ int main(int argc, char **argv)
 					 *	a signal, treat it like
 					 *	a normal timeout.
 					 */
-					select(0, NULL, NULL, NULL, &tv);
+					select(0, NULL, NULL, NULL, &fr_time_delta_to_timeval(psec));
```
arr2036: This all looks ok...

PR #2721, opened 2019-05-25
radmin: Update to use fr_time_delta_t

jpereira:


src/lib/util/print.c
```diff
@@ -838,3 +838,38 @@ char *fr_asprintf(TALLOC_CTX *ctx, char const *fmt, ...)
...
+ *
+ * @param[in] ctx	to allocate buffer in.
+ * @param[in] fp	to write the result of fmt string.
+ * @param[in] fmt	string.
+ * @param[in] ...	variadic argument list.
+ * @return
+ *   - On success, the number of bytes written is returned (zero indicates nothing was written).
+ *   - On error, -1 is returned, and errno is set appropriately
+ */
+int fr_fprintf(TALLOC_CTX *ctx, FILE *fp, char const *fmt, ...)
```
arr2036: This doesn't need to take a talloc ctx.  If you want it to be efficient, you should allocate a thread specific memory pool and use that instead... But I think fr_vasprintf already does that for the intermediary buffers, so you won't save  that much CPU time.

arr2036:
Looks fine other than that one issue

jpereira:
@arr2036 done!

src/bin/radmin.c
```diff
@@ -452,16 +452,11 @@ static int cmd_terminate(UNUSED FILE *fp, UNUSED FILE *fp_err, UNUSED void *ctx,
 
 static int cmd_uptime(FILE *fp, UNUSED FILE *fp_err, UNUSED void *ctx, UNUSED fr_cmd_info_t const *info)
 {
-	struct timeval now;
+	fr_time_t now;
```
arr2036: ah and that's an fr_time_delta_t not an fr_time_t, and maybe rename it to uptime.

jpereira:
@arr2036 done

src/bin/radmin.c
```diff
@@ -452,16 +452,11 @@ static int cmd_terminate(UNUSED FILE *fp, UNUSED FILE *fp_err, UNUSED void *ctx,
...
-	now = fr_time_to_timeval(fr_time());
-	fr_timeval_subtract(&now, &now, &start_time);
+	uptime = fr_time() - start_time;
 
-	fprintf(fp, "Uptime: %u.%06u seconds\n",
-		(unsigned int) now.tv_sec,
-		(unsigned int) now.tv_usec);
-
-	return 0;
+	return fr_fprintf(fp, "Uptime: %pVs seconds\n", fr_box_time_delta(uptime));
```
arr2036: This previously returned 0, now it'll return the number of bytes written?

src/lib/util/print.c
```diff
@@ -838,3 +838,38 @@ char *fr_asprintf(TALLOC_CTX *ctx, char const *fmt, ...)
...
+ *
+ * @param[in] ctx	to allocate buffer in.
+ * @param[in] fp	to write the result of fmt string.
+ * @param[in] fmt	string.
+ * @param[in] ...	variadic argument list.
+ * @return
+ *   - On success, the number of bytes written is returned (zero indicates nothing was written).
+ *   - On error, -1 is returned, and errno is set appropriately
+ */
+int fr_fprintf(FILE *fp, char const *fmt, ...)
```
arr2036: should be a ssize_t

src/lib/util/print.c
```diff
@@ -838,3 +838,38 @@ char *fr_asprintf(TALLOC_CTX *ctx, char const *fmt, ...)
...
+	if (fd < 0) {
+		fr_strerror_printf("Invalid 'fp'");
+		return -1;
+	}
+
+	va_start(ap, fmt);
+	buf = fr_vasprintf(NULL, fmt, ap);
+	va_end(ap);
+
+	ret = (int)write(fd, buf, strlen(buf));
```
arr2036: No need to cast

PR #2727, opened 2019-05-31
rlm_cache: Fix missing 'key_len' set in memcached driver

jpereira:


arr2036:
Was that all that was broken?

jpereira:
yep! it was the reason for that memcache errors that I showed you.

PR #2728, opened 2019-06-01
doxygen: Replace <name@addres> by (name@address)

jpereira:


jpereira:
It could help with your previous commit @arr2036

Issue #2732, opened 2019-06-06
Occasional crash with 3.0.19


pauldekkers:
# Issue type

- Defect - Crash or memory corruption.

# Defect
## How to reproduce the issue

There's no trigger I'm aware of, or no special configuration. It just happens after a couple of weeks. It does happen occasionally. It happened with earlier releases too, I never got to capturing the event: but now I have a core dump :-)

(While it happened with earlier releases, somewhere above 3.0.15 I think, the exact same configuration did run without issues somewhere in the past. And while this is a busy server, the crashes are really occasional.)

I'm using the 3.0.19 docker container these days, on an Ubuntu 18.04 LTS host. (Previously it crashed with self-compiled releases and no containers too.) Looking at system graphs, I see no issues with resources (memory, disk spikes) whatsoever at the time of the crash. (It was actually at a quiet time, in the middle of the night).

This is a proxy server (it does no EAP, just RADIUS/UDP and TLS, sqlite and rlm_cache_rbtree).

## Full backtrace from LLDB or GDB

```
# gdb /usr/sbin/freeradius /tmp/core.1
GNU gdb (Ubuntu 8.1-0ubuntu3) 8.1.0.20180409-git
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
<http://www.gnu.org/software/gdb/documentation/>.
For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /usr/sbin/freeradius...Reading symbols from /usr/lib/debug/.build-id/34/960d4eec3b9f81521a5e219541668773d33c98.debug...done.
done.
[New LWP 86]
[New LWP 88]
[New LWP 1]
[New LWP 92]
[New LWP 85]
[New LWP 90]
[New LWP 84]
[New LWP 87]
[New LWP 89]
[New LWP 83]
[New LWP 91]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[...]

PR #2733, opened 2019-06-07
threads: If process() is not set, don't call it! #2732

jpereira:


jpereira:
As described in https://github.com/FreeRADIUS/freeradius-server/issues/2732, the process() is coming NULL.

```
(gdb) bt
#0  0x0000000000000000 in ?? ()
#1  0x0000560a22119575 in request_handler_thread (arg=0x560a257ccc60) at src/main/threads.c:826
#2  0x00007f65e913b6db in start_thread (arg=0x7f65b17fa700) at pthread_create.c:463
#3  0x00007f65e89ae88f in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb)
```
then, it will avoid the crash.

alandekok:
The issue is that `request->process` should *never* be `NULL`.  If it is `NULL`, then it means that the request is no longer valid.

We shouldn't really be adding work-arounds to bugs.  We should track down the cause of the issue, and fix it.

A patch like this just makes it more difficult to track down the cause of the bug.  The server will keep running, and look like it's OK.  But it won't be.  It will gradually get more and more broken.  And then fail with an even more esoteric bug.

A better solution is to add checks that the request is valid.  If it isn't valid, the server should exit with a descriptive message.

jpereira:
Actually, you're right. I got it your point Alan!👍

PR #2735, opened 2019-06-08
Prefer HAVE_EVP_SHA3_512 macro for SHA3 code blocks

herwinw:
This macro is more descriptive and readable. It also resolves an
inconsistency in xlat_func.c, where both versions were used for related
code.


herwinw:
commit 129a35db0159335e4a987ba48f512d16f3e49c5e makes this one obsolete

PR #2736, opened 2019-06-08
Register blake2 xlat functions

herwinw:
This resolves a compiler warning as well, since the methods were defined
but never used.


arr2036:
In future can you start signing your commits, thanks!

PR #2739, opened 2019-06-11
Enable rlm_cache in authenticate section

fdurand:
Enable rlm_cache in authentication section.

I use rlm_cache in authentication section to set a value based on a the first authentication status (Combining authentication for MSCHAP/PEAP requests)  and take a decision in post-auth.


arr2036:
Use the override i.e. ``cache.authorize``.  There is no authenticate method because the cache module doesn't authenticate anything.

PR #2740, opened 2019-06-11
Remove unused sd_watchdog_event()

jpereira:


jpereira:
Since we added the `sd_watchdog_enabled()`, that function is unused causing warnings.

```
src/lib/server/main_loop.c:66:13: warning: function 'sd_watchdog_event' is not needed and will not be emitted [-Wunneeded-internal-declaration]
static void sd_watchdog_event(fr_event_list_t *our_el, UNUSED fr_time_t now, void *ctx)
```

PR #2743, opened 2019-06-14
rlm_cache: Move from AVPs to methods

jpereira:
The new methods to interact with cache module are:

cache.status
cache.add
cache.lookup
cache.delete
cache.merge
cache.merge.new


arr2036:
@jpereira you were told explicitly *NOT* to remove the old mechanism, it is unacceptable to break all the existing deployments using the cache module.

PR #2748, opened 2019-06-18
eapol_test: Build the stable 2.8 version

jpereira:


jpereira:
The purpose is because today we are facing build issues due to the HEAD of hostap.git is broken. then, this will help us to build using the latest stable hostap.

PR #2749, opened 2019-06-19
xlat_eval.c: verify if it's expected type

jpereira:


jpereira:
```
Debug : (0)      EXPAND %{session-state:[#]}

Thread 1 "unit_test_modul" received signal SIGSEGV, Segmentation fault.
0x00007ffff7820fb2 in xlat_eval_pair_real (ctx=0x608000006d00, out=0x7fffffffb9f0, request=0x6140000002a0, vpt=0x611000043760)
    at src/lib/server/xlat_eval.c:591
591			if (vpt->tmpl_da->flags.virtual) return xlat_eval_pair_virtual(ctx, out, request, vpt);
(gdb) b src/lib/server/xlat_eval.c:591
Breakpoint 1 at 0x7ffff7820f16: file src/lib/server/xlat_eval.c, line 591.
(gdb)
``

arr2036:
@jpereira before submitting PRs you need to spend some time understanding what the actual issue is.  It's really easy to just monkey patch over symptoms, it's harder to understand why something is happening and come up with the correct fix.

Why would the particular vp_tmpl_t not have a dictionary attribute?  What else could we check?

arr2036:
Please use the macros you just added :p

PR #2752, opened 2019-06-21
radmin: Fix the parser

jpereira:


jpereira:
as we have talked before @alandekok , now is starting and working.

```radmin> stats network 0 self 
count.in        1
count.out       0
count.dup       0
count.dropped   0
count.sockets   6
radmin>```

File: raddb/sites-available/control-socket [hunk 1]
```diff
@@ -12,26 +17,27 @@
...
 #
 #	See also the "radmin" program, which is used to communicate
 #	with the server over the control socket.
 #
-#	$Id$
-#
 ######################################################################
 server control {
-	namespace = control
+	namespace = "control"
```
alandekok: This doesn't need quotes.  None of the other virtual servers do this.

jpereira:
done @alandekok

PR #2755, opened 2019-06-24
radmin.c: Fix typo

jpereira:


jpereira:
@alandekok Just to mute the Jenkis.

alandekok:
I've pushed a bunch more fixes.  It should now work with `radmin -d raddb -s control`.

PR #2763, opened 2019-06-28
Implement query_timeout on socket read for rlm_postgresql

terryburton:
Use the async interface provided by libpq so what can control our own blocking. In this way we are able to implement a configurable inactivity timeout since the blocking interface does not expose such a feature.

### DIFF
--- a/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
@@ -297,7 +297,7 @@ static CC_HINT(nonnull) sql_rcode_t sql_query(rlm_sql_handle_t *handle, UNUSED r
 {
 	rlm_sql_postgres_conn_t *conn = handle->conn;
 	struct timeval timeout = {config->query_timeout, 0};
-	int sockfd;
+	int sockfd, r;
 	fd_set read_fd;
 	ExecStatusType status;
 	int numfields = 0;
@@ -326,8 +326,14 @@ static CC_HINT(nonnull) sql_rcode_t sql_query(rlm_sql_handle_t *handle, UNUSED r
 	while (PQisBusy(conn->db)) {
 		FD_ZERO(&read_fd);
 		FD_SET(sockfd, &read_fd);
-		if (!select(sockfd + 1, &read_fd, NULL, NULL, config->query_timeout ? &timeout : NULL)) {
+		r = select(sockfd + 1, &read_fd, NULL, NULL, config->query_timeout ? &timeout : NULL);
+		if (r == 0) {
 			ERROR("rlm_sql_postgresql: Socket read timeout after %d seconds", config->query_timeout);
+			return RLM_SQL_RECONNECT;
+		}
+		if (r < 0) {
+			if (errno == EINTR) continue;
+			ERROR("rlm_sql_postgresql: Failed in select: %s", fr_syserror(errno));
 			return RLM_SQL_RECONNECT;
 		}
 		if (!PQconsumeInput(conn->db)) {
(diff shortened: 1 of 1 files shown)


src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
```diff
@@ -296,14 +296,46 @@ static CC_HINT(nonnull) sql_rcode_t sql_query(rlm_sql_handle_t *handle, UNUSED r
...
+	}
+
+	/*
+	 * We try to avoid blocking by waiting until the driver indicates that
+         * the result is ready or our timeout expires
+	 */
+	while (PQisBusy(conn->db)) {
+		FD_ZERO(&read_fd);
+		FD_SET(sockfd, &read_fd);
+		if (!select(sockfd + 1, &read_fd, NULL, NULL, config->query_timeout ? &timeout : NULL)) {
```
arr2036: Should probably check for -1 return code here too, mostly for EAGAIN and EINTR.  Not 100% sure how v3.0.x handles signals.  No point in making the change for v4.0.x, but just to be safe for v3.0.x

terryburton: What is the expected behaviour here for EINTR amd EAGAIN? At the moment we would stay in the loop and retry the select (unless a handler is defined) which seems pretty standard.

It might be a bit drastic to fail the request on receipt of a signal. EAGAIN is interesting as it's sometimes used as a proxy for ENOMEM, so is it best to repeat or give up...

Linux man 2 select has the following which suggests repeating in both cases, but is that best here:

> On some other UNIX systems, select() can fail with the error EAGAIN if the system fails to allocate kernel-internal resources, rather than ENOMEM as Linux does.  POSIX specifies this error for poll(2), but not for select().  Portable programs may wish to check for EAGAIN and loop, just as with EINTR.

arr2036: EINTR should definitely be repeated.  EAGAIN/ENOMEM require previously used resources being released in order to make progress otherwise we'd essentially busy loop, so we probably want to bail out of the loop in that case.

Regarding continuing in the loop body, I'm just a bit paranoid about calling PQconsumeInput where no data is available.

src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
```diff
@@ -326,10 +326,16 @@ static CC_HINT(nonnull) sql_rcode_t sql_query(rlm_sql_handle_t *handle, UNUSED r
...
 		FD_ZERO(&read_fd);
 		FD_SET(sockfd, &read_fd);
-		if (!select(sockfd + 1, &read_fd, NULL, NULL, config->query_timeout ? &timeout : NULL)) {
+		r = select(sockfd + 1, &read_fd, NULL, NULL, config->query_timeout ? &timeout : NULL);
+		if (r == 0) {
 			ERROR("rlm_sql_postgresql: Socket read timeout after %d seconds", config->query_timeout);
 			return RLM_SQL_RECONNECT;
 		}
+		if (r < 0) {
+			if (errno == EINTR) continue;
```
terryburton: Continuing on EINTR, otherwise failing seems to be the majority case elsewhere in the code.

arr2036: Yes, that's certainly what I usually do when using select.  I don't think a new release is imminent. Could pull in the changes and then ask for volunteers to test on the user's list :)

terryburton: I'll be around for the next few days, so if it breaks things then I'll own it :-)

arr2036: Ah, and you need to sign your commits.

arr2036: or at least the last commit.

terryburton: Oops. Thought I'd set that to auto on the repo. Force-pushed.

PR #2774, opened 2019-07-08
scripts/asciidoc/pandoc-filter: Add checks for JSON.pm

jpereira:


configure.ac
```diff
@@ -105,6 +106,11 @@ else
   AC_SUBST(PANDOC_ENGINE)
 fi
 
+# pandoc and asciidoctor is defined? then check it.
+if test "x$ac_cv_path_PANDOC" != "x" && test "x$ac_cv_path_ASCIIDOCTOR" != "x"; then
+  AX_PROG_PERL_MODULES(JSON, , AC_MSG_ERROR([The documentation build requires the JSON.pm module]))
```
arr2036: Not an error, just a  warning.  Probably in the same format as the other two warnings, i.e.

``AC_MSG_WARN([Perl JSON module not found - Please install if you want build the docs])``

jpereira: done

PR #2776, opened 2019-07-09
Check if we could disable -Wdocumentation for some third headers

jpereira:


jpereira:
As can be seen in https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/556075866 the clang in Linux do not like that.

src/modules/rlm_couchbase/couchbase.h
```diff
@@ -26,11 +26,13 @@
  */
 RCSIDH(couchbase_h, "$Id$")
 
-DIAG_OFF(error)
+#ifdef HAVE_WDOCUMENTATION
```
arr2036: Do the same here as you did fo rlm_mruby.h and rlm_unbound.c.  It's not the header that's optionally included, just the macros.

jpereira: done

PR #2787, opened 2019-07-15
rlm_csv: Fix comparison of unsigned expression

jpereira:


jpereira:
It will fix the build https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/558788681

```
src/modules/rlm_csv/rlm_csv.c:824:11: error: result of comparison of unsigned expression < 0 is always false [-Werror,-Wtautological-unsigned-zero-compare]
        if (slen < 0) {
            ~~~~ ^ ~
src/modules/rlm_csv/rlm_csv.c:840:12: error: result of comparison of unsigned expression < 0 is always false [-Werror,-Wtautological-unsigned-zero-compare]
                if (slen < 0) {
                    ~~~~ ^ ~
2 errors generated.
```

arr2036:
Thanks :)

PR #2789, opened 2019-07-16
src/lib/server/cf_util.c: Check if 'CONF_ITEM' is valid

jpereira:


src/lib/server/cf_util.c
```diff
@@ -527,6 +531,8 @@ char const *_cf_filename(CONF_ITEM const *ci)
  */
 bool cf_item_is_section(CONF_ITEM const *ci)
 {
+	if (!ci) return NULL;
```
alandekok: return `false`

src/lib/server/cf_util.c
```diff
@@ -539,6 +545,8 @@ bool cf_item_is_section(CONF_ITEM const *ci)
  */
 bool cf_item_is_pair(CONF_ITEM const *ci)
 {
+	if (!ci) return NULL;
```
alandekok: return `false`

src/lib/server/cf_util.c
```diff
@@ -551,6 +559,8 @@ bool cf_item_is_pair(CONF_ITEM const *ci)
  */
 bool cf_item_is_data(CONF_ITEM const *ci)
 {
+	if (!ci) return NULL;
```
alandekok: return `false`

jpereira: @alandekok done. My `ctrl-c/ctrl-v` ....  <sigh>

alandekok:
Thinking about this, it may be worth using `CC_HINT(nonnull)` instead.  The difficulty is that the compilers have a lot of false positives when they complain about it.

jpereira:
We have several cases that are walking through a linked-list. so, the compiler never will catch it using `CC_HINT(nonnull)`. (e.g I faced before and fixed it in  https://github.com/FreeRADIUS/freeradius-server/commit/2069ef00b285d5bd408b3caaacefa46449567e40)

PR #2790, opened 2019-07-17
Add fr_box_tainted()/fr_box_tainted_len() to respect taint values

jpereira:
The new both macros have an extra field to set the "taint" state.

### DIFF
--- a/src/lib/server/xlat_tokenize.c
+++ b/src/lib/server/xlat_tokenize.c
@@ -512,7 +512,7 @@ static ssize_t xlat_tokenize_expansion(TALLOC_CTX *ctx, xlat_exp_t **head,
 		/*
 		 *	Box is so we get \t \n etc..
 		 */
-		fr_strerror_printf("Invalid char '%pV' in expression", fr_box_strvalue_len(q, 1));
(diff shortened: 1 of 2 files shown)


jpereira:
It's necessary to complete the commit 012865619f8001b0903fa129ac44cff22e91fadd and fix the xlat tests. (even the tests still failing, but related to rlm_csv)

fyi @arr2036

alandekok:
My $0.02 is that I'd rather call it `fr_box_tainted()`.  That would be clearer than using '`strvalue2`.

jpereira:
@alandekok done. actually, sounds better.

alandekok:
I think the commits are now missing the actual `fr_box_tainted_with_len()` functions?

jpereira:
It's appearing for me.

image

even I rebased just to keep above of your last commit in rlm_csv

jpereira:
@alandekok As we have `fr_box_strvalue() / fr_box_strvalue_len()`. then, I just created `fr_box_tainted()` and `fr_box_tainted_len()`.

i.e: the `fr_box_tainted_with_len()` is there as `_fr_box_tainted_with_len()`

PR #2791, opened 2019-07-17
FIx rlm_csv.c

jpereira:


jpereira:
@alandekok it's expected to fail the build due to the 'taint' string things talked in #2790

src/modules/rlm_csv/rlm_csv.c
```diff
@@ -612,11 +612,6 @@ static int mod_instantiate(void *instance, CONF_SECTION *conf)
 		}
 	}
 
-	if (!inst->key) {
```
alandekok: I'm not sure why this would be deleted.  If there's no key, it will cause run-time errors.

src/tests/keywords/unit_test_module.conf
```diff
@@ -35,6 +35,7 @@ modules {
 	}
 
 	csv {
+		key = "field1"
```
alandekok: The `key` here is similar to the `key` in rlm_files.  It's not a field name, it should be the same as the argument to the `map` e.g.

`key = "%{tolower:%{User-Name}}"`

jpereira: done

jpereira:
Part fixed in https://github.com/FreeRADIUS/freeradius-server/commit/c14778d2e9d163cce3507f8d3ccc7b37aa86d30b

PR #2795, opened 2019-07-19
Fix the %{explode:&ref <delim>} xlat function

jpereira:
As the documentation says `%{explode:&ref <delim>}`, let's inform
the user that the delim should be a single char

### DIFF
--- a/src/lib/server/xlat_func.c
+++ b/src/lib/server/xlat_func.c
@@ -1709,7 +1709,7 @@ static ssize_t xlat_func_explode(TALLOC_CTX *ctx, char **out, size_t outlen,
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


src/lib/server/xlat_func.c
```diff
@@ -1711,6 +1711,8 @@ static ssize_t xlat_func_explode(TALLOC_CTX *ctx, char **out, size_t outlen,
 
 	if (*p == '\0') goto arg_error;
 
+	if (strlen(p) > 1) goto arg_error;
```
alandekok: perhaps `if (p[1]) goto arg_error`.  You don't care about the length of the string.  You care that there's *anything* after the first character.

jpereira: done

PR #2797, opened 2019-07-20
tmpl: Fix the CHECK_ZEROED() check

jpereira:


jpereira:
It will fix the `master` build related to that error https://travis-ci.org/FreeRADIUS/freeradius-server/builds/561263188?utm_source=github_status&utm_medium=notification

PR #2802, opened 2019-07-23
Doc updates

jpereira:


raddb/mods-available/always
```diff
@@ -10,8 +10,6 @@
 #
 #  The `always` module simply returns the same result, always, without
 #  doing anything else.  It is here for use in complex policies.
-#
-
```
alandekok: I really don't like deleting these trailing comments.  The example configs use this a *lot*

raddb/mods-available/dhcpv4
```diff
@@ -28,7 +28,7 @@
 #
 
 #
-#  # Default instance
+#  ## Default instance
```
alandekok: We should decide if it's `==` or `##`, and use one version everywhere.

raddb/mods-available/rest
```diff
@@ -3,6 +3,8 @@
 #  :toc:
 #
 #  $Id$
+#
+#
```
alandekok: Why two blank comments here?

jpereira: it was a typo.

raddb/clients.conf
```diff
@@ -1,6 +1,6 @@
 #  -*- text -*-
 #
-#  :toc:
+#  :toc: left
```
alandekok: If it's possible, this should go into either a style sheet, or maybe a command-line option.

doc/all.mk
```diff
@@ -102,7 +102,7 @@ doc/raddb/%.adoc: raddb/%
 
 doc/%.html: doc/%.adoc
 	@echo HTML $^
-	${Q}$(ASCIIDOCTOR) $< -b html5 -o $@ $<
+	${Q}$(ASCIIDOCTOR) $< -b html5 -T ./scripts/asciidoc/freeradius.template -o $@ $<
```
alandekok: This requires the gem `tilt`:

```
asciidoctor: FAILED: 'tilt' could not be loaded
You must have the tilt gem installed (gem install tilt) to use custom backend templates
```

Seems like a good idea, tho.  I would be happier if asciidoc would ignore the argument if the gem wasn't present.

raddb/clients.conf
alandekok: `asciidoc-a toc=left FILE`

Maybe we just want to set this as the default for all of the files.

doc/all.mk
jpereira: the `tilt` could be checked as we do for `asciidoctor` and `pandoc`?

raddb/clients.conf
jpereira: it was a typo. :(

raddb/mods-available/dhcpv4
jpereira: yes, I noticed that the `#` is a _markdown_ stuff. I will update to use the `=` levels.

doc/all.mk
alandekok: Likely best to check at the `configure` stage.

But checking the template, it should really be applied at the PDF stage.   The file `scripts/asciidoc/freeradius.template` is LaTex, and not HTML.

We will likely need a separate HTML / CSS template for the HTML output.  Right now, the colors are red and black, which doesn't match the default FreeRADIUS look.

raddb/mods-available/dhcpv4
alandekok: There are still a lot of `##` instead of `==`

alandekok: Maybe update the `==` as a separate commit?

PR #2805, opened 2019-07-24
Update doc/upgrade/

jpereira:


doc/howto/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+List with some usual howtos for FreeRADIUS.
+
+## Topics
+
+* [Ascend](ascend.md)
+* [Bay](bay.md)
+* [Cisco](cisco.md)
+* modules
+	* [RADIUS-LDAP-eDirectory](modules/RADIUS-LDAP-eDirectory.md)
```
arr2036: Don't usually use tabs as it's a PITA to edit in any web-based editor.  Also... why are we using markdown for half the documentation and asciidoc for the other half?

jpereira: 1. The reason for the tabs, It is because of the `Mixed List` syntax. Take a look in `Mixed List` 
 at https://en.support.wordpress.com/markdown-quick-reference/

2. I am just reusing the current markdown files that are converted by pandoc.

alandekok: A good chunk of the markdown is old documentation.  So that can be gradually fixed.

In general, stand-alone pages are simpler in markdown.  Multi-page sites are better in Asciidoc

doc/all.mk
```diff
@@ -106,10 +106,24 @@ doc/raddb/%.adoc: raddb/%
...
 	${Q}./scripts/asciidoc/conf2adoc -a ${top_srcdir}/asciidoc -o $@ < $^
 
+#
+#	Converting *.adoc to *.html
+#
 doc/%.html: doc/%.adoc
 	@echo HTML $^
 	${Q}$(ASCIIDOCTOR) $< -a "toc=left" -b html5 -o $@ $<
-	${Q}perl -p -i -e 's/\.adoc/\.html/g' $@
+	${Q}perl -p -i -e 's/\.adoc/\.html/g; s/\/.html"/\/"/g' $@
```
jpereira: @alandekok That extra regex `s/\/.html"/\/"/g` is necessary because when we have hiperlinks like:

`[How-To](howto/) - How-To guides for performing common tasks.`

The `asciidoctor` convert appending a `\.html` to the end of link becoming `<a href="howto/.html">$text</a>`. so, the idea is basically to cleanup that `/\/.html`.

doc/all.mk
```diff
@@ -45,10 +45,10 @@ CONF_FILES := $(filter-out %~,$(wildcard raddb/*conf raddb/mods-available/* radd
 BASE_ADOC_FILES := $(wildcard doc/*/*.adoc) $(wildcard doc/*/*/*.adoc)
 AUTO_ADOC_FILES := $(patsubst raddb/%,doc/raddb/%.adoc,$(CONF_FILES))
 AUTO_ADOC_FILES += $(patsubst raddb/%.md,doc/raddb/%.adoc,$(shell find raddb -name "*\.md" -print))
-AUTO_ADOC_FILES += $(patsubst doc/%.md,doc/%.adoc,$(wildcard doc/*/*/*.md))
+AUTO_ADOC_FILES += $(patsubst doc/%.md,doc/%.adoc,$(wildcard doc/*.md doc/*/*.md doc/*/*/*.md doc/*/*/*/*.md))
 ADOC_FILES	:= $(BASE_ADOC_FILES) $(AUTO_ADOC_FILES)
 PDF_FILES := $(patsubst doc/%.adoc,doc/%.pdf,$(ADOC_FILES))
-HTML_FILES := $(patsubst doc/%.adoc,doc/%.html,$(ADOC_FILES))
+HTML_FILES := $(patsubst doc/%.adoc,doc/%.html,$(ADOC_FILES)) $(subst %README.adoc,index.html,$(ADOC_FILES))
```
jpereira: @alandekok here, basically we are adding the `index.html` (symbolik-links to README.html) to be resolved during `install.doc` and `clean.doc`

doc/howto/cisco.md
```diff
@@ -71,30 +66,25 @@ You must reboot after entering this command to take effect. If not, you
...
+    aaa new-model
+    aaa authentication ppp radppp if-needed radius
+    aaa authorization network radius none
+    aaa accounting network wait-start radius
 
 to get the Cisco to talk to a radius server.
 
 With IOS 11.3
-^^^^^^^^^^^^^
+-------------
```
alandekok: If we're going to use Markdown, we should just use `##` as the section titles everywhere.  Using random different formatting is confusing.

doc/README.md
```diff
@@ -4,7 +4,7 @@ This directory contains documentation for FreeRADIUS.  The
 documentation is available under the Creative Commons Non-Commercial
 license, as given in the `LICENSE` file in this directory.
 
-[Introduction](introduction/) - Introduction to FreeRADIUS and RADIUS
+[Introduction](introduction/aaa) - Introduction to FreeRADIUS and RADIUS
```
alandekok: I would really prefer to leave the "main page" as just `introduction/`.  That allows the main page to have generic text, instead of being specific to AAA.

doc/all.mk
alandekok: That makes sense.  To make it a bit clearer, you can do:

```
${Q}perl -p -i -e 's,\.adoc,\.html,g; s,/.html",/",g' $@
```

which separates `/` as regex delimiter from `/` as filename delimiter

doc/README.md
jpereira: @alandekok But, we don't have any README inside the directory. we 've only the `aaa.md` file there. so, maybe create an `index.html` into the introduction listing the content of directory? in this case, referring to `aaa`?

alandekok: The point is that we're fixing the documentation.  We're not just rearranging it.  This means adding documentation where it's missing.  In this case, a "main page" that serves as an introduction, and which points to the other documentation.

jpereira: All right. I made all the changes and you could see the current changes in https://www.jorgepereira.com.br/temp/freeradius-doc/

PR #2817, opened 2019-07-28
Fix FramedInterfaceId type in postgresql/queries.conf

PaulosV:
This is an addendum to commit fb15efe88b1f06fed89fc7225cd24ceea52dbbbf,
which fixed the PostgreSQL schema file to include the correct
FramedInterfaceId type, but did not modify the queries which were
still casting all the instances to inet. That caused the following
error to show up:

```
rlm_sql_postgresql: Status: PGRES_FATAL_ERROR
rlm_sql_postgresql: 22P02: INVALID TEXT REPRESENTATION
(2) sql: rlm_sql_postgresql: ERROR:  invalid input syntax for type inet: "168:9a03:e9a9:56fa"
```

Edit: this error is also present in branch 3.0.x, where I have originally found it.

### DIFF
--- a/raddb/mods-config/sql/main/postgresql/queries.conf
+++ b/raddb/mods-config/sql/main/postgresql/queries.conf
@@ -253,7 +253,7 @@ accounting {
 					NULLIF('%{Framed-IP-Address}', '')::inet, \
 					NULLIF('%{Framed-IPv6-Address}', '')::inet, \
 					NULLIF('%{Framed-IPv6-Prefix}', '')::inet, \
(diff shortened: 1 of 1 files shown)


PaulosV:
@alandekok thanks for merging and porting this issue, I did not get to porting it at the time, wanted to do it in the evening. A typo slipped into the port, I put a comment on it. https://github.com/FreeRADIUS/freeradius-server/commit/a2a1f5708745af6d66aac933831fcda1eb42a8af#r34474609

PR #2825, opened 2019-07-30
doc: Define {basedir} dinamically

jpereira:
It's necessary to set up the HTML links correctly
avoiding hardcode paths.


alandekok:
Cherry-picked and edited in commit da57df4b

You can use GNU Make rules to find `BASEDIR`.  And the `_assets` name is just habit in asciidoc.  It's not required.  I've renamed it to `templates` so that it's a little nicer.  We generally don't use filenames that begin with `_`

PR #2831, opened 2019-08-01
doc: append the build version to the footer

jpereira:


jpereira:
@alandekok The changes append the below information and also include the asciidoctor version/build-path as an HTML comment being helpful to track-down generated docs.

image

ps: The https://www.jorgepereira.com.br/doc/ is running the changes.

alandekok:
I've pushed a patch that does this a slightly different way.

PR #2834, opened 2019-08-01
travis: Let's validate also the doc build

jpereira:
1. we can't use the distributed versions of asciidoctor and pandoc due to be too old.
2. the package libjson-perl is required by scripts/asciidoc/conf2adoc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,7 @@ addons:
     # For newer libkqueue
     - sourceline: ppa:rjvbertin/misc
     packages:
+    - libjson-perl
     - autoconf
     - build-essential
     - clang-7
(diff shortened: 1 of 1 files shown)


alandekok:
Hmm... the travis rules are getting complex enough that they should likely be encapsulated in a "make" rule.

e.g. The travis tests could do:

`if [ "${DO_BUILD}" = 'no' ]; then make tests.documentation; fi`

with all of the rules for making doxygen, etc. in the `tests.documentation` rule.  We already have many `tests.FOO` targets, so this rule would follow that pattern.

jpereira:
@alandekok as we talked through the Slack. I added the new targets and also the Travis calls as can be seen in https://github.com/FreeRADIUS/freeradius-server/pull/2836

PR #2836, opened 2019-08-02
tweaks in doc/all.mk

jpereira:


doc/all.mk
```diff
@@ -161,8 +161,8 @@ doc/%.pdf: doc/%.md
 		-V papersize=letter \
 		--template=./scripts/asciidoc/freeradius.template -o $@ $<
 
-.PHONY: asciidoc html pdf clean clean.doc
-asciidoc: $(ADOC_FILES)
+.PHONY: adoc html pdf clean clean.doc
+adoc: $(ADOC_FILES)
```
alandekok: This should just be left as "asciidoc".  The file contents are "asciidoc", not "adoc".  We just use the "adoc" extension because having 100 files named "...asciidoc" is annoying.

jpereira: done

alandekok:
I've pulled this over with some manual changes:

* `all.doc` and `tests.doc` targets, to match existing naming patterns
* other minor tweaks

PR #2842, opened 2019-08-02
Fix doc/unlang/string_double.adoc

jpereira:


jpereira:
fix the below message:

```
asciidoctor: WARNING: string_double.adoc: line 21: unterminated table block
asciidoctor: WARNING: string_double.adoc: line 21: unterminated table block```

PR #2844, opened 2019-08-03
Fix an issue whereby the username is occasionaly not properly termina…

ed-at-work:
…ted.


arr2036:
Fixed in dee78b44119168e0cc5714602f8f7449a2e661aa

PR #2850, opened 2019-08-05
travis: enable the option 'retries' for apt-get

jpereira:
It will enable capability to retry during network issue.


jpereira:
It will drive the Travis to retry during errors like https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/567899667#L384

PR #2854, opened 2019-08-05
doc: Set the font-color to black

jpereira:


alandekok:
I've pushed some other changes.  After trying a bunch of things, I think just changing the background color looks the cleanest.

PR #2856, opened 2019-08-05
Fix the Oracle support

jpereira:


arr2036:
May as well add support for 18 too?

jpereira:
@arr2036 done

PR #2862, opened 2019-08-07
Fix mongo build

jpereira:


arr2036:
Should be using -isystem instead of -I which is probably why you're getting warnings.

jpereira:
@arr2036 install the lib and test locally that you'll see the error. e.g: `brew install libmongoc`

Basically, the problem is because we are looking for the libraries/headers assuming that everything is installed in the same path as often as in Linux. 

e.g:

```
-I/usr/include/libbson-1.0 -I/usr/include/libmongoc-1.0
```

But, we should be able to allow the user to inform each path if they want. even if want to build on Mac.

e.g:

bson.h

```
[jpereira@sugarloaf freeradius-server-v3.0.x.git]$ find /usr/local/Cellar/mongo-c-driver/ -iname bson.h
/usr/local/Cellar/mongo-c-driver//1.14.0/include/libbson-1.0/bson/bson.h
/usr/local/Cellar/mongo-c-driver//1.14.0/include/libbson-1.0/bson.h
[jpereira@sugarloaf freeradius-server-v3.0.x.git]$
```
mongoc.h

```
[jpereira@sugarloaf freeradius-server-v3.0.x.git]$ find /usr/local/Cellar/mongo-c-driver/ -iname mongoc.h
/usr/local/Cellar/mongo-c-driver//1.14.0/include/libmongoc-1.0/mongoc/mongoc.h
/usr/local/Cellar/mongo-c-driver//1.14.0/include/libmongoc-1.0/mongoc.h
[jpereira@sugarloaf freeradius-server-v3.0.x.git]$
```

arr2036:
Talking about -Wdocumentation, -isystem should disable warnings

jpereira:
@arr2036 It fixed the warning. pr updated.

PR #2864, opened 2019-08-08
doc: Don't show dynamic doc/*.adoc files

jpereira:


alandekok:
This ignores *all* `adoc` files, including ones we want to add manually.  It might be OK for the `doc/raddb` directory.  It is *not* OK for the top-level `doc` directory.

jpereira:
I converted the doc/howto/*.md to *.adoc as can be seen here https://github.com/FreeRADIUS/freeradius-server/pull/2865

PR #2866, opened 2019-08-09
doc: Add topnav bar with the sections

jpereira:


arr2036:
Please try and do this within the asciidoctor framework first.  There are plugins which split the HTML content out into separate pages, but allow for common TOC trees on the LHS.

arr2036:
https://github.com/atolcd-contrib/asciidoctor-extensions-lab/blob/c3592fe12e396e6fe21b9c03ed444c6fef6dc0b2/lib/multipage-html5-converter.rb

arr2036:
The different pages need to be combined into a single document via includes, the multipage convertor will then split them out based on what you configure, but leave a common TOC.  We might not need a top navbar at all if this works well, and it allows us to render the doc site as an ebook or a PDF.

PR #2867, opened 2019-08-09
Add Oracle session pool to rlm_sql oracle driver

pobept:

### DIFF
--- a/raddb/mods-config/sql/driver/oracle
+++ b/raddb/mods-config/sql/driver/oracle
@@ -1,0 +1,17 @@
+#  -*- text -*-
+#
+#  SQL driver configuration for Oracle
+#
+#  Should be included inside of a SQL module's configuration section
+#
+# $Id$
+#
+oracle {
+	spool {
+		min = ${...pool.min}
+		max = ${...pool.max}
+		inc = ${...pool.spare}
+		timeout = 0
+		statement_cache_size = 64
+	}
+}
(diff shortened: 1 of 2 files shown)


arr2036:
Can you give some more background on these changes for people who don't daily interact with instaclient and oracle databases? Will it impact the ability to establish connections asynchronously in future?

alandekok:
I'm happy to disallow using clients below 10g.  That's 10 years old.

alandekok:
If you could sign the commits and squash them together, I think we can pull it into "master" and to v3.

pobept:
> Can you give some more background on these changes for people who don't interact with instaclient and oracle databases daily? Will it impact the ability to establish connections asynchronously in future?

The changes concern only the driver for Oracle, and do not break the architecture of the application, they allow you to get a connection to the database given faster, because for example, an authentication step is scrolling through.
On the other hand, these changes are necessary to enable the query caching mechanism with binding variables.

pobept:
> If you could sign the commits and squash them together, I think we can pull it into "master" and to v3.

Ok, I will sign the changes and re-create MR for v4 and v3

pobept:
In addition, I have a working v4 driver for Oracle and Postgres with xlat templates and binding variables. but they require a bit of refactoring

arr2036:
I'd be interested to see what you did there, even if you just push the code up to a public branch in your repo.

--

In regards to this PR, is it that multiple sessions are multiplexed over a single network connection? Or does the new API manage both sessions and connections (are they the same thing?).

pobept:
> In regards to this PR, is it that multiple sessions are multiplexed over a single network connection? Or does the new API manage both sessions and connections (are they the same thing?).

Sessions and connections are pooled by OCI. Documentation says that, OCI multiplex several OCI sessions over a few physical connections. 

Testing shows that when using the OCI Session pool with parameters min = 1, max = 30, inc = 2 and the number of application threads 30, the same 30 connections are opened to the database.

PR #2868, opened 2019-08-09
Add Oracle session pool to rlm_sql oracle driver

pobept:

### DIFF
--- a/raddb/mods-config/sql/driver/oracle
+++ b/raddb/mods-config/sql/driver/oracle
@@ -1,0 +1,17 @@
+#  -*- text -*-
+#
+#  SQL driver configuration for Oracle
+#
+#  Should be included inside of a SQL module's configuration section
+#
+# $Id$
+#
+oracle {
+	spool {
+		min = ${...pool.min}
+		max = ${...pool.max}
+		inc = ${...pool.spare}
+		timeout = 0
+		statement_cache_size = 64
+	}
+}
--- a/src/modules/rlm_sql/drivers/rlm_sql_oracle/rlm_sql_oracle.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_oracle/rlm_sql_oracle.c
@@ -15,6 +15,7 @@
  *   along with this program; if not, write to the Free Software
  *   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  *
+ * @copyright 2019 Robert Biktimirov (pobept@gmail.com)
  * @copyright 2000,2006 The FreeRADIUS server project
  * @copyright 2000 David Kerry (davidk@snti.com)
  */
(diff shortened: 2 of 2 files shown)


arr2036:
This isn't a critique of your patch, but this is only a subset of what's required in v4.

What should be happening here, is the driver should inform the parent rlm_sql instance that the driver is capable of multiplexing all outbound requests over a single connection.  The parent rlm_sql instance should then, instead of creating a thread pool, just call the connection function of the driver on thread instantiate, and store the connection in thread instance data.

Acquiring and releasing of connections should be wrapped in an internal function in rlm_sql to abstract the difference between pool based connections, and multiplexing connections.  The release would likely be a noop in the multiplexing case, as cleanup is already handled by a different callback.

Unless there's a very good reason not to, I'd then hardcode sessMin = 0, sessMax = ∞ (or the maximum allowed i.e. UINT32_MAX), sessIncr = 10.  If these values pose a problem in a production environment, i.e. introducing significant latency when opening new sessions, or sessMax pre-allocating structures and using then we could revisit those values, and maybe make them configurable.

There's some new syntax coming in v4 that will limit the number of outstanding yielded requests per-thread, per-module, which is why it's not ideal to enforce a maximum in the driver itself.

I don't think the changes are too complex, but if you'd rather we make them and merge your patch in manually because you're not familiar with the code, then that's a possibility.  If you want to attempt them yourself, then we're happy to provide feedback.

pobept:
Thanks for the detailed answer.

Implementation issues:
Ok, we notified the parent rlm_sql instance about the possibility of creating a connection pool on the driver side.
Receiving and releasing connections are enclosed in separate abstracted functions.
We also need to implement restriction and control mechanisms (amount of use, reconnecting, timeouts, idle time, ..), which are already very well implemented in thread pool api.
Is it necessary to re-implement the functionality, if only a very small part of the databases supported in FreeRadius provide the ability to create a connection pool through their api.

Regarding multiplexing connections:
"driver is capable of multiplexing all outbound requests over a single connection" is not entirely true, documentation says that, OCI multiplex several sessions over a few physical connections (MxN).
In my test environment, M = H.

I would like to understand this issue and get feedback.

arr2036:
The requirements for multiplexed connections and non-multiplexed connections are different.

In general, the connection pool API is there to handle non-multiplexed connections, as we need to manage large numbers of handles to be able to achieve processing parallelism for requests.  An example of this is PostgreSQL.  Even with their async API, they do not allow multiplexing of queries on the same connection, you need distinct connections and distinct connection handles for each query.

If multiple handles are not needed because multiple requests can execute using the same resource handle, as appears to be the case with Oracle, we can remove a lot of the complexity by not using the connection pool, and instead spawn a single connection or handle per thread.  By doing this we give the underlying library the ability to batch up requests better, and increase performance.  Redis is a great example of the kinds of performance boosts you can get by ensuring the request processing pipeline remains full. https://redis.io/topics/pipeliningset

Note: We won't see these kinds of performance improvements immediately, because the rest of the rlm_sql module needs to be converted to support async I/O.  This work is in preparation for that change.  The reason why I'm suggesting we make this change now, is it's relatively low hanging fruit.

OK, so lets deal with those requirements individually:
- Amount of use (lifetime) was a "nice to have" feature leftover from the v1/v2 days where we were more worried about memory leaks in the libraries we were using to communicate with databases.  I've been working with FreeRADIUS for about 12 years at this point, and have never needed to use this feature, so I think it's safe to drop for thread-specific connections.
- If we need to manage connections ourselves instead of this being abstracted by the oracle library, i.e. we need to detect cases where the connection is dead and needs to be re-established, then there's a simple library in src/lib/server/connection.c.  It can be plumbed into the worker thread's event loop (which I believe is provided in the thread instantiation callback for the module), and asynchronously manage the connection state, setup timers, etc...  rlm_sql just needs to query that API to figure out if the connection is up or down, and take appropriate action.  As the connections are thread-specific, no locking or synchronisation needs to be done, it's all pretty simple.  You shouldn't provide an open callback here, and instead, signal when the connection is open using fr_connection_signal_open, probably in the init function after calling OCILogin.
- We should really implement idle timeout.  I'm happy to take a look at this if you want, or if you want to get familiar with the code it's a relatively good short mini-project.  Basically, we'd need to add acquire and release functions to connection.c, and have them start and stop an idle timeout timer.  i.e. when the connection is released, the idle timer starts, when the connection is acquired the idle timer stops.  If the idle timer fires, it should call connection_state_failed, which will re-initialise the connection.  The initial timer should be started in fr_connection_signal_open, or _connection_writable, if the return from the open callback indicates the connection was established correctly.

PR #2873, opened 2019-08-12
Fix wrong 'return'

jpereira:


alandekok:
Already pushed a different fix, thanks.

PR #2874, opened 2019-08-12
Add Oracle session pool to rlm_sql oracle driver v3

pobept:
backport from v4


arr2036:
You don't need to keep closing the PRs.  You can just force push to the branch if your repo if you have changes.

PR #2877, opened 2019-08-12
doc: Convert doc/schemas/logstash/README.{md,adoc}

jpereira:


arr2036:
Could you just run find and convert all remaining md files

PR #2883, opened 2019-08-13
doc: Don't append extras '/'

jpereira:


jpereira:
It will fix to create things like: `<a href='..//path'>.....` (extras slashes)

PR #2887, opened 2019-08-14
SQL logging-related changes for MySQL, PostgreSQL, Cassandra, MSSQL and Oracle

terryburton:
SQL logging-related changes for MySQL, PostgreSQL, Cassandra, MSSQL and Oracle.

- Increase the precision of the post-auth logging to whatever the database will support.
- Don't use CURRENT_TIMESTAMP / NOW() constructions. The logs can be replayed after a substantial delay.
- Use Event-Timestamp (provided or synthesised) for accounting rather than %S.
- Increase robustness of Event-Timestamp by defaulting to %l.
- Fixes of MSSQL: Schema creation and accounting probably haven't worked in recent years.
- Fixes of Oracle: Schema creation probably hasn't worked in recent years.
- Fix for SQLite acctsessiontime calculation.

Tested so far as deploying each schema, then having FR emit the SQL for an accounting-on, accounting-start and post auth query (in debug mode) and running these manually these against the respective database. There may be some things that slip through but the basis for the new %{integer:Event-Timestamp}-based inserts and calculations has been tested.


arr2036:
I agree with the majority of these changes, however, it would be nice to deduplicate the expansion syntax for timestamps.  Could you look at moving that into a dollar expansion?

arr2036:
Things like ``TO_DATE('1970-01-01','YYYY-MM-DD') + NUMTODSINTERVAL(%{%{integer:Event-Timestamp}:-%l},'SECOND')`` really should only appear once, if only to minimise the desire to scratch out one's eyes.

terryburton:
Yes, more readable now.

Also having event_timestamp_epoch broken out as a separate variable makes it easy for the user to change the basis for accounting, e.g. using the request receipt time rather than the given Event-Timestamp, which was an idea proposed by @alandekok.

alandekok:
The `$variable` expansion isn't recursive:

```
event_timestamp_epoch = "%{%{integer:Event-Timestamp}:-%l}"
event_timestamp = "TO_TIMESTAMP($event_timestamp_epoch)"
```

Won't work.  The `event_timestamp` variable will just expand to the assigned string, including the `$`.

It should instead be:

```
event_timestamp = "TO_TIMESTAMP(%{%{integer:Event-Timestamp}:-%l})"
```

We can look at making `$variable` expansion recursive, but that's a separate issue.

terryburton:
Is that a V4 limitation? I expanded on V3 and it seemed to work. I'll take a look later.

terryburton:
Chains of ${} expansions do appear to expand correctly when compiling the config using V4 HEAD. In fact there's prior precedent, e.g the accounting-off query expanding `${..accounting-on.query}`.

Let me know if I'm missing the point.

However, I found a couple of issues with V4:

1. Event-Timestamps do not appear to be synthesised when they are missing from requests.
2. `%l` is expanding to a date string rather than epoch seconds.

Not sure whether the first is intentional or a bug.

mcnewton:
Some of the expansions have changed in v4, try `%c` instead.

They're in `xlat_eval_one_letter`: https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/server/xlat_eval.c#L244

alandekok:
The `%l` expansion is wrong.  It's a 32-bit number as per the documentation and v3.  I'll push a fix.

alandekok:
As for adding Event-Timestamp, that should be in the `default` virtual server, in the `recv Accounting-Request` section.  I'll add it in.

arr2036:
*sigh* please don't.  Just use ``%{integer:%l}``.

arr2036:
All the timestamp expansions should be date types, because it's trivial to convert them to other formats, and... it's the correct type.  A lot of the one letter expansions are duplicative and should be removed.  There's a module just for outputting the time in different formats.

terryburton:
Thanks.

`%c` is current time. '%l' was based on the request reception time and had precedent in the existing queries.conf for Cassandra.

However, since it's just a fallback to when %{Event-Timestamp} is not available (which in a lot of cases means that someone broke their config because it's synthesised by the server when not present), and they should be close in value (the accounting precision is only to 1 second by definition anyway), I think either `%{integer:%l}` or `%c` should suffice.

I guess the question (for @arr2036) is was it intended for `%l` to change from rendering as an integer to a string as this might break existing query.conf files?

terryburton:
@arr2036 Your comment answers my question. I'll update as you suggest early next week.

terryburton:
or not. I'll see how the dust settles :-)

arr2036:
The compromise we came to was to move the rlm_date stuff into a new xlat which took an attribute reference and a time format.  I'll let you know when that's done.

raddb/mods-config/sql/main/postgresql/queries.conf
```diff
@@ -27,6 +27,25 @@
...
+# request reception time recorded by this server, for example if you
+# distrust the provided Event-Timestamp.
+#event_timestamp_epoch = "%l"
+
+event_timestamp_epoch = "%{%{integer:Event-Timestamp}:-%l}"
+
+# event_timestamp is the SQL snippet for converting an epoch timestamp
+# to an SQL date.
+
+event_timestamp = "TO_TIMESTAMP($event_timestamp_epoch)"
```
alandekok: That should be:

event_timestamp = "TO_TIMESTAMP(${event_timestamp_epoch})"

You need the `{}` around all variable names in order to expand them.

terryburton: You just beat me to it. I didn't realise that your earlier comments referred to Travis failures which I've just noticed.

I'll push a fix.

alandekok:
travis is failing because the changes use `$foo` instead of `${foo}`.  Once that's fixes, travis should pass, and I can merge it.

PR #2888, opened 2019-08-14
Minor doc tweaks

jpereira:


jpereira:
@alandekok I know, but we are readily inserting the _copyrights_ with https://github.com/FreeRADIUS/freeradius-server/blob/master/doc/templates/docinfo.html#L1

e.g:

```[jpereira@sugarloaf freeradius-server.git]$ curl -s https://www.jorgepereira.com.br/doc/ | grep "Network RADIUS" | head -2
// Copyright (C) 2019 Network RADIUS SAS.  Licenced under CC-by-NC 4.0.
// Development of this documentation was sponsored by Network RADIUS SAS.
[jpereira@sugarloaf freeradius-server.git]$```

arr2036:
Why do we bother having copyright and licensing notices in the .c files, the server prints out the licensing information when it starts up, isn't that enough?!?!?

alandekok:
The docinfo stuff inserts a *visible* copyright into the output HTML page.  The original source files MUST have a copyright and a license.

PR #2889, opened 2019-08-14
Fix wrong include of 'string.h'

jpereira:


jpereira:
Related to https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/571986939

PR #2891, opened 2019-08-14
Fix comparison of unsigned expression

jpereira:


jpereira:
It fixes the https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/572057953

PR #2892, opened 2019-08-15
doc: Rename README.adoc to index.adoc

jpereira:


arr2036:
Can you fix the references too

PR #2894, opened 2019-08-15
doc: Fix the topic level0 for Unlang docs

jpereira:
The *.adoc files should have the first title starting with '=' (called as level0 in the asciidoc documentation)

### DIFF
--- a/doc/unlang/break.adoc
+++ b/doc/unlang/break.adoc
@@ -1,5 +1,4 @@
-
-== The break Statement
+= The break Statement
 
 .Syntax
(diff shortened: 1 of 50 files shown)


arr2036:
If they should start with '=' then you need to fix all the heading levels, not just add '= Unlang' to the start of the document.  The title of the document isn't 'Unlang'.

alandekok:
The title for each page should be the unlang keyword, or whatever other concept is currently at the top of the file.

jpereira:
It was updated.

PR #2897, opened 2019-08-15
Fix the rlm_sql_cassandra driver build

jpereira:


jpereira:
It fixes 04026a0df54c9752540fe2cc5e67ac1e57522f29 and other issues.

src/modules/rlm_sql/drivers/rlm_sql_cassandra/rlm_sql_cassandra.c
```diff
@@ -818,7 +817,7 @@ do {\
 			       cass_cluster_set_max_concurrent_creation(inst->cluster, inst->spawn_max));
 	}
 
-	if (delay) cass_cluster_set_reconnect_wait_time(inst->cluster, fr_time_delta_to_msec(inst->spawn_retry_delay));
+	cass_cluster_set_reconnect_wait_time(inst->cluster, fr_time_delta_to_msec(inst->spawn_retry_delay));
```
arr2036: use FR_CONF_OFFSET_IS_SET, and check the bool value

src/modules/rlm_sql/drivers/rlm_sql_cassandra/rlm_sql_cassandra.c
```diff
@@ -231,7 +231,7 @@ static const CONF_PARSER driver_config[] = {
 
 	{ FR_CONF_OFFSET("spawn_threshold", FR_TYPE_UINT32, rlm_sql_cassandra_t, spawn_threshold) },
 	{ FR_CONF_OFFSET("spawn_max", FR_TYPE_UINT32, rlm_sql_cassandra_t, spawn_max) },
-	{ FR_CONF_OFFSET("spawn_retry_delay", FR_TYPE_TIME_DELTA, rlm_sql_cassandra_t, spawn_retry_delay) },
+	{ FR_CONF_OFFSET("spawn_retry_delay", FR_TYPE_TIME_DELTA, rlm_sql_cassandra_t, spawn_retry_delay), .dflt = "2.0" },
```
arr2036: use FR_CONF_OFFSET_IS_SET

src/modules/rlm_sql/drivers/rlm_sql_cassandra/rlm_sql_cassandra.c
jpereira: @arr2036 Check the old code https://github.com/FreeRADIUS/freeradius-server/commit/04026a0df54c9752540fe2cc5e67ac1e57522f29#diff-627b3e727957fd2ac985c189ed299e11L822 so, create an extra variable? because I configured the `dflt` value as `2.0`, therefore it always will be _true_.

alandekok: Users can set `spawn_retry_delay=""`.

A different way to fix this is to set

```
{ FR_CONF_OFFSET("spawn_retry_delay", FR_TYPE_TIME_DELTA | FR_TYPE_REQUIRED, rlm_sql_cassandra_t, spawn_retry_delay), .dflt = "2.0" },
```
which means it's always set.  And, that if users set `spawn_retry_delay=""`, then it's disallowed.

jpereira: @alandekok got it. ready updated.

PR #2898, opened 2019-08-15
Add Cassandra SQL driver to be built by Jenkis

jpereira:
* Add Cassandra SQL driver to be built by Jenkis
* rlm_sql_cassandra: Some functions were deprecated in 2.9

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,6 +119,7 @@ services:
 - redis-server
 before_install:
 - npm i -g @antora/cli@2.0 @antora/site-generator-default@2.0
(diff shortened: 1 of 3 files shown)


arr2036:
If they're deprecated then you need to fixup the conf parser entries to show that too

jpereira:
@arr2036 Updated using `FR_CONF_DEPRECATED()`

jpereira:
The `rlm_sql_cassandra` is ready built as can be seen here https://travis-ci.org/FreeRADIUS/freeradius-server/jobs/572807934#L6201

PR #2900, opened 2019-08-16
Add 'docsite' target

jpereira:


arr2036:
I think docsite would be better

jpereira:
@arr2036 done

PR #2903, opened 2019-08-17
The 'docsite' need 'asciidoc' to build raddb asciidoc files

jpereira:


arr2036:
Docsite should have a dependency on all adoc files and also the target required to build the raddb adoc files.  Then you can remove the .phony target, and antora will only be run if adoc source files have changed.

jpereira:
done

PR #2904, opened 2019-08-20
doc: Add 'INSTALL' to howto/ section

jpereira:


arr2036:
It'd be 'Installation' as the link title

jpereira:
done @arr2036

PR #2905, opened 2019-08-20
Ignore the dynamic *.adoc in doc/antora/modules/raddb

jpereira:


alandekok:
We should probably just add all of the output files explicitly.  That directory includes both auto-generated files, and manually generated files.  The wildcard "ignore all adoc files" rule means that we can easily skip commiting files that we want to commit.

PR #2910, opened 2019-08-21
rlm_dhcpv4: Don't leak memory on error conditions.

jpereira:


arr2036:
Not technically a memory leak, but sure...

PR #2913, opened 2019-08-22
Allow parsing of 'float64' in configuration files

nchaigne:
This small patch adds support of type 'float64' in configuration file.

Rationale: precision of float32 is not always sufficient, even for what would seem to be very simple numbers.
For example: "88.005" when printed with the default precision shows "88.004997".


alandekok:
OK, github is fucking up.  I already merged this over an hour ago in commit b7c6b98012f.

This latest merge commit does nothing.

PR #2914, opened 2019-08-22
mod_readme2adoc: the xref should include the module raddb

jpereira:


alandekok:
It doesn't have to include the local module name.  It works fine without it.

jpereira:
weird, I tried a few minutes ago and I noticed that it was showing the wrong behavior. but, now after the update, it is showing correctly. I will close it.

PR #2918, opened 2019-08-23
rlm_attr_filter: Don't leak memory on error conditions.

jpereira:


src/modules/rlm_attr_filter/rlm_attr_filter.c
```diff
@@ -303,6 +303,7 @@ static rlm_rcode_t CC_HINT(nonnull(1,2)) attr_filter_common(void const *instance
 	 */
 	if (!found) {
 		rad_assert(!output);
+		fr_pair_list_free(&output);
```
alandekok: ?  The the line above asserts that `output == NULL`.  In which case there's nothing to free.

PR #2919, opened 2019-08-23
doc: Add sanity checks to verify if the rlm_$name FR_CONF_ is documented

jpereira:

PR #2920, opened 2019-08-23
Add missing config options in raddb/

jpereira:
It helps #2919 involving the modules:

rlm_attr_filter
rlm_cache
rlm_rest
rlm_sometimes


raddb/mods-available/rest
```diff
@@ -113,6 +113,21 @@ rest {
 	#
 #	multiplex = yes
 
+	#
+	#  fail_body_decode:: Force body decoding to fail for debugging purposes.
+	#
+#	fail_body_decode = no
```
arr2036: These are only available in debug builds and should not be included in the documentation

raddb/mods-available/rest
```diff
@@ -113,6 +113,21 @@ rest {
...
+
+	#
+	#  fail_header_decode:: Force header decoding to fail for debugging purposes.
+	#
+#	fail_header_decode = no
```
arr2036: These are only available in debug builds and should not be included in the documentation

raddb/mods-available/cache
```diff
@@ -174,6 +174,11 @@ cache {
 	#
 	add_stats = no
 
+	#
+	#  max_entries:: Maximum number of retries.
+	#
+#	max_entries = 0
```
arr2036: That's not the maximum number of retries, that's the maximum number of entries allowed in the rbtree driver

PR #2942, opened 2019-08-30
Fix fr_fprintf() to use the file stream instead of fd

jpereira:
Just use the fputs() instead of get the fileno()/write()


jpereira:
It is impacting the `radmin -e 'uptime'` (the only place that uses it)

PR #2954, opened 2019-09-05
Packaging fixes and "make test"

terryburton:
Fixes for Debian.
Perform a "make test" during package build for both Debian and Redhat.

### DIFF
--- a/debian/control.in
+++ b/debian/control.in
@@ -31,7 +31,9 @@ Build-Depends: @BUILDDEPS@ debhelper (>= 9),
  python-dev,
  samba-dev,
  dh-systemd,
- unixodbc-dev
+ unixodbc-dev,
+ asciidoctor,
+ pandoc
 Section: net
 Priority: optional
 Maintainer: Josip Rodin <joy-packages@debian.org>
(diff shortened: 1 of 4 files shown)


alandekok:
I've fixed the debian patch in  commit a4a6825899ac.  I've also pulled over the builddep change.

I'm less sure about adding `make test` to the package builds.  We can always run travis / jenkins before packaging.  I'm not sure how adding `make test` helps us here.

terryburton:
> I'm less sure about adding `make test` to the package builds. We can always run travis / jenkins before packaging. I'm not sure how adding `make test` helps us here.

It is simple to skip tests where necessary, but I agree that the check step is not strictly due to existing CI. From a release perspective, having tests run in environments that closely resemble distros' runtime might help to detect fragility.

alandekok:
The main reason to run the tests in the "package build" process is when the packaging does heavy patching of the server core.  Since we're not doing that, I think we're OK with dropping these.

We can always have a separate integration test.  Which patches the server, builds it, and then runs the tests.

My concern is that adding this to the "make package" target is mixing two independent things.

PR #2955, opened 2019-09-05
mongo: Fix examples

geaaru:
ObjectId() and Date() commands are not accepted by the low-level mongo JSON engine.
For the field 'date' it's needed to use $date command.
For replacing ObjectId() it is possible to use sha256 or any others hashing command.

### DIFF
--- a/raddb/mods-config/sql/main/mongo/queries.conf
+++ b/raddb/mods-config/sql/main/mongo/queries.conf
@@ -57,40 +57,40 @@ sql_user_name = "%{User-Name}"
 #
 authorize_check_query = "db.mobiles.aggregate([ \
 	{ \
-	  '$match': { \
-	    'calling_station_id': '%{Calling-Station-id}', \
-	    'auth_blocked': 'false' \
-	  } \
+		'$match': { \
+			'calling_station_id': '%{Calling-Station-id}', \
+			'auth_blocked': 'false' \
+		} \
 	}, \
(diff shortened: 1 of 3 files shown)


raddb/mods-config/sql/main/mongo/queries.conf
```diff
@@ -56,41 +56,41 @@ sql_user_name = "%{User-Name}"
...
-	{ \
-	  '$project': { \
-	    '_id': 0, \
-	    'attribute': '$attributes.k', \
-	    'value': '$attributes.v', \
-	    'op': ':=' \
-	  } \
-	} \
-    ])" \
+    { \
```
alandekok: Can you drop these changes from the patch?  They seem to be just whitespace changes with nothing else.

We try to use tabs everywhere in the config files instead of many spaces.

geaaru: Ok, I fix all query with tabs.

geaaru: it's a bit unsupportable a mixed solution with tabs and spaces, now are used only tabs.

I found also some more fixes for sqlippool queries and some comma issues. Let me know if is ok now. thanks

geaaru: Maybe could be more correct divide commit between fix and indentation.

PR #2958, opened 2019-09-06
jenkins: The 'QA_RPATHS' is already in 'make rpm' target

jpereira:


jpereira:
It is ready in https://github.com/FreeRADIUS/freeradius-server/blob/master/Makefile#L381

PR #2959, opened 2019-09-06
detect changes to package manifests and pass no-cache flag to docker …

mattrose:
This will look for changes to the freeradius.spec file or any changes in the debian folder, and if it finds them, passes the --no-cache flag to docker, so that any new dependencies added to those package manifest files will be added to the docker build images


scripts/jenkins/Jenkinsfile
```diff
@@ -1,5 +1,25 @@
...
+    def changeLogSets = currentBuild.changeSets
+    for (int i = 0; i < changeLogSets.size(); i++) {
+        def entries = changeLogSets[i].items
+        for (int j = 0; j < entries.length; j++) {
+            def entry = entries[j]
+            def files = new ArrayList(entry.affectedFiles)
+            for (int k = 0; k < files.size(); k++) {
+                def file = files[k]
+                if (file.path =~ /(^debian\/.*)|(redhat\/freeradius.spec)/) {
+                    echo "file ${file.path} matches"
```
arr2036: Matches what?

mattrose: You know what, that line was just so I could debug it when I was starting out.  I'll pull it.

I also need to sign the commit

mattrose: Or, maybe I'll just add some text to clarify

arr2036: Yes, I do know what cos I'm reading the script, nobody will no what when they're looking at the log output which was of course my point :)

mattrose: fixed now

arr2036: Maybe some comments about what the no-cache flag does, and why we're only applying it to a subset of builds?

mattrose:
Added a comment to describe the function

scripts/jenkins/Jenkinsfile
```diff
@@ -1,5 +1,33 @@
...
+    def string = ""
+    def changeLogSets = currentBuild.changeSets
+    for (int i = 0; i < changeLogSets.size(); i++) {
+        def entries = changeLogSets[i].items
+        for (int j = 0; j < entries.length; j++) {
+            def entry = entries[j]
+            def files = new ArrayList(entry.affectedFiles)
+            for (int k = 0; k < files.size(); k++) {
+                def file = files[k]
+                if (file.path =~ /(^debian\/.*)|(redhat\/freeradius.spec)/) {
```
jpereira: @mattrose Probably you should do the same test as is doing for Debian due to existing others files into Redhat/ directory instead of only the `*.spec`

scripts/jenkins/Jenkinsfile
```diff
@@ -1,5 +1,33 @@
...
+                    echo "changes in file ${file.path}, passing --no-cache flag to docker build"
+                    string = " --no-cache "
```
jpereira: Due to being into two of `for (...)` levels. Probably should be better to just `return " --no-cache"` and get off from the `for()` instead of running through all entries even if ready matched.

scripts/jenkins/Jenkinsfile
mattrose: The freeradius.spec file is the only file that depends on the environment it's being built in.  All other files in that directory are patches or ancillary files that get added to the package as it's being built, and we don't really want to rebuild the whole image if there's only a change in them.  I would have made a similar change and only rebuilt on modifications to `debian/control.in` but I'm not as sure of the other files in that directory as I am of the redhat directory

arr2036:
hm that merge commit is huge.  You probably need to pull and rebase from master

PR #2967, opened 2019-09-09
Fix missing free() in $INCLUDE

jpereira:


src/lib/server/cf_file.c
```diff
@@ -800,22 +800,21 @@ static int process_include(CONF_SECTION *this, char const *ptr, char *buff[stati
 		if (stat(my_directory, &stat_buf) < 0) {
 			ERROR("%s[%d]: Failed reading directory %s: %s", filename, *lineno,
 			      my_directory, fr_syserror(errno));
-			talloc_free(my_directory);
-			return -1;
+			goto error;
```
arr2036: Goto label should be here

src/lib/server/cf_file.c
```diff
@@ -800,22 +800,21 @@ static int process_include(CONF_SECTION *this, char const *ptr, char *buff[stati
...
 		}
 #endif
 		dir = opendir(my_directory);
 		if (!dir) {
 			ERROR("%s[%d]: Error reading directory %s: %s",
 			      filename, *lineno, value,
 			      fr_syserror(errno));
+		error:
```
arr2036: not in the middle of the cases where the error could occur

src/lib/server/cf_file.c
```diff
@@ -853,8 +852,7 @@ static int process_include(CONF_SECTION *this, char const *ptr, char *buff[stati
 			 */
 			if (cf_file_include(this, buff[2], CONF_INCLUDE_FROMDIR, buff, true) < 0) {
 				closedir(dir);
-				talloc_free(my_directory);
-				return -1;
+				goto error;
```
arr2036: just squash the commits...

jpereira:
done @arr2036

arr2036:
Goto label still not at the first incidence of the error.

-Arran

jpereira:
As I said, take a look at the code and you will see that the first incidence is inside a `#ifdef ... #endif` condition. either way, I just fixed the missing `free()` and kept as the old approach.

arr2036:
@jpereira you didn't say that.  Or at least I didn't get a notification of you saying that...

PR #2984, opened 2019-09-13
return empty string when no changes are detected

mattrose:
The last one inserted "null" instead of an empty string when no changes were detected.

@jpereira Does this look OK to you?


jpereira:
Looks good!

PR #2988, opened 2019-09-13
rlm_python: PyObject functions should return 'Py_RETURN_NONE' instead of 'NULL'

jpereira:


src/modules/rlm_python/rlm_python.c
```diff
@@ -230,13 +230,13 @@ static PyObject *mod_log(UNUSED PyObject *module, PyObject *args)
 	char *msg;
 
 	if (!PyArg_ParseTuple(args, "is", &status, &msg)) {
-		return NULL;
+		Py_RETURN_NONE;
 	}
 
 	fr_log(&default_log, status, __FILE__, __LINE__, "%s", msg);
 	Py_INCREF(Py_None);
```
arr2036: Need to remove this too, else we leak nones.

jpereira: done

PR #2995, opened 2019-09-16
rlm_python: Load the libpython using correct extension

jpereira:


arr2036:
No patches for v3.0.x

PR #2996, opened 2019-09-16
various jenkinsfile fixes

mattrose:
Mainly so that the updaterepo jenkins job will pick up the correct build number when creating the various package repositories.

Also this ensures that all checkout jobs inherit from the jenkins configuration

### DIFF
--- a/scripts/jenkins/Jenkinsfile
+++ b/scripts/jenkins/Jenkinsfile
@@ -46,20 +46,20 @@ def buildClosures(arg) {
                 stage("build-${platform}") {
                     rebuild = detectChanges()
                     docker.build("${platform}-master-test","${rebuild} -f ./scripts/docker/build-${platform}/Dockerfile.deps ./scripts/docker/build-${platform}").inside {
-                        checkout([$class: 'GitSCM', branches: [[name: '*/master']], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}/build"]], submoduleCfg: [], userRemoteConfigs: [[url: 'https://github.com/freeradius/freeradius-server']]])
+                        checkout([$class: 'GitSCM', branches: scm.branches, userRemoteConfigs: scm.userRemoteConfigs, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}/build"],[$class: 'CloneOption', depth: 0, noTags: false, reference: '', shallow: false]], submoduleCfg: []])
                         sh "cat /etc/os-release || cat /etc/redhat-release"
+                        def commit_num = readFile("./build-number").trim()
                         dir("${platform}/build") {
                             if (platform.contains("centos")) {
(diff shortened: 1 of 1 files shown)


mattrose:
Pushed again because I forgot to sign the commit.

scripts/jenkins/Jenkinsfile
```diff
@@ -46,20 +46,20 @@ def buildClosures(arg) {
...
+                                sh "sed -i -e \"s/^Release:.*\$/Release: ${commit_num}/\" redhat/freeradius.spec"
                                 sh 'RADIUSD_VERSION_STRING=$(cat VERSION) make rpm'
                                 sh "mv rpmbuild/RPMS/x86_64/*.rpm .."
                             } else {
                                 sh "apt-get install -y unixodbc-dev"
-                                sh label: '', script: '''version=$(dpkg-parsechangelog | grep "^Version: " | awk \'{print $2}\')
-                                commit_num=$(git describe --tags --long --match "release_*"  --match "branch_*" | sed -e \'s/^.*-\\([0-9]*\\)-g[0-9a-f]*/\\1/\')
-                                commit_msg="$(git log --oneline -1 $GIT_COMMIT)"
-                                dch -b -v ${version}-${commit_num} "$commit_msg"'''
+                                def version = sh (script: "dpkg-parsechangelog | grep '^Version: ' | awk '{print \$2}'", returnStdout: true).trim()
```
jpereira: Maybe: `awk '/Version/ { print $2}'`

mattrose: Yeah, 6 of one, a half-dozen of the other...

scripts/jenkins/Jenkinsfile
```diff
@@ -46,20 +46,20 @@ def buildClosures(arg) {
                 stage("build-${platform}") {
                     rebuild = detectChanges()
                     docker.build("${platform}-master-test","${rebuild} -f ./scripts/docker/build-${platform}/Dockerfile.deps ./scripts/docker/build-${platform}").inside {
-                        checkout([$class: 'GitSCM', branches: [[name: '*/master']], doGenerateSubmoduleConfigurations: false, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}/build"]], submoduleCfg: [], userRemoteConfigs: [[url: 'https://github.com/freeradius/freeradius-server']]])
+                        checkout([$class: 'GitSCM', branches: scm.branches, userRemoteConfigs: scm.userRemoteConfigs, extensions: [[$class: 'CleanBeforeCheckout'],[$class: 'RelativeTargetDirectory', relativeTargetDir: "${platform}/build"],[$class: 'CloneOption', depth: 0, noTags: false, reference: '', shallow: false]], submoduleCfg: []])
                         sh "cat /etc/os-release || cat /etc/redhat-release"
+                        def commit_num = readFile("./build-number").trim()
```
arr2036: Hm, what's the reason for doing it this way instead of running the git describe command directly in the Jenkins file

arr2036: I don't see a problem with it... just interested.  Anyway, will merge this now.

mattrose: Basically I didn't want to run that command more than once in the Jenkinsfile, to avoid future problems, and I had to do it outside of the docker image and the parallel builds to create the file that the updaterepo job reads.  I was hoping to be able to pass that in to the docker images as a variable, but that proved impossible. I had to write that file anyway, so I just had the builds read it, rather than run the whole big command again.

arr2036: OK, fair enough.  The command is still accessible within the Jenkins file which is the main thing.

mattrose: I did a lot of work trying to get it so that things were only set in one place if possible, just to reduce the risk of copypasta errors later on.

PR #3005, opened 2019-09-20
Fix the error message for 'uknown address family'

jpereira:


src/main/conffile.c
```diff
@@ -1331,7 +1331,7 @@ static inline int fr_item_validate_ipaddr(CONF_SECTION *cs, char const *name, PW
 
 
 		default:
-			cf_log_err(&(cs->item), "Unknown address family passed for parsing IP address.", ipaddr->af);
+			cf_log_err(&(cs->item), "Unknown address \"%d\" family passed for parsing IP address.", ipaddr->af);
```
arr2036: cf_log_err(&(cs->item), "Unknown address family (%d) passed for parsing IP address", ipaddr->af);

arr2036: you're specifying the address family, not the address

jpereira: done

PR #3013, opened 2019-10-01
Add methods cache.{store,load,status,clear,ttl} for rlm_cache

jpereira:
Now the rlm_cache has the capability of handling the cache entries using methods.


jpereira:
@arr2036 can you take a look here?

PR #3019, opened 2019-10-02
Add Magic module

DomAmato:
Magic is a new authentication protocol built around the blockchain. This module is the minimal requirement to authenticate via a local or hosted gateway.

### DIFF
--- a/raddb/mods-available/magic
+++ b/raddb/mods-available/magic
@@ -1,12 +1,11 @@
-# When the Login-Time attribute is set to some value, and the
-# user has been permitted to log in, a Session-Timeout is
-# calculated based on the remaining time.  See "doc/README".
+# Config for connecting to a Magic Authentication Gateway
 #
 magic {
+    # Whether the socket connection is local (unix) or remote
     is_local = true
     # The ip or url of the magic gateway
     # socket-address = 127.0.0.1
-    socket-address = tmp/magicsock
-	# The port that the magic gateway is listening on
-	port = 12345
+    socket-address = tmp/magicsock 
+    # The port that the magic gateway is listening on
+    port = 12345
 }
--- a/raddb/mods-available/magic
+++ b/raddb/mods-available/magic
@@ -1,0 +1,12 @@
+# When the Login-Time attribute is set to some value, and the
+# user has been permitted to log in, a Session-Timeout is
+# calculated based on the remaining time.  See "doc/README".
+#
+magic {
+    is_local = true
+    # The ip or url of the magic gateway
+    # socket-address = 127.0.0.1
+    socket-address = tmp/magicsock
+	# The port that the magic gateway is listening on
+	port = 12345
+}
--- a/src/modules/rlm_magic/rlm_magic.c
+++ b/src/modules/rlm_magic/rlm_magic.c
@@ -17,10 +17,10 @@
 /**
  * $Id$
  * @file rlm_magic.c
- * @brief Example module code.
+ * @brief Middleware to pass along an authentication attempt to a Magic Gateway for verification
  *
  * @copyright 2013 The FreeRADIUS server project
- * @copyright 2013 your name (your name\@address)
+ * @copyright 2019 Magic Foundation (hello@magic.co)
  */
 RCSID("$Id$")
 
--- a/src/modules/rlm_magic/rlm_magic.c
+++ b/src/modules/rlm_magic/rlm_magic.c
@@ -1,0 +1,249 @@
+/*
+ *   This program is is free software; you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation; either version 2 of the License, or (at
+ *   your option) any later version.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, write to the Free Software
+ *   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/**
+ * $Id$
+ * @file rlm_magic.c
+ * @brief Example module code.
+ *
+ * @copyright 2013 The FreeRADIUS server project
+ * @copyright 2013 your name (your name\@address)
+ */
+RCSID("$Id$")
+
+#include <freeradius-devel/radiusd.h>
+#include <freeradius-devel/modules.h>
+#include <freeradius-devel/rad_assert.h>
+
+/*
+ *	Define a structure for our module configuration.
+ *
+ *	These variables do not need to be in a structure, but it's
+ *	a lot cleaner to do so, and a pointer to the structure can
+ *	be used as the instance handle.
+ */
+typedef struct
+{
+	unsigned short port;
+	char const *socket_address;
+	bool is_local;
+} rlm_magic_t;
+
+/*
+ *	A mapping of configuration file names to internal variables.
+ */
+static const CONF_PARSER module_config[] = {
+	{"port", FR_CONF_OFFSET(PW_TYPE_SHORT, rlm_magic_t, port), "12345"},
+	{"is_local", FR_CONF_OFFSET(PW_TYPE_BOOLEAN, rlm_magic_t, is_local), false},
+	{"socket-address", FR_CONF_OFFSET(PW_TYPE_STRING, rlm_magic_t, socket_address), "127.0.0.1"},
+	CONF_PARSER_TERMINATOR};
+
+/*
+ *	Do any per-module initialization that is separate to each
+ *	configured instance of the module.  e.g. set up connections
+ *	to external databases, read configuration files, set up
+ *	dictionary entries, etc.
+ */
+static int mod_instantiate(UNUSED CONF_SECTION *conf, void *instance)
+{
+	rlm_magic_t *inst = instance;
+	DEBUG2("Creating Magic module");
+
+	if (inst->is_local)
+	{
+		#ifndef HAVE_SYS_UN_H
+			ERROR("Trying to use local unix sockets on a system that does not support it");
+			return -1;
+		#endif
+		DEBUG2("Using local Unix socket %s for authentication", inst->socket_address);
+	}
+	else
+	{
+		DEBUG2("Using remote server %s:%i for authentication", inst->socket_address, inst->port);
+	}
+
+	return 0;
+}
+
+/*
+ *	Don't currently do anything here
+ */
+static rlm_rcode_t CC_HINT(nonnull) mod_authorize(UNUSED void *instance, UNUSED REQUEST *request)
+{
+	return RLM_MODULE_NOOP;
+}
+
+/*
+ *	Authenticate the user with the given password.
+ */
+static rlm_rcode_t CC_HINT(nonnull) mod_authenticate(void *instance, REQUEST *request)
+{
+	// The instance variables
+	rlm_magic_t *inst = instance;
+
+	//Socket related vars
+	char buffer[BUFSIZ];
+	int sockfd;
+
+	DEBUG2("Attempting Magic Authentication");
+
+	if (inst->is_local)
+	{
+		sockfd = fr_socket_client_unix(inst->socket_address, false);
+		if (sockfd == -1)
+		{
+			fr_perror("socket");
+			return RLM_MODULE_FAIL;
+		}
+	}
+	else
+	{
+		fr_ipaddr_t address_info;
+		address_info.af = AF_INET;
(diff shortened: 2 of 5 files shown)


arr2036:
What's the advantage in talking over a raw TCP socket instead of just calling out to a REST API?

arr2036:
So the major issues I can see:
- This module doesn't use the connection pool API, so you're creating and destroying connections every time a user logs in, which is incredibly inefficient.
- With zero technical documents being published by magic.co, we can't determine if this will be a useful module, and worth the effort to include it in the server core and maintain it.

My recommendation would be to implement this module in Perl or python and distribute it directly from magic.co as part of the alpha package.  If the protocol gets fleshed out beyond submitting small JSON blobs over a raw TCP connection, and ends up being something you couldn't have just implemented easier with a REST API (and existing rlm_rest module) then we'd definitely reconsider.

DomAmato:
I'll shed a bit more light on some of the reasoning it was made into a module instead of using an interpreted language. 

One of the goals was to get our authentication mechanism working on openWRT which the packages for python and perl are not compiled for and take up a ton of space that the routers usually cannot afford to have installed without a swap file and extra storage space added. We had a module written in python but because of the aforementioned reasons and the fact that it doesn't look like freeradius 3 will ever support python 3 we decided not to go down that path.

This is very much an alpha thing and ideally we don't want to add more maintenance on your end but also maintaining our own fork isn't ideal for us either. 

There will be updates and maybe at that time we will submit another PR with more features fleshed out but before we publish more information we have been working with the SEC regarding the legality so obviously things are slow while we finish that work first.

I will take a look at the rlm_rest module in the meantime but can you shed light on why this module needs to use the connection_pool api when so many others don't? I am not opposed to using it but it seems like if that is a bar set for this module that it should be more consistently enforced on other modules that already exist in the code base.

alandekok:
Most modules which do outbound connections use the connection API:

```
$ git grep -l fr_connection_get src/modules/
src/modules/rlm_cache/drivers/rlm_cache_memcached/rlm_cache_memcached.c
src/modules/rlm_couchbase/mod.c
src/modules/rlm_couchbase/rlm_couchbase.c
src/modules/rlm_krb5/rlm_krb5.c
src/modules/rlm_ldap/ldap.c
src/modules/rlm_mschap/auth_wbclient.c
src/modules/rlm_redis/rlm_redis.c
src/modules/rlm_rediswho/rlm_rediswho.c
src/modules/rlm_rest/rlm_rest.c
src/modules/rlm_smsotp/rlm_smsotp.c
src/modules/rlm_sql/rlm_sql.c
src/modules/rlm_sqlhpwippool/rlm_sqlhpwippool.c
src/modules/rlm_sqlippool/rlm_sqlippool.c
src/modules/rlm_yubikey/validate.c
```

There main reason is that opening connections can be expensive, especially for databases.  There may be not just database communication overhead, but also TLS.

The connection API allows a connection to be opened once, and then re-used many times for different requests.  In our experience, this re-use substantially increases the throughput of the server, and also lowers latency.

arr2036:
> I'll shed a bit more light on some of the reasoning it was made into a module instead of using an interpreted language.
> 
> One of the goals was to get our authentication mechanism working on openWRT which the packages for python and perl are not compiled for and take up a ton of space that the routers usually cannot afford to have installed without a swap file and extra storage space added. We had a module written in python but because of the aforementioned reasons and the fact that it doesn't look like freeradius 3 will ever support python 3 we decided not to go down that path.

OK, in that case if you're targeting a specific platform then there's nothing to stop you from building your own packages and making them available.  Your code does not need to be in the main source tree in order for you to do this.

> This is very much an alpha thing and ideally we don't want to add more maintenance on your end but also maintaining our own fork isn't ideal for us either.

You don't have to.  You'd be maintaining only your specific module.

> There will be updates and maybe at that time we will submit another PR with more features fleshed out but before we publish more information we have been working with the SEC regarding the legality so obviously things are slow while we finish that work first.
> 
> I will take a look at the rlm_rest module in the meantime but can you shed light on why this module needs to use the connection_pool api when so many others don't? 

AFAIK every module that ships with FreeRADIUS that manages handles representing connections to an external database, either uses the connection API to manage those connections, has a thread local connection, or has a connection pool managed by another library.

Taking a quick scan through v3.x clients of the connection pool API are rlm_cache, rlm_couchbase, rlm_ippool, rlm_krb5, rlm_ldap, rlm_redis, rlm_rest, rlm_sql*, rlm_yubikey.

rlm_unbound uses libunbound which manages its own connections. rlm_securid uses the securid library of connection management.  rlm_opendirectory uses Apple's OpenDirectory framework.

> I am not opposed to using it but it seems like if that is a bar set for this module that it should be more consistently enforced on other modules that already exist in the code base.

That statement is based on a false premise.

raddb/mods-available/magic
```diff
@@ -0,0 +1,12 @@
+# When the Login-Time attribute is set to some value, and the
+# user has been permitted to log in, a Session-Timeout is
+# calculated based on the remaining time.  See "doc/README".
+#
+magic {
+    is_local = true
+    # The ip or url of the magic gateway
+    # socket-address = 127.0.0.1
```
alandekok: everything else uses `server` for the configuration item which contains the name of the external server.  If you make your module "unusual" to configure, then people will be confused by it.

You should just use `server`, and drop the `is_local` setting.  If the `server` starts with a `/`, then it's a local Unix socket.  Otherwise it's not.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+ *   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/**
+ * $Id$
+ * @file rlm_magic.c
+ * @brief Example module code.
+ *
+ * @copyright 2013 The FreeRADIUS server project
+ * @copyright 2013 your name (your name\@address)
```
alandekok: Adding your actual name would be good. :)

arr2036: That's fixed up in  a later commit, but the commits should be squashed.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+ *	to external databases, read configuration files, set up
+ *	dictionary entries, etc.
+ */
+static int mod_instantiate(UNUSED CONF_SECTION *conf, void *instance)
+{
+	rlm_magic_t *inst = instance;
+	DEBUG2("Creating Magic module");
+
+	if (inst->is_local)
+	{
```
alandekok: Formatting code consistently is a huge thing for us.  All of the code in the server uses the same formatting.  This practice ensures that we can read the code years later.  If instead the code was randomly formatted by random people, it would become un-maintainable.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+	else
+	{
+		DEBUG2("Using remote server %s:%i for authentication", inst->socket_address, inst->port);
+	}
+
+	return 0;
+}
+
+/*
+ *	Don't currently do anything here
```
alandekok: Just omit functions which don't do anything.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+	rlm_magic_t *inst = instance;
+
+	//Socket related vars
+	char buffer[BUFSIZ];
+	int sockfd;
+
+	DEBUG2("Attempting Magic Authentication");
+
+	if (inst->is_local)
+	{
```
alandekok: more formatting comments.  We really are completely fanatical about this.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+			return RLM_MODULE_FAIL;
+		}
+	}
+	else
+	{
+		fr_ipaddr_t address_info;
+		address_info.af = AF_INET;
+		struct hostent *hostent;
+		
+		hostent = gethostbyname(inst->socket_address);
```
alandekok: This is a non-starter for run-time use.  DNS lookups should be done when the server starts up, and cached.  If people change DNS while the server is running, too bad.

The alternative is to make the server dependent on DNS for *every single packet*.  So if DNS goes down, this line will *lock* the server for 30 seconds.  And very bad things will happen.

Some modules do allow hostnames for server identifiers.  This is generally done only when the third-party APIs don't really support IP addresses.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+			fprintf(stderr, "error: inet_addr(\"%s\")\n", *(hostent->h_addr_list));
+			return RLM_MODULE_FAIL;
+		}
+		address_info.ipaddr.ip4addr.s_addr = in_addr;
+
+		sockfd = fr_socket_client_tcp(NULL, &address_info, inst->port, false);
+	}
+	// Create our message which needs to be in this format
+	// '{"address": "username", "password": "password", "sessionId": ""}\n'
+	char *socket_message = talloc_array(request, char, BUFSIZ);
```
alandekok: There's no need to do multiple `strlcat`.  Just one massive `talloc_asprintf(request, " ...", args);` will do.

It's easier to understand, and less code.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if (shutdown(sockfd, SHUT_WR) != 0)
+	{
+		perror("shutdown socket");
+		return RLM_MODULE_FAIL;
+	}
+	talloc_free(socket_message);
+
+	// read from the socket
+	int nbytes_read = read(sockfd, buffer, BUFSIZ);
```
alandekok: If you're only reading 1 byte from the buffer, there's no need to read `BUFSIZE` bytes.  That will generally block the server on the `read()` call.  The only way to stop that is if the other end either writes `BUFSIZE` data, OR writes less than `BUFSIZE` bytes, and then closes the socket.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	return RLM_MODULE_FAIL;
+}
+
+#ifdef WITH_ACCOUNTING
+/*
+ *	Massage the request before recording it or proxying it
+ */
+static rlm_rcode_t CC_HINT(nonnull) mod_preacct(UNUSED void *instance, UNUSED REQUEST *request)
```
alandekok: more "delete functions which aren't used".

src/modules/rlm_magic/rlm_magic.c
DomAmato: is it safe to leave them in if at some point it is likely to do something? I suppose its not really more or less work but was wondering if there was some adverse reason for leaving them there

src/modules/rlm_magic/rlm_magic.c
DomAmato: understandable

src/modules/rlm_magic/rlm_magic.c
DomAmato: makes sense, this would then happen in the mod_instantiate function correct?

DomAmato:
I think we as a team need to do a review of the networking architecture for magic and then come back to this. This module was a MVP we needed to prove the technology as viable but given your feedback I think there are some conversations we will have internally about how to move forward. Thanks for all the suggestions and feedback on this module!

src/modules/rlm_magic/rlm_magic.c
alandekok: Yes, the instantiate module should do that lookup and cache the IP address in the instance data.  There's a `fr_pton()` function which handles all of the DNS lookups, and IPv4 / IPv6 issues.  The `gethostbyname()` API is just not useful.

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+	strlcpy(socket_message, "{\"address\": \"", BUFSIZ);
+	strlcat(socket_message, request->username->vp_strvalue, BUFSIZ);
+	strlcat(socket_message, "\", \"password\": \"", BUFSIZ);
+	strlcat(socket_message, request->password->vp_strvalue, BUFSIZ);
+	strlcat(socket_message, "\", \"sessionId\": \"\"}\n", BUFSIZ);
+
+	// Write to the socket
+	if (write(sockfd, socket_message, BUFSIZ) == -1)
+	{
+		perror("write");
```
jpereira: you should release the memory here also: e.g `talloc_free(socket_message);`

src/modules/rlm_magic/rlm_magic.c
```diff
@@ -0,0 +1,249 @@
...
+		return RLM_MODULE_FAIL;
+	}
+	if (shutdown(sockfd, SHUT_WR) != 0)
+	{
+		perror("shutdown socket");
```
jpereira: and here: e.g `talloc_free(socket_message);`

PR #3020, opened 2019-10-02
grammar and spelling fixes for modules

alanbuxey:


raddb/mods-available/yubikey
```diff
@@ -235,7 +235,7 @@ yubikey {
 			#
 			#  spread:: Enable LRU (Least Recently Used).
 			#
-			#  The `yubico` library uses `curl` multihandles. If the multihandle is reused
+			#  The `yubico` library uses `curl` multi-handles. If the multi-handle is reused
```
arr2036: It's actually 'multi handle'

arr2036: https://curl.haxx.se/libcurl/c/libcurl-multi.html

raddb/mods-available/detail
```diff
@@ -100,7 +100,7 @@ detail {
 	#  locking:: Whether or not we should lock the detail file
 	#  before writing to it.
 	#
-	#  Mainly used if the detail file reader is be reading this file.
+	#  Mainly used if the detail file reader is to be reading this file.
```
arr2036: Mainly used if the detail file reader will be reading this file.

raddb/mods-available/yubikey
alanbuxey: agreed, I should have checked how CURL defines it rather than the rest of the world ;-)

raddb/mods-available/detail
alanbuxey: 'will be' sounds too certain. what if it's not going to be reading the file just yet? 
I chose 'is to be' since the previous wording suggested that was the intention of the text :)

PR #3022, opened 2019-10-02
Add tests for radmin/control-socket

jpereira:
New unit tests to validate the radmin commands.


jpereira:
Sample.

```
[jpereira@sugarloaf freeradius-server.git]$ make clean.test.radmin
[jpereira@sugarloaf freeradius-server.git]$ make test.radmin
RADMIN src/tests/radmin/config/test.conf
Starting RADMIN test server... Executing: /bin/sh -c build/bin/local/radiusd -Pxxxl build/tests/radmin/radius.log -d src/tests/radmin/config -n test -D /Volumes/Users/jpereira/Devel/FreeRADIUS/freeradius-server.git/share/dictionary/
RADMIN-TEST 00-help
RADMIN-TEST 01-show-client-all-proto
RADMIN-TEST 02-show-client-tcp-proto
RADMIN-TEST 03-show-client-udp-proto
RADMIN-TEST 04-show-config-item1
RADMIN-TEST 05-set-module-handled-updated
RADMIN-TEST 06-show-module-handled-status
gmake[1]: Entering directory '/Volumes/Users/jpereira/Devel/FreeRADIUS/freeradius-server.git'
gmake[1]: Leaving directory '/Volumes/Users/jpereira/Devel/FreeRADIUS/freeradius-server.git'
```

src/tests/radmin/all.mk
```diff
@@ -0,0 +1,148 @@
...
+#
+#	Test name
+#
+TEST := test.radmin
+
+#
+#	Config settings
+#
+PORT               := 12341
+BIN_PATH           := $(BUILD_DIR)/bin/local
```
alandekok: A lot of this is already defined in Make.inc

TESTBINDIR - directory where the binary files are located
TESTBIN - path (including jlibtool arguments) for running program

src/tests/radmin/all.mk
```diff
@@ -0,0 +1,148 @@
...
+		exit $$ret; \
+	fi
+
+.PHONY: clean.test.radmin
+clean.test.radmin:
+	${Q}rm -f $(RADMIN_OUTPUT_DIR)/*.ok \
+		  $(RADMIN_OUTPUT_DIR)/*.log \
+		  $(RADMIN_OUTPUT_DIR)/*.skip \
+		  $(RADMIN_CONFIG_PATH)/test.conf \
+		  $(RADMIN_SOCKET_FILE)
```
alandekok: there's no need to remove individual output files.  If all output files are in a directory, just remove the directory.

arr2036:
Be nice to move the all.mk contents to its own file.

src/tests/radmin/all.mk
```diff
@@ -0,0 +1,148 @@
...
+
+clean.test: clean.test.radmin
+
+#
+#	Build the configuration files.
+#
+$(RADMIN_CONFIG_PATH)/dictionary:
+	${Q}echo "# test dictionary not install.  Delete at any time." > $@
+
+$(RADMIN_CONFIG_PATH)/test.conf: $(RADMIN_CONFIG_PATH)/dictionary
```
alandekok: This should just be a static configuration file in src/tests/radmin

The main server binary takes the directory from `-d whatever`, and creates a configuration item of `raddbdir = whatever`.  The test of the configuration items can be taken from that:

```
testdir = ${raddbdir}/config
logdir = ${raddbdir}/../../build/src/tests/radmin
```
etc.  That simplifies things a fair bit.

src/tests/radmin/all.mk
```diff
@@ -0,0 +1,148 @@
...
+
+#
+#	Run the radmin commands against the radiusd.
+#
+$(RADMIN_OUTPUT_DIR)/%.ok: $(RADMIN_CONFIG_PATH)/test.conf | test.radmin.kill $(RADMIN_CONFIG_PATH)/radiusd.pid
+	$(eval RADMIN_TEST      := $(notdir $(patsubst %.ok,%,$@)))
+	$(eval RADMIN_TEST_IN   := $(RADMIN_CMDS_PATH)/$(RADMIN_TEST).in)
+	$(eval RADMIN_TEST_RES  := $(RADMIN_CMDS_PATH)/$(RADMIN_TEST).res)
+	$(eval RADMIN_TEST_OUT  := $(RADMIN_OUTPUT_DIR)/$(RADMIN_TEST).out)
+	$(eval RADMIN_TEST_SKIP := $(RADMIN_OUTPUT_DIR)/$(RADMIN_TEST).skip)
```
alandekok: Why do we need special "skip" functionality?  I think this can be deleted.

alandekok:
The tests seem good, but there's a lot of new boilerplate makefile rules which are different from everything else in the server.  Is it not possible to use a similar layout to `src/tests/unit/all.mk` or `src/tests/keywords/all.mk` ?

Those files use a common layout, names, and comments for the test rules.  That makes the tests pretty simple to understand.

the final rule to run each test ends up last in the `all.mk` file.  It has the most complexity, but ends up being 10-30 lines of shell scripting.

If we need a separate set of rules to start / stop the server, they could go in a separate makefile.  There's already `src/tests/Makefile` which is a little specific to running the auth tests.  But it could be leveraged for the radmin tests. too.

i.e. there's no need to duplicate all of that code in the radmin tests.  Instead, split the start / stop / kill rules into a separate makefile, and then include that file from multiple places.

PR #3030, opened 2019-10-07
Some fixes for radmin tests

jpereira:


src/tests/radmin/all.mk
```diff
@@ -40,8 +40,8 @@ $(OUTPUT)/%: $(DIR)/% test.radmin.radiusd_kill test.radmin.radiusd_start
 		echo "--------------------------------------------------"; \
 		echo "$(RADIUSD_RUN)"; \
 		echo "$(RADMIN_BIN) -q -f $(RADMIN_SOCKET_FILE) > $(FOUND) < $<"; \
-		$(MAKE) $(TEST).radiusd_kill; \
-		exit 1;\
+		$(MAKE) test.radmin.radiusd_kill; \
```
alandekok: why is this preferred?  I'm happy to leave it as $(TEST)

jpereira: I would like too. but, does not expand.

```
[jpereira@sugarloaf freeradius-server.git]$ make Q= test.radmin | grep radiusd_kill
if [ -f build/tests/radmin/radiusd.pid ]; then ret=0; if ! ps `cat build/tests/radmin/radiusd.pid` >/dev/null 2>&1; then rm -f control-socket; echo "FreeRADIUS terminated during test called by test.radmin.radiusd_kill"; echo "GDB output was:"; cat "build/tests/radmin/gdb.log"; echo "--------------------------------------------------"; tail -n 40 "build/tests/radmin/gdb.log"; echo "Last entries in server log (build/tests/radmin/gdb.log):"; ret=1; fi; if ! kill -TERM `cat build/tests/radmin/radiusd.pid` >/dev/null 2>&1; then ret=1; fi; exit ; fi
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
		/usr/local/bin/gmake .radiusd_kill; \
```

jpereira: Even I've tried to use double reference but without success. e.g: `$$(TEST).radiusd_kill`

alandekok: Ah OK.  More boilermake magic.

src/tests/radmin/all.mk
```diff
@@ -54,6 +54,5 @@ $(OUTPUT)/%: $(DIR)/% test.radmin.radiusd_kill test.radmin.radiusd_start
 		touch $@;\
 	fi
 
-$(TEST): $(FILES)
+$(TEST): ${BUILD_DIR}/bin/radmin $(FILES)
```
alandekok: That should just edit the build rules, like src/tests/unit/all.mk:

```
#
#	Run the radmin commands against the radiusd.
#
$(OUTPUT)/%: ${BUILD_DIR}/bin/radmin $(DIR)/% test.radmin.radiusd_kill test.radmin.radiusd_start
```

jpereira: done

PR #3045, opened 2019-10-12
Example coa-relay site

terryburton:
I'll work this up into a v4 patch once it lands.

### DIFF
--- a/raddb/sites-available/coa-relay
+++ b/raddb/sites-available/coa-relay
@@ -1,0 +1,285 @@
+# -*- text -*-
+######################################################################
+#
+#  Sample site for receiving a CoA or Disconnect-Request packet for a user
+#  and/or session, looking up matching sessions in the radacct table, then
+#  relaying the packet towards the owning NAS for each session.
+#
+#  The process is as follows:
+#
+#    - A CoA/Disconnect-Request is received by FreeRADIUS.
+#    - The radacct table is searched for active sessions that match each of
+#      the provided identifier attributes: User-Name, Acct-Session-Id. The
+#      search returns the owning NAS and Acct-Unique-Id for the matching
+#      session/s.
+#    - The original CoA/Disconnect-Request content is written to a detail file
+#      with custom attributes representing the NAS and Acct-Session-Id.
+#    - A detail reader follows the file and originates CoA/Disconenct-Requests
+#      containing the original content, relayed to the corresponding NAS for
+#      each session using the custom attributes.
+#
+#  This simplifies scripting directly against a set of NAS devices since a
+#  script need only send a single CoA/Disconenct to FreeRADIUS which will
+#  then:
+#
+#    - Lookup all active sessions belonging to a user, in the case that only a
+#      User-Name attribute is provided in the request
+#    - Handle routing of the request to the correct NAS, in the case of a
+#      multi-NAS setup
+#
+#  For example, to Disconnect a specific session:
+#
+#    $ echo 'Acct-Session-Id = "769df3 312343"' | \
+#      radclient 127.0.0.1 disconnect testing123
+#
+#  To perform a CoA update of all active sessions belonging to a user:
+#
+#    $ cat <<EOF | radclient 127.0.0.1 coa testing123
+#      User-Name = bob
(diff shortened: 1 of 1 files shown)


raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+#
+server coa {
+	#  When a packet is received, it is processed through the
+	#  recv-coa section.  This applies to *both* CoA-Request and
+	#  Disconnect-Request packets.
+	recv-coa {
+
+		# Record the original request type (CoA or Disconnect)
+		#
+		switch Packet-Type {
```
alandekok: This could just be

```
update request {
  CoA-Packet-Type := "%{Packet-Type}"
}
```

raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+					update request {
+
+						COA-Packet-DST-IP-Address := "%{1}"
+						COA-Acct-Session-Id := "%{2}"
+
+						# Set the CoA/Disconnect port
+						COA-Packet-DST-Port := 1700
+
+						# NASs don't usually want to
+						# see User-Name, and
```
alandekok: Some NASes do want to see `User-Name` in a CoA / Disconnect packet.  Those ones will NAK the request if it doesn't contain a `User-Name`

raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+					disconnect:Event-Timestamp = "%l"
+					disconnect:Message-Authenticator = 0x00
+				}
+			}
+
+			case "CoA-Request" {
+				update {
+
+					#  Include given attributes
+					coa: += request:[*]
```
alandekok: The disconnect should include these, too.  The packet contents are generally identical.

raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+
+		ok
+	}
+}
+
+
+#  The CoA packet is in the "proxy-request" attribute list.
+#  The CoA reply (if any) is in the "proxy-reply" attribute list.
+#
+server originate-coa {
```
alandekok: Maybe rename this to something else, as there's already an `originate-coa` virtual server.

raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+	pre-proxy {
+		update proxy-request {
+
+			#  Specify the source IP address for the request
+			NAS-IP-Address = 10.1.2.3
```
alandekok: That's the IP address of the NAS, which is not the same as the source IP that the packet is being sent from.  I think this line was left over from earlier testing, and can likely be deleted.

raddb/sites-available/coa-relay
```diff
@@ -0,0 +1,289 @@
...
+
+#
+#  Homeserver CoA / Disconnect endpoints
+#
+#  See proxy.conf for more details on configuring a home_server and
+#  home_server_pool.
+#
+home_server coa-nas1 {
+	type = coa
+	ipaddr = 10.0.0.1
```
alandekok: This should be `192.0.2.1`, which is the canonical "documentation IP network".  Along with a comment "please edit this..."

raddb/sites-available/coa-relay
terryburton: Thanks. I wrote this out noddy fashion after initially struggling with syntax.

raddb/sites-available/coa-relay
terryburton: I've now left User-Name in with a comment for clearing it.

terryburton:
> Can you just squash all of the commits together? I'll pull it in then.

Done. Thanks.

arr2036:
What’s the rationale for the custom attributes?

terryburton:
> What’s the rationale for the custom attributes?

If you put the data into pre-existing attributes, which includes Packet-Type={CoA-Request,Disconnect-Request}, then the detail reader processes the request through the recv-coa section (rather than accounting) which does not act on the coa list.

Since you need to indicate the intended packet type through another attribute it seemed to make sense to keep the remaining looked-up attributes separate from the provided attributes, and apply them when generating the CoA.

PR #3048, opened 2019-10-14
MySQL sqlippools: Find and allocate stored procedure

terryburton:
I'll amend and test with v4 once it lands.


arr2036:
the v4 version should be almost identical? Why not make the changes now and revise them later if needed?

terryburton:
> the v4 version should be almost identical? Why not make the changes now and revise them later if needed?

PR #3051

PR #3049, opened 2019-10-15
The 'RADIUSD_BIN' should be global

jpereira:


jpereira:
It should be global to be accessed outside. e.g: https://github.com/FreeRADIUS/freeradius-server/blob/master/src/tests/eapol_test/all.mk#L77

PR #3050, opened 2019-10-15
Fix typo in the 'eap' value needed by eapol_test

jpereira:


arr2036:
Huh? Why would we want to make the aka prime test run the wrong eap method?

jpereira:
I used wrong branch. nevermind. <sigh>

PR #3056, opened 2019-10-18
Add tests for DHCPv6 / rfc3319

jpereira:


src/tests/unit/dhcpv6_rfc3319.txt
```diff
@@ -0,0 +1,36 @@
...
+load dhcpv6
+dictionary-load dhcpv6
+
+#
+#  3.1  SIP Servers Domain Name List
+#
+
+#  domain = mysip.com
+encode-pair SIP-Server-Domain-Name-List = mysip.com
+data 00 15 00 09 6d 79 73 69 70 2e 63 6f 6d
```
alandekok: RFC 3319 Section 3.1 says:

The option length is followed by a sequence of labels, encoded
   according to Section 3.1 of RFC 1035 [5], quoted below:

      "Domain names in messages are expressed in terms of a sequence of
      labels.  Each label is represented as a one octet length field
      followed by that number of octets.  Since every domain name ends
 with the null label of the root, a domain name is terminated by a
      length byte of zero.  The high order two bits of every length
      octet must be zero, and the remaining six bits of the length field
      limit the label to 63 octets or less.  To simplify
      implementations, the total length of a domain name (i.e., label
      octets and label length octets) is restricted to 255 octets or
      less."

      RFC 1035 encoding was chosen to accommodate future
      internationalized domain name mechanisms.

i.e. this is *not* data type `string` or `octets`.  It's data type DNS resource record.  The encoding here is *not* just the bare string, but the DNS resource record encoding.

This test should be a failed test, because we don't currently support encoding DNS resource records.

src/tests/unit/dhcpv6_rfc3319.txt
```diff
@@ -0,0 +1,36 @@
...
+
+#  ipv6 = 2001:0db8:85a3:0000:0000:8a2e:0370:1234
+encode-pair SIP-Server-Domain-Name-List = 0x2001db885a38a2e3701234
+data 00 15 00 0b 20 01 db 88 5a 38 a2 e3 70 12 34
+
+#  array with domain & ipv6
+#  arg[1] = mysip.com
+#  arg[2] = 2001:0db8:85a3:0000:0000:8a2e:0370:1234
```
arr2036: FQDNs can encode IPv6 addresses?

src/tests/unit/dhcpv6_rfc3319.txt
```diff
@@ -0,0 +1,36 @@
...
+#  3.1  SIP Servers Domain Name List
+#
+
+#  domain = mysip.com
+encode-pair SIP-Server-Domain-Name-List = mysip.com
+data 00 15 00 09 6d 79 73 69 70 2e 63 6f 6d
+
+#  ipv6 = 2001:0db8:85a3:0000:0000:8a2e:0370:1234
+encode-pair SIP-Server-Domain-Name-List = 0x2001db885a38a2e3701234
+data 00 15 00 0b 20 01 db 88 5a 38 a2 e3 70 12 34
```
arr2036: That's not the correct output.  DNS attributes should be encoded in DNS wire format as per the RFC

arr2036: Don't assume the encoder is correct.  The point is to come up with tests that actually check the output is right.  There's lots of places where it's not complete.  I got through writing the encoder and then realised we needed major work on other areas of the server like splitting the protocols out into their own dictionaries, and being able to create grouping attributes.

arr2036:
`data` has now been changed to `match` and I think @alandekok almost has the DNS label encoding working.

arr2036:
Looks good to me so far

jpereira:
But, we have a problem. the tests are failing because the Travis doesn't have any ipv6 configured interface. then, we couldn't resolve any AAAA. so, It will be necessary to add some extra pokes trough the Travis script to set up a dummy ipv6.

agreed?

arr2036:
you still need an ipv6 entry for localhost I guess.  I'm not sure we even really want to test DNS resolution.  It's handled by a completely separate API, not by the encoder or decoder itself.  If we really wanted to test DNS resolution in isolation, adding a 'cast' keyword to unit test framework for casting one type to another would be a better way to do that explicitly.

jpereira:
yes, I am working on it and trying to figure how to do that simple with Jenkins.

alandekok:
We should use IPs and not host names for the tests.  DNS resolution may not work, or may be slow.

PR #3057, opened 2019-10-18
rlm_linelog : Don't seek if the linelog file is /dev/stdout 

tuyan:
This pull request disables seek to the end of the file if the file is /dev/stdout. 

Helps to linelog to /dev/stdout which is a symlink to /proc/self/fd/1


alandekok:
I've pushed a slightly more generic fix.

PR #3058, opened 2019-10-21
Update ldap uri

jpereira:


arr2036:
The URI needs to be escaped to function.  It's one of the weird quirks of slapd

PR #3060, opened 2019-10-22
travis: Enable IPv6

jpereira:


.travis.yml
```diff
@@ -118,6 +118,8 @@ services:
 - postgresql
 - redis-server
 before_install:
+# Add IPv6 interfaces: https://github.com/travis-ci/travis-ci/issues/8361
+- if [ "${TRAVIS_OS_NAME}" == "linux" ]; then sudo sh -c 'echo 0 > /proc/sys/net/ipv6/conf/all/disable_ipv6'; fi
```
arr2036: We only run on Linux, so you can drop the OS check, but otherwise seems fine

jpereira: done

PR #3062, opened 2019-10-22
Create a separate rlm_python3 module.

aren:
python3 support for v3.0.x. Implemented as a separate, python3-only module (rlm_python3).

This code is based off of the now-closed #1686 (which was based on the existing rlm_python but intended for v4).


alandekok:
Thanks.  I can't merge the changes until the build errors are fixed, though.  Once that's done, we can pull it in a release 3.0.20.

aren:
I'll get those fixed shortly

aren:
I believe this PR is ready and the build errors are fixed.

PR #3080, opened 2019-10-30
Update the option-code 'Status-Code' in dictionary.rfc3315

jpereira:
Using latest list in https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#dhcpv6-parameters-5


jpereira:
Using latest list in https://www.iana.org/assignments/dhcpv6-parameters/dhcpv6-parameters.xhtml#dhcpv6-parameters-5

alandekok:
I've pushed a slightly different fix to decode the struct.

PR #3088, opened 2019-10-31
The 'Option-Request' expects only a valid option number

jpereira:


arr2036:
This is a really bad idea.  If you look at the DHCPv4 code we do this dynamically because otherwise it can very easily get out of sync with the dictionaries.  I believe @alandekok is going to push a proper fix for this.

PR #3093, opened 2019-10-31
The 'Status-Code-Value' should be 'uint16'

jpereira:


jpereira:
https://tools.ietf.org/html/rfc8415#section-21.13

```
      status-code          The numeric code for the status encoded in
                           this option.  A 2-octet field containing an
                           unsigned integer.```

alandekok:
Oops.  that was a typo.

PR #3099, opened 2019-11-01
Split out the dhcpv6.txt by dhcpv6_rfcXXX.txt

jpereira:
1. Each test entry was moved to the respective rfc. e.g: dhcpv6_rfcXXX.txt
2. Added the `decode-pair -` for all tests.
3. The `dhcpv6.txt` still there.

### DIFF
--- a/src/tests/unit/all.mk
+++ b/src/tests/unit/all.mk
@@ -33,7 +33,16 @@ FILES  := \
 	dhcpv6_rfc3315.txt \
 	dhcpv6_rfc3319.txt \
 	dhcpv6_rfc3646.txt \
+	dhcpv6_rfc4242.txt \
+	dhcpv6_rfc4280.txt \
+	dhcpv6_rfc4580.txt \
+	dhcpv6_rfc5678.txt \
+	dhcpv6_rfc5908.txt \
+	dhcpv6_rfc5970.txt \
 	dhcpv6_rfc6355.txt \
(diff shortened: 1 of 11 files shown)


arr2036:
The fact that a test used options _from_ a particular RFC does not make them tests _for_ a particular RFC.  The test suite does not exist to advertise our compliance with a particular RFC, the test cases exist to exercise the encoder and decoder.

As agreed with @alandekok, the exception to this are, if an organisation is sponsoring development of a particular feature/implementation of a standard and has requested evidence of compliance, or the RFC has complex structures or types which exercise the encoder or decoder in a "useful" way, i.e. a way that may reveal defects or trigger corner cases.

Splitting out the generic test cases that existed before you started working on the test suite, into separate RFC files because they used test options from those RFCs does not qualify for the exception above.

Splitting those tests out makes it harder to work on the encoder and decoder, because it's impossible to get a good overview of the coverage we already have in the unit tests.

Because the dhcpv6 unit test file was large, I've split out the tests into separate files by type.
Because of the confusion over  when test cases should be split into separate files, we will at some point, redo the type specific tests with an internal dictionary only used for the unit tests.

jpereira:
Ok, btw, the current tests by types do not cover up the `decode-pair -`.

arr2036:
OK, if you could submit a PR to add those, then that'd be useful

jpereira:
Ok

PR #3100, opened 2019-11-01
Add tests for DHCPv6 / rfc8415

jpereira:


arr2036:
Use `match-regex` to deal with the different output between macOS and Linux.

PR #3103, opened 2019-11-05
Helpful to print the fullpath of unit test

jpereira:


jpereira:
It's hard to figure out which file is failing.

e.g:

```
match failed at line 11 of src/tests/unit/protocols/dhcpv6
  got      : Attr-54 = 0x0001001020010db885a3000000008a2e03707334
  expected : MOS-Address-IS = 2001:0db8:85a3:0000:0000:8a2e:0370:7334
             ^ differs here```

PR #3107, opened 2019-11-07
Allow more verbose

jpereira:
perl -p -i -e 's/\t\@/\t\${Q}/' $(find . -name "all.mk" -print)


arr2036:
The majority of the commands you removed the @ from were echo or printf, which is wrong.  We don't want:

```
echo DOXYGEN /path/to/doxygen
DOXYGEN /path/to/doxygen
```

in the output...

src/lib/all.mk
```diff
@@ -19,7 +19,7 @@ src/freeradius-devel: | src/include/${1}
 
 src/include/${1}:
 	$${Q}[ -e $$@ ] || ln -sf $${top_srcdir}/src/lib/${1} $$@
-	@echo LN-SF src/lib/${1} $$@
+	${Q}echo LN-SF src/lib/${1} $$@
```
arr2036: These all need to be escaped $${Q}

arr2036: But still, you're emitting the echo command... so not necessary to begin with

jpereira:
Got it @arr2036 . I updated and fixed some extra things like: `${Q}${Q}command....` and now we are using `@echo ...` only to print the first target message.

PR #3115, opened 2019-11-11
use the correct OID

jpereira:


alandekok:
Fixed via commit b406242c

PR #3116, opened 2019-11-11
DEBUG: treat lengthy Tunnel-Password as fatal

cipherboy:
When the server is running in debug mode, treat a Tunnel-Password whose
length exceeds 249 characters as being a fatal error. This'll help to
diagnose connection failures caused by the otherwise silent truncation.

This is inside a giant block to reuse existing logic:

```c
if ((rad_debug_lvl) ||
      (compat_mode_str && (strcmp(compat_mode_str, "cistron") == 0))) {

....

}
```

It might be worthwhile (but a potentially breaking change) to always execute these checks. Note that it appears that `radclient` might do truncation of its own, but the server doesn't when computing the response, causing communication issues.

I'm guessing this was a side effect of d1cdce1b07d4630832f4e834d22ddc5d06557c5a (that the truncation was made silent) -- but the original function `make_tunnel_passwd` seems to have moved and I don't have a great place to put this logic any more. 

If this isn't the correct to put this check, do tell me.

### DIFF
--- a/src/modules/rlm_files/rlm_files.c
+++ b/src/modules/rlm_files/rlm_files.c
@@ -215,6 +215,18 @@ static int getusersfile(TALLOC_CTX *ctx, char const *filename, rbtree_t **ptree,
 			 */
 			for (vp = fr_cursor_init(&cursor, &entry->reply); vp; vp = fr_cursor_next(&cursor)) {
 				/*
+				 * If we're dealing with Tunnel-Password,
+				 * note that we're truncating it to 249 bytes.
+				 */
+				if (strcmp(vp->da->name, "Tunnel-Password") == 0) {
+					if (strlen(vp->data.strvalue) > 249) {
+						ERROR("[%s]:%d Refusing to accept Tunnel-Password with length longer than 249 characters.",
+						      filename, entry->lineno);
+						return -1;
+					}
+				}
+
+				/*
 				 *	If it's NOT a vendor attribute,
 				 *	and it's NOT a wire protocol
 				 *	and we ignore Fall-Through,


src/modules/rlm_files/rlm_files.c
```diff
@@ -214,6 +214,18 @@ static int getusersfile(TALLOC_CTX *ctx, char const *filename, rbtree_t **ptree,
 			 *	worth doing.
 			 */
 			for (vp = fr_cursor_init(&cursor, &entry->reply); vp; vp = fr_cursor_next(&cursor)) {
+				/*
+				 * If we're dealing with Tunnel-Password,
+				 * note that we're truncating it to 249 bytes.
+				 */
+				if (strcmp(vp->da->name, "Tunnel-Password") == 0) {
+					if (strlen(vp->data.strvalue) > 249) {
```
cipherboy: As an aside, I think this is technically wrong as we'd likely need to handle the format like:

```
    Tunnel-Password = <something-short>
    Tunnel-Password += <something-long-but-under-249>
```

But that seems like an edge case. If I had a better place to check the computed value of `Tunnel-Password` on startup, that'd be nicer and more correct.

alandekok: The above snippet is for two tunnel passwords, not for two strings which add up to >249 characters.

alandekok:
The issue is that the long Tunnel-Password can come from anywhere, not just from rlm_files.  So this is the wrong place to put it.

Radclient also does similar truncation, because it has the same limits.

Similarly, any other attribute of type `string` or `octets` will also get silently truncated to 253 octets.  This is unfortunately a limitation of RADIUS.

One solution is to create a "sanity check" function.  That function can then do sanity checks against all attributes, no matter where they come from.  It can run before the reply is sent.  There can even be a configuration item which sets whether or not this function is run.

cipherboy:
Apologies for the delay. I've had other work that has been a higher priority; this is just an item on our [QE wishlist](https://bugzilla.redhat.com/show_bug.cgi?id=1463673) that I've finally had cycles to take a look at.

> The issue is that the long Tunnel-Password can come from anywhere, not just from rlm_files.

Yeah, I agree with that.

But if you have two sets of inputs: those that are static, that you read and parse up front, and those that are dynamic that you read as necessary. Parsing `rlm_files` happens on server startup, right? So you'd want _some_ validation to happen then. My assertion is that should be fatal, because it indicates misconfiguration. When you get a request and go to reply -- if you then occur say, a database lookup and that returns too long of a Tunnel-Password -- and the value there is too long, there you _should_ truncate it and log that this value was too long, but otherwise continue to function. This shouldn't be fatal, because the server started successfully (and probably didn't read Tunnel-Password until a request came in, but I could be wrong) -- so it should be a warning to preserve the current behavior. An admin can later read the logs and truncate Tunnel-Password as necessary.

I also don't know how common lengthy Tunnel-Passwords are. I'm guessing not that common. 

IOW, a fatal config sanity checker (what this is attempting to be, for the type I happen to care about at present) -- and a non-fatal reply sanity-checker.

My 2c on the matter.

alandekok:
There are a few modules which load static configurations (e.g. attr_filter, passwd, among others).  Attributes can be read dynamically from many back-ends, SQL, LDAP, exec, redis, etc.

There are attributes other than Tunnel-Password which have length limitations.  I think adding a check in one module for one attribute is a bit of a band-aid, and not a complete solution to the problem.

I'm not even sure that long attributes should be fatal.  The user may want to store intermediate results in an attribute for later editing.  This is allowed by the server, and arguably also common.

Long Tunnel-Passwords are rare, because they're hard to manage for one.  Another is that most people using RADIUS are aware of length limitations, and avoid long attributes of any type.

It is very difficult to come up with a solution which works everywhere, and doesn't get it wrong.

The design of the server lets people do things, even it it might be wrong.  We've found that approach is preferable to forbidding people from doing things, just because it might be wrong.

I think in the end the simplest approach here is just "don't do that".  If more validation is needed, it should be in the form of a module which sanity checks the attributes, and prints warning messages in debug mode.

cipherboy:
@alandekok, Thanks for you reply. As I don't have the cycles to work on this right now, we've decided to document it as a known issue and move on.

Sorry for the noise. I'll work on submitting some minor fixes to the Python 3 module build system as I get time.

 - Alex

PR #3118, opened 2019-11-13
Update unit tests for rfc6355

jpereira:
Add tests for the Server-ID-DUID attributes

### DIFF
--- a/src/tests/unit/protocols/dhcpv6/rfc6355.txt
+++ b/src/tests/unit/protocols/dhcpv6/rfc6355.txt
@@ -45,17 +45,6 @@ match 00 01 00 02 00 04
 decode-pair -
 match Client-ID-DUID = Client-ID-DUID-UUID
 
-#
-#  Server Identifier
-#
-
-#  Define the DUID-LLT + DUID
-encode-pair Server-ID-DUID = Server-ID-DUID-UUID
-match 00 02 00 02 00 04
-
-decode-pair -
-match Server-ID-DUID = Server-ID-DUID-UUID
-
 #  Encode a 16-octet UUID
 encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f
 match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f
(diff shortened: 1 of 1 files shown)


src/tests/unit/protocols/dhcpv6/rfc6355.txt
```diff
@@ -45,6 +45,26 @@ match 00 01 00 02 00 04
...
+encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f1112
+match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f
+
+#  But if we decode something that's too long, we ignore the extra data
+decode-pair -
+match Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f
+
+#  Or we pad the UUID to 16 octets if it's too short
+encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d
+match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 00 00
```
arr2036: If the UUID is expected to be 16 octets and it's shorter, then it's malformed and must be created as an unknown attribute

arr2036: Unless the RFC specifically mentions different behaviour?

alandekok: This is encoding, not decoding.  So we can't create an unknown attribute in the packet.

The RFC says that the field is 16 octets.  If the admin sets a UUID which is shorter, then the encoder pads out the field to 16 octets.

arr2036:
> On 13 Nov 2019, at 05:58, Alan DeKok <notifications@github.com> wrote:
> 
> ﻿
> @alandekok commented on this pull request.
> 
> In src/tests/unit/protocols/dhcpv6/rfc6355.txt:
> 
> > @@ -45,6 +45,26 @@ match 00 01 00 02 00 04
>  decode-pair -
>  match Client-ID-DUID = Client-ID-DUID-UUID
>  
> +#  Encode a 16-octet UUID
> +encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f
> +match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f
> +
> +#  UUID-Value field is 16 octets.  So we truncate if if it's too long, and pad it with zeros if it's too short.
> +encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f1112
> +match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 0e 0f
> +
> +#  But if we decode something that's too long, we ignore the extra data
> +decode-pair -
> +match Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d0e0f
> +
> +#  Or we pad the UUID to 16 octets if it's too short
> +encode-pair Client-ID-DUID = Client-ID-DUID-UUID, Client-ID-DUID-UUID-Value = 0x000102030405060708090a0b0c0d
> +match 00 01 00 12 00 04 00 01 02 03 04 05 06 07 08 09 0a 0b 0c 0d 00 00
> This is encoding, not decoding. So we can't create an unknown attribute in the packet.
> 
> The RFC says that the field is 16 octets. If the admin sets a UUID which is shorter, then the encoder pads out the field to 16 octets.
> 
This kind of behaviour can lead to hard to debug issues. My strong preference would be that If the admin has specified an invalid UUID then the attribute should be skipped or the encoder should fail. This should be the same for any fixed size string or octets attributes.

Adding padding hasn’t made the value valid, it’s just masked the problem.

alandekok:
Other DHCP clients / servers just pad the fields.  So it seems reasonable, and common practice.

arr2036:
Yeah that just means they're doing the wrong thing too :)

PR #3122, opened 2019-11-13
Add tests for DHCPv6 packets "client <-> server"

jpereira:


arr2036:
Can you add the test packets from: 

- https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=DHCPv6.pcap
- https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=dhcpv6.pcap

It's a good idea to mix in real-world data with the invented packets when doing these kinds of tests.

jpereira:
Both links are pointing to the same file.

btw, I didn't invent any flow/packets. The current new tests in packet.txt were extracted from the communication between *my-notebook <-> router-ipv6*.

therefore, the only _missing_ flow is the `Release (client) <-> Reply (server)`. (I will add it now)

arr2036:
I switched the link to the wireshark testing corpus.   Your test data doesn't cover PD, and a number of other different modes/operations.

jpereira:
added steps 5. (release) and 6. (reply).

PR #3123, opened 2019-11-13
Update copyright header for DHCPv6 unit tests

jpereira:


arr2036:
These should all be CC-BY to encourage interoperability testing, and the only test files that were developed exclusively under contract were the DHCPv6 tests.  For all other tests it is not appropriate to change the copyright notices without first doing some investigative work to figure out who authored them. If it was Alan then he's welcome to change the copyright to NR, if it was me, then it'll be a mix of contract work, and work outside of contract.

For example, even though the EAP-AKA and EAP-AKA' modules were developed under contract, the base encoder/decoder for EAP-SIM, EAP-AKA, EAP-AKA' was developed on my own time, because I didn't want to extend the poor quality code that was already there.  There's a clear split between the state machine and the protocol encoder/decoder.  The tests here deal exclusively with the protocol side, so I retain the copyright.

arr2036:
My preference would be to change the copyright on the DHCPv6 code to NR as that's most definitely being developed under contract, and either leave the rest with no copyright, or change it to the "The FreeRADIUS Project".  But we'd need agreement from @alandekok.

PR #3124, opened 2019-11-14
Add coa-relay site

terryburton:
I'm not sure whether the "Do not include Proxy-State in CoA/Disconnect-Requests" patch is required or desirable.

We don't need Proxy-State in the requests that we originate as we're not proxying. FR adds Proxy-State during IO so I don't think we can filter the attribute in the configuration.

However, the RFCs seem to indicate that NASs should accept a CoA/Disconnect and treat the Proxy-State as opaque data, so hopefully it's harmless in practise.

Perhaps we need a means to indicate whether we're proxying or originating requests?

Drop the patch if necessary.

### DIFF
--- a/raddb/mods-available/radius
+++ b/raddb/mods-available/radius
@@ -107,6 +107,17 @@ radius {
 #	no_connection_fail = no
 
 	#
+	#  originate::
+	#
+	#  Sometimes we are creating a request that is not for the purpose of
+	#  proxying another request, in which case we do not want to add a
+	#  Proxy-State attribute.
+	#
+	#  In some cases, such as originating a CoA or Disconnect request,
+	#  including Proxy-State may confuse the receiving NAS.
+#	originate = no
+
+	#
 	#  status_checks { ... }:: For "are you alive?" queries.
 	#
 	#  If the home server does not respond to proxied packets, the
--- a/raddb/sites-available/coa-relay
+++ b/raddb/sites-available/coa-relay
@@ -1,0 +1,283 @@
+# -*- text -*-
+######################################################################
+#
+#  This virtual server simplifies the process of sending CoA-Request or
+#  Disconnect-Request packets to a NAS.
+#
+#  This virtual server will receive CoA-Request or Disconnect-Request
+#  packets that contain *minimal* identifying information.  e.g. Just
+#  a User-Name, or maybe just an Acct-Session-Id attribute.  It will
+#  look up that information in a database in order to find the rest of
+#  the session data.  e.g. NAS-IP-Address, NAS-Identifier, NAS-Port,
+#  etc.  That information will be added to the packet, which will then
+#  be sent to the NAS.
+#
+#  This process is useful because NASes require the CoA packets to
+#  contain "session identification" attributes in order to to do CoA
+#  or Disconnect.  If the attributes aren't in the packet, then the
+#  NAS will NAK the request.  This NAK happens even if you ask to
+#  disconnect "User-Name = bob", and there is only one session with a
+#  "bob" active.
+#
+#  Using this virtual server makes the CoA or Disconnect process
+#  easier.  Just tell FreeRADIUS to disconnect "User-Name = bob", and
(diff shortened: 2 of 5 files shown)


alandekok:
Many NAS implementors read the requirements of RFC 3576 badly, and will reject CoA packets that contain `Proxy-State`.  However, the "Do not include Proxy-State in CoA/Disconnect-Requests" patch removes Proxy-State for *all* CoA packets, which isn't what we want.

RFC 8559 gives suggestions as to what to do, but the suggestions are necessarily vague.  We should really just use the `filter` section in the CoA proxy virtual server which removes attributes that the NAS doesn't like.

I'm also not sure why there's a need to update the `rlm_radius` module to add a `recv coa` method.  That shouldn't be necessary.  We will be deleting all of the `recv-coa / send coa / pre-proxy / post-proxy` methods from all of the modules at some point.

terryburton:
> Many NAS implementors read the requirements of RFC 3576 badly, and will reject CoA packets that contain `Proxy-State`. However, the "Do not include Proxy-State in CoA/Disconnect-Requests" patch removes Proxy-State for _all_ CoA packets, which isn't what we want.
> 
> RFC 8559 gives suggestions as to what to do, but the suggestions are necessarily vague. We should really just use the `filter` section in the CoA proxy virtual server which removes attributes that the NAS doesn't like.

Filtering didn't work because the Proxy-State was re-added when calling the radius client module. Hence I neutered the code to provoke the discussion. Perhaps there's another way to avoid this that I missed.

> I'm also not sure why there's a need to update the `rlm_radius` module to add a `recv coa` method. That shouldn't be necessary. We will be deleting all of the `recv-coa / send coa / pre-proxy / post-proxy` methods from all of the modules at some point.

In a `recv CoA-Request` or `recv Disconnect-Request` section the plain call (no section/method override) to the `radius-coa.192.0.2.1` module fails. I still get "`The "radius" module does not have a '(null)' method.`" which this fixed.

alandekok:
Ah, yes.  The rlm_radius module should probably have a flag to control adding `Proxy-State`.  It should add it by default for most packets, but potentially skip it for CoA packets.  Because NASes are dumb. :(

alandekok:
I've also pushed changes to fix the `The "radius" module does not have a '(null)' method.` message.  The `recv CoA-Request` section should just call the `radius` method, and not `radius.coa`.

terryburton:
> I've also pushed changes to fix the `The "radius" module does not have a '(null)' method.` message. The `recv CoA-Request` section should just call the `radius` method, and not `radius.coa`.

I presume you mean call the radius module without supplying the method, which it already does. Where I call `radius-coa-192.0.2.1` (an instance of the `radius` module) are you seeing `radius.coa`. Perhaps I named things confusingly.

To which patch do you refer? (Maybe 2c7be06ffcfb644b7e3c28dd33690349b9d38083.) That was included during my testing and I can't see immediately how that helps.

The function that we want to invoke is `mod_process` which is only referenced from the "section-based" methods:

```
extern module_t rlm_radius;
module_t rlm_radius = {
...
	.methods = {
		[MOD_PREACCT]		= mod_process,
		[MOD_ACCOUNTING]	= mod_process,
		[MOD_AUTHORIZE]		= mod_process,
		[MOD_AUTHENTICATE]     	= mod_process,
	},
};
```

We're calling from the `recv CoA-Request` section which isn't mapped.

If my patch is hitting a legacy approach then maybe we could assign a wildcard `method_name`:

```
        .method_names = (module_method_names_t[]){
                { "recv",       CF_IDENT_ANY,   mod_process },
// OR?
                { CF_IDENT_ANY, CF_IDENT_ANY,   mod_process },

                MODULE_NAME_TERMINATOR
        }
```

I might be missing the point in which case let me know and I'll dig further later this week... (It's late and I only have access to a Chromebook.)

alandekok:
Commit 2c7be06 fixes the error message but not the problem.

We really need a wildcard `method_name`: like you suggested.

   { "recv",       CF_IDENT_ANY,   mod_process },

PR #3126, opened 2019-11-15
Add 'partial_dns_label' decode for dhcpv6

jpereira:
It is a continuation of 88be676

### DIFF
--- a/src/protocols/dhcpv6/decode.c
+++ b/src/protocols/dhcpv6/decode.c
@@ -409,6 +409,7 @@ static ssize_t decode_option(TALLOC_CTX *ctx, fr_cursor_t *cursor, fr_dict_t con
 	ssize_t			rcode;
 	fr_dict_attr_t const	*da;
 	fr_dhcpv6_decode_ctx_t	*packet_ctx = decoder_ctx;
+	VALUE_PAIR *vp;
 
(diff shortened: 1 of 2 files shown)


src/protocols/dhcpv6/decode.c
```diff
@@ -438,26 +439,34 @@ static ssize_t decode_option(TALLOC_CTX *ctx, fr_cursor_t *cursor, fr_dict_t con
...
+	 *	RFC 4704 says "FQDN", unless it's a single label, in which case it's a
+	 *	partial name, and we omit the trailing zero.
+	 */
+	for (vp = fr_cursor_head(cursor); vp; vp = fr_cursor_next(cursor)) {
+		if (vp->da->type == FR_TYPE_STRING &&
+			vp->da->flags.subtype == FLAG_ENCODE_PARTIAL_DNS_LABEL) {
+			fr_pair_value_bstrncpy(vp, vp->vp_strvalue+1, vp->vp_length-1);
+		}
+	}
+
```
alandekok: This decoding really belongs in the `fr_struct_from_network()` function.  It should have a `decode_value()` callback, just like the `encode_value()` callback in `fr_struct_to_network()`.

That way it works for arrays of DNS labels inside of a struct, too

PR #3138, opened 2019-11-19
Fix formatting in dictionary files

jpereira:
$ make dictionary.format


share/dictionary/control/dictionary
```diff
@@ -1,7 +1,8 @@
-## -*- text -*-
+# -*- text -*-
 # Copyright (C) 2019 The FreeRADIUS Server project and contributors
 # This work is licensed under CC-BY version 4.0 https://creativecommons.org/licenses/by/4.0
 # Version $Id$
```
arr2036: We need two text markers?

Issue #3141, opened 2019-11-19
rlm_sql_mysql: ERROR 1406 (Data too long for column 'nasportid' at row 1): 


Hadronx:
# Issue type

- Defect - Unexpected behaviour 

# Defect description
## How to reproduce issue
Use a switch Cisco 2960-x as autenticator 
 
## Output of ``[freeradius] -X`` showing issue occurring
[...]
(27) sql: SQL-User-Name set to 'host/user1'

(27) sql: EXPAND INSERT INTO radacct (acctsessionid,		acctuniqueid,		username, realm,			nasipaddress,		nasportid, nasporttype,		acctstarttime,		acctupdatetime, acctstoptime,		acctsessiontime, 	acctauthentic, connectinfo_start,	connectinfo_stop, 	acctinputoctets, acctoutputoctets,	calledstationid, 	callingstationid, acctterminatecause,	servicetype,		framedprotocol, framedipaddress,	framedipv6address,	framedipv6prefix, framedinterfaceid,	delegatedipv6prefix) VALUES ('%{Acct-Session-Id}', '%{Acct-Unique-Session-Id}', '%{SQL-User-Name}', '%{Realm}', '%{NAS-IP-Address}', '%{%{NAS-Port-ID}:-%{NAS-Port}}', '%{NAS-Port-Type}', FROM_UNIXTIME(%{integer:Event-Timestamp}), FROM_UNIXTIME(%{integer:Event-Timestamp}), NULL, '0', '%{Acct-Authentic}', '%{Connect-Info}', '', '0', '0', '%{Called-Station-Id}', '%{Calling-Station-Id}', '', '%{Service-Type}', '%{Framed-Protocol}', '%{Framed-IP-Address}', '%{Framed-IPv6-Address}', '%{Framed-IPv6-Prefix}', '%{Framed-Interface-Id}', '%{Delegated-IPv6-Prefix}')
[...]

PR #3143, opened 2019-11-19
Increase the 'nasportid' SQL field for 'varchar(32)'. ref #3141

jpereira:


mcnewton:
Can you make them consistent with Oracle? May as well all be the same.

doc/ChangeLog
```diff
@@ -2,6 +2,7 @@ FreeRADIUS 3.0.21 Thu 14 Nov 2019 12:00:00 EDT urgency=low
 	Feature improvements
 
 	Bug fixes
+	* Increase the 'nasportid' SQL field for 'varchar(30)'. #3141
```
mcnewton: and here, and the commit description

jpereira:
@mcnewton done

PR #3144, opened 2019-11-19
Increase the 'nasportid' SQL field for 'varchar(32)'. ref #3141

jpereira:


jpereira:
Backporting from v3. But, I noticed that exists a previous increase to `50`. so, should we keep all being `32`?

herwinw:
In Postgresql, there is no real advantage on specifying a field as `VARCHAR(x)` over `TEXT` (unless you actually want to limit the size of the fields). If this is true for the other databases as well, I can imagine an easier fix.

Slightly offtopic, but defining `nasipaddress` as a string of length 15 is not really compatible with IPv6

PR #3155, opened 2019-11-22
The dhcpv6 'packet_*.txt' should decode only the hex

jpereira:
It was updated using the tool scripts/pcap2decode-proto.py

e.g:

```
./scripts/pcap2decode-proto.py -p dhcpv6 \
	-f ../tcpdump.git/tests/dhcpv6-ia-pd.pcap \
	> src/tests/unit/protocols/dhcpv6/packet_ia-pd.txt
```


jpereira:
The tool can be found in https://github.com/FreeRADIUS/freeradius-server/pull/3154

PR #3156, opened 2019-11-23
Generate OID-containing attributes for unknown X509v3 extensions

lytboris:
Sometimes we need to process a custom X509v3 extension embedded by CA (e.g. MS CA) to authenticate a user properly.

This patch allows one to do that: just create custom attribute in raddb/dictionary and you're done.

### DIFF
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -2360,9 +2360,37 @@ int cbtls_verify(int ok, X509_STORE_CTX *ctx)
 				if (*p == ' ') *p = '-';
 			}
 
-			X509V3_EXT_print(out, ext, 0, 0);
-			len = BIO_read(out, value , sizeof(value) - 1);
-			if (len <= 0) continue;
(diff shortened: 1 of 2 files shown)


src/main/tls.c
```diff
@@ -2360,9 +2360,37 @@ int cbtls_verify(int ok, X509_STORE_CTX *ctx)
...
+#else
+				ASN1_STRING *srcasn1p;
+				srcasn1p = X509_EXTENSION_get_data(ext);
+				srcp = ASN1_STRING_data(srcasn1p);
+#endif
+				int j, asn1len;
+				const char hexdigits[] = "0123456789abcdef";
+				asn1len = ASN1_STRING_length(srcasn1p);
+				/* 5 comes from '0x' + .. + \0 */
+				for (j = 0; j < asn1len && dstp < (value + sizeof(value) - 5); j++, srcp++) {
```
alandekok: My only $0.02 here is that there's a function which does this.  `fr_bin2hex(value + 2, src , asn1len);`

PR #3158, opened 2019-11-25
Update dhcpv6 tests with payload description

jpereira:


jpereira:
It was updated using the latest version of https://github.com/FreeRADIUS/freeradius-server/pull/3154

e.g:

```
$ ./scripts/pcap2decode-proto.py -p dhcpv6 -f ../../tcpdump.git/tests/dhcpv6-domain-list.pcap > src/tests/unit/protocols/dhcpv6/packet_domain-list.txt
```

PR #3161, opened 2019-11-26
Import .pcap RADIUS tests from wireshark project

jpereira:
Copied from https://wiki.wireshark.org/SampleCaptures and converted
with:

	./scripts/pcap2decode-proto.py -p radius -f ~/radius_localhost.pcapng \
		> src/tests/unit/protocols/radius/packet_wireshark01.txt

### DIFF
--- a/src/tests/unit/protocols/radius/packet_wireshark01.txt
+++ b/src/tests/unit/protocols/radius/packet_wireshark01.txt
@@ -1,0 +1,797 @@
+#  -*- text -*-
+#  Copyright (C) 2019 Network RADIUS SARL <legal@networkradius.com>
+#  This work is licensed under CC-BY version 4.0 https://creativecommons.org/licenses/by/4.0
+#
+#  Version $Id$
+#
+#  Test vectors for RADIUS protocol
+#
+#  Based on:
+#  - https://wiki.wireshark.org/SampleCaptures
+#  - https://wiki.wireshark.org/SampleCaptures?action=AttachFile&do=get&target=radius_localhost.pcapng
+#
+proto radius
+proto-dictionary radius
+
+#
+#  1.
+#
+# [ RADIUS ]
+#   code      = Access-Request
+#   id        = 103
+#   len       = 87
+#   authenticator= 40b664dbf5d681b2adbd1769515118c8
+#   \attributes\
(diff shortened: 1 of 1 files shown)


src/tests/unit/protocols/radius/packet_wireshark01.txt
```diff
@@ -0,0 +1,797 @@
...
+#    |  type      = NAS-Port
+#    |  len       = 6
+#    |  value     = 123
+#    |[ Message-Authenticator ]
+#    |  type      = Message-Authenticator
+#    |  len       = 18
+#    |  value     = aa5dcc5ab695fb6baf1cec851f601e22
+#
+decode-proto 01 2b 00 4b 6f 75 df 88 62 88 8f 90 d4 f6 ff 19 ae 4f 38 fb 01 07 73 74 65 76 65 02 12 1a 6c 45 1b c3 a8 45 7e 4a 50 39 e3 6c 48 43 b6 04 06 c0 a8 00 1c 05 06 00 00 00 7b 50 12 aa 5d cc 5a b6 95 fb 6b af 1c ec 85 1f 60 1e 22
+match Packet-Type = Access-Request, Packet-Authentication-Vector = 0x6f75df8862888f90d4f6ff19ae4f38fb, User-Name = "steve", User-Password = "\214\202LѷU?dZ\026>\307l\\\301=", NAS-IP-Address = 192.168.0.28, NAS-Port = 123, Message-Authenticator = 0xaa5dcc5ab695fb6baf1cec851f601e22
```
arr2036: We need to know the secret

arr2036:
We need the secret and the ability to set it for a test file

src/tests/unit/protocols/radius/packet_wireshark01.txt
jpereira: The secret is `testing123`

jpereira:
The secret is `testing123`. But, probably it will be necessary to do some extra pokes on the unit_test to handle it. Or perhaps not?

It was extracted from https://wiki.wireshark.org/SampleCaptures

image

alandekok:
We can add a separate way to set the secret.  Likely via a new attribute.  I spent some time looking at ways to pass things from the test harness to the test tools.  The options here are limited.

PR #3166, opened 2019-11-28
Add deb-build-deps target to Makefile

mattrose:
This target will use `mk-build-deps` to create a dependencies package for the freeradius-server build dependencies and uses apt to install all of the build dependencies.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,14 +313,12 @@ dist-tag: freeradius-server-$(RADIUSD_VERSION_STRING).tar.gz freeradius-server-$
 #	Build a debian package
 #
 debian/control: debian/rules
-	debian/rules $?
+	debian/rules $@
 
-freeradius-build-deps_$(RADIUSD_VERSION_STRING)_all.deb: debian/control
-	mk-build-deps $?
-
+# freeradius-build-deps_$(RADIUSD_VERSION_STRING)_all.deb: debian/control
 .PHONY: deb-build-deps
-deb-build-deps: freeradius-build-deps_$(RADIUSD_VERSION_STRING)_all.deb
-	apt-get install -y ./$?
+deb-build-deps: debian/control
+	mk-build-deps -t 'apt-get -o Debug::pkgProblemResolver=yes --no-install-recommends -y' -ir $?
 
 .PHONY: deb
 deb:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -310,6 +312,16 @@ dist-tag: freeradius-server-$(RADIUSD_VERSION_STRING).tar.gz freeradius-server-$
...
 #
+debian/control: debian/rules
+	debian/rules $?
+
+freeradius-build-deps_$(RADIUSD_VERSION_STRING)_all.deb: debian/control
+	mk-build-deps $?
+
+.PHONY: deb-build-deps
+deb-build-deps: freeradius-build-deps_$(RADIUSD_VERSION_STRING)_all.deb
+	apt-get install -y ./$?
```
mcnewton: You can do this all in one step; simpler, and saves problems if the package name changes in future:

    mk-build-deps -ir debian/control

I'm slightly indifferent, but I'm not too sure we should have the Makefile installing (or trying to install) packages on the system? It's also encouraging someone to run `make` as root.

arr2036: @mattrose can you make the change @mcnewton suggested and then I'll merge the PR

mattrose: That change brings up the apt confirmation dialog.  Which means we then can't automate the procedure.

jpereira: @mattrose try:

`export DEBIAN_FRONTEND=noninteractive; mk-build-deps -ir debian/control`

or

`DEBIAN_FRONTEND=noninteractive mk-build-deps -ir debian/control`

jpereira: Ah! It is a makefile. the second option should work.

mcnewton: TBH we should set `DEBIAN_FRONTEND` before all apt-get calls. But I'm surprised it stops for input, I've never seen it ask for confirmation before. Recent change in behaviour?

jpereira: Adding `export DEBIAN_FRONTEND=noninteractive` at the 314 should work fine.

mattrose: DEBIAN_FRONTEND doesn't get rid of the initial confirmation dialog, it just suppresses any interactive prompts from the packages themselves.

mattrose: The only way I can see to change that behaviour is to edit the apt config file itself, which I don't want to do

mcnewton: Should be able to fix with the `-t` option then

mattrose: hmm, I'll give that a try

PR #3170, opened 2019-12-02
Fix Libressl 3.0.2 build

ndowens:
This patch fixes systems with atleast LibreSSL version 3.0.2 for freeradius-server version 3.0.20
Signed-off-by: Nathan Owens <ndowens04@gmail.com>

### DIFF
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -1579,7 +1579,7 @@ done:
 	return 0;
 }
 
-#if OPENSSL_VERSION_NUMBER < 0x10100000L || defined(LIBRESSL_VERSION_NUMBER)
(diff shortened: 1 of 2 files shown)


alandekok:
What was wrong with the previous code?  i.e. did it refuse to compile with older LibreSSL versions ?

We've had people say that the code is OK, so I want to be sure I understand why the change is necessary.

ndowens:
I know it is not compiling with 3.0.2, maybe will update it to atleast for
that version

ndowens:
Updated patch I made, to include 2.8.0 LibreSSL and up since newest freeradius-server didn't work on 2.8.0 either.

ndowens:
Ok this updated patch don't work, so give me a bit to refix

Zirias:
Great to see that now suddenly, checking for just libressl precense isn't a problem any more. Well, at least this works now.

PR #3172, opened 2019-12-02
Fix leak in xlat_tokenize_attribute()

jpereira:


src/lib/server/xlat_tokenize.c
```diff
@@ -338,6 +339,7 @@ static inline ssize_t xlat_tokenize_attribute(TALLOC_CTX *ctx, xlat_exp_t **head
 		if (err == ATTR_REF_ERROR_INVALID_LIST_QUALIFIER) {
 			fr_strerror_printf("Unknown expansion function or invalid list qualifier");
 		}
+		talloc_free(vpt);
```
arr2036: vpt will be NULL as the function just errored?

PR #3179, opened 2019-12-04
Pica8 dictionary

fdurand:
Pica8 dictionary

### DIFF
--- a/share/dictionary.pica8
+++ b/share/dictionary.pica8
@@ -1,0 +1,18 @@
+# -*- text -*-
+# Copyright (C) 2019 The FreeRADIUS Server project and contributors
+#
+# dictionary.pica8 Vendor specfic attributes for pica8
+#
+#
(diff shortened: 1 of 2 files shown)


jpereira:
you are missing to include it in `share/dictionary`

share/dictionary.pica8
```diff
@@ -0,0 +1,18 @@
...
+# Pica8 RADIUS attributes
+# For general information please visit:
+# http://www.pica8.com
+#
+
+VENDOR		Pica8				35098
+
+BEGIN-VENDOR	Pica8
+
+ATTRIBUTE	Pica8			Pica8-AVPair	1	string
```
jpereira: @fdurand it should be:

`ATTRIBUTE		Pica8-AVPair	1	string`

alandekok:
I've manually pulled it over, thanks.  We prefer to avoid intermediate commits of "add / fix / fix"

PR #3186, opened 2019-12-06
Fix rlm_python build with Python 3.6.x

jpereira:
We don't need `'-Wl,-stack_size,1000000'`


jpereira:
It will fix the issue.

```
[jpereira@sugarloaf freeradius-server.git]$ make 1> /dev/null
ld: -stack_size option can only be used when linking a main executable
clang-9: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [build/lib/local/rlm_python.la] Error 1
```

PR #3184, opened 2019-12-05
Fix the 'python_path' (rlm_python/Python2)

jpereira:
The PySys_SetPath() reset the entire python path causing problems to use the existing libraries.

Therefore, we should split the 'python_path' paths and append into 'sys.path'. It is the correct way to keep the default paths allowing to use existing libraries

PR #3189, opened 2019-12-09
Remove unnecessary src/modules/rlm_python3/radiusd_test.py

jpereira:
1. Already exist the correct file raddb/mods-available/python3/example.py
2. It's not Python3 format.


jpereira:
It should be removed to avoid user mistakes as can be seen in https://github.com/FreeRADIUS/freeradius-server/pull/3184#issuecomment-563378500 (probably the `radiusd_test.py` was pushed by mistake.)

PR #3194, opened 2019-12-10
Don't call 'instantiate' or 'detach' if they're not declared.

jpereira:
It's related to the discussion in #3185.


jpereira:
It fixes the false/negative error.

```
Tue Dec 10 19:34:14 2019 : Error: python_function_load - Module 'example' not found
Tue Dec 10 19:34:14 2019 : Error: python_error_log:200, Exception type: <class 'ModuleNotFoundError'>, Exception value: No module named 'example'
Tue Dec 10 19:34:14 2019 : Error: python_function_load - Failed to import python function 'example.instantiate'
Tue Dec 10 19:34:14 2019 : Error: python_error_log:188, Unknown error
Tue Dec 10 19:34:14 2019 : Error: raddb//mods-enabled/python3[1]: Instantiation failed for module "python3"```

PR #3199, opened 2019-12-10
Just call Py_DECREF()

jpereira:


arr2036:
Looks like you have some scan issues to fix

jpereira:
Argghh!! Now is fixed @arr2036

src/modules/rlm_python3/rlm_python3.c
```diff
@@ -221,46 +219,23 @@ static void python_error_log(void)
...
-		Py_DecRef(pExcTraceback);
-	}
-	if (pStr1) {
-		Py_DecRef(pStr1);
-	}
-	if (pStr2) {
-		Py_DecRef(pStr2);
-	}
+	Py_DECREF(pExcType);
+	Py_DECREF(pExcValue);
```
arr2036: What about the rest of the objects?

arr2036: Never mind, it's all been moved around

PR #3202, opened 2019-12-11
Don't leak memory when release the client

jpereira:


alandekok:
I don't think there's a need for this.  the talloc hierarchy will take care of freeing things appropriately.

The `_client_free()` function is only there because the `pending` packets have to be free'd *before* the other fields in the `client` structure.

jpereira:
You're right! thanks

Issue #3203, opened 2019-12-12
Docker - command from docker hub examples


RubyHuntsman:
# Issue type
- Defect - Crash or memory corruption.

# Defect
It's not posible to run container by folowing docker run command with port forward

```text
docker run --name my-radius -p 1812-1813:1812-1813/udp freeradius/freeradius-server
```
## The same docker run command with -X param (last lines of output)

```text
Listening on auth address * port 1812 bound to server default
Listening on acct address * port 1813 bound to server default
Listening on auth address :: port 1812 bound to server default
Listening on acct address :: port 1813 bound to server default
Listening on auth address 127.0.0.1 port 18120 bound to server inner-tunnel
Listening on proxy address * port 48381
Listening on proxy address :: port 51642
Ready to process requests
```

jpereira:
@RubyHuntsman first of all, you could get more information about the first steps to use the container in https://github.com/FreeRADIUS/freeradius-server/tree/v3.0.x/scripts/docker

therefore, _maybe_ that extra parameters requested in https://github.com/FreeRADIUS/freeradius-server/pull/3206 could simplify the first contact.

mcnewton:
Checked this and it all works as expected. I've updated the README to make it clearer that the default image is useless without supplying a local config.

PR #3206, opened 2019-12-13
Add a Docker parameter to set up the Client and Secret. ref #3203

jpereira:
It will simplify the first user contact with our images.

### DIFF
--- a/doc/ChangeLog
+++ b/doc/ChangeLog
@@ -27,6 +27,7 @@ FreeRADIUS 3.0.21 Thu 14 Nov 2019 12:00:00 EDT urgency=low
 	* The rlm_python should append the 'python_path' paths in 'sys.path',
 	  It fixes the expected behavior to use the existing Python modules.
 	  Fixes #3180
+	* Docker images now support parameters to inform the client address. #3203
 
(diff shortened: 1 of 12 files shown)


mcnewton:
This looks like it will forcibly overwrite any `clients.conf` copied in by the user, no? Which means there's no way to use more than one client (mostly unconfigurable) any more.

TBH I'm not really sure what the point of this is. To make use of the Docker images the user has to supply a working `raddb` config anyway, which may as well contain a `clients.conf` file in it.

jpereira:
Actually, it is a good point @mcnewton . thank you!

alandekok:
You could write the clients to a file `raddb/clients-docker.conf`

and then update the main `clients.conf` file to do:

```
$-INCLUDE clients-docker.conf`
```

mcnewton:
It doesn't really solve much because you still have to create the rest of the config. Unless we added extra options to create users file entries, etc. That could get complicated very quickly.

jpereira:
Maybe use the Docker `VOLUME` parameter to persist externally the config directory. maybe like https://hub.docker.com/_/postgres does.

PR #3214, opened 2019-12-17
Add capability to verify undocumented 'FR_CONF.*' in src/*

jpereira:
We are currently crossing all 'FR_CONF.*' in src/modules/rlm_${mod} vs raddb/${mod} and now
also 'FR_CONF.*' in src/*


jpereira:
It will stop failing as soon as we merge previous PRs.

PR #3222, opened 2019-12-24
Fix radmin to ignore command as an empty line.

jpereira:
When we are in interactive mode:

we ignore <enter> commands. e.g:

```
radmin> <enter>
```

therefore, we should ignore also empty lines avoiding to
break the server. e.g:

```
radmin>           <enter>
radmin>           <enter>
radmin> help<enter>
```

### DIFF
--- a/src/modules/proto_control/radmin.c
+++ b/src/modules/proto_control/radmin.c
@@ -491,7 +491,7 @@ static void radmin_free(char *line)
 	if (!use_readline) return;
 #endif
 
-	free(line);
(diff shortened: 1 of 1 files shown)


jpereira:
btw, It does not affect the `-e ` parameter.

e.g:

```
$ build/bin/local/radmin -q -f build/tests/radmin/control-socket.sock -e "show debug level"
8
$ build/bin/local/radmin -q -f build/tests/radmin/control-socket.sock -e "show profile status"
off
$
```

PR #3223, opened 2019-12-24
Fix float point exception in radmin

jpereira:

PR #3224, opened 2019-12-24
Add tests for 'radmin> stats network'

jpereira:


jpereira:
It needs #3223 first.

PR #3235, opened 2020-01-06
Remove ambiguity from radsniff manpage

terryburton:


arr2036:
Why is this an ambiguity?

terryburton:
Error may be the right term. `response-filter` and `request-filter` are switched in the parameters for the respective options.

Plain "filter" aligns with the adoc manpage in v4, although I note that it doesn't differentiate between what I presume are BPF filters for PCAP (for `-f`) vs RADIUS attribute filters (for `-r` and `-R`).

arr2036:
Maybe just refer to the filter as 'attribute-filter' to differentiate from the BPF filter.

terryburton:
> Maybe just refer to the filter as 'attribute-filter' to differentiate from the BPF filter.

Done, and also PR for v4. Thanks.

arr2036:
Perfect, thanks! All merged.

PR #3237, opened 2020-01-07
4.0.x - rest - curl option CURLOPT_CAINFO

nchaigne:
Port to "master" of https://github.com/FreeRADIUS/freeradius-server/pull/2167

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -22,7 +22,18 @@ rest {
 		#
 		#  .Set options about the certificate.
 		#
+		#  Certificate Authorities:
+		#  "ca_file" (libcurl option CURLOPT_ISSUERCERT).
+		#    File containing a single CA, which is the issuer of the server
+		#    certificate.
+		#  "ca_info_file" (libcurl option CURLOPT_CAINFO).
+		#    File containing a bundle of certificates, which allow to handle
+		#    certificate chain validation.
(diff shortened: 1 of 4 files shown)


arr2036:
This introduces configuration item ambiguity.  `ca_file` in other contexts typically means the certificates used to build the chain for the certificate being offered, or the chain to validate the certificate being presented.  That _seems_ more like what ca_info_file does.  The original fix in v3.0.x should have been to swap the curl options used, and add a "certificate_issuer_file" config item to allow the user to specify which cert in the chain was actually used.

nchaigne:
We discussed that at the time on the mailing list for the v3 pull request. The goal was to not change existing configuration, so people using the curl module would not be impacted. The new configuration item name was a compromise.

But maybe with v4 it's acceptable to rename (and break things) ?

alandekok:
Yes, v4 implements breaking changes.  We should "do it the right way" here.

nchaigne:
Alright then. Agreed on the following names ?
"ca_file" for "CURLOPT_CAINFO" (the bundle of certificates) - which is shown as "CAfile" in curl verbose output
"certificate_issuer_file" for "CURLOPT_ISSUERCERT"

arr2036:
It's ok, I'm putting together a patch that documents the rest of the cert configuration options too

nchaigne:
Good documentation :)
Thanks!

PR #3238, opened 2020-01-07
Fix Doxygen comment

jpereira:


alandekok:
There's a ton of them.  Arran is in the middle of reworking the documentation in that file, and will have a fix shortly.

jpereira:
Ah! Ok, 

btw. weird, I've found just one.

```
[jpereira@sugarloaf freeradius-server.git]$ git co v4/doxygen1
Switched to branch 'v4/doxygen1'
[jpereira@sugarloaf freeradius-server.git]$ grep "@parma" -r src/ -l
```

PR #3241, opened 2020-01-08
MySQL schema extensions for processing radacct: Data usage per period

terryburton:
Provides a table, index and SP for extracting per-user data usage within
arbitrary periods using the standard radacct table entries.


terryburton:
If this is accepted I'll do the same for PostgreSQL (then the other DBs as a lower priority) and cherry-pick for v3.0.x.

terryburton:
Typos. Third time lucky...

PR #3244, opened 2020-01-14
Update dictionary.fortinet

spaetow:
Found an updated Fortinet dictionary. Added details to FreeRADIUS


arr2036:
Master too please.  They've now diverged so it's not as simple as cherry-picking over.

spaetow:
See #3247

PR #3252, opened 2020-01-16
rlm_postgresql: Bound the total query time on non-Linux systems too

terryburton:
Non-Linux systems do not decrease the timeval passed in the timeout parameter
to select, so we keep track of the elapsed time ourselves.


arr2036:
Looks good, seems like you found the time manipulation functions :)

terryburton:
Use of the new functions makes things really clear and is a nice abstraction for accessing monotonic time where it is available.

The only thing that wasn't immediately obvious (to me) without looking into the implementation was which functions were for marshalling the time out of the internal representation, e.g. is it *intended* to do basic arithmetic on the output of fr_time(). But there was existing code that fit the bill which I stole.

PR #3253, opened 2020-01-16
rlm_postgresql: Bound the total query time on non-Linux systems too

terryburton:
Non-Linux systems do not decrease the timeval passed in the timeout parameter
to select, so we keep track of the elapsed time ourselves.

### DIFF
--- a/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
+++ b/src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
@@ -74,26 +74,6 @@ static const CONF_PARSER driver_config[] = {
 	CONF_PARSER_TERMINATOR
 };
 
-#define USEC 1000000
-static void tv_sub(struct timeval *end, struct timeval *start,
-		struct timeval *elapsed)
-{
-	elapsed->tv_sec = end->tv_sec - start->tv_sec;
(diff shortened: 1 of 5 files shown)


src/modules/rlm_sql/drivers/rlm_sql_postgresql/rlm_sql_postgresql.c
```diff
@@ -74,6 +74,26 @@ static const CONF_PARSER driver_config[] = {
 	CONF_PARSER_TERMINATOR
 };
 
+#define USEC 1000000
+static void tv_sub(struct timeval *end, struct timeval *start,
```
arr2036: Hmm, this doesn't exist already?

arr2036: Given the care for the master patch, I'm going to assume it doesn't, and humbly request this gets added to the util library.  I would just double-check and grep for uses of struct timeval, because I really thought we had timeval arithmetic functions in v3.

terryburton: So we're never short of a function like this when we need one!...

arr2036: Wow, was that three or four duplications of the same function *sigh*

PR #3255, opened 2020-01-17
Add tests for DHCPv6 / rfc7600

jpereira:
Add for section 4.9.  Provisioning 4rd Parameters to CEs

### DIFF
--- a/share/dictionary/dhcpv6/dictionary.rfc7600
+++ b/share/dictionary/dhcpv6/dictionary.rfc7600
@@ -10,8 +10,7 @@
 #
 ##############################################################################
 
-#  This is really "group" of 4RD-Map-Rule
-ATTRIBUTE	4RD					97	octets		# Complex struct see RFC 7600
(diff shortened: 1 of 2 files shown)


alandekok:
We also need tests for the `4rd` grouped attribute.  RFC 7600 says:

```
      OPTION_4RD DHCPv6 option
      contains at least one encapsulated OPTION_4RD_MAP_RULE option and
      a maximum of one encapsulated OPTION_4RD_NON_MAP_RULE option.
```

we likely need something like

```
encode-pair 4RD = { 4RD-Map-Rule-IPv4-Prefix-Len = 16, 4RD-Map-Rule-IPv6-Prefix-Len = 128, 4RD-Map-Rule-EA-Len = 4, 4RD-Map-Rule-WPK-Authorized = 1, 4RD-Map-Rule-IPv4-Prefix = 0x01020304 }
match 00 61 00 0c 00 62 00 08 10 80 04 01 01 02 03 04
```

PR #3271, opened 2020-01-24
sqlippool fixes

terryburton:


arr2036:
why do we need to fallback, the attribute name should be qualified with the protocol

arr2036:
i.e radius.frames-ip-address

terryburton:
> why do we need to fallback, the attribute name should be qualified with the protocol
> i.e radius.framed-ip-address

Principle of least surprise for the cuddly users?

I can patch the inline documentation if you would rather keep this as a strict check.

arr2036:
...and when we implement Diameter which uses identical attribute names and the module refuses to insert attributes into the request because D is before R, and so the unqualified attribute resolves to the Diameter version, not the RADIUS one, and we don't allow protocol attributes to be mixed in the same request?..

At least with this, they get the error message on startup.  Module configuration items _MUST_ use qualified attribute references because modules (unlike virtual servers) don't have an explicit namespace.

terryburton:
Got it. Thanks for the explanation of D before R. I'll update the patch.

PR #3276, opened 2020-02-04
Backport dictionary changes from v3.0.x

jpereira:
Also formated with `make dictionary.format`


arr2036:
Commits not signed...

PR #3277, opened 2020-02-04
Fix delta stringfy values

jpereira:
Everything called inside STRINGFY() will become a string.


arr2036:
What's the issue here?

alandekok:
I've pushed a different fix which is a bit simpler.  Just use `"4"` instead of `STRINGIFY(4)`

PR #3279, opened 2020-02-06
Add support to handle coredumps on FreeBSD

jpereira:


arr2036:
are you sure you need the configure checks? I’m fairly sure this is a standard header on freebsd

arr2036:
are you sure you need the configure checks? I’m fairly sure this is a standard header on freebsd

arr2036:
are you sure you need the configure checks? I’m fairly sure this is a standard header on freebsd

arr2036:
your get dumpable function is incorrect, it should return the state of the flag not whether the process is being traced. you should add the trace status in the debugger check code, that’s pretty useful.

arr2036:
your get dumpable function is incorrect, it should return the state of the flag not whether the process is being traced. you should add the trace status in the debugger check code, that’s pretty useful.

jpereira:
@arr2036 I updated the PR.

1. Removed the checks by `configure` and now we are only looking for `ifdef __FreeBSD__`
2. About the `fr_get_dumpable_flag()`, as described in the `procctl()` manual https://www.freebsd.org/cgi/man.cgi?query=procctl&sektion=2&manpath=freebsd-release-ports , the `PROC_TRACE_STATUS` keep the status:

i.e:
```
Tracing includes attachment to the
			  process using	the ptrace(2) and ktrace(2), debugging
			  sysctls, hwpmc(4), dtrace(1),	and core dumping.
			  Possible values for the data argument	are:
```

So, what do you think @arr2036 ?

jpereira:
@arr2036 

My test.

1. FreeBSD sysctl settings.

```
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$ sudo sysctl kern.corefile=%N.%P.core
kern.corefile: %N.%P.core -> %N.%P.core
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
```

2. The `allow_core` is properly set.

```
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$ grep allow_core ../v3/radiusd.conf
	allow_core_dumps = yes
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
``` 

3. I appended the below _dummy_ sigsegv into the `src/main/radiusd.c` just to validate.

```
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$ git df src/main/radiusd.c
diff --git a/src/main/radiusd.c b/src/main/radiusd.c
index 9739514509..c7e9874242 100644
--- a/src/main/radiusd.c
+++ b/src/main/radiusd.c
@@ -355,6 +355,9 @@ int main(int argc, char *argv[])
         */
        if (main_config_init() < 0) exit(EXIT_FAILURE);

+       // dummy sigsegv
+       raise (SIGSEGV);
+
        /*
         *  This is very useful in figuring out why the panic_action didn't fire.
         */
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
```

4. I started the service as can be seen. (non-root)

```
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$ ./build/bin/local/radiusd -d ../v3 -D share/ -l stdout
Wed Feb 12 02:48:44 2020 : Info: Starting - reading configuration files ...
Wed Feb 12 02:48:44 2020 : Info: Core dumps are enabled
Segmentation fault (core dumped)
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
```

4. Here is the `core` file with the _fake intentional crash_.

```
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$ gdb ./build/bin/local/radiusd radiusd.77930.core -ex "bt"
Reading symbols from ./build/bin/local/radiusd...
[New LWP 101197]
Core was generated by `./build/bin/local/radiusd -d ../v3 -D share/ -l stdout'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  thr_kill () at thr_kill.S:3
3	thr_kill.S: No such file or directory.
#0  thr_kill () at thr_kill.S:3
#1  0x00000008009f0534 in __raise (s=0xb) at /usr/src/lib/libc/gen/raise.c:52
#2  0x0000000000250cea in main (argc=0x7, argv=0x7fffffffe910) at src/main/radiusd.c:359
(gdb) quit
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
[vagrant@vm-freebsd01 freeradius-server-v3.0.x.git]$
```

jpereira:
Fixed conflicts and rebased. It might be good to have it on next 3.0.21 @alandekok @arr2036

alandekok:
I think this is good to go.

arr2036:
No, it's wrong.  get_dumpable should return the current status of the procctl control, and the code that's in there should be used for detecting whether a debugger is connected.

src/lib/debug.c
```diff
@@ -502,6 +523,29 @@ static int fr_get_dumpable_flag(void)
 	if (ret != 1) return 0;
 	return 1;
 }
+#elif defined(__FreeBSD__) && defined(PROC_TRACE_STATUS)
+static int fr_get_dumpable_flag(void)
+{
+	int status;
+
+	if (procctl(P_PID, getpid(), PROC_TRACE_STATUS, &status) == -1) {
```
alandekok: should be PROC_TRACE_CTL?

src/lib/debug.c
```diff
@@ -502,6 +523,29 @@ static int fr_get_dumpable_flag(void)
 	if (ret != 1) return 0;
 	return 1;
 }
+#elif defined(__FreeBSD__) && defined(PROC_TRACE_STATUS)
```
alandekok: should be PROC_TRACE_CTL

PR #3282, opened 2020-02-07
Update dictionary.wimax attributes

jpereira:
It was based on the image "[0105] FIG. 9" in https://patents.google.com/patent/WO2009155120A2/en


jpereira:
image

PR #3285, opened 2020-02-08
Added ability to rlm_python show a stacktrace from errors.

jpereira:
Backport from 66e2b9935d51dba8f88b7c09bc5afc8d2c9c288e


arr2036:
you missed the void from the argument list for python_error_log, and this commit needs to be signed.

arr2036:
Or you redeclared python_error_log, not sure which

jpereira:
@arr2036 fixed

PR #3293, opened 2020-02-13
Check first if the func is set

jpereira:


alandekok:
I'm not sure why these are necessary.  The whole point of the trunk code is to send / receive requests.  i.e. mux / demux

what would be better is to update `fr_trunk_alloc()` to check for `func->request_mux` and `func->request_demux`, then fail if those functions aren't provided.  Similar checks can be done for any function which is absolutely required.

PR #3298, opened 2020-02-20
Fix and improve cui/postgresql

jirutka:
### fix postgresql schema for cui
I'm getting the following error with the current schema:

    rlm_sql_postgresql: 22001: STRING DATA RIGHT TRUNCATION

Moreover, in PostgreSQL the text type is highly preferred above varchar. Schema for radacct already uses it, so it seems that schema for cui is just outdated.

I also fixed formatting.

### use real upsert (ON CONFLICT) instead of workaround in postgresql cui 

Clause ON CONFLICT performs better than the RULE workaround and it's guaranted to be atomic (the workaround is not).

It requires PostgreSQL 9.5 or newer; this version was released in 2016 and the previous versions are already unsupported, so it shouldn't be a problem for anyone using latest FreeRADIUS versions.

**There's a problem with backward compatibility because ON CONFLICT doesn't work when there's a RULE ON INSERT defined on the table.** It results in the following error in radius.log:

    Error: rlm_sql_postgresql: 0A000: FEATURE NOT SUPPORTED

### DIFF
--- a/raddb/mods-config/sql/cui/postgresql/queries.conf
+++ b/raddb/mods-config/sql/cui/postgresql/queries.conf
@@ -14,7 +14,7 @@ post-auth {
 		VALUES \
 			('%{%{Packet-Src-IPv6-Address}:-%{Packet-Src-IP-Address}}', '%{Calling-Station-Id}', \
 			'%{User-Name}', '%{reply:Chargeable-User-Identity}') \
-		ON CONFLICT (clientipaddress, callingstationid, username) \
+		ON CONFLICT ON CONSTRAINT ${..cui_table}_pkey \
 		DO UPDATE SET cui = EXCLUDED.cui, lastaccounting = '-infinity'::timestamp"
 
 }
--- a/raddb/mods-config/sql/cui/postgresql/queries.conf
+++ b/raddb/mods-config/sql/cui/postgresql/queries.conf
@@ -5,12 +5,17 @@
 #  $Id$
 
 post-auth {
+	#  Note: Clause ON CONFLICT (aka upsert) was added in PostgreSQL 9.5.
+	#  If you're using an older version, you should upgrade, or use queries.conf
+	#  and schema.sql from FreeRADIUS 3.0.20 or older.
 	query = "\
 		INSERT INTO ${..cui_table} \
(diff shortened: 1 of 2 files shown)


alandekok:
I think this mostly looks OK.  I'm not sure why some things were randomly upper-cased / lower-cased, or whitespaces were added / deleted.  Changes in PRs should be functional.  We shouldn't mix reformatting changes with functional changes.

jirutka:
> I'm not sure why some things were randomly upper-cased / lower-cased, or whitespaces were added / deleted. 

It was inconsistent before.

>  Changes in PRs should be functional. We shouldn't mix reformatting changes with functional changes.

You’re very right, sorry about that. I just rebased it and splitted formatting changes into a separate commit.

terryburton:
This approach looks good to me.

I recently validated that TEXT vs VARCHAR(N) doesn't lead to a loss of index performance.

The ON CONFLICT syntax for upsert is consistent with code that we recently committed:

https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/mods-config/sql/main/postgresql/process-radacct.sql#L106

The only difference as you can see is that we previously named the primary key then referred to it by name in the CONSTRAINT rather than repeat the column list. This makes it more difficult to end up with an incomplete update in the event that the primary key is changed:

https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/mods-config/sql/main/postgresql/process-radacct.sql#L53

@jirutka If you agree that's better then can you please update and retest?

jirutka:
> The only difference as you can see is that we previously named the primary key then referred to it by name in the CONSTRAINT rather than repeat the column list. This makes it more difficult to end up with an incomplete update in the event that the primary key is changed

Okay, pushed another commit with this change. I just used a more compact syntax for defining the named primary key contraint.

terryburton:
> I just used a more compact syntax for defining the named primary key contraint.

@jirutka Nice, thanks.

mcnewton:
@jirutka are you able to sign your commits, please? Then this can be merged. Thanks!

jirutka:
Sure, I’ve rebased it against the current master and signed all the commits.

mcnewton:
Great, thanks!

PR #3303, opened 2020-03-03
Fix __lsan_default_suppressions() return on FreeBSD

jpereira:


jpereira:
It fixes.

```
src/lib/util/debug.c:160:2: error: non-void function '__lsan_default_suppressions' should return a value [-Wreturn-type]
        return
        ^
```

src/lib/util/debug.c
```diff
@@ -162,6 +162,8 @@ char const CC_HINT(used) *__lsan_default_suppressions(void)
 		"leak:libSystem_atfork_child"
 #elif defined(__linux__)
 		"leak:kqueue"
+#else
```
alandekok: to be safe, I would change this to `#elif defined(__FreeBSD__)` so that the change fixes it only for FreeBSD.

i.e. make the smallest change which fixes the problem.

jpereira: Got it. make sense. done

PR #3304, opened 2020-03-03
Fix incompatible pointer to integer conversion

jpereira:


jpereira:
It fixes.

```
[vagrant@vm-freebsd01 freeradius-server.git]$ make -j4
CC src/lib/util/debug.c
CC src/lib/tls/utils.c
LINK build/lib/libfreeradius-ldap.la
src/lib/util/debug.c:156:2: error: non-void function '__lsan_default_suppressions' should return a value [-Wreturn-type]
        return
        ^
LINK build/lib/libfreeradius-json.la
1 error generated.
gmake: *** [scripts/boiler.mk:690: build/objs/src/lib/util/debug.lo] Error 1
gmake: *** Waiting for unfinished jobs....
src/lib/tls/utils.c:197:20: error: arithmetic on a pointer to the function type 'char *(int, int)' is a GNU extension [-Werror,-Wpointer-arith]
        *out = mktime(&t) + timezone;
                          ^ ~~~~~~~~
src/lib/tls/utils.c:197:7: error: incompatible pointer to integer conversion assigning to 'time_t' (aka 'long') from 'char *(*)(int, int)'
      [-Werror,-Wint-conversion]
        *out = mktime(&t) + timezone;
             ^ ~~~~~~~~~~~~~~~~~~~~~
2 errors generated.
gmake: *** [scripts/boiler.mk:690: build/objs/src/lib/tls/utils.lo] Error 1
[vagrant@vm-freebsd01 freeradius-server.git]$```

alandekok:
Hmm.. `timezone()` is a function on the local system.  it shouldn't be cast to `time_t`.  The proper fix is to call `timegm()` instead of `mktime()`

PR #3307, opened 2020-03-03
Fix missing check

jpereira:
it should be used only with USE_READLINE_HISTORY

### DIFF
--- a/src/modules/proto_control/radmin.c
+++ b/src/modules/proto_control/radmin.c
@@ -828,8 +828,9 @@ int main(int argc, char **argv)
 
 	char const	*raddb_dir = RADIUS_DIR;
 	char const	*dict_dir = DICTDIR;
+#ifdef USE_READLINE_HISTORY
 	char 		history_file[PATH_MAX];
-
+#endif
 	TALLOC_CTX	*autofree = talloc_autofree_context();
 
 	char *commands[MAX_COMMANDS];


src/modules/proto_control/radmin.c
```diff
@@ -828,8 +828,9 @@ int main(int argc, char **argv)
 
 	char const	*raddb_dir = RADIUS_DIR;
 	char const	*dict_dir = DICTDIR;
+#ifdef USE_READLINE_HISTORY
 	char 		history_file[PATH_MAX];
```
mcnewton: What about line 1290?

jpereira: take a look at 1066.

mcnewton: That's irrelevant. Put `#undef USE_READLINE_HISTORY` above `int main()`, and try to recompile it.

jpereira: Well, if we are compiling against a system without libreadline. It is supposed to declare the stubs functions:

```
#ifndef USE_READLINE_HISTORY
#	define add_history(line)
#	define stifle_history(line)
#	define read_history()
#	define write_history(history_file)
#endif
```

Therefore, we shouldn't keep `char history_file[PATH_MAX];` to avoid `unused` build errors.

mcnewton: Ah sorry. Macros not in uppercase :-(

PR #3308, opened 2020-03-03
Add %c and %C xlats from v4 for expanding the current epoch time

terryburton:


alandekok:
`%c` and `%C` were already there in commit 33f8e17fb94.  I did miss adding it to the list of one-character expansions.  I'll do that now.

terryburton:
Thanks.

PR #3309, opened 2020-03-03
fixes for radmin build on FreeBSD without libreadline

jpereira:


alandekok:
I've pushed a fix with some edits.  the `cmd_copy()` function needs to be run after every input line, so that's about the only major change.

jpereira:
Great!

PR #3319, opened 2020-03-12
jlibtool: Fix sigsegv in --mode=execute without parameters

jpereira:


jpereira:
Don't crash with:

```
[jpereira@jorge-sugarloaf freeradius-server.git]$ build/make/jlibtool --mode=execute
Segmentation fault: 11
[jpereira@jorge-sugarloaf freeradius-server.git]$
```

PR #3321, opened 2020-03-12
Use new API for fr_dict_enum_by_value()

jpereira:


jpereira:
[use fr_dict_dict_enum_by_value() in more places](https://trello.com/c/EUehLo5F/116-use-frdictdictenumbyvalue-in-more-places)

PR #3325, opened 2020-03-18
Remove MINGW32 ifdef's

jpereira:


jpereira:
Just to complete https://github.com/FreeRADIUS/freeradius-server/commit/77d36acbd25a4d31634edd4455c11c709645e806

PR #3330, opened 2020-03-19
Add reload capability to systemd unit file. ref#3315

jpereira:


arr2036:
We don't handle HUP in master

PR #3333, opened 2020-03-23
v3.0.x Timestamp-related updates to queries

terryburton:
NOTE: These changes haven't been extensively tested with v3 however they have sat in v4 for a long time (tested prior to commit) and do not rely on new server functionality.

Probably only fair to let me manager of the upcoming release decide whether they need to bake for longer...


alandekok:
My only real concern is changing the schema.  Will the old queries be compatible with the new schema, and vice versa?  Such use _should_ be discouraged, but it's good to know what the side effects are when that happens.

terryburton:
> My only real concern is changing the schema. Will the old queries be compatible with the new schema, and vice versa? Such use _should_ be discouraged, but it's good to know what the side effects are when that happens.

Testing with recent versions of MySQL and Oracle (for which there are schema changes) shows that inserting a timestamp with fractional seconds into a field without fractional seconds leads to truncation and a successful insert. The other way around doesn't matter.

So I think we're on safe ground.

PR #3337, opened 2020-03-25
More compile fixes when not having uint128_t

herwinw:
The typecast fixes did make me long for a language with operator overloading.

### DIFF
--- a/src/lib/util/uint128.h
+++ b/src/lib/util/uint128.h
@@ -62,7 +62,7 @@
 /** Create a 128 bit integer value with n bits high
  *
  */
-static uint128_t uint128_gen_mask(uint8_t bits)
+static inline uint128_t uint128_gen_mask(uint8_t bits)
 {
 	uint128_t ret;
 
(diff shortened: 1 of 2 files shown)


src/lib/util/uint128.h
```diff
@@ -62,7 +62,7 @@
 /** Create a 128 bit integer value with n bits high
  *
  */
-static uint128_t uint128_gen_mask(uint8_t bits)
+UNUSED static uint128_t uint128_gen_mask(uint8_t bits)
```
alandekok: it's better to just `ifdef` them out... I thought it already did that?

arr2036: the compiler definitely will not warn about unused functions defined in headers.

arr2036: maybe change them to static inline if you're seeing issues

herwinw: `inline` did the trick, I thought that was a C++ thing that didn't exist in C

arr2036:
I don't think you can just do the assignment.  There needs to be something to explicitly copy the low 64bit to a 64bit integer.

arr2036:
Closed by b271e352f5edffbdfe5ff18bc17be9bb15371cf1

PR #3338, opened 2020-03-25
revised *_encodable() functions to be as consistent as possible

jejones3141:
src/protocols/{dhcpv*,radius}/encode.c have functions for iterating over encodable attributes. Here we rewrite them to simplify them and make them as alike as possible (dhcpv6 only encodes boolean attributes if they're true, and radius doesn't check ancestry).

### DIFF
--- a/src/protocols/dhcpv4/encode.c
+++ b/src/protocols/dhcpv4/encode.c
@@ -56,7 +56,7 @@ static inline VALUE_PAIR *next_encodable(fr_cursor_t *cursor, void *encoder_ctx)
 	VALUE_PAIR	*vp;
 	fr_dhcpv4_ctx_t	*packet_ctx = encoder_ctx;
 
-	while ((vp = fr_cursor_next(cursor))) if (is_encodable(packet_ctx->root, vp)) break;
+	do { vp = fr_cursor_next(cursor); } while (vp && !is_encodable(packet_ctx->root, vp));
 	return fr_cursor_current(cursor);
 }
 
--- a/src/protocols/dhcpv6/encode.c
+++ b/src/protocols/dhcpv6/encode.c
@@ -77,7 +77,7 @@ static inline VALUE_PAIR *next_encodable(fr_cursor_t *cursor, void *encoder_ctx)
 	VALUE_PAIR		*vp;
 	fr_dhcpv6_encode_ctx_t	*packet_ctx = encoder_ctx;
 
-	while ((vp = fr_cursor_next(cursor))) if (is_encodable(packet_ctx->root, vp)) break;
+	do { vp = fr_cursor_next(cursor); } while (vp && !is_encodable(packet_ctx->root, vp));
 	return fr_cursor_current(cursor);
 }
 
--- a/src/protocols/radius/encode.c
+++ b/src/protocols/radius/encode.c
@@ -42,18 +42,12 @@ static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
 			      fr_da_stack_t *da_stack, unsigned int depth,
 			      fr_cursor_t *cursor, void *encoder_ctx);
 
-
-/** Determine if the current attribute is encodable, or find the first one that is
- *
- * @param cursor to iterate over.
- * @return encodable VALUE_PAIR, or NULL if none available.
- */
-static inline VALUE_PAIR *first_encodable(fr_cursor_t *cursor)
+static inline bool is_encodable(VALUE_PAIR const *vp)
 {
-	VALUE_PAIR *vp;
+	if (!vp) return false;
+	if (vp->da->flags.internal) return false;
 
-	for (vp = fr_cursor_current(cursor); vp && vp->da->flags.internal; vp = fr_cursor_next(cursor));
-	return fr_cursor_current(cursor);
+	return true;
 }
 
 /** Find the next attribute to encode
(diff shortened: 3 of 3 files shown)


src/protocols/radius/encode.c
```diff
@@ -42,35 +42,34 @@ static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
 			      fr_dict_attr_t const **tlv_stack, unsigned int depth,
 			      fr_cursor_t *cursor, void *encoder_ctx);
 
+static inline bool is_encodable(VALUE_PAIR const *vp)
+{
+	return vp && !vp->da->flags.internal;
```
alandekok: That should really also check for `fr_dict_parent_common()`, in case some manages to add DHCPv4 attributes to a RADIUS packet.

src/protocols/dhcpv4/encode.c
```diff
@@ -36,11 +36,7 @@
 
 static inline bool is_encodable(fr_dict_attr_t const *root, VALUE_PAIR *vp)
 {
-	if (!vp) return false;
-	if (vp->da->flags.internal) return false;
-	if (!fr_dict_parent_common(root, vp->da, true)) return false;
-
-	return true;
+	return vp && !vp->da->flags.internal && fr_dict_parent_common(root, vp->da, true);
```
mcnewton: A small comment for me personally (for a few of these changes) is that I find the more verbose version easier to read.

alandekok: I'd prefer to use the previous code.  Experience shows it's easier to read.

src/protocols/radius/encode.c
jejones3141: Easy enough to add; will do. Also, while looking this afternoon to see who calls fr_dict_parent_common() in general, it looks like src/lib/eap_aka_sim/encode.c has yet another batch of _*encodable() functions. I will give it the same treatment as the ones under protocol/*.

src/protocols/dhcpv4/encode.c
```diff
@@ -70,13 +66,10 @@ static inline VALUE_PAIR *next_encodable(fr_cursor_t *cursor, void *encoder_ctx)
...
 {
-	VALUE_PAIR	*vp;
+	VALUE_PAIR	*vp = fr_cursor_current(cursor);
 	fr_dhcpv4_ctx_t	*packet_ctx = encoder_ctx;
 
-	vp = fr_cursor_current(cursor);
-	if (is_encodable(packet_ctx->root, vp)) return vp;
-
-	return next_encodable(cursor, encoder_ctx);
+	return is_encodable(packet_ctx->root, vp) ? vp : next_encodable(cursor, encoder_ctx);
```
alandekok: We try to avoid ` ... ? ... : ...`, except in function arguments.  It makes things difficult to read.

src/protocols/dhcpv6/encode.c
```diff
@@ -52,18 +52,13 @@ static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
...
-
 	/*
 	 *	False bools are represented by the absence of
 	 *	an option, i.e. only bool attributes which are
 	 *	true get encoded.
 	 */
-	if ((vp->da->type == FR_TYPE_BOOL) && !vp->vp_bool) return false;
-
-	return true;
+	return vp && !vp->da->flags.internal && fr_dict_parent_common(root, vp->da, true) &&
```
alandekok: similar comments here.  4 lines of `if A do B` are easier read than `if A1 &&  A2 && A3 && A4`

src/protocols/dhcpv4/encode.c
arr2036: Yeah, same

src/protocols/radius/encode.c
arr2036: eap_aka_sim/encode.c has different logic.  It's inefficient to use fr_dict_parent_common.  Should just take an `fr_dict_t *` and compare that to da->dict.

src/protocols/dhcpv4/encode.c
jejones3141: That appears to be the consensus, and the more verbose version makes sense if things change--less error-prone and easier to read diffs. So be it.

jejones3141: I'm convinced. Back to the previous code.

src/protocols/dhcpv4/encode.c
jejones3141: OK. Thanks to you and mcnewton; these should ultimately go into the coding standards.

jejones3141:
Thank you all. I will close this pull request and take your advice.

jejones3141:
Reopening per Alan's suggestion.

src/protocols/radius/encode.c
jejones3141: I spoke too soon; RADIUS is at least nonobvious. From the fr_<foo>_ctx_t * that the other encodability checking functions cast the void * they're handed to, you can get to a  fr_dict_t * to compare with da->dict (if that's not the one to compare, please let me know), but fr_radius_ctx_t doesn't  look like it has anything ultimately traceable to a dictionary.

alandekok: You will have to add a `fr_dict_t *dict` to `fr_radius_ctx_t`

there's a `fr_dict_t const *dict_radius;` in `src/protocols/radius/attrs.h`, so you can just initialize it from that:

`packet_ctx = dict_radius`

and that should work.

arr2036: The pair encode functions should all take an `fr_dict_t *`.  That just needs to be passed around to where it's needed. The `fr_*_ctx_t*` arg for the encodability checks can removed.

For RADIUS, DHCPv4, DHCPv6 the functions should be identical and should probably be moved to `static inline` functions in a common header file.  The EAP-SIM-AKA encodability functions can use the common functions too, but they still need that check for bools.

arr2036: ah no, maybe that was just the decode functions.  Really the encode functions need it too.

PR #3340, opened 2020-03-26
Updates dictionary.bskyb

shaunmiles:
* adds new attributes required by the business


arr2036:
Hmm, interesting.  Did they finally finish integration with the Cloud, or if this a new project?

shaunmiles:
The integration happened in a different manner. The Cloud is now fully with Sky and has been rebranded as Sky Wifi.

PR #3344, opened 2020-04-01
Allow use of password preparation methods with rlm_eap_pwd

michael-dev:
RFC 5931 and RFC 8146 specify multiple password preparation methods, in order to enable the reuse of different password hashs stored on server side.

This patch adds support for control:NT-Password and a generic interface using EAP-Pwd-Password-Hash, EAP-Pwd-Password-Salt and EAP-Pwd-Password-Prep attributes to support them all without needing to implement each hash function in rlm_eap_pwd.

Tested against eapol_test from wpa_supplicant/hostapd.

Signed-off-by: Michael Braun <michael-dev@fami-braun.de>

### DIFF
--- a/src/modules/rlm_eap/types/rlm_eap_pwd/rlm_eap_pwd.c
+++ b/src/modules/rlm_eap/types/rlm_eap_pwd/rlm_eap_pwd.c
@@ -41,11 +41,93 @@ USES_APPLE_DEPRECATED_API	/* OpenSSL API has been deprecated by Apple */
 #define MPPE_KEY_LEN    32
 #define MSK_EMSK_LEN    (2*MPPE_KEY_LEN)
 
+/* EAP-PWD can use different preprocessing (prep) modes to mangle the password
+ * before proving to both parties that they both know the same (mangled) password.
+ *
+ * The server advertises a preprocessing mode to the client. Only "none" is
+ * mandatory to implement.
+ *
+ * What is a good selection on the preprocessing mode?
+ *
+ * a) the server uses a hashed password
+ * b) the client uses a hashed password
+ *
+ * a | b | result
+ * --+---+---------------------------------------
+ * n | n | none
+ * n | y | hint needed (cannot know automatically)
+ * y | n | select by hash given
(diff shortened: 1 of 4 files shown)


src/modules/rlm_eap/types/rlm_eap_pwd/rlm_eap_pwd.c
```diff
@@ -153,18 +240,282 @@ static int send_pwd_request (pwd_session_t *session, EAP_DS *eap_ds)
...
+			/* salt len is 1 byte */
+			RDEBUG("EAP-Pwd-Password-Salt too long (more than 255 octets)");
+			goto out;
+		}
+		rad_assert(pw->vp_length <= sizeof(session->salt));
+
+		session->salt_present = 1;
+		session->salt_len = pw->vp_length;
+		memcpy(session->salt, pw->vp_octets, pw->vp_length);
+	}
```
alandekok, line 454: Is there a missing `goto success` here?

michael-dev: PWD_PASSWORD_SALT can only be used in combination with PWD_PASSWORD_HASH - a salt without a hash does not make sense, though a hash does not always require a salt. So  no goto success missing here.

n0emis:
Are there any news on this PR? This would be a highly appreciated feature...

alandekok:
I've pulled this over manually after resolving some merge conflicts.  Please double-check.

n0emis:
@alandekok Cloud you then close this PR as merged or is there a reason to keep it open?

alandekok:
> @alandekok Cloud you then close this PR as merged or is there a reason to keep it open?

I was hope you could double-check it, and report that it works for you, as expected.

n0emis:
Yes, it worked as expected, although I would prefer if the additional config options would be in the provided example configuration.

alandekok:
> I would prefer if the additional config options would be in the provided example configuration.

It *is* possible to submit a PR....

PR #3350, opened 2020-04-03
Python3 fixes

alejandro-perez:
While I was including the novel rlm_python3 in our CI, I've found some issues that are addressed by this PR (one per commit).

Please, note that the last commits updates the `src/modules/rlm_python3/configure` file as a result of the additions made to `src/modules/rlm_python3/configure.ac`.

### DIFF
--- a/src/modules/rlm_python3/configure
+++ b/src/modules/rlm_python3/configure
@@ -589,6 +589,15 @@ targetname
 mod_cflags
 mod_ldflags
 PYTHON3_CONFIG_BIN
+pkgpyexecdir
+pyexecdir
+pkgpythondir
+pythondir
+PYTHON_PLATFORM
+PYTHON_EXEC_PREFIX
+PYTHON_PREFIX
+PYTHON_VERSION
+PYTHON
 CPP
 OBJEXT
 EXEEXT
@@ -648,7 +657,8 @@ CFLAGS
 LDFLAGS
 LIBS
 CPPFLAGS
-CPP'
+CPP
+PYTHON'
 
 
 # Initialize some variables set by options.
@@ -1266,6 +1276,7 @@ Some influential environment variables:
   CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
               you have headers in a nonstandard directory <include dir>
   CPP         C preprocessor
+  PYTHON      the Python interpreter
 
 Use these variables to override the choices made by `configure' or to help
 it to find libraries and programs with nonstandard names/locations.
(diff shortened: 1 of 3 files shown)


alejandro-perez:
I have tested this for:
* Centos 6, 7, and 8
* Debian 8, 9 and 10
* Ubuntu 16.04, 18.04 and 20.04
* Alpine 3.9 and 3.11

src/modules/rlm_python3/rlm_python3.c
```diff
@@ -41,8 +41,15 @@ RCSID("$Id$")
 #include <link.h>
 #endif
 
+/* Since version 3.8, the "m" suffix is no longer available.
```
jpereira, line 44: The comments could follow the other comments.

e.g:

```
/**
  * Since version 3.8, the "m" suffix is no longer available.
  * https://bugs.python.org/issue36707
  */
```

alejandro-perez: Done. I've also updated the comments of the configure.ac to prefix them with `dnl`

alejandro-perez:
It seems these changes are only needed after 307678b268cf8898a65632147ecc40d37ea3f9d3, that is, after having switched over python3-config. That's why Ubuntu 20.04 packagers haven't hit this yet.

alejandro-perez:
Hi guys,
if this hasn¡t been merged because more testing is required (something reasonable), would it be a good idea if I brought it to the dev list?

alandekok:
We just haven't had time to look at it.  We'll take a look.

src/modules/rlm_python3/configure
```diff
@@ -1,51 +1,51 @@
 #! /bin/sh
 # From configure.ac Revision.
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.69.
-#
-#
-# Copyright (C) 1992-1996, 1998-2012 Free Software Foundation, Inc.
-#
+# Generated by GNU Autoconf 2.63.
```
jpereira, line 4: Can you update the `configure` using a version >= 2.69?

jpereira:
the patch looks good. just could be good to update the `configure` using the `autoconf >= 2.69`

src/modules/rlm_python3/configure
alejandro-perez: Yep. Would any autoconf >= 2.69 work,or is there any flavour that you prefer (ie. the one with centos7, the one from debian 10....).

alejandro-perez:
I've rebased the last commit to use autoconf 2.69 and change commit message

src/modules/rlm_python3/configure.ac
```diff
@@ -58,7 +59,13 @@ if test x$with_[]modname != xno; then
 			'`
 		AC_MSG_NOTICE([Sanitized cflags were \"${mod_cflags}\"])
 
-		python3_ldflags=`${PYTHON3_CONFIG_BIN} --ldflags`
+		dnl # From python 3.8, --embed is required
+		dnl # https://bugs.python.org/issue36721
+		if test x"$PYTHON_VERSION" = x"3.8"; then
```
arr2036, line 64: So we want it to break again with 3.9?

src/modules/rlm_python3/configure.ac
```diff
@@ -84,6 +84,7 @@ if test x$with_[]modname != xno; then
 
 		targetname="rlm_python3"
 	fi
+	AC_CHECK_FUNCS([dl_iterate_phdr])
```
arr2036, line 87: Where is this actually used?

src/modules/rlm_python3/configure.ac
alejandro-perez: Well, I didn't know whether that was a change for good, or just for this version.

src/modules/rlm_python3/configure.ac
alejandro-perez: https://github.com/FreeRADIUS/freeradius-server/blob/4b614008d6f25b4a7ed628d85119ee7db543dd46/src/modules/rlm_python3/rlm_python3.c#L976

src/modules/rlm_python3/configure.ac
arr2036: The comment states `From python 3.8, --embed is required`

alejandro-perez: My bad. Probably I meant "For Python 3.8" on that comment.
Having said that, it seems the change is here to stay

alejandro-perez: Comparing version strings can be tricky. There is this macro AX_COMPARE_VERSION from the autoconf-archive. Is it ok to use those?

arr2036: Yeah that's fine

src/modules/rlm_python3/configure.ac
arr2036: Need to run `autoheader` too to update the macro definitions.

alejandro-perez: I typically run `autoreconf` which includes autoheader, so that should be done?

alejandro-perez:
I've pushed a new commit that uses AX_COMPARE_VERSION to check whether Python version is >= 3.8.
I've regenerated `configure` using `autoreconf`

src/modules/rlm_python3/configure.ac
arr2036: Just didn't see any changes in this PR to `autoconf.h.in` which there should be with a new function check.

alejandro-perez: If you mean `config.h.in`, that's because it was there already, so this PR hasn't modified that.

arr2036: Weird, OK.  I guess this must have just been omitted in a previous PR.

PR #3355, opened 2020-04-07
rlm_radius_udp: The recv_buff should be > 0

jpereira:
It will avoid to pass invalid parameters for setsockopt() like:

"radius - Failed setting 'recv_buf': EINVAL: Invalid argument"

### DIFF
--- a/src/modules/rlm_radius/rlm_radius_udp.c
+++ b/src/modules/rlm_radius/rlm_radius_udp.c
@@ -745,7 +745,11 @@ static fr_connection_state_t conn_init(void **h_out, fr_connection_t *conn, void
 	talloc_set_destructor(h, _udp_handle_free);
 
 #ifdef SO_RCVBUF
+#  ifdef __APPLE__
+	if (h->inst->recv_buff_is_set && h->inst->recv_buff > 0) {
+#  else
 	if (h->inst->recv_buff_is_set) {
+#  endif
 		int opt;
 
 		opt = h->inst->recv_buff;


arr2036:
@jpereira you mentioned on slack you thought this was the correct fix, so let's go over why it's not.

1. We only know that setting recv_buff to 0 works on Linux, we don't know if it works on FreeBSD, and we know it definitely doesn't work on macOS.  As FreeBSD and macOS tend to have similar behaviour, the more likely behaviour is that our of our supported operating systems Linux will allow a recv_buff of 0, so it's better just to set it for Linux unless you want to double check behaviour on FreeBSD and ensure it works as expected there too.

2. You're silently ignoring a configuration item, and essentially changing its value at a data sink (by ignoring it).  This creates two issues.  First, it's confusing to debug.  Everywhere else in the code recv_buff remains set at 0, so the assumption is that value was actually applied to the socket. There's a mismatch between the config and the state of the socket, but nothing to indicate that.
Second, the server doesn't give any indication to the user that their config item was ignored and was never applied.
`FR_INTEGER_BOUND_CHECK` and friends modify the values of config items to be within a sane range, AND emit a warning for the user so that they can correct the value.  With your fix there were no bounds checks when `replicate` was true.

arr2036:
I didn't get it first time either.  Needs:
- https://github.com/FreeRADIUS/freeradius-server/commit/975b0530fafb02b29ea41d6d2e336f616035b0e7
- https://github.com/FreeRADIUS/freeradius-server/commit/7ca95179daf31166c7a6adf78552cdc81ad7ba98

PR #3358, opened 2020-04-08
Generalize/replace the "hack" in dhcpv4_encode_xlat() with a new

jejones3141:
function, fr_cursor_filter_current() (the name in initial discussion
was fr_cursor_intersect_current(), but it's unrelated to the intersect
functions). It takes two parameters, a cursor and an "evaluation
function", and looks at successive items, starting at the current
cursor position, until either it runs out or finds one that the
evaluation function approves of. If it runs out, it returns NULL;
otherwise, it returns a pointer to the approved item.

### DIFF
--- a/src/lib/util/cursor.c
+++ b/src/lib/util/cursor.c
@@ -402,6 +402,60 @@ void fr_cursor_merge(fr_cursor_t *cursor, fr_cursor_t *to_append)
 	}
 }
 
+/** Return the first item that satisfies an evaluation function.
+ *
+ * @param[in] cursor	to operate on
+ * @param[in] eval	evaluation function
+ * @param[in] uctx	context for the evaluation function
+ * @return the first item satisfying eval, or NULL if no such item exists
+ */
+void *fr_cursor_filter_head(fr_cursor_t *cursor, fr_cursor_eval_t eval, void *uctx)
+{
+	void *item;
+
+	item = fr_cursor_head(cursor);
+	if (eval(item, uctx)) return item;
+
+	return fr_cursor_filter_next(cursor, eval, uctx);
+}
+
+/** Return the next item, skipping the current item, that satisfies an evaluation function.
+ *
+ * @param[in] cursor	to operate on
+ * @param[in] eval	evaluation function
+ * @param[in] uctx	context for the evaluation function
+ * @return the next item satisfying eval, or NULL if no such item exists
+ */
+void *fr_cursor_filter_next(fr_cursor_t *cursor, fr_cursor_eval_t eval, void *uctx)
+{
+	void *item;
+
+	do {
+		item = fr_cursor_next(cursor);
+	} while (item && !eval(item, uctx));
+
+	return item;
+}
+
+/** Return the next item, starting with the current item, that satisfies an evaluation function.
+ *
+ * @param[in] cursor    to operate on
+ * @param[in] eval      evaluation function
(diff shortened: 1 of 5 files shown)


src/lib/util/cursor.c
```diff
@@ -402,6 +402,21 @@ void fr_cursor_merge(fr_cursor_t *cursor, fr_cursor_t *to_append)
 	}
 }
 
+/** Return the next item from a cursor that also satisfies an evaluation function.
+ *
+ * @param[in] cursor	to operate on
+ * @param[in] eval	evaluation function
+ * @return the next item satisfying eval, or NULL if no such item exists
+ */
+void *fr_cursor_filter_current(fr_cursor_t *cursor, bool (*eval)(void *))
```
arr2036, line 411: To be consistent with the rest of the cursor code there should be two functions `fr_cursor_filter_head` which finds the first instance of an item in the list that matches the filter, and `fr_cursor_filter_next` which finds the next instance of an item in the list that matches the filter.

arr2036, line 411: Can you also define a type for the callback.

arr2036: and add test cases :)

jejones3141: I'll make it consistent.

jejones3141: Yes, I can and will.

arr2036: You should also add an additional `void *uctx` arg and pass it to the filter function.

src/lib/util/cursor.c
```diff
@@ -402,6 +402,56 @@ void fr_cursor_merge(fr_cursor_t *cursor, fr_cursor_t *to_append)
...
+ * @param[in] cursor	to operate on
+ * @param[in] eval	evaluation function
+ * @param[in] uctx	context for the evaluation function
+ * @return the next item satisfying eval, or NULL if no such item exists
+ */
+void *fr_cursor_filter_next(fr_cursor_t *cursor, fr_cursor_eval_t eval, void *uctx)
+{
+	void *item;
+
+	do { item = fr_cursor_next(cursor); } while (item && !eval(item, uctx));
```
alandekok, line 433: Can you put this on multiple lines?  It's clearer that way

src/lib/util/cursor.c
```diff
@@ -402,6 +402,56 @@ void fr_cursor_merge(fr_cursor_t *cursor, fr_cursor_t *to_append)
...
+}
+
+/** Return the next item, starting with the current item, that satisfies an evaluation function.
+ *
+ * @param[in] cursor    to operate on
+ * @param[in] eval      evaluation function
+ * @param[in] uctx	context for the evaluation function
+ * @return the next item satisfying eval, or NULL if no such item exists
+ */
+void *fr_cursor_filter_current(fr_cursor_t *cursor, fr_cursor_eval_t eval, void *uctx)
```
alandekok, line 445: same here..

thanks.

src/lib/util/cursor.c
jejones3141: Done.

src/lib/util/cursor.c
jejones3141: Of course.

src/lib/util/cursor.c
jejones3141: Will do.

PR #3366, opened 2020-04-10
rlm_radius: Fix potential double-free

jpereira:


arr2036:
It's not a potential double free, but it's not a useful free :)

PR #3368, opened 2020-04-11
Added tls authentication and expanded src/lib/curl functionatlity

stephenmac98:
Added fr_curl_easy_tls_init and fr_curl_response_certinfo to src/lib/curl/base.c and FR_CURL_SET_OPTION to src/lib/curl/base.h

### DIFF
--- a/src/lib/curl/base.c
+++ b/src/lib/curl/base.c
@@ -29,7 +29,16 @@
 #include <curl/curl.h>
 #include <talloc.h>
 
-static uint32_t instance_count = 0;
+
+
+static 			int 		instance_count;
+static fr_dict_t	const 		*dict_freeradius; /*internal dictionary for server*/
+
+extern fr_dict_autoload_t rlm_imap_dict[];
+fr_dict_autoload_t rlm_imap_dict[] = {
+	{ .out = &dict_freeradius, .proto = "freeradius" },
+	{ NULL }
+};
 
 /** Initialise global curl options
  *
@@ -52,7 +61,7 @@ int fr_curl_init(void)
 	 *	the free function will also free the
 	 *	memory allocated during SSL init.
 	 */
-	if (fr_openssl_init() < 0) return -1;
+	if (tls_init() < 0) return -1;
 #endif
 
 	ret = curl_global_init(CURL_GLOBAL_ALL);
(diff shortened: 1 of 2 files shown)


src/lib/curl/base.c
```diff
@@ -29,7 +29,16 @@
 #include <curl/curl.h>
 #include <talloc.h>
 
-static uint32_t instance_count = 0;
+
+
+static 			int 		instance_count;
```
arr2036, line 34: There's no reason to add this whitespace or change instance count back to an `int`.

src/lib/curl/base.c
```diff
@@ -52,7 +61,7 @@ int fr_curl_init(void)
 	 *	the free function will also free the
 	 *	memory allocated during SSL init.
 	 */
-	if (fr_openssl_init() < 0) return -1;
+	if (tls_init() < 0) return -1;
```
arr2036, line 64: This should not be changed to tls_init

src/lib/curl/base.c
```diff
@@ -77,10 +86,120 @@ int fr_curl_init(void)
 void fr_curl_free(void)
 {
 	if (--instance_count > 0) return;
-
 #ifdef WITH_TLS
-	fr_openssl_free();
+	tls_free();
```
arr2036, line 90: This should not be changed to tls_free

src/lib/curl/base.c
```diff
@@ -77,10 +86,120 @@ int fr_curl_init(void)
...
+	SET_OPTION(CURLOPT_SSL_VERIFYPEER, (conf->tls_check_cert == true) ? 1L : 0L);
+	SET_OPTION(CURLOPT_SSL_VERIFYHOST, (conf->tls_check_cert_cn == true) ? 2L : 0L);
+
+	return 0;
+error:
+	return -1;
+}
+
+
+int fr_curl_response_certinfo(REQUEST *request, void *handle)
```
arr2036, line 116: this can be changed to type of fr_curl_io_request_t and renamed randle.  Line on 118 can be removed

src/lib/curl/base.h
```diff
@@ -33,6 +33,20 @@ extern "C" {
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
 
+#include <freeradius-devel/server/base.h>
+#include <freeradius-devel/server/module.h>
+#include <freeradius-devel/server/rad_assert.h>
+#include <freeradius-devel/curl/base.h>
```
arr2036, line 39: Curl base.h doesn't need to include itself

src/lib/curl/base.h
```diff
@@ -33,6 +33,20 @@ extern "C" {
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
 
+#include <freeradius-devel/server/base.h>
+#include <freeradius-devel/server/module.h>
+#include <freeradius-devel/server/rad_assert.h>
```
arr2036, line 38: It doesn't have any asserts

src/lib/curl/base.h
```diff
@@ -33,6 +33,20 @@ extern "C" {
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
 
+#include <freeradius-devel/server/base.h>
```
arr2036, line 36: It shouldn't include every other header in the server library, especially as it's not linked against it

src/lib/curl/base.h
```diff
@@ -33,6 +33,20 @@ extern "C" {
...
+#include <freeradius-devel/server/module.h>
```
arr2036, line 37: It doesn't have any module structures

PR #3369, opened 2020-04-13
Added easy tls initialization to curl

stephenmac98:
Added fr_curl_easy_tls_init and fr_curl_response_certinfo to src/lib/curl/base.c and FR_CURL_SET_OPTION and fr_curl_tls_config to src/lib/curl/base.h

### DIFF
--- a/src/lib/curl/base.h
+++ b/src/lib/curl/base.h
@@ -32,6 +32,16 @@ extern "C" {
 #include <curl/curl.h>
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
+#include <freeradius-devel/server/module.h>
+
+#define FR_CURL_SET_OPTION(_x, _y)\
+do {\
+	if ((ret = curl_easy_setopt(randle->candle, _x, _y)) != CURLE_OK) {\
+		char const *_option;\
+		_option = STRINGIFY(_x);\
+		goto error;\
+	}\
+} while (0)
 
 /** Uctx data for timer and I/O functions
  *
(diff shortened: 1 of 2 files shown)


src/lib/curl/base.h
```diff
@@ -55,6 +65,33 @@ typedef struct {
...
+	char const		*tls_ca_file;
+	char const		*tls_ca_issuer_file;
+	char const		*tls_ca_path;
+	char const		*tls_random_file;
+	bool			tls_check_cert;
+	bool			tls_check_cert_cn;
+	bool			tls_extract_cert_attrs;
+} fr_curl_tls_t;
+
+static CONF_PARSER fr_curl_tls_config[] = {
```
arr2036, line 81: This should go in base.c, and the declaration should go in the header.  When you move it to base.c remove the `static` keyword.  When you add the declaration to base.h add the `extern` keyword.

Arrays like this should not go in headers, because the memory will be duplicated wherever they're used.  Moving the array into a source file and exporting the symbol, means there's only one copy, which is referenced by the code that uses it.

src/lib/curl/base.h
```diff
@@ -32,6 +32,16 @@ extern "C" {
 #include <curl/curl.h>
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
+#include <freeradius-devel/server/module.h>
+
+#define FR_CURL_SET_OPTION(_x, _y)\
+do {\
+	if ((ret = curl_easy_setopt(randle->candle, _x, _y)) != CURLE_OK) {\
+		char const *_option;\
```
arr2036, line 40: Looks like you removed the debug statement, which is why CIT is throwing an error and saying _option isn't used.

The original log message was:

`REDEBUG("Failed setting curl option %s: %s (%i)", option, curl_easy_strerror(ret), ret);`

src/lib/curl/base.h
```diff
@@ -32,6 +32,16 @@ extern "C" {
 #include <curl/curl.h>
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
+#include <freeradius-devel/server/module.h>
+
+#define FR_CURL_SET_OPTION(_x, _y)\
+do {\
+	if ((ret = curl_easy_setopt(randle->candle, _x, _y)) != CURLE_OK) {\
```
arr2036, line 39: Add a locally scoped `int _ret` var too and use it in the assignment and log message.

src/lib/curl/base.h
```diff
@@ -32,6 +32,16 @@ extern "C" {
 #include <curl/curl.h>
 #include <freeradius-devel/server/request.h>
 #include <freeradius-devel/util/event.h>
+#include <freeradius-devel/server/module.h>
```
arr2036, line 35: This should be moved to base.c

src/lib/curl/base.h
arr2036: If you're interested, data like this goes in the text memory segment of the library or executable. https://en.wikipedia.org/wiki/Code_segment.

arr2036: edit: If you're interested, memory like this goes in the data segment of the library or binary - https://en.wikipedia.org/wiki/Data_segment

arr2036: Grr, I was right first time, starts off in text, and gets copied to data.

arr2036:
You also need to sign your commits - https://help.github.com/en/github/authenticating-to-github/signing-commits

PR #3373, opened 2020-04-14
Fix the missing libfreeradius-io.a dependency

jpereira:


jpereira:
```Failed executing 'build/make/jlibtool --quiet --mode=execute ./build/bin/local/radwho -h'
dyld: Symbol not found: _fr_worker_request_add
  Referenced from: /Users/jpereira/Devel/FreeRADIUS/freeradius-server.git/build/lib/local//.libs/libfreeradius-server.dylib
  Expected in: flat namespace```

PR #3380, opened 2020-04-16
Add initial code for dlopen() from within make.

jejones3141:


src/build/dlopen.c
```diff
@@ -0,0 +1,61 @@
...
+
+/** Register function(s) with make.
+ * 
+ * @return non-zero value on success, or zero on failure.
+ * @note gmk_add_function() "returns" void, so we can't really say whether it
+ *       succeeded or failed. Thus the return of the constant 1.
+ */
+int mk_dlopen_gmk_setup(void)
+{
+	gmk_add_function("mk_dlopen", &mk_dlopen, 1, 1, 0);
```
alandekok, line 59: There's no reason to have the exported name `mk_dlopen`.  It can just be `dlopen`

jejones3141:
You're right; will make that change.

jejones3141:
Changed and pushed, along with correcting my blunder of omitting stdio.h.
Squashed, of course.

src/build/dlopen.c
```diff
@@ -0,0 +1,62 @@
...
+int plugin_is_GPL_compatible;
+
+/** GNU make-callable layer around dlopen().
+ * 
+ * @return NULL on failure; otherwise, a string. 
+ * 
+ * @note argv[0] should point to the name of the file to open, minus the extension.
+ * 
+ */
+char * mk_dlopen(UNUSED char const *nm, UNUSED unsigned int argc, char **argv)
```
arr2036, line 34: argc should be checked and error if it's not 1

arr2036: (might SEGV if argv[0] is NULL)

src/build/dlopen.c
```diff
@@ -0,0 +1,62 @@
...
+	
+	extended_file_name = gmk_alloc(1 + strlen(DL_EXTENSION) + strlen(argv[0]));
+	sprintf(extended_file_name, "%s%s", argv[0], DL_EXTENSION);
+
+	handle = dlopen(extended_file_name, RTLD_LAZY);
+
+	if (!handle) return NULL;
+	
+	result = gmk_alloc(2);
+	strcpy(result, "1");
```
arr2036, line 48: can you please `dlclose` the handle.

src/build/dlopen.c
```diff
@@ -0,0 +1,62 @@
...
+	void	*handle;
+	char	*extended_file_name;
+	char	*result;
+	
+	extended_file_name = gmk_alloc(1 + strlen(DL_EXTENSION) + strlen(argv[0]));
+	sprintf(extended_file_name, "%s%s", argv[0], DL_EXTENSION);
+
+	handle = dlopen(extended_file_name, RTLD_LAZY);
+
+	if (!handle) return NULL;
```
arr2036, line 45: If the handle is NULL, call dlerror() and pass the message to make's error function

arr2036: `extended_file_name` is never freed.

src/build/dlopen.c
```diff
@@ -0,0 +1,62 @@
...
+char * mk_dlopen(UNUSED char const *nm, UNUSED unsigned int argc, char **argv)
+{
+	void	*handle;
+	char	*extended_file_name;
+	char	*result;
+	
+	extended_file_name = gmk_alloc(1 + strlen(DL_EXTENSION) + strlen(argv[0]));
+	sprintf(extended_file_name, "%s%s", argv[0], DL_EXTENSION);
+
+	handle = dlopen(extended_file_name, RTLD_LAZY);
```
arr2036, line 43: Should be `RTLD_LAZY | RTLD_LOCAL` otherwise you pollute the global symbol table and may get symbol conflicts.

src/build/dlopen.c
arr2036: @alandekok mentioned that make restricts the number of arguments, so you can ignore those comments.

src/build/dlopen.c
arr2036: Returning "1" isn't overly useful.  It'd be better if we could get the path name to the dynamic library.

On Solaris, FreeBSD and Linux this information is available from `dlinfo()`, querying `RTLD_DI_ORIGIN`.

On macos you can iterate over the loaded images, and find the one that matches the pointer returned by `dlopen()`. `_dyld_get_image_name()` will apparently return the absolute path.

Example code is available here: https://stackoverflow.com/questions/20481058/find-pathname-from-dlopen-handle-on-osx man page is available here: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html

jejones3141: I will dlclose() the handle and return the path name.

jejones3141: Shall I not concern myself with Windows for the time being?

alandekok: @jejones3141 We don't run on Windows.  :(  There was some attempt to port it to Windows ~10 years ago.  But most of that has been stripped out.

src/build/dlopen.c
alandekok: > If the handle is NULL, call `dlerror()` and pass the message to make's error function

There is no documented error API for `make`.  :(  the best we can hope for is `fprintf()`

arr2036:
```
char buff[1024];
char *exp;

snprintf(buff, sizeof(buff), "error dlopen: %s", dlerror());

exp = gmk_expand(buff);
gmk_free(exp);
```

It's easy enough to turn that into a function we can call for logging.  Add one for error, warn, info.

src/build/dlopen.c
arr2036: Yeah, honestly you could strip the windows case out of the extension macro.

jejones3141:
I now have it set up to return the full absolute path of the library--was
surprised to find that RTLD_DI_ORIGIN gives back the absolute path of the
directory the library is in, and added a separate make-callable layer
around dlerror(). Now that the main function returns a filename rather than
a handle or something that maps to a handle, I'm inclined to change the
name to reflect that. dlpath, perhaps?

jejones3141:
Experimenting with the function I had that split out calling dlerror(), so
one could put
$(X $(dlerror))
in one's makefile for X in {info, warn, error} showed no output at all with
$(info $(dlerror)) added after a call to find a nonexistent library. Either
there's some issue with when things are evaluated, or somewhere between the
two GNU make itself does some dl*() call (which will make it forget a
previous error). I think we're still stuck with fprintf().

alandekok:
> `$(X $(dlerror))` ... showed no output at all

This is GNU make idiocy.  `$(dlerror)` is presumably the expansion of a variable called `dlerror`.

`$(dlerror )` is a call to the `dlerror` function, with no arguments.

Note the space after `dlerror`?  Yeah.  It's that stupid.

src/build/dlopen.c
```diff
@@ -0,0 +1,130 @@
...
+#include <gnumake.h>
+
+#ifdef __APPLE__
+#  define DL_EXTENSION ".dylib"
+#else
+#  include <linux/limits.h>
+#  define DL_EXTENSION ".so"
+#endif
+
+char		*mk_dlpath(char const *nm, unsigned int argc, char **argv);
```
alandekok, line 20: These functions can be `static`, as they don't need to be exported.

jejones3141: True; will make that change.

src/build/dlopen.c
```diff
@@ -0,0 +1,130 @@
...
+		strcat(dl_pathname, "/");
+		strcat(dl_pathname, extended_file_name);
+	}
+	return dl_pathname;
+}
+#endif
+
+char *mk_dlerror(UNUSED char const *nm, UNUSED unsigned int argc, UNUSED char **argv)
+{
+	char *msg = dlerror();
```
alandekok, line 112: The issue may be that the string returned from `dlerror()` changes for every call to the `dl*` functions.  So if GNU make calls `dlopen()` itself after the call to `mk_dlopen()`, then the error string is trashed.

You may need to call `dlerror()` from the `mk_dlpath()` function, and then cache the string yourself.

Or verify that your `mk_dlerror()` function is actually being called from `$(dlerror )`.  Maybe via `fprintf(stderr, ...)`

GNU make _shouldn't_ generally be calling `dlopen()` itself, but who knows what it does. :(

jejones3141: True. I've corrected the "what if dlopen() returns NULL" issue, and confirmed that I'm not fprintf-ing dlerror(), and strangely now

```
$(info $(dlpath $(BUILD_DIR)/lib/.libs/no-such-file))
$(info $(dlerror ))

gives me 

```
<blank line because I handed it a bogus value>
build/lib/.libs/no-such-file.so: cannot open shared object file: No such file or directory
```

which is what I thought would happen (but didn't when I didn't leave a space in $(dlerror)). Confirmed that it's really calling dlerror().

PR #3386, opened 2020-04-19
Quiet the compiler

jpereira:


jpereira:
It fixes the https://travis-ci.org/github/FreeRADIUS/freeradius-server/jobs/676918188#L2827

PR #3393, opened 2020-04-20
Add SSL support for delta CRLs

jpereira:


raddb/sites-available/tls
```diff
@@ -515,41 +523,42 @@ home_server tls {
...
-	       #  the check succeeds.
-	       #
-	       #  In 2.1.10 and later, this check can be done
-	       #  more generally by checking the value of the
-	       #  TLS-Client-Cert-Common-Name attribute.  This check
-	       #  can be done via any mechanism you choose.
-	       #
-	#	check_cert_cn = %{User-Name}
-	#
+		#
```
alandekok, line 529: can you redo this without the whitespace changes?  That makes it harder to see which changes are important, and which are just whitespace.

arr2036:
It's not clear from the config item what "use_deltas" does, it's not clear from the docs either, what are they and why would we ever want to disable them?

raddb/sites-available/tls
jpereira: Done @alandekok

PR #3396, opened 2020-04-21
rlm_always: xlat for setting module status and rcode via unlang

terryburton:
Parity with: radmin -e 'set module status db_status {alive,dead,notfound,...}'

Introduces a module option `register_xlat` so that existing rlm_always modules do not register xlats with names that potentially conflict with xlats registered by other modules.

If this is along the right lines then I'll work up the same for v4.

### DIFF
--- a/raddb/mods-available/always
+++ b/raddb/mods-available/always
@@ -32,6 +32,26 @@
 #
 #  mpp = <integer>
 #
+#  An xlat based on the instance name can be called to change the status
+#  returned by the instance.
+#
+#  Force the module status to be alive or dead:
+#
+#  %{db_status:alive}
+#  %{db_status:dead}
+#
+#  Update the rcode returned by an alive module (a dead module returns fail):
+#
+#  %{db_status:ok}
+#  %{db_status:fail}
+#  %{db_status:notfound}
+#  ...
+#
+#  The above xlats expand to the current status of the module. To fetch the
+#  current status without affecting it call the xlat with an empty argument:
+#
+#  %{db_status:}
+#
 always reject {
 	rcode = reject
 }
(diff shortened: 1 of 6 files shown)


arr2036:
- xlat name should be the same name as the module.  Realistically we're not going to have more xlats registered for the always modules.
- status argument doesn't seem to be necessary.  We avoid arguments in xlats as much as possible, especially in v3.
- xlat should return existing status so that it can be restored easily if necessary.
- blank argument should just result in the existing status being returned.
- disagree with synonyms, they tend to be more confusing in the long run.

arr2036:
But yes, this is definitely useful.

terryburton:
> * Module names are necessarily unique.  There's no reason for the toggle.

The potential clash is between rlm_always instance names (our and user-controlled) and registered xlat names (mostly ours, but potential for third-party modules).

It's worth dropping the register_xlat option if we can and I don't think we have any out of the box clashes. So we could register xlats for all rlm_always instances and tell people to change their existing instance names if they happen to collide with an xlat registered by some other module.

> * status argument doesn't seem to be necessary.  We avoid arguments in xlats as much as possible, especially in v3.
> * xlat should return existing status so that it can be restored easily if necessary.
> * blank argument should just result in the existing status being returned.

Good. I think that a nicer interface.

> * disagree with synonyms, they tend to be more confusing in the long run.  Unless @alandekok already added them in radmin, in which case I guess we should be consistent _sigh_.

They actually not synonyms. `alive` and `dead` related to the "forcing" state of the module, i.e. a module can be forced dead (returning fail) whilst retaining it's original rcode so that it isn't lost when the module is forced `alive`. That's what's presented through radmin so we should probably aim for the same as it's potentially useful.

terryburton:
* register_xlat toggle removed
* new style interface: status prefix removed; blank to get status
* synonyms retained for parity with radmin

PR #3398, opened 2020-04-21
resource-check virtual server for triggering control flow updates...

terryburton:
... based on the status of some external resource.

### DIFF
--- a/raddb/sites-available/resource-check
+++ b/raddb/sites-available/resource-check
@@ -1,0 +1,140 @@
+# -*- text -*-
+######################################################################
+#
+#  This virtual server provides an example of how to dynamically amend the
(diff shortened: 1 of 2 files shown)


raddb/sites-available/resource-check
```diff
@@ -0,0 +1,140 @@
...
+#    always db_online {
+#        # Default to online
+#        rcode = ok
+#    }
+#
+#  Now trigger the resource checks by sending a server-status request to this
+#  virtual server, as follows:
+#
+#    echo "Message-Authenticator = 0x00" | \
+#      radclient -r 1 -t 3 -q 127.0.0.1:18122 status testing123
```
jpereira, line 37: Here the port could be the default `1812`

PR #3399, opened 2020-04-22
exec: Fix minor leaks

jpereira:
It's always good to release all allocated memory when something goes wrong.


src/lib/server/exec.c
```diff
@@ -826,6 +833,7 @@ int fr_exec_wait_start(REQUEST *request, fr_value_box_t *vb, VALUE_PAIR *env_pai
 				close(to_child[1]);
 			}
 			fr_strerror_printf("Failed opening pipe to read from child - %s", fr_strerror());
+			talloc_free(envp);
```
arr2036, line 836: When you've got a bunch of similar cleanup cases it's better just to add a goto and put an error label at the top of the function.

jpereira: It's true. let me update it.

PR #3400, opened 2020-04-22
rlm_radius: Fix minor memory leak in encode

jpereira:
It's always good to release all allocated memory when something goes wrong.


src/modules/rlm_radius/rlm_radius_udp.c
```diff
@@ -2693,6 +2694,7 @@ static rlm_rcode_t mod_enqueue(void **rctx_out, void *instance, void *thread, RE
 		fr_assert(!u->rr && !u->packet);	/* Should not have been fed to the muxer */
 		fr_trunk_request_free(&treq);		/* Return to the free list */
 		talloc_free(r);
+		talloc_free(u);
```
arr2036, line 2697: Sure if you want a double free.. Look at where that memory is parented from/

PR #3410, opened 2020-04-29
v3.0.x New MariaDB sqlippools SP that emulates SKIP LOCKED using "user locks"

terryburton:
If we're careful we can emulate the SKIP LOCKED pragma by using user locks with
names based on the framedipaddress.

Testing shows greatly improved throughput largely due to preventing thread
starvation under highly concurrent load.

This will no longer be necessary if/when MariaDB adds support for SKIP LOCKED and most of the world has upgraded it (or MySQL > 8.0).

### DIFF
--- a/raddb/mods-config/sql/ippool/mysql/procedure-no-skip-locked.sql
+++ b/raddb/mods-config/sql/ippool/mysql/procedure-no-skip-locked.sql
@@ -1,0 +1,148 @@
+--
+-- A stored procedure to reallocate a user's previous address, otherwise
+-- provide a free address.
+--
+-- NOTE: This version of the SP is intended for MySQL variants that do not
+--       support the SKIP LOCKED pragma, i.e. MariaDB and versions of MySQL
+--       prior to 8.0. It should be a lot faster than using the default SP
+--       without the SKIP LOCKED pragma under highly concurrent workloads
+--       and not result in thread starvation.
+--
+--       It is however a *useful hack* which should not be used if SKIP
+--       LOCKED is available.
+--
+-- WARNING: This query uses server-local, "user locks" (GET_LOCK and
+--          RELEASE_LOCK), without the need for a transaction, to emulate
+--          row locking with locked-row skipping. User locks are not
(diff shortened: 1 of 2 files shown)


raddb/mods-config/sql/ippool/mysql/procedure-no-skip-locked.sql
```diff
@@ -0,0 +1,150 @@
...
+--       without the SKIP LOCKED pragma under highly concurrent workloads
+--       and not result in thread starvation.
+--
+--       It is however a *useful hack* which should not be used if SKIP
+--       LOCKED is available.
+--
+-- WARNING: This query uses server-local, "user locks" (GET_LOCK and
+--          RELEASE_LOCK), without the need for a transaction, to emulate
+--          row locking with locked-row skipping. User locks are not
+--          supported on clusters such as Galera.
```
alandekok, line 17: No cluster...

raddb/mods-config/sql/ippool/mysql/procedure-no-skip-locked.sql
```diff
@@ -0,0 +1,150 @@
...
+--
+-- Using this SP reduces the usual set dialogue of queries to a single
+-- query:
+--
+--   START TRANSACTION; SELECT FOR UPDATE; UPDATE; COMMIT;  ->  CALL sp()
+--
+-- The stored procedure is executed on an database instance within a single
+-- round trip which often leads to reduced deadlocking and significant
+-- performance improvements especially on multi-master clusters, perhaps even
```
alandekok, line 26: cluster?

Perhaps this paragraph should be removed.

terryburton: Good catch. Copy-paste....

terryburton: Pushed. I retained the text explaining why performance is improved without the erroneous reference to clusters.

terryburton:
Thanks.

PR #3416, opened 2020-05-01
tmpl: Fix minor leak

jpereira:
Let's release all allocated memory when something goes wrong.


src/lib/server/tmpl.c
```diff
@@ -2051,6 +2052,7 @@ ssize_t _tmpl_to_atype(TALLOC_CTX *ctx, void *out,
 		talloc_free(tmp_ctx);
 		if (ret < 0) {
 			RPEDEBUG("Failed copying data to output box");
+			talloc_free(*vb_out);
```
alandekok, line 2055: should be `TALLOC_FREE`, which also sets `*vb_out = NULL`

jpereira: yep, I updated it.

PR #3442, opened 2020-05-08
GitHub Actions CI workflow

terryburton:
This is the result of a brief experiment with configuring a GitHub Actions workflow for CI.

Initial goal is to enable GitHub Actions based CI builds mirroring the current Travis CI configuration as far as possible. Coverity scanning is not integrated. MySQL and PostgreSQL fixtures are upgraded to the current Docker image releases.

The matrix is also extended to build in a Ubuntu 18.04 environment with GCC 10 and clang 10.

Builds appear in the Actions tab and should not require any specific authorisations or configuration.

PR #3440 and PR #3441 fixes issues identified by this. Pull these first otherwise most builds will fail. There are some outstanding build/test issues that GCC 10 is revealing.

Main conclusion is that the build times are approximately the same as Travis, however the job start time is quicker and parallelism is wider so you can reasonably add more configurations to the test matrix.

### DIFF
--- a/.github/workflows/freeradius.yml
+++ b/.github/workflows/freeradius.yml
@@ -1,0 +1,187 @@
+name: GitHub Actions CI
+
+on:
+  push:
+  pull_request:
+
+env:
+  ASAN_OPTIONS: symbolize=1 detect_leaks=1 detect_stack_use_after_return=1
+  LSAN_OPTIONS: fast_unwind_on_malloc=0:malloc_context_size=50
+  KQUEUE_DEBUG: yes
+  M_PERTURB: "0x42"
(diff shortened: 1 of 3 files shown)


arr2036:
If you want to maintain this then please use your own fork so we don't keep getting notifications. IMHO there's no reason to investigate GitHub actions when we have two existing CI systems.  If we want a broader matrix across more operating systems then we'll use Jenkins and expand our internal Jenkins capacity.

GitHub actions doesn't solve the two major problems we have with Public CI systems:

- On test failure we still won't have the ability to replicate tests on a machine which we can get console access to.
- We don't have the ability to run tests on (Free|Open)BSD systems.  Which means we don't know if I/O, timer, or process reaping issues are due to libkqueue, epoll, or our code.

PR #3444, opened 2020-05-09
arp: Check that filter is not truncated when reformatting

terryburton:
Silences a warning from gcc > 7.1.

### DIFF
--- a/src/modules/proto_arp/proto_arp_ethernet.c
+++ b/src/modules/proto_arp/proto_arp_ethernet.c
@@ -198,7 +198,10 @@ static int mod_open(fr_listen_t *li)
 		filter = "arp";
 	} else {
 		filter = buffer;
-		snprintf(buffer, sizeof(buffer), "arp and %s", filter);
+		if (snprintf(buffer, sizeof(buffer), "arp and %s", filter) >= (int)sizeof(buffer) ) {
+			PERROR("Filter length exceeds buffer");
+			return -1;
+		};
 	}
 
 	if (fr_pcap_apply_filter(thread->pcap, filter) < 0) {


src/modules/proto_arp/proto_arp_ethernet.c
```diff
@@ -198,7 +198,10 @@ static int mod_open(fr_listen_t *li)
 		filter = "arp";
 	} else {
 		filter = buffer;
-		snprintf(buffer, sizeof(buffer), "arp and %s", filter);
+		if (snprintf(buffer, sizeof(buffer), "arp and %s", filter) >= (int)sizeof(buffer) ) {
```
jpereira, line 201: Probably the `snprintf()` never will return great than `sizeof(buffer)`. so, maybe:

```
char const *pre_filter = "arp and %s";
size_t len = (strlen(pre_filter) + strlen(filter));

if (len >= sizeof(buffer)) {
    PERROR("Filter length exceeds buffer");
    return -1;
}

snprintf(buffer, sizeof(buffer), "%s %s",pre_filter, filter);
...
```

jpereira: Probably the `snprintf()` never will return great than `sizeof(buffer)`. so, maybe:

```
...
char const *pre_filter = "arp and";
size_t len = (strlen(pre_filter) + 1 /*<space>*/ + strlen(filter) + 1 /*\n*/);

if (len >= sizeof(buffer)) {
    PERROR("Filter length exceeds buffer");
    return -1;
}

snprintf(buffer, sizeof(buffer), "%s %s",pre_filter, filter);
...
```

jpereira, line 201: Probably the `snprintf()` never will return great than `sizeof(buffer)`. so, maybe:

```
...
char const *pre_filter = "arp and";
size_t len = (strlen(pre_filter) + 1 /*<space>*/ + strlen(filter) + 1 /*\n*/);

if (len >= sizeof(buffer)) {
    PERROR("Filter length exceeds buffer");
    return -1;
}

snprintf(buffer, sizeof(buffer), "%s %s",pre_filter, filter);
...
```

PR #3446, opened 2020-05-11
Add the ability to move signed/unsigned integers into dbuffs.

jejones3141:
We provide for either wire format or the format of the computer
running the code, and add generic macros that will provide the
appropriate size based on the type of the argument. The names
are now fr_dbuff_encode* and fr_dbuff_net_encode*.

### DIFF
--- a/src/lib/util/dbuff.h
+++ b/src/lib/util/dbuff.h
@@ -29,6 +29,7 @@ RCSIDH(dbuff_h, "$Id$")
 extern "C" {
 #  endif
 
+#include <freeradius-devel/missing.h>
 #include <freeradius-devel/util/debug.h>
 #include <freeradius-devel/util/net.h>
 #include <stdbool.h>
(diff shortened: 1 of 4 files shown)


src/lib/util/dbuff.h
```diff
@@ -410,6 +411,68 @@ static inline ssize_t fr_dbuff_net_from_uint64(fr_dbuff_t *dbuff, uint64_t num)
...
+#define FR_DBUFF_NET_FROM_UINT64_RETURN(_dbuff, _num) FR_DBUFF_RETURN(fr_dbuff_net_from_uint64, _dbuff, _num)
+
+#define fr_dbuff_net_encode(dbuff, value) \
+	_Generic((value), \
+		int16_t		: fr_dbuff_net_encode_int16(dbuff, value), \
+		int32_t		: fr_dbuff_net_encode_int32(dbuff, value), \
+		int64_t		: fr_dbuff_net_encode_int64(dbuff, value), \
+		uint16_t	: fr_dbuff_net_encode_uint16(dbuff, value), \
+		uint32_t	: fr_dbuff_net_encode_uint32(dbuff, value), \
+		uint64_t	: fr_dbuff_net_encode_uint64(dbuff, value) \
```
arr2036, line 466: You should cast `value` explicitly to an integer of the correct width to avoid warnings.  `_Generic` is a bit strange in that the compiler appears to evaluate every possible type for a given use.

jejones3141:
Will do; thanks.

PR #3449, opened 2020-05-13
radiusd: Allow '-l <file>' to catch stdout and stderr

jpereira:
Instead of redirect to /dev/null.

### DIFF
--- a/src/lib/util/log.c
+++ b/src/lib/util/log.c
@@ -801,7 +801,11 @@ int fr_log_init(fr_log_t *log, bool daemonize)
 		dup2(devnull, STDERR_FILENO);
 		log->print_level = false;
 
-	} else if (fr_debug_lvl) {
+	} else if (fr_debug_lvl
+#ifdef WITH_VERIFY_PTR
+	 || log->file
+#endif
+	) {
 		/*
 		 *	If we're debugging, allow STDOUT and STDERR to
 		 *	go to the log file.


alandekok:
I don't think this is the right approach.  The server should not be producing messages on `stdout` or `stderr`.  It should be using the standard logging facilities.

However... for developers, it would still be useful to be able to add debug statements for stdout / stderr.  But this is really only useful for developers and debugging.

One solution would be to leverage `WITH_VERIFY_PTR`, which should *only* be enabled for developers.  And then maybe change the patch to:

```
   } else if (fr_debug_lvl
#ifdef WITH_VERIFY_PTR
 || log->file
#endif
) {
```

That way the default builds work as normal, and developers can still use stdout / stderr.

This is admittedly a but fugly.  But it

jpereira:
Good point. I updated the PR as you suggested. Thanks

arr2036:
there’s legitimate reasons to have stdout/stderr go through the normal logging infrastructure. libldap (for example) dumps debug output directly to stdout. it’s annoying to lose this data. there’s a couple of techniques we can use, i’ll take a look tomorrow

mcnewton:
As much as logging to stdout feels wrong, there's at least a couple of cases where it could be useful. Running inside Docker is one where it expects logs on stdout, otherwise you end up having to bodge some kind of syslog (and possibly fake init) daemon inside the container. I think systemd also likes logs on stdout, but I'd be less concerned about that, people should be using syslog there.

(Trying to run squid proxy inside docker at the moment, and it's being a royal pain to get the access logs via Docker because it won't log them to stdout after dropping privs.)

alandekok:
> As much as logging to stdout feels wrong, there's at least a couple of cases where it could be useful. 

We can always do `-l stdout`, which works.  This PR is more about logging to a file, *and* also allowing `stdout` and `stderr` to do things.  I must admit I've found the current behavior annoying at times.

mcnewton:
Ah OK. I couldn't see why that had seemingly been removed and needed adding again.

PR #3453, opened 2020-05-15
radsniff: Add new test suite

jpereira:
New target 'make test.radsniff'


alandekok:
I've pulled it over manually with some changes on commit 8706eeb384

- remove the output file on error.  This allows you to do `make test.radsniff` repeatedly.  If the tests failed, they will still run the next time
- make the tests depend on the radsniff binary
- I didn't include the 2M packet.  I'm not sure it's necessary?

if we run with ASAN and LSAN, then any issues / leaks will show up even with a small  pcap file.

PR #3456, opened 2020-05-15
radsniff: Fix the tests

jpereira:


jpereira:
Now is fine.

```
[jpereira@jorge-sugarloaf freeradius-server.git]$ make test.radsniff
RADSNIFF-TEST INPUT=radius-auth+acct+coa-100pkts.pcap ARGV="-q -c 10"
[jpereira@jorge-sugarloaf freeradius-server.git]$
```

src/tests/radsniff/all.mk
```diff
@@ -29,7 +29,7 @@ $(OUTPUT)/%.txt: $(DIR)/%.txt $(OUTPUT)/%.pcap $(TESTBINDIR)/radsniff
 	$(Q)if ! $(TESTBIN)/radsniff $(ARGV) -I $(PCAP_IN) -D share/dictionary 1> $(FOUND) 2>&1; then         \
 		echo "FAILED";                                                                                \
 		cat $(FOUND);                                                                                 \
-		echo "RADSNIFF: $(TESTBIN)/radsniff $(ARGV) -I $(PCAP_IN) -D share/dictionary" -xx;           \
```
alandekok, line 32: I added `-xx` so that we could see the output of radsniff.  Otherwise running that command produces *zero* output.

src/tests/radsniff/all.mk
```diff
@@ -15,7 +15,7 @@ $(eval $(call TEST_BOOTSTRAP))
 #
 .PRECIOUS: $(OUTPUT)/%.pcap
 $(OUTPUT)/%.pcap: $(DIR)/%.pcap.gz
-	$(Q)gzcat $< >$@
```
alandekok, line 18: any system which has gzip should have gzcat, no?

jpereira: No. 

```
[root@devbox-freeradius-server-ubuntu1910 freeradius-server.git]# gzcat
bash: gzcat: command not found
[root@devbox-freeradius-server-ubuntu1910 freeradius-server.git]#```

That's the reason for `gunzip -c`

mcnewton: it's just zcat - but that can probably get confused with non-gzip zcat on older systems.

PR #3463, opened 2020-05-20
Make fr_radius_encode_pair() a one-line wrapper around a dbuff version.

jejones3141:


arr2036:
I'll hold off merging this until the rest of the encoder elements are done.  Maybe use this PR to push up commits for each area of the encoder, and then squash it all down at the end.

jejones3141:
Will do; thanks.

arr2036:
@jejones3141 looks like you tripped LGTM CI.  Can you take a look and see what it's complaining about.

jejones3141:
Found it, and will send up a correction. (Also corrected a doxygen mistake.)

src/protocols/radius/encode.c
```diff
@@ -255,40 +262,34 @@ static ssize_t encode_tunnel_password(uint8_t *out, size_t outlen,
...
 }
 
-static ssize_t encode_tlv_hdr_internal(uint8_t *out, size_t outlen,
+static ssize_t encode_tlv_hdr_internal(fr_dbuff_t *dbuff,
 				       fr_da_stack_t *da_stack, unsigned int depth,
 				       fr_cursor_t *cursor, void *encoder_ctx)
 {
 	ssize_t			slen;
-	uint8_t			*p = out;
+	size_t			init_remaining = fr_dbuff_remaining(dbuff);
```
arr2036, line 270: I added `FR_DBUFF_MAX_NO_ADVANCE`.  Maybe it'd be better to use that here to initialise a `fr_dbuff_t` on the stack, then we don't have to initialise a new dbuff for every call for encode_tlv_hdr[_internal].

You can then return at the end with:

`return fr_dbuff_advance(dbuff, fr_dbuff_used(tlv))`

src/protocols/radius/encode.c
```diff
@@ -710,12 +709,46 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
 	return out_p - out;
 }
 
-static ssize_t attr_shift(uint8_t const *start, uint8_t const *end,
+/** Encodes the data portion of an attribute into a dbuff
+ *
+ * Fully migrating away from the original encode_value() will require a revision of
+ * fr_struct_to_network() and revising the fr_encode_value_t type.
+ */
+static ssize_t encode_value_dbuff(fr_dbuff_t *dbuff,
```
arr2036, line 717: I don't think `encode_value` is used outside of this compilation unit, so there's no reason to wrap it?

jejones3141:
I wrapped it so that dbuff-using functions wouldn't have to switch back to
ptr/len to call encode_value(). encode_value() itself has to keep using
ptr/len for now because it is passed to fr_struct_to_network() which
expects that--one can't write a layer around it as one can for
fr_value_box_to_network(). It will have to wait until all callers of
fr_struct_to_network() use dbuffs.

jejones3141:
Thank you. I will look into where I could use FR_DBUFF_MAX_NO_ADVANCE(),
and make sure that line in src/protocols/radius/encode.c is right.

src/protocols/radius/encode.c
jejones3141: Sorry; I thought replying to the email would show up here. The ptr/len version of encode_value() has to stay around because it's passed as a parameter to fr_struct_to_network(), which others use (and which, because it takes function pointer parameters, isn't easy to wrap); I wrote the one layer around it rather than doing it in each caller in src/protocols/radius/encode.c.

src/protocols/radius/encode.c
jejones3141: Thank you. I'll look for places I can use FR_DBUFF_MAX_NO_ADVANCE to advantage.

src/protocols/radius/encode.c
```diff
@@ -25,24 +25,46 @@
...
+	FR_DBUFF_RETURN(encode_concat, _dbuff, _da_stack, _depth, _cursor, _ctx)
+#define ENCODE_RFC_HDR_RETURN(_dbuff, _da_stack, _depth, _cursor, _ctx) \
+	FR_DBUFF_RETURN(encode_rfc_hdr, _dbuff, _da_stack, _depth, _cursor, _ctx)
+#define ENCODE_WIMAX_HDR_RETURN(_dbuff, _da_stack, _depth, _cursor, _ctx) \
+	FR_DBUFF_RETURN(encode_wimax_hdr, _dbuff, _da_stack, _depth, _cursor, _ctx)
+#define ENCODE_VSA_HDR_RETURN(_dbuff, _da_stack, _depth, _cursor, _ctx) \
+	FR_DBUFF_RETURN(encode_vsa_hdr, _dbuff, _da_stack, _depth, _cursor, _ctx)
+#define ENCODE_TLV_HDR_RETURN(_dbuff, _da_stack, _depth, _cursor, _ctx) \
+	FR_DBUFF_RETURN(encode_tlv_hdr, _dbuff, _da_stack, _depth, _cursor, _ctx)
+#define ENCODE_EXTENDED_HDR_RETURN(_dbuff, _da_stack, _depth, _cursor, _ctx) \
```
alandekok, line 47: If these macros are only used once, there's no need to have them as macros.

src/protocols/radius/encode.c
alandekok: It's been 4+ weeks.  Can these commits be updated to use the new API?

The idea behind DBUFF was to avoid these manual length calculations.

src/protocols/radius/encode.c
```diff
@@ -303,14 +319,16 @@ static ssize_t encode_tlv_hdr_internal(uint8_t *out, size_t outlen,
 		vp = fr_cursor_current(cursor);
 	}
 
-	return p - out;
+	return init_remaining - fr_dbuff_remaining(dbuff);
```
alandekok, line 322: This should just be `return fr_dbuff_used(dbuff)`

here, and other places.

src/protocols/radius/encode.c
```diff
@@ -303,14 +319,16 @@ static ssize_t encode_tlv_hdr_internal(uint8_t *out, size_t outlen,
...
 }
 
-static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
+static ssize_t encode_tlv_hdr(fr_dbuff_t *dbuff,
 			      fr_da_stack_t *da_stack, unsigned int depth,
 			      fr_cursor_t *cursor, void *encoder_ctx)
 {
-	ssize_t			slen;
+	ssize_t	slen;
+	uint8_t	*hdr_field = dbuff->p;
```
alandekok, line 330: this is a nitpick: we typically would use `attr` here for radius.  We don't generally use `field` suffixes, as they're pretty much redundant.

src/protocols/radius/encode.c
```diff
@@ -724,7 +772,7 @@ static ssize_t attr_shift(uint8_t const *start, uint8_t const *end,
 	 */
 
 	while (check_len > (255 - hdr_len)) {
-		total += hdr_len;
+		new_hdr_count++;
```
alandekok, line 775: I'm not sure why `total` was changed to `new_hdr_count` and then a later multiplication.  It should just use `total` as before

src/protocols/radius/encode.c
```diff
@@ -735,7 +783,7 @@ static ssize_t attr_shift(uint8_t const *start, uint8_t const *end,
 	 *	"encode_value" function to take into account the header
 	 *	lengths.
 	 */
-	if ((ptr + ptr[1] + total) > end) return (ptr + ptr[1]) - start;
+	if (fr_dbuff_advance(dbuff, new_hdr_count * hdr_len) < 0) return ptr[1];
```
alandekok, line 786: returning `ptr[1]` changes the meaning of this return code.  It should return the number of bytes necessary to encode this data.

The previous code returned the total amount of data encoded since `start`.  this change just returns the total amount of data in this particular attribute.   So at the minimum, returning `ptr[1]` is different from what it was before.

src/protocols/radius/encode.c
```diff
@@ -1387,10 +1419,10 @@ ssize_t fr_radius_encode_pair(uint8_t *out, size_t outlen, fr_cursor_t *cursor,
 			 *	using a different scheme than the "long
 			 *	extended" one.
 			 */
-			ret = encode_concat(out, outlen, &da_stack, 0, cursor, encoder_ctx);
+			ENCODE_CONCAT_RETURN(dbuff, &da_stack, 0, cursor, encoder_ctx);
```
alandekok, line 1422: It's not clear why macros are useful here.

src/protocols/radius/encode.c
jejones3141: `fr_dbuff_used(dbuff)` includes bytes already written to dbuff, so just doing that doesn't work. What is needed is no-advance dbuffs having start set to the parent's p (or more accurately the same thing as their p is set to; the `FR_DBUFF_NO_ADVANCE_RESERVE()` macro has an edge case); actually, I'm not sure why that's not done for all children. Then you can write

```
fr_dbuff_t work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
/* Do what's needed using work_dbuff */
return fr_dbuff_advance(dbuff, fr_dbuff_used(&work_dbuff));
```

alandekok: Yeah, that's the pattern which should be used everywhere.

src/protocols/radius/encode.c
jejones3141: Yes, they can, once FR_DBUFF_MAX_NO_ADVANCE() gives you a dbuff with matching start and p.

src/protocols/radius/encode.c
jejones3141: Agreed; I will change it back.

PR #3470, opened 2020-05-26
radsniff: Add test suite

jpereira:


src/tests/radsniff/all.mk
```diff
@@ -6,42 +6,57 @@
...
 $(OUTPUT)/%.pcap: $(DIR)/%.pcap.gz
 	$(Q)gunzip -c $< > $@
 
-$(OUTPUT)/%.txt: $(DIR)/%.txt $(OUTPUT)/%.pcap $(TESTBINDIR)/radsniff
-	$(eval TARGET   := $(patsubst %.txt,%,$(notdir $@)))
-	$(eval FOUND    := $@)
+#
+#	Run the radsniff commands
+#
+$(OUTPUT)/%.txt: $(DIR)/%.txt | $(TESTBINDIR)/radsniff $(PCAP_IN)
```
alandekok, line 33: Why change to  `| $(TESTBINDIR)/radsniff `?  That's an *order* dependency.  i.e. the `radsniff` program has to exist, but the rule isn't re-run if `radsniff` changes.

Just delete the `|` character.

jpereira: yep, done.

PR #3471, opened 2020-05-27
Add fr_dbuff_rewind() function and tests for it

jejones3141:
It is added to let encoding functions back out of writes that prove unnecessary, e.g.
a header for which no data will fit in the buffer.


arr2036:
This would prevent us ever doing stream parsing, because we might rewind past the start of the buffer.  You can just clone the input dbuff with `FR_DBUFF_NO_ADVANCE`, and then advance the original dbuff before you return from the function

jejones3141:
OK.

PR #3478, opened 2020-06-03
Fix '-fsanitize=' handle

jpereira:
It will avoid having ' -fsanitize=,whatever' instead of '-fsanitize=whatever' or '-fsanitize=whatever,'

### DIFF
--- a/configure
+++ b/configure
@@ -13440,17 +13440,17 @@ done
 
     devldflags="$devldflags"
 
-    fsanitizeflags="$fsanitizeflags,address"
+    fsanitizeflags="address,$fsanitizeflags"
   fi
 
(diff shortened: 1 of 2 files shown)


arr2036:
This not only fails to fix the actual issue, but rather impressively, breaks the argument sanitization that strips the preceding comma at the front of the argument string.

arr2036:
@jpereira this is a repeating theme, your enthusiasm is great, and your work on testing the server is really helpful, but you have to be more diligent when suggesting fixes.  Understand the code first.  I don't have stats for the number of your PRs that are closed, but I'm guessing it's above 50%, which means 50% of the time you took creating them is not productive time, and 50% of the time we took reviewing them is is not productive.

Just take more time before suggesting fixes, understand how the code works, and how your patch alters that.  Ask questions if you don't understand something, if you don't understand the answer ask more questions :)

PR #3483, opened 2020-06-09
v3.0.x: proto_dhcp: Server must echo DHCP-Client-Identifier if given

terryburton:
As per RFC 6842.

### DIFF
--- a/src/modules/proto_dhcp/dhcpd.c
+++ b/src/modules/proto_dhcp/dhcpd.c
@@ -355,6 +355,15 @@ static int dhcp_process(REQUEST *request)
 		if (relay) relay->vp_ipaddr = vp->vp_ipaddr;
 	}
 
+	/*
+	 *	RFC 6842: If there's a DHCP-Client-Identifier ("uid") in the
+	 *	request then echo this in the reply.
+	 */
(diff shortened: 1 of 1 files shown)


src/modules/proto_dhcp/dhcpd.c
```diff
@@ -355,6 +355,15 @@ static int dhcp_process(REQUEST *request)
 		if (relay) relay->vp_ipaddr = vp->vp_ipaddr;
 	}
 
+	/*
+	 *	RFC 6842: If there's a DHCP-Client-Identifier ("uid") in the
+	 *	request then echo this in the reply.
+	 */
+	vp = fr_pair_find_by_num(request->packet->vps, 61, DHCP_MAGIC_VENDOR, TAG_ANY); /* DHCP-Client-Identifier */
+	if (vp) {
```
alandekok, line 363: I'd do:

`if (vp && !fr_pair_find_by_num(request->reply->vps, 61, DHCP_MAGIC_VENDOR, TAG_ANY))`

i.e. if it's in the packet and not in the reply, add it to the reply.

terryburton: That works. Pushed.

PR #3492, opened 2020-06-11
Convert dhcpv4 encoding to use dbuff

jejones3141:


src/protocols/dhcpv4/encode.c
```diff
@@ -73,28 +77,26 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
...
-		memcpy(p, &vp->vp_ipv6addr, sizeof(vp->vp_ipv6addr));
-		p += sizeof(vp->vp_ipv6addr);
+		fr_dbuff_bytes_in(dbuff, vp->vp_ip.prefix);
+		fr_dbuff_memcpy_in(dbuff, (uint8_t const *)&vp->vp_ipv6addr, sizeof(vp->vp_ipv6addr));
 		break;
 
 	case FR_TYPE_IPV6_ADDR:
-		if (outlen < sizeof(vp->vp_ipv6addr)) {
+		len = sizeof(vp->vp_ipv6addr);
+		if (fr_dbuff_remaining(dbuff) < (size_t)len) {
```
arr2036, line 96: This is redundant.  As are the other manual length checks.  You should use the macros which cause the function to return if we're out of space.

jejones3141: Thank you; I'll eliminate those in this and in dhcpv6.

src/protocols/dhcpv4/encode.c
```diff
@@ -106,39 +108,40 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
 	fr_proto_da_stack_build(da_stack, vp ? vp->da : NULL);
 
 	FR_PROTO_STACK_PRINT(da_stack, depth);
-	FR_PROTO_HEX_DUMP(out, (p - out), "Value");
+	FR_PROTO_HEX_DUMP(data_ptr, len, "Value");
 
-	return p - out;
+	return len;
```
arr2036, line 113: Manually recording the len value makes this much more complex than it needs to be.  The pattern should be take a copy of the dbuff when entering the function, and set `start = p`.  Then advance the parent at the end.  Feel free to add a debuff function to do that, or you can just advance it by fr_dbuff_used

arr2036: I don't think there'll be a large performance penalty for this.  In the majority of cases the fr_dbuff_t structures will all be on the stack, and they'll likely be close enough together that at least some of the time they'll be in the same 64 byte cache line so copying between them  should only be slightly slower than a register to register copy.

If you think about it, it's very similar to what we do anyway (setting p to out, and q to p + len), it's just the structure involved is larger.

src/protocols/dhcpv4/encode.c
```diff
@@ -73,28 +74,17 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
 	case FR_TYPE_ETHERNET:
 	case FR_TYPE_STRING:
 	case FR_TYPE_OCTETS:
-		len = fr_value_box_to_network(&need, p, outlen, &vp->data);
-		if (len < 0) return -2;
+		if (fr_value_box_to_network_dbuff(&need, dbuff, &vp->data) < 0) return -2;
```
arr2036, line 77: This return value is wrong, but it was always wrong.  Should return need as a negative integer, not -2.

src/protocols/dhcpv4/encode.c
jejones3141: Agreed. Switched to non-advancing dbuff.

src/protocols/dhcpv4/encode.c
jejones3141: Corrected.

src/protocols/dhcpv4/encode.c
jejones3141: They're changed to use the macros.

src/protocols/dhcpv4/encode.c
```diff
@@ -106,39 +96,40 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
 	fr_proto_da_stack_build(da_stack, vp ? vp->da : NULL);
 
 	FR_PROTO_STACK_PRINT(da_stack, depth);
-	FR_PROTO_HEX_DUMP(out, (p - out), "Value");
+	FR_PROTO_HEX_DUMP(data_ptr, init_remaining - fr_dbuff_remaining(dbuff), "Value");
 
-	return p - out;
+	return init_remaining - fr_dbuff_remaining(dbuff);
```
alandekok, line 101: this should be `fr_dbuff_used()` as with the RADIUS encoder.

src/protocols/dhcpv4/encode.c
```diff
@@ -151,13 +142,13 @@ static uint8_t *extend_option(uint8_t *start, uint8_t *end, uint8_t *p, int len)
...
 		 */
-		sublen = 255 - out[1];
+		sublen = 255 - hdr_field[1];
 		if (sublen > len) sublen = len;
 
 		/*
 		 *	Add in the data left at the current pointer.
 		 */
-		out[1] += sublen;
+		hdr_field[1] += sublen;
```
alandekok, line 151: we should just use `out` here.  there's no need to rename the variables to `hdr_field`.  it doesn't match the style of the rest of the code.

arr2036:
Can you squash these two commits together so we can review

jejones3141:
> Can you squash these two commits together so we can review

Which two?

PR #3493, opened 2020-06-12
Convert DHCPv6 encoding to use dbuff

jejones3141:
Note: the conversion is not complete; fr_struct_to_network() expects a ptr/len value
encoder, which means one can't write a simple dbuff layer around it, and ARP,
DHCPv4, and DHCPv6 all use it. encode_value() thus remains as is for now, with a
dbuff-flavored wrapper.

### DIFF
--- a/src/protocols/dhcpv6/encode.c
+++ b/src/protocols/dhcpv6/encode.c
@@ -28,6 +28,7 @@
 #include <stdint.h>
 #include <stddef.h>
 #include <talloc.h>
+#include <freeradius-devel/util/dbuff.h>
 #include <freeradius-devel/util/pair.h>
 #include <freeradius-devel/util/types.h>
 #include <freeradius-devel/util/proto.h>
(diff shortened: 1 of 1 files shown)


jejones3141:
I found an issue and have pushed a correction: encode_option() used FR_DBUFF_MAX() but not as a function parameter, which meant that the difference in remaining bytes was incorrect. Unit tests didn't notice the issue (and with the change they still all pass).

src/protocols/dhcpv6/encode.c
```diff
@@ -501,57 +521,50 @@ static inline ssize_t encode_array(uint8_t *out, size_t outlen,
...
+static ssize_t encode_tlv(fr_dbuff_t *dbuff,
 			  fr_da_stack_t *da_stack, unsigned int depth,
 			  fr_cursor_t *cursor, void *encoder_ctx)
 {
-	ssize_t			slen;
-	uint8_t			*p = out, *end = p + outlen;
+	fr_dbuff_t		work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
 	VALUE_PAIR const	*vp = fr_cursor_current(cursor);
 	fr_dict_attr_t const	*da = da_stack->da[depth];
+	ssize_t			len;
```
arr2036, line 552: Should be slen as it's a signed length...

PR #3494, opened 2020-06-12
.NET Core module

bcr:
First cut of a .NET Core module. Still missing some functionality.


src/modules/rlm_dotnet/coreclrhost.h
```diff
@@ -0,0 +1,133 @@
+// Retrieved from https://github.com/dotnet/coreclr/blob/master/src/coreclr/hosts/inc/coreclrhost.h
+
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
```
alandekok, line 4: This may cause issues with the GPL license.  I'll have to check.

bcr: OK -- I think I can reproduce the functionality of this file pretty easily if needed.

arr2036:
We're not accepting new modules into v3.0.x.  If you'd like to rework this patch for master branch we'd consider it for inclusion.

PR #3499, opened 2020-06-17
Bump to use latest libkqueue

jpereira:


arr2036:
We now use the NR packages which are the latest version of libkqueue.

PR #3500, opened 2020-06-17
travis: Disable ptrace protection

jpereira:
It will allow the Travis processes be debugged.


jpereira:
By default, the Travis enables the `ptrace_scope` denying to call the `gdb` as can be seen here https://travis-ci.org/github/FreeRADIUS/freeradius-server/jobs/699451872#L5047

Therefore, it would help us to troubleshooting errors that happens only on Travis.

More info about `ptrace_scope` in https://linux-audit.com/protect-ptrace-processes-kernel-yama-ptrace_scope/ https://www.kernel.org/doc/Documentation/security/Yama.txt

PR #3503, opened 2020-06-19
Add fr_dbuff_bits_in() and associated tests

jejones3141:
This will let us convert fr_struct_to_network() and its callers to use dbuffs

### DIFF
--- a/src/lib/util/dbuff.h
+++ b/src/lib/util/dbuff.h
@@ -54,6 +54,7 @@ struct fr_dbuff_s {
 		uint8_t const *p_i;		//!< Immutable position pointer.
 		uint8_t *p;			//!< Mutable position pointer.
 	};
+	uint8_t		bit_offset;		//!< Offset within a byte; always in [0, 8)
 	bool		is_const;		//!< The buffer this dbuff wraps is const.
 	bool		adv_parent;		//!< Whether we advance the parent
 						///< of this dbuff.
@@ -72,6 +73,7 @@ struct fr_dbuff_s {
 	.start	= (_dbuff)->start, \
 	.end	= (_dbuff)->end, \
 	.p	= (_dbuff)->p, \
+	.bit_offset = (_dbuff)->bit_offset, \
 	.is_const = (_dbuff)->is_const, \
 	.adv_parent = false, \
 	.parent = (_dbuff) \
@@ -86,6 +88,7 @@ struct fr_dbuff_s {
 	.p	= ((size_t)(_dbuff) > (_reserve)) && ((_dbuff)->end - (_reserve)) >= ((_dbuff)->p) ? \
 			(_dbuff)->p : \
 			(_dbuff)->end - (_reserve), \
+	.bit_offset = (_dbuff)->bit_offset, \
 	.is_const = (_dbuff)->is_const, \
 	.adv_parent = _adv_parent, \
 	.parent = (_dbuff) \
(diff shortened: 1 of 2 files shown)


src/lib/util/dbuff.h
```diff
@@ -323,8 +343,9 @@ static inline ssize_t CC_HINT(nonnull) _fr_dbuff_advance(fr_dbuff_t *dbuff, size
  */
 static inline ssize_t fr_dbuff_advance(fr_dbuff_t *dbuff, size_t inlen)
 {
-	size_t freespace = fr_dbuff_remaining(dbuff);
+	size_t freespace = fr_dbuff_remaining(dbuff);;
```
arr2036, line 346: double colon

src/lib/util/dbuff.h
```diff
@@ -347,6 +368,7 @@ static inline ssize_t fr_dbuff_memcpy_in(fr_dbuff_t *dbuff, uint8_t const *in, s
 
 	fr_assert(!dbuff->is_const);
 
+	_fr_dbuff_bit_flush(dbuff);
```
arr2036, line 371: Wouldn't it be better to do this in the advance function?

src/lib/util/dbuff.h
jejones3141: *slaps forehead* Thanks; I'll fix that.

src/lib/util/dbuff.h
jejones3141: It would. I will make sure that fr_dbuff_bits_in() can work with that.

arr2036: you set remaining bits after calling advance so it should be fine

alandekok: The dictionary parsing code ensures that normal data types starts on a byte boundary.  The first `bit[]` field must also start on a byte boundary.  Then 0 or more other `bit[]` fields.  The final `bit[]` field must end on a byte boundary.

So the rest of the code can assume this is true, and blow up if it isn't true.

arr2036: I'm not sure this is the right approach in general.  Manually specifying the length is annoying.

```
fr_dbuff_bits_in(dbuff, BIT(MY_BIT_MASK_A, value), BIT(MY_BIT_MASK_B, value), BIT(MY_BIT_MASK_C, value))
```

You `|` the masks together, and take the high bit as the length of the bits you're packing. The maximum number of bits would be 64.

I think there'll be cases where we extract things like 4bit fields from structs into their own attributes, and need to re-encode them.  So being able to call fr_dbuff_bits_in and have it continue at a bit offset would still be useful.

arr2036: Seems like there's different requirements here.  I'll probably poke the function once it's pulled in, so don't worry about it for now

src/lib/util/dbuff.h
jejones3141: Fixed.

src/lib/util/dbuff.h
jejones3141: OK. Aside from the double semicolon, I've left it unchanged.

jejones3141: Pushed a change to add FR_DBUFF_BITS_IN_RETURN()

jejones3141:
Per discussion with Alan, I'm leaving the bits_in() function for fr_struct_to_network() to use, and as part of converting fr_struct_to_network() to use dbuff, we'll convert bits_in() to use dbuff, only writing complete bytes to dbuffs, so that there's no need for fr_dbuff_bits_in().

PR #3507, opened 2020-06-24
Align V3 dhcp_sqlippool config with V4

ndptech:
Rather than putting returned addresses into Framed-IP-Address and then
copying into DHCP-Your-IP-Address, place the IP address directly into
the correct attribute.


arr2036:
Commits need to be signed with your gpg key

PR #3510, opened 2020-07-01
curl version req update

stephenmac98:
Updated the libcurl dependency for imap

### DIFF
--- a/src/lib/curl/m4/libcurl_check_config.m4
+++ b/src/lib/curl/m4/libcurl_check_config.m4
@@ -212,8 +212,8 @@ if (x) ;
 	      fi
 
 	      # RTSP, IMAP, POP3 and SMTP were added in
-	      # 7.20.0 (0x071400 == 463872)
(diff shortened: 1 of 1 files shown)


arr2036:
You need to make the changes to the m4 script and then re-build the configure file with autoconf, otherwise your changes will be overwritten the next time someone runs autoconf.

src/lib/curl/m4/libcurl_check_config.m4
```diff
@@ -212,8 +212,8 @@ if (x) ;
 	      fi
 
 	      # RTSP, IMAP, POP3 and SMTP were added in
-	      # 7.20.0 (0x071400 == 463872)
-	      if test $_libcurl_version -ge 463872; then
+	      # 7.24.0 (0x071800 == 464896)
+	      if test $_libcurl_version -ge 464896 then
```
arr2036, line 216: You stripped the semicolon, which broken the configure script :(

Issue #1756, opened 2016-09-25
EAP-FAST and TLS 1.2


alandekok:
# Issue type
- [x] Feature request.
# Defect/Feature description

EAP-FAST requires `disable_tlsv1_2 = yes`, because the PRF label used in FAST is forbidden by the RFCs.  And, the PRF function in TLS 1.2 is different from the PRF function in earlier versions of TLS.

The solution is to re-implement the TLS 1.2 PRF function in FreeRADIUS, and then make EAP-FAST use it.  For example code, see `tls_connection_prf()` in the hostap repo.

jimdigriz:
Supported by 70dbb66

jimdigriz:
Fixed by https://github.com/FreeRADIUS/freeradius-server/commit/6ae4f72727957909ed9e4b5e80df7be6030dd737

Can now be closed.

PR #3516, opened 2020-07-06
v3.0.x: TLSv1.3 support for EAP-{TLS,TTLS,PEAP}

jimdigriz:
Picking up from where https://github.com/FreeRADIUS/freeradius-server/pull/2449 left off and aligning to [draft-ietf-emu-eap-tls13 (TLS)](https://tools.ietf.org/html/draft-ietf-emu-eap-tls13) and [draft-ietf-emu-tls-eap-types (TTLS/PEAP)](https://tools.ietf.org/html/draft-ietf-emu-tls-eap-types) we have this patch set.

TLSv1.3 unit tests disabled for now (though they work) as a patched `eapol_test` from https://gitlab.com/jimdigriz/hostap/-/compare/master...tls13 is required.

### DIFF
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -3301,6 +3301,16 @@ post_ca:
 
 	SSL_CTX_set_options(ctx, ctx_options);
 
+	/*
+	 *	TLS 1.3 introduces the concept of early data (also known as zero
+	 *	round trip data or 0-RTT data). Early data allows a client to send
+	 *	data to a server in the first round trip of a connection, without
+	 *	waiting for the TLS handshake to complete if the client has spoken
+	 *	to the same server recently. This doesn't work for EAP, so we
+	 *	disable early data.
+	 *
+	 *	OpenSSL 1.1.1 and later set this as the default
+	 */
 #if OPENSSL_VERSION_NUMBER >= 0x10100000L && OPENSSL_VERSION_NUMBER < 0x10101000L
 	SSL_CTX_set_max_early_data(ctx, 0);
 #endif
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -858,10 +858,10 @@ int tls_handshake_send(REQUEST *request, tls_session_t *ssn)
 		record_minus(&ssn->clean_in, NULL, written);
 
 		/* Get the dirty data from Bio to send it */
-		err = BIO_read(ssn->from_ssl, ssn->dirty_out.data,
-			       sizeof(ssn->dirty_out.data));
+		err = BIO_read(ssn->from_ssl, ssn->dirty_out.data + ssn->dirty_out.used,
+			       sizeof(ssn->dirty_out.data) - ssn->dirty_out.used);
 		if (err > 0) {
-			ssn->dirty_out.used = err;
+			ssn->dirty_out.used += err;
 		} else {
 			if (!tls_error_io_log(request, ssn, err,
 					      "Failed in " STRINGIFY(__FUNCTION__) " (SSL_write)")) {
@@ -3092,6 +3092,13 @@ post_ca:
 	ctx_options |= SSL_OP_NO_SSLv3;
 
 	/*
+	 *	EAP (hopefully) does not have middlebox deployments
+	 */
+#ifdef SSL_OP_ENABLE_MIDDLEBOX_COMPAT
+	ctx_options &= ~SSL_OP_ENABLE_MIDDLEBOX_COMPAT;
+#endif
+
+	/*
 	 *	SSL_CTX_set_(min|max)_proto_version was included in OpenSSL 1.1.0
 	 *
 	 *	This version already defines macros for TLS1_2_VERSION and
@@ -3119,9 +3126,9 @@ post_ca:
 			 *	Pick the maximum one we know about.
 			 */
 #ifdef TLS1_4_VERSION
-			max_version = TLS1_2_VERSION; /* NOT a typo! EAP methods for TLS 1.4 are NOT finished */
+			max_version = TLS1_3_VERSION; /* NOT a typo! EAP methods for TLS 1.4 are NOT finished */
 #elif defined(TLS1_3_VERSION)
-			max_version = TLS1_2_VERSION; /* NOT a typo! EAP methods for TLS 1.3 are NOT finished */
+			max_version = TLS1_3_VERSION;
 #elif defined(TLS1_2_VERSION)
 			max_version = TLS1_2_VERSION;
 #elif defined(TLS1_1_VERSION)
@@ -3293,6 +3300,10 @@ post_ca:
 	}
 
 	SSL_CTX_set_options(ctx, ctx_options);
+
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && OPENSSL_VERSION_NUMBER < 0x10101000L
+	SSL_CTX_set_max_early_data(ctx, 0);
+#endif
 
 	/*
 	 *	TODO: Set the RSA & DH
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -3092,6 +3092,11 @@ post_ca:
 	ctx_options |= SSL_OP_NO_SSLv3;
 
 	/*
+	 *	If set then dummy Change Cipher Spec (CCS) messages are sent in
+	 *	TLSv1.3. This has the effect of making TLSv1.3 look more like TLSv1.2
+	 *	so that middleboxes that do not understand TLSv1.3 will not drop
+	 *	the connection. This isn't needed for EAP-TLS, so we disable it.
+	 *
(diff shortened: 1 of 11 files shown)


jimdigriz:
> I'd just drop this commit. The session resumption cache includes TLS attributes, i.e. expiry, cert information, etc. If we don't cache those attributes, then the user could get back online _after_ his certificate has expired. Which is a security violation.

Which commit? Nothing was reference. Do you mean 'tls: fix session resumption' (8cf675f)?

jimdigriz:
> > I'd just drop this commit. The session resumption cache includes TLS attributes, i.e. expiry, cert information, etc. If we don't cache those attributes, then the user could get back online _after_ his certificate has expired. Which is a security violation.
>> I'd just drop this commit. The session resumption cache includes TLS attributes, i.e. expiry, cert information, etc. If we don't cache those attributes, then the user could get back online after his certificate has expired. Which is a security violation.
>
> Which commit? Nothing was reference. Do you mean 'tls: fix session resumption' ([8cf675f](https://github.com/FreeRADIUS/freeradius-server/commit/8cf675f76c48a11928444bb6f1de0c616002f79a))?

Removed, and reworked the unit test so to tickle an attribute so that session resumption works.

jimdigriz:
The only outstanding task I know of is to [discard any early_data](https://tools.ietf.org/html/draft-ietf-emu-eap-tls13-10#section-2.1.1)

arr2036:
This is meant to be a patch series for TLS 1.3.  Strip out the commits not related to TLS 1.3.

jimdigriz:
> This is meant to be a patch series for TLS 1.3. Strip out the commits not related to TLS 1.3.

Done

jimdigriz:
Don't merge this, I have found some bugs...

jimdigriz:
Okay, added `SSL_CTX_set_max_early_data(..., 0)` (to handle https://github.com/openssl/openssl/commit/c39e4048b538ec76313c264e860cfb5cd677a9ac) and I tickled out the bugs I found.

The other patches I'll include in another PR along with a fix for https://github.com/FreeRADIUS/freeradius-server/issues/1756

src/modules/rlm_eap/libeap/eap_tls.c
```diff
@@ -160,15 +164,36 @@ int eaptls_success(eap_handler_t *handler, int peap_flag)
...
+				     tls_session->ssl, tls_session->label,
+				     context, context_size);
 	} else if (handler->type != PW_EAP_FAST) {
 		RWDEBUG("Not adding MPPE keys because there is no PRF label");
 	}
 
-	eaptls_gen_eap_key(handler->request->reply, tls_session->ssl,
-			   handler->type);
+	rad_assert(context_tls13[0] == handler->type);	/* expected by eaptls_gen_eap_key */
+	eaptls_gen_eap_key(handler->request->reply, tls_session->ssl, context_tls13, context_tls13_size);
```
alandekok, line 195: shouldn't this be `context` and `context_size`, instead of hard-coded tls13 contexts?

src/main/tls.c
```diff
@@ -3091,6 +3091,13 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client)
 	ctx_options |= SSL_OP_NO_SSLv2;
 	ctx_options |= SSL_OP_NO_SSLv3;
 
+	/*
+	 *	EAP (hopefully) does not have middlebox deployments
+	 */
+#ifdef SSL_OP_ENABLE_MIDDLEBOX_COMPAT
```
alandekok, line 3097: I'd add a comment about "If set then dummy Change Cipher Spec (CCS) messages are sent in TLSv1.3. This has the effect of making TLSv1.3 look more like TLSv1.2 so that middleboxes that do not understand TLSv1.3 will not drop the connection.  This isn't needed for EAP-TLS, so we disable it"

src/main/tls.c
```diff
@@ -3294,6 +3301,10 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client)
 
 	SSL_CTX_set_options(ctx, ctx_options);
 
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && OPENSSL_VERSION_NUMBER < 0x10101000L
+	SSL_CTX_set_max_early_data(ctx, 0);
```
alandekok, line 3305: I'd add a comment "TLS 1.3 introduces the concept of early data (also known as zero round trip data or 0-RTT data). Early data allows a client to send data to a server in the first round trip of a connection, without waiting for the TLS handshake to complete if the client has spoken to the same server recently.  This doesn't work for EAP, so we disable early data"

src/modules/rlm_eap/libeap/mppe_keys.c
```diff
@@ -150,38 +150,50 @@ static void PRF(unsigned char const *secret, unsigned int secret_len,
...
-void eaptls_gen_mppe_keys(REQUEST *request, SSL *s, char const *prf_label)
+void eaptls_gen_mppe_keys(REQUEST *request, SSL *s, char const *label, uint8_t const *context, UNUSED size_t context_size)
 {
 	uint8_t out[4 * EAPTLS_MPPE_KEY_LEN];
 	uint8_t *p;
-	size_t prf_size;
+	size_t len;
 
-	prf_size = strlen(prf_label);
+	len = strlen(label);
```
alandekok, line 159: I'm not sure why variable names were changed, but whatever.

src/modules/rlm_eap/libeap/mppe_keys.c
```diff
@@ -219,18 +232,21 @@ void eapttls_gen_challenge(SSL *s, uint8_t *buffer, size_t size)
 	memcpy(p, s->s3->client_random, SSL3_RANDOM_SIZE);
 	p += SSL3_RANDOM_SIZE;
 	memcpy(p, s->s3->server_random, SSL3_RANDOM_SIZE);
+	p += SSL3_RANDOM_SIZE;
```
alandekok, line 235: I don't think this line is necessary.  Compilers may complain about unused assignments

src/modules/rlm_eap/types/rlm_eap_ttls/ttls.c
```diff
@@ -415,8 +415,11 @@ static VALUE_PAIR *diameter2vp(REQUEST *request, REQUEST *fake, SSL *ssl,
 				return NULL;
 			}
 
-			eapttls_gen_challenge(ssl, challenge,
-					      sizeof(challenge));
+			/*
+			 *	TLSv1.3 exports a different key depending on the length
+			 *	requested so ask for *exactly* what the spec requires
+			 */
+			eapttls_gen_challenge(ssl, challenge, vp->vp_length + 1);
```
alandekok, line 422: https://tools.ietf.org/html/rfc5281#section-11.2.4 says 16 octets of challenge, followed by 1 octet of `ident`.

We might want to check the `ident`, too...

src/modules/rlm_eap/libeap/mppe_keys.c
jimdigriz: It is no longer a PRF in TLS 1.3

src/modules/rlm_eap/types/rlm_eap_ttls/ttls.c
jimdigriz: The comment above this (vintage deKok circa 2003): "We're a little forgiving in that we have loose checks on the length, and we do NOT check the Id (first octet of the response to the challenge). But if the client gets the challenge correct, we're not too worried about the Id."

I suspect this is because this code is in the middle of the VP iterator and may not have seen the challenge yet. Do you want me to break the check out to actually look at the ID?

src/modules/rlm_eap/libeap/mppe_keys.c
jimdigriz: Resolved: https://github.com/FreeRADIUS/freeradius-server/pull/3516/commits/5deea57dd182cf575ef1d4c3a0288428aaed0ff4

src/main/tls.c
jimdigriz: Resolved by https://github.com/FreeRADIUS/freeradius-server/pull/3516/commits/3abc3162cebb9df99eb579cd20fbe296283b1718

src/main/tls.c
jimdigriz: Resolved by https://github.com/FreeRADIUS/freeradius-server/pull/3516/commits/becf52a42962070854e771019b08b97197b14588

src/modules/rlm_eap/libeap/eap_tls.c
jimdigriz: Resolved by https://github.com/FreeRADIUS/freeradius-server/pull/3516/commits/eced45c8153bc0d091450e4533ec49dddcddc86d

PR #3525, opened 2020-07-17
Convert fr_struct_to_network() to use dbuff

jejones3141:
Since this function can take a value encoder appropriate for the protocol using it,
this entails converting encode_value for DHCPv6 and RADIUS, the two protocols that
pass in their encode_value functions.


arr2036:
There's no way to review a PR this large in a sensible fashion, please rework it so that there are discreet sets of changes.

jejones3141:
OK.

PR #3526, opened 2020-07-17
Convert EAP-AKA encoding to use dbuff

jejones3141:


src/lib/eap_aka_sim/encode.c
```diff
@@ -97,18 +99,18 @@ static bool is_eap_aka_encodable(void *item, void *uctx)
...
 	uint32_t		iv[4];
+	fr_dbuff_t		work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
 
 	/*
 	 *	One IV per packet
 	 */
 	if (packet_ctx->iv_included) return 0;
 
-	CHECK_FREESPACE(outlen, 4 + AKA_SIM_IV_SIZE);	/* AT_IV + Length + Reserved(2) + IV */
+	FR_DBUFF_CHECK_REMAINING_RETURN(&work_dbuff, 4 + AKA_SIM_IV_SIZE);	/* AT_IV + Length + Reserved(2) + IV */
```
arr2036, line 113: This is now redundant because we do length checks everywhere else.

src/lib/eap_aka_sim/encode.c
```diff
@@ -347,25 +345,16 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
...
 			fr_strerror_printf("%s: AKA-RES Length must be between 4-16 bytes, got %zu bytes",
 					   __FUNCTION__, vp->vp_length);
 			return PAIR_ENCODE_FATAL_ERROR;
 		}
 
-		CHECK_FREESPACE(outlen, vp->vp_length + 2);
-
-		memcpy(p, &res_len, sizeof(res_len));			/* RES Length (bits, big endian) */
-		p += sizeof(res_len);
+		FR_DBUFF_CHECK_REMAINING_RETURN(&work_dbuff, vp->vp_length + 2);
```
arr2036, line 354: again, now redundant

src/lib/eap_aka_sim/encode.c
jejones3141: Good point. I will get rid of it.

src/lib/eap_aka_sim/encode.c
jejones3141: I will eliminate this one as well (and note when the corrected version is pushed); thanks.

src/lib/eap_aka_sim/encode.c
jejones3141: Redundant check removed, and the change pushed.

src/lib/eap_aka_sim/encode.c
jejones3141: Also removed, and change pushed.

jejones3141:
fr_dbuff_remaining() calls removed

src/lib/eap_aka_sim/encode.c
```diff
@@ -701,70 +639,68 @@ static ssize_t encode_rfc_hdr(uint8_t *out, size_t outlen, fr_da_stack_t *da_sta
 	 */
 	da = da_stack->da[depth];
 
-	p += 2;	/* Leave space for attr + len */
+	fr_dbuff_advance(&work_dbuff, 2);
```
arr2036, line 642: Need to check the return to make sure the buffer was actually advanced by 2

arr2036: And really this should be an extension call

src/lib/eap_aka_sim/encode.c
```diff
@@ -584,20 +532,19 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
...
-	uint16_t		actual_len;
+	size_t			actual_len;
+	fr_dbuff_t		len_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
+	fr_dbuff_t		work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
 	fr_dict_attr_t const	*da = da_stack->da[depth];
 	fr_assert(da->flags.array);
 
-	p += 2;
-	value = p;	/* Space for actual length */
+	fr_dbuff_advance(&work_dbuff, 2);
```
arr2036, line 547: Should be an extend call

src/lib/eap_aka_sim/encode.c
```diff
@@ -176,22 +177,22 @@ static ssize_t encode_encrypted_value(uint8_t *out, size_t outlen,
...
-
 	/*
 	 *	Usually in and out will be the same buffer
 	 */
-	if (unlikely(out != in)) memcpy(out, in, inlen);
-	p += inlen;
+	if (unlikely(fr_dbuff_start(&work_dbuff) != in)) {
+		FR_DBUFF_MEMCPY_IN_RETURN(&work_dbuff, in, inlen);
+	} else {
+		FR_DBUFF_ADVANCE_RETURN(&work_dbuff, inlen);
```
arr2036, line 186: should be extend

src/lib/eap_aka_sim/encode.c
```diff
@@ -785,25 +721,29 @@ static inline ssize_t encode_tlv_internal(uint8_t *out, size_t outlen,
...
 	if (!da->flags.extra && da->flags.subtype) {
-		slen = encode_encrypted_value(value, end - value, value, p - value, encoder_ctx);
+		ssize_t	value_len = fr_dbuff_used(&work_dbuff) - 2;
+
+		slen = encode_encrypted_value(&value_dbuff, fr_dbuff_marker_current(&value_start),
+					      value_len, encoder_ctx);
 		if (slen < 0) return PAIR_ENCODE_FATAL_ERROR;
 
-		p = value + slen;
+		fr_dbuff_advance(&work_dbuff, slen - value_len);
```
arr2036, line 730: should be extend

src/lib/eap_aka_sim/encode.c
```diff
@@ -825,18 +765,16 @@ static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
...
 		if (len < 0) return len;
-
-		p += len;
-		outlen -= len;
 	}
+	tl_dbuff = FR_DBUFF_NO_ADVANCE(&work_dbuff);
 
-	if (outlen < 4) return 0;
-	if (outlen > SIM_MAX_ATTRIBUTE_VALUE_LEN) outlen = SIM_MAX_ATTRIBUTE_VALUE_LEN;
+	FR_DBUFF_ADVANCE_RETURN(&work_dbuff, 2);
```
arr2036, line 773: should be extend

src/lib/eap_aka_sim/encode.c
```diff
@@ -1057,7 +991,7 @@ ssize_t fr_aka_sim_encode(REQUEST *request, VALUE_PAIR *to_encode, void *encode_
 		eap_packet->type.data = realloced;
 	}
 
-	return len;
+	return 0;
```
arr2036, line 994: Should still return the encoded length

src/lib/eap_aka_sim/encode.c
```diff
@@ -845,28 +783,30 @@ static ssize_t encode_tlv_hdr(uint8_t *out, size_t outlen,
...
-	size_t			attr_len;
 
 	fr_da_stack_t		da_stack;
 	fr_dict_attr_t const	*da = NULL;
 	fr_aka_sim_encode_ctx_t	*packet_ctx = encoder_ctx;
 
-	if (!cursor || !out) return PAIR_ENCODE_FATAL_ERROR;
-
-	CHECK_FREESPACE(outlen, 4);		/* Attributes lengths are always multiples of 4 */
+	if (!cursor || !fr_dbuff_marker(&m, &work_dbuff)) return PAIR_ENCODE_FATAL_ERROR;
```
arr2036, line 809: Not sure why fr_dbuff_marker is included in this condition?

src/lib/eap_aka_sim/encode.c
```diff
@@ -386,17 +373,8 @@ static ssize_t encode_value(uint8_t *out, size_t outlen,
 	 */
 	case FR_CHECKCODE:
 	{
```
arr2036, line 375: Remove bracketing if there are no local variables

src/lib/eap_aka_sim/encode.c
```diff
@@ -701,70 +639,68 @@ static ssize_t encode_rfc_hdr(uint8_t *out, size_t outlen, fr_da_stack_t *da_sta
...
 		 */
-		sublen = end - p;
-		if (sublen > SIM_MAX_ATTRIBUTE_VALUE_LEN) sublen = SIM_MAX_ATTRIBUTE_VALUE_LEN;
 
 		/*
 		 *	Determine the nested type and call the appropriate encoder
 		 */
 		if (da_stack->da[depth + 1]->type == FR_TYPE_TLV) {
-			slen = encode_tlv_hdr(p, sublen, da_stack, depth + 1, cursor, encoder_ctx);
+			slen = encode_tlv_hdr(&FR_DBUFF_MAX(&work_dbuff, SIM_MAX_ATTRIBUTE_VALUE_LEN),
```
arr2036, line 696: using FR_SBUFF_MAX like this is wrong.  It should be initialised outside of the loop and used to restrict the maximum length of all children.

src/lib/eap_aka_sim/encode.c
```diff
@@ -992,56 +935,47 @@ ssize_t fr_aka_sim_encode(REQUEST *request, VALUE_PAIR *to_encode, void *encode_
 	}
 	fr_cursor_head(&cursor);	/* Reset */
 
-	MEM(p = buff = talloc_zero_array(eap_packet, uint8_t, 1024));	/* We'll shrink this later */
-	end = p + talloc_array_length(p);
+	MEM(buff = talloc_zero_array(eap_packet, uint8_t, 1024));	/* We'll shrink this later */
+	fr_dbuff_init(&dbuff, buff, talloc_array_length(buff));
```
arr2036, line 939: should now be an extensible dbuff, and should probably start at 512

src/lib/eap_aka_sim/encode.c
jejones3141: Looking at it, no real need; in theory it might return NULL, but that's pretty paranoid. Hoisting it out of the `if`.

src/lib/eap_aka_sim/encode.c
jejones3141: Done.

src/lib/eap_aka_sim/encode.c
jejones3141: Hmm.... the pre-dbuff version looked like it was bounding the lengths of the individual children, but it works on unit tests with their sum constrained to that value; changing to work that way.

src/lib/eap_aka_sim/encode.c
jejones3141: Done.

PR #3527, opened 2020-07-17
Change internal_encode() to use FR_DBUFF_NO_ADVANCE()

jejones3141:


src/protocols/internal/encode.c
```diff
@@ -57,38 +57,31 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
...
 
 	FR_PROTO_STACK_PRINT(da_stack, depth);
 
-	enc_field = dbuff->p;
+	enc = work_dbuff.p;
 
 	/*
 	 *	Zero out first encoding byte
 	 */
-	fr_dbuff_memset(dbuff, 0, 1);
```
arr2036, line 76: Why was this removed?

src/protocols/internal/encode.c
```diff
@@ -175,6 +172,7 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
 	 */
 	case FR_TYPE_EXTENDED:	/* Just another type of TLV */
 	case FR_TYPE_TLV:
+	{
```
arr2036, line 175: Unnecessary bracketing?

src/protocols/internal/encode.c
```diff
@@ -57,38 +57,31 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
 			       fr_da_stack_t *da_stack, unsigned int depth,
 			       fr_cursor_t *cursor, void *encoder_ctx)
 {
-	uint8_t				*enc_field, *len_field, *value_field, *value_end;
```
arr2036, line 60: I really dislike fields being renamed for no purpose.

src/protocols/internal/encode.c
jejones3141: Looking at the code, it's now `FR_DBUFF_BYTES_IN_RETURN(&work_dbuff, 0);`. If that's not there, please let me know.

src/protocols/internal/encode.c
jejones3141: It is indeed unneccessary; out it goes.

src/protocols/internal/encode.c
jejones3141: Quite understandable; back they go.

src/protocols/internal/encode.c
arr2036: Ah, ok, apologies, missed that.

src/protocols/internal/encode.c
jejones3141: Gone; the corrected version is pushed.

src/protocols/internal/encode.c
jejones3141: Names restored and the result pushed.

jejones3141:
Closing; a version that not only uses FR_DBUFF_NO_ADVANCE() but also follows other requested changes (avoiding fr_dbuff_remaining() and FR_DBUFF_CHECK_REMAINING_RETURN()) is in fr_struct_to_network_dbuff.

PR #3531, opened 2020-07-21
Small docs change "blacklist" -> "blocklist"

tashian:
Per best current practice in https://tools.ietf.org/id/draft-knodel-terminology-01.html

### DIFF
--- a/doc/antora/modules/raddb/pages/sites-available/default.adoc
+++ b/doc/antora/modules/raddb/pages/sites-available/default.adoc
@@ -157,17 +157,17 @@ Useful range of values: 5 to 600
 
 
 
-nak_lifetime:: Time for which blacklisted
+nak_lifetime:: Time for which blocklisted
 clients are placed into a NAK cache.
 
 If a dynamic client is disallowed, it is
-placed onto a "NAK" blacklist for a period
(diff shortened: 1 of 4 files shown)


tashian:
There is also many `master`/`slave` references in the FreeRADIUS codebase that could be clarified as `leader` / `follower`, but I did not make those changes as they are more delicate.

alandekok:
The document you reference is simply an opinion piece by a random person.  It is not an IETF "best current practices" standard, and you should not misrepresent the situation by claiming it is one.

tashian:
You're right, it's a proposed BCP. Thanks @alandekok

arr2036:
My primary concern is understandability and consistency of the documentation.  If a datastore the documentation refers to uses the terms `primary`/`replica` we'll use those terms in out documentation.

If a protocol describes a key from which other keys are derived as a `master key` then we're going to use that term in the configuration items and documentation for a module implementing that protocol, we're NOT going to refer to it as a `leader key` or `primary key`.

FWIW - If you're aiming to be non-offensive I would also suggest you not promote terms like `leader` and `follower`, cult survivors and those who have escaped oppressive regimes will thank you.

PR #3532, opened 2020-07-22
ldap: Add references for Active Directory 'Nested groups'

jpereira:


raddb/mods-available/ldap
```diff
@@ -43,7 +43,11 @@ ldap {
 	#  Group membership can be queried by using the above "ldapsearch" string,
 	#  and adding "memberof" qualifiers.  For ActiveDirectory, use:
 	#
-	#    ldapsearch ... '(&(objectClass=user)(sAMAccountName=user)(memberof=CN=group,${base_dn}))'
+	#    ldapsearch ... '(&(objectCategory=Person)(sAMAccountName=user)(memberof=CN=group,${base_dn}))'
```
mcnewton, line 46: Changing from `objectClass=user` to `objectCategory=person` will break machine authentication, no?

alandekok: Yes.  I'd prefer to leave the examples alone, but add comments which document using `objectCategory=Person`

mcnewton: I don't think the change to `objectCategory` is even necessary - with ref to https://social.technet.microsoft.com/wiki/contents/articles/5392.active-directory-ldap-syntax-filters.aspx#Filter_on_objectCategory_and_objectClass just adding the memberOf oid should tell AD to search the nested groups. I don't recall needing any other changes to making nested group searches work with AD before.

I'd suggest adding a comment suggesting using that oid for nested group searches (and possibly pointing to the TechNet article), and as you say leave everything else alone. Same on below changes, too.

jpereira: Answer your question @mcnewton, It won't break.

1. The `jpereira` is a member of the below groups.

image

2. The below groups are members of the group `WiFi`

image

3. Therefore, the search is working as expected.

e.g:

```
# ldapsearch -LLL -H ldap://192.168.2.254 -D jane.doe@sugarloaf.lan -w HeyH0Password -x -LLL -b "CN=Users,DC=sugarloaf,DC=lan" '(&(objectCategory=Person)(sAMAccountName=jpereira)(memberof=CN=Devel,CN=users,DC=sugarloaf,DC=lan))' memberOf
dn: CN=Jorge Pereira,CN=Users,DC=sugarloaf,DC=lan
memberOf: CN=Fileserver,CN=Users,DC=sugarloaf,DC=lan
memberOf: CN=Devel,CN=Users,DC=sugarloaf,DC=lan
#
```

I've just suggested using `objectCategory=person` instead of `objectClass=user` to avoid the user misunderstands the phrase: _Where 'user' is the user as above, and 'group' is the group you are querying for._ and do something like:

`ldapsearch ... '(&(objectClass=MyUsername)(sAMAccountName=MyUsername)(memberof=CN=group,${base_dn}))'`

either way, I dropped such commit.

mcnewton: Yes, user accounts will still work. The test to do is to try it with a computer account and see if that still works.

jpereira: Well, the *same query* is working on the production server as well. But, okay. The changes was dropped.

arr2036:
Can this be pulled into mater too

PR #3538, opened 2020-07-27
Sqlite memleak

ndptech:
Fix to handle memory leak observed when DHCP used SQLite as the back end and the server was subjected to a heavy load.
SQLITE_BUSY responses were being treated by reconnecting - without clearing resources for former connections.


alandekok:
You should rebase your branch first. That will get rid of the extra commits.

PR #3543, opened 2020-07-28
Refinement to DHCP de-dup

ndptech:
Added in the message code.  Without this, certain clients request packet
would match their discover packet and so get treated as a duplicate.


alandekok:
Fixed slightly differently.

PR #3547, opened 2020-07-31
rlm_smtp with unit test

stephenmac98:
mod_authorize sends emails
It requires the request attributes:
SMTP-Sender-Email, SMTP-Recipients, SMTP-Mail-Header, SMTP-Mail-Body
It optionally accepts:
SMTP-Attachments, User-Name, User-Password

mod_authenticate confirms an email can be sent
It requires the request attributes:
User-Name, User-Password

### DIFF
--- a/raddb/mods-available/smtp
+++ b/raddb/mods-available/smtp
@@ -1,0 +1,276 @@
+#  -*- text -*-
+#
+#
+#  $Id$
+
+#######################################################################
+#
+#  = SMTP Module
+#
+#  The `smtp` module validates a users name and password against an SMTP server.
+#  It should be called from an `authenticate` section.
+#
+#  The module can optionally perform a tls handshake, enabled with require_cert
+#
+
+smtp {
+	#
+	#  tls { ... }:: Configure the tls related items which control
+	#  how FreeRADIUS connects to an SMTP server.
+	#
+	tls {
+		#
+		#  .Certificate validation options
+		#
+		#  Specifies how the certificate(s) presented by the
+		#  SMTP server are validated, and which certificates
+		#  (if any) to send to that SMTP server.
+		#
+		#  The options here behave the same as the options in
+		#  other `tls` sections in the server.
+		#
+
+		#
+		#  certificate_file:: PEM formatted file containing the certificate we present to the SMTP server
+		#
+		#  Specifies a certificate and any intermediary CAs we should send to the SMTP server.
+		#
+		#  This file should usually contain the client certificate file first, then any
+		#  intermediary signing CAs, shallowest (direct signee of the certificate_file)
+		#  to deepest (signed directly by the root CA).
+		#
+#		certificate_file     = /path/to/radius.pem
+
+		#
+		#  ca_file:: PEM formatted file containing the chain
+		#  to validate the SMTP server's certificate.
+		#
+		#  Any certificate chain MUST be in order from server
+		#  certificate (first in the file) to intermediary CAs (second) to
+		#  Root CA (last in the file) as per RFC 4346 Section 7.4.2 (see certificate_list)
+		#
+		#  Providing a complete certificate chain here is the
+		#  most common way of validating the certificate
+		#  presented by an SMTP server.
+		#
+#		ca_file	             = "${certdir}/cacert.pem"
+
+		#
+		#  ca_issuer_file:: PEM formatted file containing the
+		#  CA that signed the SMTP server's certificate.
+		#
+		#  Specifies the certificate which directly signed
+		#  the certificate presented by the SMTPs server.
+		#
+		#  This configuration option can be used to prevent
+		#  certificates passing validation which were signed
+		#  by other intermediary CAs, or root CAs, in the
+		#  trusted certificate chain.
+		#
+#		ca_issuer_file     = "${certdir}/caissuer.pem"
+
+		#
+		#  ca_path:: A directory containing multiple root CA certs named by their hash.
+		#
+		#  See the OpenSSL documentation for more details:
+		#  - https://www.openssl.org/docs/man1.1.0/man3/SSL_CTX_set_default_verify_paths.html
+		#  - https://www.openssl.org/docs/man1.1.1/man1/c_rehash.html
+		#
+		#  This configuration option should only be used when the SMTP server being contacted
+		#  is not known ahead of time (using a URL from an external source), and/or the CA used
+		#  to sign the SMTP server certificate is unknown.
+		#
+#		ca_path	             = "${certdir}"
+
+		#
+		#  private_key_file:: PEM formatted file containing the private key for the specified `certificate_file`
+		#
+		#  This item must be specified if `certificate_file` is being used.
+		#
+#		private_key_file     = /path/to/radius.key
+
+		#
+		#  private_key_password:: Password used to decrypt the `private_key_file`.
+		#
+#		private_key_password = "supersecret"
+
+		#
+		#  random_file:: Source of random data used for various cryptographic functions.
+		#
+#		random_file          = /dev/urandom
+
+		#
+		#  check_cert:: Server certificate verification requirements.
+		#
+		#  May be one of:
+		#
+		#  [options="header,autowidth"]
+		#  |===
+		#  | Option | Description
+		#  | `no`   | Server certificate can be signed by any CA or be self-signed.
+		#  | `yes`  | Server certificate must be issued by one of the trusted CAs.
+		#  |===
+		#
+		#  Default is `yes`
+		#
+#		check_cert = no
+
+		#
+		#  request_cert:: Options for controlling how the
+		#  module requests TLS to the SMTP server.
+		#
+		#  May be one of:
+		#
+		#  [options="header,autowidth"]
+		#  |===
+		#  | Option | Description
+		#  | `demand`   | Require ssl validation to accept login
+		#  | `never`  | Do not try to establish ssl connection
+		#  | `allow`  | Try to establish ssl connection, continue even if it cannot
+		#  |===
+		#
+		#  default is 'allow'
+		#
+#		require_cert = allow
+
+		#
+		#  check_cert_cn:: Server certificate CN verification requirements.
+		#
+		#  May be one of:
+		#
+		#  [options="header,autowidth"]
+		#  |===
+		#  | Option | Description
+		#  | `no`   | Server certificate CN can be any value.
+		#  | `yes`  | Server certificate CN matches the host in the URI.
+		#  |===
+		#
+		#  Default is `yes`
+		#
+#		check_cert_cn = no
+
+		#
+		#  extract_cert_attrs:: Extract OIDs from presented certificates as OIDs.
+		#
+		#  Default is `no`
+		#
+#		extract_cert_attrs = no
+	}
+
+	#
+	#  uri:: URI which will be used for connecting to the SMTP server.
+	#
+	#  The `smtp` module uses Curl (https://curl.haxx.se/libcurl/) to implement
+	#  the underlying mail protocols.  The URI should be in the form of:
+	#
+	#  `SCHEME://IP:PORT/`
+	#
+
+	#	`SCHEME` can be one of the protocols supported by Curl.
+	#	i.e. ftp, ftps, gopher, http, https, imap, imaps, ldap, ldaps,
+	#	     pop3, pop3s, rtsp, scp, sftp, smb, smbs, smtp, smtps.
+	#	If no `SCHEME` is given, it defaults to `smtp`.
+	#
+	#	`IP` is the IP address of the server.  It can be an IPv4 address,
+	#	IPv6 address, hostname, etc.
+	#
+	#	`PORT` is optional, and will normally be chosen to be correct
+	#	for the given `SCHEME`.
+	#
+	#  For more information, see the CURL documentation at
+	#
+	#	https://ec.haxx.se/cmdline/cmdline-urls
+	#
+	uri = "smtp://192.0.20.1/"
+
+	#
+	#  timeout:: How long the module will wait before giving up on the response
+	#  from the SMTP server.
+	#
+	timeout = 5s
+
+	#
+	#  template_directory:: The source directory where all file attachments are pulled from
+	#  All file attachments should be their relative path from this location, without a leading /
+	#
+	template_directory = raddb/mods_config/smtp
+
+	#
+	#  Attachments:: attachments with their relative path from template_directory
+	#  There can be no leading / or ..
+	#
+	attachments = &SMTP-Attachments[*]
+
+	#
+	#  envelope_address:: This is the address used to send the mail,
+	#  Sent to the receiving server as MAIL FROM:<envelope_address>
+	#  sender_address will be defaulted to if this is not set
+	#
+	envelope_address = "postmaster@localhost"
+
+	#
+	#  sender_address:: This are the addresses displayed in the FROM: element of the header
+	#  This can be different than the provided envelope_address,
+	#  If envelope_address is set, this can be formatted however you want it to appear to the receiver
+	#  If envelope_address is not set, the first element in sender_address will be used as the envelope address
+	#
+	sender_address = &SMTP-Sender-Address[*]
+
+	#
+	#  recipients:: Email addresses to be set as recipients for the email
+	#  If recipients is set bcc, cc, and to will not be automatically added to the email
+	#  And so they must be included here.
+	#
+#	recipients = &SMTP-Recipients[*]
+#	recipients = &SMTP-TO[*]
+#	recipients = &SMTP-CC[*]
+#	recipients = &SMTP-BCC[*]
+
+	#
+	#  to:: Email addresses to be set in the TO: header
+	#  These addresses will be added as envelope recipients only if recipients is not set
+	#
+#	to = &SMTP-TO[*]
+	#
+	#  cc:: Email addresses to be set in the CC: header
+	#  These addresses will be added as envelope recipients only if recipients is not set
+	#
+#	cc = &SMTP-CC[*]
+
+	#
+	#  bcc:: Comma separated list of emails.
+	#  the local part may contain commas, the domain may not (RFC 2821)
+	#  Therefore, first comma after the @ represents a new address
+	#  not listed in the header of the email
+	#  if recipients is not set, these emails will be added to the envelope recipients
+	#
+#	bcc = &SMTP-BCC[*]
+
+	#
+	#  set_date:: Adds a Date: to the header, set to the time the request is received
+	#  Formatted as "Fri, 07 Aug 2020 00:57:37 -0400, (EDT)"
+	#  May be one of:
+	#
+	#  [options="header,autowidth"]
+	#  |===
+	#  | Option | Description
+	#  | `no`   | A Date: should be specified in the header, or left to the receiving mta
+	#  | `yes`  | A Date is formatted and added (recommended)
+	#  |===
+	#
+	#  Default is `yes`
+	#
+#	set_date = yes
+
+	#
+	#  header:: This is where any other headers can be added to the email
+	#  FROM, TO, and CC, should not be added here if their respective elements were set in the config
+	#  If no DATE header is provided, one will be added showing time that the request was received(recommended)
+	#  Non-standard mail headers may be set. Adhere to your MTA's documentation
+	#
+	header {
+		subject = "email subject"
+		Message-ID = "950124.162336@example.com"
+#		X-Originating-IP = "192.0.20.1"
+	}
+}
--- a/scripts/travis/exim-setup.sh
+++ b/scripts/travis/exim-setup.sh
@@ -1,0 +1,133 @@
+#!/bin/sh -e
+#
+# ### This is a script to setup an exim smtp server for testing rlm_smtp
+#
+
+#
+# Declare the important path variables
+#
+
+# Base Directories
+BASEDIR=$(git rev-parse --show-toplevel)
+BUILDDIR="${BASEDIR}/build/ci/exim4"
+
+# Directories for exim processes
+RUNDIR="${BUILDDIR}/run"
+MAILDIR="${BUILDDIR}/mail"
+MAILDELIVERYDIR="${BUILDDIR}/mail_delivery_test"
+LOGDIR="${BUILDDIR}/eximlog"
+SPOOLDIR="${BUILDDIR}/spool"
+CERTDIR="${BUILDDIR}/certs"
+CERTSRCDIR="${BASEDIR}/raddb/certs"
+PASSWORD="whatever"
+
+# Important files for running dovecot
+CONF="${BUILDDIR}/exim.conf"
+
+# Used for supporting files
+TRAVISDIR="${BASEDIR}/scripts/travis"
+
+#
+# Prepare the directories and files needed for running exim
+#
+
+# Stop any currently running exim instance
+echo "Checking for a running exim instance"
+if [ -e "${RUNDIR}/exim.pid" ]
+then
+	echo "Stopping the current exim instance"
+	kill "$(cat ${RUNDIR}/exim.pid)"
+	rm -r "${BUILDDIR}"
+fi
+
+# Create the directories
+mkdir -p "${BUILDDIR}" "${RUNDIR}" "${MAILDELIVERYDIR}" "${MAILDIR}" "${LOGDIR}" "${SPOOLDIR}" "${CERTDIR}"
+
+# Create the certificate
+echo "Generating the certificates"
+openssl pkcs8 -in ${CERTSRCDIR}/rsa/server.key -passin pass:${PASSWORD} -out ${CERTDIR}/server.key
+cp ${CERTSRCDIR}/rsa/server.pem ${CERTDIR}/server.pem
+cp ${CERTSRCDIR}/rsa/ca.pem ${CERTDIR}/ca.pem
+
+# Create exim.conf file
+echo "Generating the exim configuration file"
+touch "${CONF}"
+
+# Build exim.conf
+echo "
+#
+# Set the user to run as - use -DEXIMUSER=user -DEXIMGROUP=group
+# rather than defining them here.
+#
+#EXIMUSER = username
+#EXIMGROUP = groupname
+LISTEN=127.0.0.1
+#
+#
+#  Where all the config files, logs, etc are. See also the
+#  "keep_environment" setting below.
+#
+MAIL_DIR = ${MAILDIR}
+pid_file_path = ${RUNDIR}/exim.pid
+log_file_path = ${LOGDIR}/%s
+spool_directory = ${SPOOLDIR}
+exim_user = EXIMUSER
+exim_group = EXIMGROUP
+daemon_smtp_ports = 2525 : 2465
+local_interfaces = LISTEN
+deliver_drop_privilege
+keep_environment = ${BASEDIR}
+tls_advertise_hosts = *
+tls_certificate = ${CERTDIR}/server.pem
+tls_privatekey = ${CERTDIR}/server.key
+tls_verify_certificates = ${CERTDIR}/ca.pem
+#tls_dhparam = ${CERTDIR}/dh
+tls_on_connect_ports = 2465
+tls_verify_hosts = *
+tls_require_ciphers = \${if =={\$received_port}{2525}\
+                           {NORMAL:%COMPAT}\
+                           {SECURE128}}
+received_header_text =
+acl_smtp_rcpt = accept
+begin acl
+begin routers
+#
+#  Only one router - we'll send everything to the \"local_delivery\"
+#  transport.
+#
+local_delivery:
+  driver = accept
+  transport = local_delivery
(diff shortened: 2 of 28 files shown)


mcnewton:
Given that a lot of the SMTP-* attributes are likely to stay the same, it might be nice to give a layer of indirection to the in the module configuration rather than having to set them on every request? I can imagine in most use cases at least the from/recipients are always going to be identical.

e.g. let you do something like
```
smtp {
  ...
  from_address = &SMTP-Sender-Email

  # maybe this could be recipient= given multiple times
  recipients = &SMTP-Recipients[*]

  subject = "This is a test message"

  body = "User %{User-Name} just failed to log in"
}
```

For the most basic alerting messages that would then be a lot simpler to configure, as there'd be no `update {}` sections required in unlang.

I also think that the headers `From`, `To`, `Subject` and `Date` should be automatically added based on the sender/recipient addresses (and then maybe have a "I know what I'm doing" module option to override that). In my experience most people have no clue how e-mail works and you'll likely end up with loads of messages being sent that are missing key headers. People especially won't have any idea of the difference between the sender address (envelope from) and the from address (header from), for example.

All sent messages should also add a unique `Message-Id` header by default (does curl do it if it's missing?).

Looks like it's still enforcing MIME on every message, that's likely to be a problem if people e.g. want to send directly into some monitoring or other automated systems, so it should really be configurable to send plain text only.

Being picky, mod_authenticate is only checking the username/password are accepted by the SMPT server, right, not actually doing any sort of MAIL FROM/RCPT TO or VRFY checks? (Not that those would confirm a mail could actually be sent, either.)

Nice module though. Exim config looks good... vaguely familiar :laughing:

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+		RDEBUG2("A handle could not be allocated for the request");
+		return RLM_MODULE_FAIL;
+	}
+
+	/* Initialize the mail_ctx to perform the email */
+	mail_ctx = talloc_zero(randle, fr_mail_ctx);
+	mail_ctx->request 		= request;
+	mail_ctx->randle 		= randle;
+	mail_ctx->time_status 		= UNLOADED; /* prevent duplication if there is a user supplied a DATE: element */
+	mail_ctx->time 			= fr_time(); /* time the request was received. Used to set DATE: if none is supplied */
```
arr2036, line 364: Minor nit with this.  Generally when you're allocating and immediately initialising a struct, don't use the zero or clearing allocation function, just initialise a structure with a compound literal. i.e.

``
*mail_ctx = (fr_mail_ctx){
	.request = request,
	.randle = randle,
	.time_status = UNLOADED,
	.time = fr_time()
}
```

Elements you don't explicitly specify will be initialised to zero, and you save writing over the struct twice (once for zeroing it out, and once for filling it).

File: src/modules/rlm_smtp/rlm_smtp.c [hunk 2]
```diff
@@ -0,0 +1,614 @@
...
+	if (!randle){
+		RDEBUG2("A handle could not be allocated for the request");
+		return RLM_MODULE_FAIL;
+	}
+
+	/* Initialize the mail_ctx to perform the email */
+	mail_ctx = talloc_zero(randle, fr_mail_ctx);
+	mail_ctx->request 		= request;
+	mail_ctx->randle 		= randle;
+	mail_ctx->time_status 		= UNLOADED; /* prevent duplication if there is a user supplied a DATE: element */
```
arr2036, line 363: If there's only two states, just use a bool

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+		vp = fr_cursor_next(&uctx->cursor);
+	}
+	RDEBUG2("Ititialized %d attachment(s)", attachments_set);
+	return attachments_set;
+}
+
+/*
+ * Free the curl slists
+ */
+static int _free_mail_slists(fr_mail_ctx *uctx)
```
arr2036, line 261: It's better to state what you're freeing in the function name i.e. `_free_mail_ctx`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+	VALUE_PAIR 		*vp;
+	int 			attachments_set = 0;
+
+	vp = fr_cursor_iter_by_da_init(&uctx->cursor, &uctx->request->packet->vps, attr_smtp_attachment_file);
+
+	/* Initialize the buffer to write the file path */
+	fr_sbuff_init_talloc(uctx, &uctx->path_buffer, &uctx->sbuff_ctx, 0, SIZE_MAX);
+
+	/* Write the initial path to the buffer */
+	fr_sbuff_in_bstrncpy(&uctx->path_buffer, inst->template_dir, strlen(inst->template_dir));
```
arr2036, line 212: `fr_sbuff_in_strcpy`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+	/* Make sure the template_directory path ends in a "/" */
+	if (inst->template_dir[strlen(inst->template_dir)-1] != '/'){
```
arr2036, line 214: `if (fr_sbuff_is_char(&uctx->path_buffer, '/'))`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+		RDEBUG2("Adding / to end of template_dir");
+		fr_sbuff_in_bstrncpy(&uctx->path_buffer, "/", 1);
```
arr2036, line 216: `fr_sbuff_in_char(&uctx->path_buffer, '/')`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+{
+	curl_mimepart		*part;
+	REQUEST			*request = uctx->request;
+	VALUE_PAIR 		*vp;
+	int 			attachments_set = 0;
+
+	vp = fr_cursor_iter_by_da_init(&uctx->cursor, &uctx->request->packet->vps, attr_smtp_attachment_file);
+
+	/* Initialize the buffer to write the file path */
+	fr_sbuff_init_talloc(uctx, &uctx->path_buffer, &uctx->sbuff_ctx, 0, SIZE_MAX);
```
arr2036, line 209: You're leaking (ish, it will be freed eventually) memory here every time attachment_source is called, you should call this once when the uctx is created, then set a marker so you can just copy over the trailing portion of the path.  Also means you don't need to run the logic to figure out if there's a trailing slash each time.

As I mentioned in slack, you should drop a marker after copying in the template_dir string (and the trailing slash), and reset the position each time you want to process a new vp.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+			goto next;
+		}
+		if(strncmp(vp->vp_strvalue, "..", 2) == 0) {
+			RDEBUG2("Cannot access values outside of template_directory");
+			goto next;
+		}
+
+		/* Copy the filename into the buffer */
+		fr_sbuff_in_bstrncpy(&uctx->path_buffer, vp->vp_strvalue, vp->vp_length);
+		fr_sbuff_trim_talloc(&uctx->path_buffer, SIZE_MAX);
```
arr2036, line 243: Actually realised you don't need this trim call, just remove it.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+		/* Add the file attachment as a mime encoded part */
+		attachments_set++;
+		RDEBUG2("ititializing attachment: %s", uctx->path_buffer.buff);
+		part = curl_mime_addpart(mime);
+		curl_mime_encoder(part, "base64");
+		curl_mime_filedata(part, uctx->path_buffer.buff);
+	next:
+		vp = fr_cursor_next(&uctx->cursor);
+	}
+	RDEBUG2("Ititialized %d attachment(s)", attachments_set);
```
arr2036, line 254: `talloc_free(uctx->path_buffer.buff)`

src/modules/rlm_smtp/rlm_smtp.c
arr2036, line 209: Memory reallocs are still expensive, so you should pass a reasonable value here instead of `0`.  Maybe `talloc_array_length(inst->template_dir) + 128`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,614 @@
...
+	fr_sbuff_marker(&uctx->m, &uctx->path_buffer);
+
+	/* Check for any file attachments */
+	while(vp) {
+		/* Move to the end of the template directory filepath */
+		fr_sbuff_set(&uctx->path_buffer, &uctx->m);
+
+		/* Check to see if the file attachment is valid, skip it if not */
+		RDEBUG2("Trying to set attachment: %s", vp->vp_strvalue);
+		if(vp->vp_tainted) {
```
arr2036, line 228: que?

raddb/mods-available/smtp
```diff
@@ -0,0 +1,197 @@
+#  -*- text -*-
+#
+#
+#  $Id$
+
+#######################################################################
+#
+#  = SMTP Module
+#
+#  The `smtp` module validates a users name and password against an IMAP server.
```
herwinw, line 10: I guess this file has been copied from `rlm_imap` and not been fully updated, it has a lot of references to IMAP

raddb/mods-available/smtp
```diff
@@ -0,0 +1,275 @@
...
+#	recipients = "ex@ample.com,ex2@ample.com"
+
+	#
+	#  to:: Comma separated list of emails.
+	#  the local part may contain commas, the domain may not (RFC 2821)
+	#  Therefore, a comma after the @ represents a new address
+	#  listed in the TO: header of the email
+	#  if recipients is not set, these emails will be added to the envelope recipients
+	#
+#	to = "ex@ample.com,ex2@ample.com"
```
arr2036, line 228: Shouldn't be comma separated.  You're introducing domain specific formatting for emails into the FreeRADIUS configuration scheme.  Use `FR_TYPE_MULTI` for the config item, and build the list of emails from the tmpls you get.

raddb/mods-available/smtp
```diff
@@ -0,0 +1,275 @@
...
+
+	#
+	#  cc:: Comma separated list of emails.
+	#  the local part may contain commas, the domain may not (RFC 2821)
+	#  Therefore, first comma after the @ represents a new address
+	#  listed in the CC; header of the email
+	#  if recipients is not set, these emails will be added to the envelope recipients
+	#
+#	cc = "ex@ample.com,ex2@ample.com"
```
arr2036, line 237: Same here

raddb/mods-available/smtp
```diff
@@ -0,0 +1,275 @@
...
+
+	#
+	#  bcc:: Comma separated list of emails.
+	#  the local part may contain commas, the domain may not (RFC 2821)
+	#  Therefore, first comma after the @ represents a new address
+	#  not listed in the header of the email
+	#  if recipients is not set, these emails will be added to the envelope recipients
+	#
+#	bcc = "ex@ample.com,ex2@ample.com"
```
arr2036, line 246: and here

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+#define MAX_ATTRMAP	128
+
+extern fr_dict_autoload_t rlm_smtp_dict[];
+fr_dict_autoload_t rlm_smtp_dict[] = {
+	{ .out = &dict_radius, .proto = "radius" },
+	{ .out = &dict_freeradius, .proto = "freeradius"},
+	{ NULL }
+};
+
+static fr_dict_attr_t 	const 	*attr_auth_type;
```
arr2036, line 43: `const` is left associative, and should be part of the type on the LHS. i.e. `static fr_dict_attr_t const<tabs>*<variable name>`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+ * @copyright 2020 Network RADIUS SARL <legal@networkradius.com>
+ */
+RCSID("$Id$")
+
+#include <freeradius-devel/server/base.h>
+#include <freeradius-devel/server/module.h>
+#include <freeradius-devel/curl/base.h>
+
+static fr_dict_t 	const 	*dict_radius; /*dictionary for radius protocol*/
+fr_dict_t 		const 	*dict_freeradius;
```
arr2036, line 32: These should both be static with const on the left.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+static fr_dict_attr_t 	const 	*attr_smtp_to;
+static fr_dict_attr_t 	const 	*attr_smtp_cc;
+static fr_dict_attr_t 	const 	*attr_smtp_bcc;
+static fr_dict_attr_t 	const 	*attr_smtp_header;
+static fr_dict_attr_t 	const 	*attr_smtp_body;
+static fr_dict_attr_t 	const 	*attr_smtp_attachment_file;
+
+extern fr_dict_attr_autoload_t rlm_smtp_dict_attr[];
+fr_dict_attr_autoload_t rlm_smtp_dict_attr[] = {
+	{ .out = &attr_user_name, .name = "User-Name", .type = FR_TYPE_STRING, .dict = &dict_radius },
```
arr2036, line 57: The majority of these should be `FR_TYPE_TMPL` the lists of email addresses should be `FR_TYPE_TMPL | FR_TYPE_MULTI`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+	curl_mime		*mime;
+} fr_mail_ctx;
+
+/*
+ *	A mapping of configuration file names to internal variables.
+ */
+static const CONF_PARSER module_config[] = {
+	{ FR_CONF_OFFSET("uri", FR_TYPE_STRING, rlm_smtp_t, uri) },
+	{ FR_CONF_OFFSET("template_directory", FR_TYPE_STRING, rlm_smtp_t, template_dir) },
+	{ FR_CONF_OFFSET("sender_address", FR_TYPE_STRING, rlm_smtp_t, sender_address) },
```
arr2036, line 116: Anything that's dynamically expanded should be `FR_TYPE_TMPL | FR_TYPE_MULTI`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+	{ FR_CONF_OFFSET("envelope_address", FR_TYPE_STRING, rlm_smtp_t, envelope_address) },
+	{ FR_CONF_OFFSET("recipients", FR_TYPE_STRING, rlm_smtp_t, recipient_addrs) },
+	{ FR_CONF_OFFSET("TO", FR_TYPE_STRING, rlm_smtp_t, to_addrs) },
+	{ FR_CONF_OFFSET("CC", FR_TYPE_STRING, rlm_smtp_t, cc_addrs) },
+	{ FR_CONF_OFFSET("BCC", FR_TYPE_STRING, rlm_smtp_t, bcc_addrs) },
+	{ FR_CONF_OFFSET("timeout",FR_TYPE_TIME_DELTA, rlm_smtp_t, timeout) },
```
arr2036, line 122: Missed space

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+	if(vp) return vp->vp_strvalue;
+
+	/* If there is an envelope address in the config, use that, otherwise use sender_address */
+	return (inst->envelope_address ? inst->envelope_address : inst->sender_address);
+}
+
+/*
+ * 	Adds every element associated with a dict_attr to a curl_slist
+ */
+static int da_to_slist(fr_mail_ctx *uctx, struct curl_slist **out, const fr_dict_attr_t *dict_attr)
```
arr2036, line 147: Good abstraction 👍

arr2036: Hmm, but the problem is this should really be taking an array of `vp_tmpl_t`, so it's more `tmpl_to_slist`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+	/* Iterate over the VP and add the string value to the curl_slist */
+	vp = fr_cursor_iter_by_da_init(&uctx->cursor, &uctx->request->packet->vps, dict_attr);
+	while (vp) {
+		*out = curl_slist_append(*out, vp->vp_strvalue);
+		elems_added++;
+		vp = fr_cursor_next(&uctx->cursor);
+	}
+	/* Check that the elements were found */
+	if (elems_added == 0) {
+		RDEBUG2("There were no %s added to the slist", dict_attr->name);
```
arr2036, line 162: Doesn't really make sense.  `RDEBUG2` should be user interpretable.  Users don't know what slists are.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+		return (++addresses_added);
+	}
+	talloc_free(&address_buffer.buff);
+	return 0;
+}
+
+/*
+ * 	Adds every element associated with a dict_attr to a curl_slist
+ */
+static int da_to_recipients(fr_mail_ctx *uctx, struct curl_slist **out, const fr_dict_attr_t *dict_attr)
```
arr2036, line 219: Same with this, it should be operating on a tmpl not a da

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+	if (elems_added == 0) {
+		RDEBUG2("There were no %s added to the slist", dict_attr->name);
+	}
+	return elems_added;
+}
+
+/*
+ * 	Adds every element associated with a dict_attr to an sbuff
+ */
+static ssize_t da_to_sbuff (fr_mail_ctx *uctx, fr_sbuff_t *out, const fr_dict_attr_t *dict_attr)
```
arr2036, line 242: and here...

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+		FR_CURL_REQUEST_SET_OPTION(CURLOPT_PASSWORD, password->vp_strvalue);
+		RDEBUG2("Username and password set");
+	}
+
+	/* Send the envelope address */
+	FR_CURL_REQUEST_SET_OPTION(CURLOPT_MAIL_FROM, get_envelope_address(mail_ctx, inst));
+
+	/* Set the recipients */
+	mail_ctx->recipients = NULL; /* Prepare the recipients curl_slist to be initialized */
+       	if(recipients_source(mail_ctx, inst) <= 0) {
```
arr2036, line 658: Random indentation

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,897 @@
...
+		RDEBUG2("At least one recipient is required to send an email");
+		goto error;
+	}
+	FR_CURL_REQUEST_SET_OPTION(CURLOPT_MAIL_RCPT, mail_ctx->recipients);
+
+	/* Set the header elements */
+	mail_ctx->header = NULL; /* Prepare the header curl_slist to be initialized */
+       	if(header_source(t, mail_ctx, inst) != 0) {
```
arr2036, line 666: Random indentation

src/tests/modules/smtp/smtp_crln/module.conf
```diff
@@ -0,0 +1,37 @@
...
+
+		random_file		= /dev/urandom
+
+		check_cert_cn		= no
+
+		require_cert		= demand
+
+		extract_cert_attrs	= yes
+
+        }
```
arr2036, line 21: random indentation

src/tests/modules/smtp/smtp_stringparse/module.conf
```diff
@@ -0,0 +1,37 @@
...
+
+		random_file		= /dev/urandom
+
+		check_cert_cn		= no
+
+		require_cert		= demand
+
+		extract_cert_attrs	= yes
+
+        }
```
arr2036, line 21: Random indentation

src/modules/rlm_smtp/rlm_smtp.c
arr2036: Sender address should probably be dynamic too?

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	curl_mime		*mime;
+} fr_mail_ctx;
+
+/*
+ *	A mapping of configuration file names to internal variables.
+ */
+static const CONF_PARSER module_config[] = {
+	{ FR_CONF_OFFSET("uri", FR_TYPE_STRING, rlm_smtp_t, uri) },
+	{ FR_CONF_OFFSET("template_directory", FR_TYPE_STRING, rlm_smtp_t, template_dir) },
+	{ FR_CONF_OFFSET("attachments", FR_TYPE_TMPL | FR_TYPE_MULTI, rlm_smtp_t, attachments), .dflt = "&SMTP-Attachments", .quote = T_BARE_WORD},
```
arr2036, line 117: Default should be `&SMTP-Attachments[*]`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	{ FR_CONF_OFFSET("sender_address", FR_TYPE_STRING, rlm_smtp_t, sender_address) },
+	{ FR_CONF_OFFSET("envelope_address", FR_TYPE_STRING, rlm_smtp_t, envelope_address) },
+	{ FR_CONF_OFFSET("recipients", FR_TYPE_TMPL | FR_TYPE_MULTI, rlm_smtp_t, recipient_addrs), .dflt = "&SMTP-Recipients", .quote = T_BARE_WORD},
```
arr2036, line 120: Default should be `&SMTP-Recipients[*]`

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	{ FR_CONF_OFFSET("TO", FR_TYPE_TMPL | FR_TYPE_MULTI, rlm_smtp_t, to_addrs), .dflt = "&SMTP-TO", .quote = T_BARE_WORD},
```
arr2036, line 121: Same for the rest of these.  Should be looking at all instances of the attribute, not just the first.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	{ FR_CONF_OFFSET("tls", FR_TYPE_SUBSECTION, rlm_smtp_t, tls), .subcs = (void const *) fr_curl_tls_config },//!<loading the tls values
+	CONF_PARSER_TERMINATOR
+};
+
+/** Iterate over a talloced array of elements
+ *
+@verbatim
+talloc_foreach(vpt_m, tmpl_t *, vpt) {
+	tmpl_debug(vpt);
+}
```
arr2036, line 135: That's defined centrally.  Just include the `<freeradius-devel/talloc.h>` header

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	if (elems_added == 0) {
+		RDEBUG2("There were no %s elements found", dict_attr->name);
+	}
+	return elems_added;
+}
+
+/*
+ *	Adds a comma separated list of emails to a curl slist
+ */
+static int address_to_slist(void *ctx, struct curl_slist **out, const char * address_source, size_t inlen)
```
arr2036, line 183: I get what you're trying to do here, and I think it's fine to do format checks, but automatically splitting on commas makes this inconsistent with other areas of the server.

If someone wants to split then they can in theory do `%{explode:, <input>}` to get a list of addresses.  It doesn't quite work yet, but it'll work very soon once I'm done with the code I'm working on.

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+static int tmpl_arr_to_slist (fr_mail_ctx *uctx, struct curl_slist **out, tmpl_t ** const tmpl)
+{
+	REQUEST 	*request = uctx->request;
+	int 		len, count, i;
+	tmpl_t 		*current;
+
+	len = talloc_array_length(tmpl);
+	count = 0;
+
+	for(i = 0; i < len; i++) {
```
arr2036, line 263: This would be where you'd use the `talloc_foreach` macro...

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	return copied;
+}
+
+/*
+ * 	Adds a tmple to an sbuff as a comma separated lsit
+ */
+static ssize_t tmpl_to_sbuff (fr_mail_ctx *uctx, fr_sbuff_t *out, tmpl_t const *vpt)
+{
+	REQUEST			*request = uctx->request;
+
```
arr2036, line 317: This is fine for now until we have a new tmpl_expand function

src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -0,0 +1,1107 @@
...
+	fr_sbuff_in_bstrncpy(&sbuff, preposition, prep_len);
+
+	len = talloc_array_length(vpt);
+	out_len = 0;
+	for (i = 0; i < len; i++) {
+		current = vpt[i];
+		RDEBUG2("NAME: %s", current->name);
+		/* If there have already been elements added, keep them comma separated */
+		if (out_len > 0) {
+			out_len += fr_sbuff_in_bstrncpy(&sbuff, ", ", 2);
```
arr2036, line 363: just use fr_sbuff_in_bstrcpy_literal(&sbuff, ", ")

scripts/travis/exim-setup.sh
```diff
@@ -0,0 +1,133 @@
...
+#
+# Prepare the directories and files needed for running exim
+#
+
+# Stop any currently running exim instance
+echo "Checking for a running exim instance"
+if [ -e "${RUNDIR}/exim.pid" ]
+then
+	echo "Stopping the current exim instance"
+	kill "$(cat ${RUNDIR}/exim.pid)"
```
jpereira, line 39: Maybe should be good force using SIGKILL. e.g: `kill -9 "$(cat ${RUNDIR}/exim.pid)"`

arr2036: No.  That can cause the sockets to remain bound for a period and prevent exim being restarted.  You only use `kill -9` if the process is completely unresponsive.

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Converted to bool and changed to an option in the config

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: changed to _free_mail_ctx

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Changed

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Changed

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: removed

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Everything is freed in it's appropriate function. No memory leaks on travis

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: changed

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Alan told me to

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Fixed

src/modules/rlm_smtp/rlm_smtp.c
stephenmac98: Fixed

PR #3554, opened 2020-08-05
Downstream - OpenSSL DH Parameter fixes - v3.0.x

cipherboy:
OpenSSL in RHEL 8.3 introduces a breaking change in FIPS mode:
user-provided dhparams will be ignored (and dhparam generation
may fail as well), unless they are on the FIPS approved list of
parameters. However, OpenSSL since v1.1.1 will automatically select
an appropriate DH parameter set anyways, if the user did not provide
any. These will be FIPS approved.

----

You can see this documented in the [RHEL 8.3 beta release notes](
https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8-beta/html-single/8.3_release_notes/index#known-issue_security):

> OpenSSL in FIPS mode accepts only specific D-H parameters
>
> In FIPS mode, Transport Security Layer (TLS) clients that use OpenSSL return a bad dh value error and abort TLS connections to servers that use manually generated parameters. This is because OpenSSL, when configured to work in compliance with FIPS 140-2, works only with D-H parameters compliant to NIST SP 800-56A rev3 Appendix D (groups 14, 15, 16, 17, and 18 defined in RFC 3526 and with groups defined in RFC 7919). Also, servers that use OpenSSL ignore all other parameters and instead select known parameters of similar size. To work around this problem, use only the compliant groups.
>
> (BZ#1810911) 

Note that the above BZ is private; [`rh-bz#1859527`](https://bugzilla.redhat.com/show_bug.cgi?id=1859527) is the tracker against this package.

My fix was going to be two-fold:

 1) Ship a single hard-coded, FIPS-approved dhparam file, removing dynamic generation from `/etc/raddb/certs/{bootstrap,Makefile}`.
 2) Add this patch, ignoring it in FIPS mode.

The net result should be that existing deployments continue to function as expected, while new installs will be functional across split deployment scenarios (FIPS machine talking to non-FIPS RADIUS server).

### DIFF
--- a/src/main/tls.c
+++ b/src/main/tls.c
@@ -1352,6 +1352,23 @@ static int load_dh_params(SSL_CTX *ctx, char *file)
 
 	if (!file) return 0;
 
+	/*
+	 * Prior to trying to load the file, check what OpenSSL will do with it.
+	 *
+	 * Certain downstreams (such as RHEL) will ignore user-provided dhparams
+	 * in FIPS mode, unless the specified parameters are FIPS-approved.
+	 * However, since OpenSSL >= 1.1.1 will automatically select parameters
+	 * anyways, there's no point in attempting to load them.
+	 *
+	 * Change suggested by @t8m
+	 */
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+	if (FIPS_mode() > 0) {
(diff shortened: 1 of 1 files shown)


cipherboy:
Let me know if you'll take this patch (in which case I'll open a corresponding one for master) or if you have any feedback on the proposed scenario. Note that this'll eventually affect CentOS 8 as well.

PR #3558, opened 2020-08-05
redis_ippool: Use client "uid" (Option 61) as device key when this is provided

terryburton:
As per RFC 2132.

Also add more clarity and examples to the description of the module configuration item.


terryburton:
Agreed with @arr2036 that we would rework this as an example. Don't merge.

terryburton:
Reworked as examples.

PR #3563, opened 2020-08-07
V4 new dhcp schema

ndptech:
Make a schema for DHCP address pools that reflects fields needed for DHCP

The status field enables IPs in the pool to be marked as dynamic, static, declined or disabled to allow for managing static IPs within the pool and handling DHCP-Decline packets


arr2036:
Can we change `gatewayipaddress` to something more generic.  In an ISP network this would usually be the Circuit ID instead of the gateway IP address as the gateway would be too coarse to be much use for hardening.

PR #3567, opened 2020-08-11
tacacs: Fix encode and decoder

jpereira:
Several fixes and improvements as the below commit list.


share/dictionary/tacacs/dictionary
```diff
@@ -8,7 +8,7 @@
 # Version:	$Id$
 #
 
-PROTOCOL	TACACS		5
+PROTOCOL	TACACS		5	verify=lib
```
arr2036, line 11: Does the tacacs dictionary have a validation library? If it does it should be added to the same commit.  Splitting out commits is fine, but if you don't group related changes it makes it very difficult to review.

src/bin/unit_test_attribute.mk
```diff
@@ -1,5 +1,5 @@
 TARGET		:= unit_test_attribute
 SOURCES		:= unit_test_attribute.c
 
-TGT_PREREQS	:= $(LIBFREERADIUS_SERVER) libfreeradius-tacacs.a libfreeradius-util.a
+TGT_PREREQS	:= $(LIBFREERADIUS_SERVER) libfreeradius-util.a
```
arr2036, line 4: It does unless you've added test points

jpereira: https://github.com/FreeRADIUS/freeradius-server/pull/3567/commits/618bd09fe962a2758d1efaec6936720cd6f3a00b

src/protocols/tacacs/encode.c
```diff
@@ -23,191 +23,697 @@
...
+#include <freeradius-devel/util/struct.h>
 
 #include "tacacs.h"
 #include "attrs.h"
 
-int fr_tacacs_packet_encode(RADIUS_PACKET * const packet, char const * const secret, UNUSED size_t secret_len)
+/**
+ *	Encode a TACACS+ 'arg_N' fields.
+ */
+static size_t tacacs_encode_body_arg_n_len(VALUE_PAIR *vps, fr_dict_attr_t const *da, uint8_t **body)
```
arr2036, line 42: Please convert this and all other encoder functions to use sbuffs or dbuffs depending on which API you think is easier. We won't be accepting new encoder functions that use manual length checks like this.

src/protocols/tacacs/encode.c
```diff
@@ -23,191 +23,697 @@
...
+		 * subsequent packet will increment the sequence number by one. Thus clients
+		 * only send packets containing odd sequence numbers, and TACACS+ servers only
+		 * send packets containing even sequence numbers.
+		 */
+		if (packet_is_authen_start_request(packet)) { /* Start */
+			/**
+			 * 4.1. The Authentication START Packet Body
+			 *
+			 * 1 2 3 4 5 6 7 8  1 2 3 4 5 6 7 8  1 2 3 4 5 6 7 8  1 2 3 4 5 6 7 8
+			 * +----------------+----------------+----------------+----------------+
```
arr2036, line 208: If you're using the struct decoder would it make more sense to use the struct encoder too?

src/protocols/tacacs/encode.c
```diff
@@ -23,191 +23,697 @@
...
+static size_t tacacs_encode_body_arg_n_len(VALUE_PAIR *vps, fr_dict_attr_t const *da, uint8_t **body)
+{
+	VALUE_PAIR *vp;
+	fr_cursor_t	cursor;
+	size_t 	arg_cnt = 0;
+	uint8_t *p = *body;
+
+	fr_assert(body != NULL);
+
+	for (vp = fr_cursor_init(&cursor, &vps);
```
arr2036, line 51: That's be encoding internal tacacs attributes too, which are disallowed

src/protocols/tacacs/tacacs.h
```diff
@@ -50,16 +52,16 @@
...
 typedef enum {
-	TAC_PLUS_INVALID		= 0x00,
-	TAC_PLUS_AUTHEN			= 0x01,
-	TAC_PLUS_AUTHOR			= 0x02,
-	TAC_PLUS_ACCT			= 0x03
+	FR_TAC_PLUS_INVALID		= 0x00,
+	FR_TAC_PLUS_AUTHEN			= 0x01,
+	FR_TAC_PLUS_AUTHOR			= 0x02,
+	FR_TAC_PLUS_ACCT			= 0x03
 } tacacs_type_t;
```
arr2036, line 59: So you rename all the macros but leave the structures without an `fr_` prefix?

arr2036:
I honestly don't understand the order of commits here... Why would you start by enabling the tests when the test points haven't even been added yet so the tests are guaranteed to fail...

jpereira:
Because the previous tests and code were not following the RFC. therefore, I started creating the tests based on .pcap captured between a NAS & TACACS+ server.

alandekok:
I've cherry-picked the patches in a slightly different order.

PR #3576, opened 2020-08-18
Make sqlippool generic rather than protocol specific

ndptech:
Rename the queries used by sqlippool to reflect their actions rather than protocol specific packets

### DIFF
--- a/raddb/mods-config/sql/ippool/mssql/queries.conf
+++ b/raddb/mods-config/sql/ippool/mssql/queries.conf
@@ -6,14 +6,16 @@
 
 # MSSQL-specific syntax - only needed if the queries need wrapping as
 # a transaction - specifically when the SELECT and UPDATE are separate queries
-#allocate_begin = "BEGIN TRAN"
-#allocate_commit = "COMMIT TRAN"
+#alloc_begin = "BEGIN TRAN"
+#alloc_commit = "COMMIT TRAN"
+alloc_begin = ""
+alloc_commit = ""
 
 #
 #  This query tries to allocate the same IP-address to the user
 #  that they had last session
 #
-allocate_existing = "\
+alloc_existing = "\
 	WITH cte AS ( \
 		SELECT TOP(1) \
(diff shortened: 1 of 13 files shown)


raddb/mods-config/sql/ippool/mssql/queries.conf
```diff
@@ -112,13 +112,13 @@ pool_check = "\
 #allocate_commit = ""
 
 #
-#  This series of queries frees an IP number when an accounting START record arrives.
+#  This series of queries frees an IP number when an accounting START / ALIVE record arrives.
```
alandekok, line 115: Maybe not "frees" an IP address?  But extends it?

src/modules/rlm_sqlippool/rlm_sqlippool.c
```diff
@@ -125,32 +125,32 @@ static CONF_PARSER module_config[] = {
 	{ FR_CONF_OFFSET("pool_check", FR_TYPE_STRING | FR_TYPE_XLAT, rlm_sqlippool_t, pool_check), .dflt = "" },
 
 
-	{ FR_CONF_OFFSET("extend_begin", FR_TYPE_STRING | FR_TYPE_XLAT, rlm_sqlippool_t, extend_begin), .dflt = "START TRANSACTION" },
+	{ FR_CONF_OFFSET("extend_begin", FR_TYPE_STRING | FR_TYPE_XLAT, rlm_sqlippool_t, extend_begin), .dflt = "" },
```
alandekok, line 128: TBH you could just remove the `.dflt = ""`, and it would be fine.

alandekok:
Can you rebase these on top of the "master" branch, instead of merging it?  That way the history looks cleaner.

But overall this is very good, thanks.

arr2036:
@ndptech  can you write up a series of generic method names for ippool modules that we can discuss and standardise on.  I don't believe we currently support overrides based on section names, but we should support overrides for generic methods like `sqlippool.update`.  We should add these methods as well as section bindings to both ippool modules.

Maybe `alloc`, `update`, `release`, `bulk-release`, `query`, `bulk-query`.

PR #3577, opened 2020-08-18
Amend redis_ippool to handle DHCP Request correctly

ndptech:
A DHCP Request should update the address previously assigned in response
to a DHCP Discover.
Former behaviour used the same code as Discover which is less efficient
and also uses the "offer" time rather than the "lease" time for the
setting of the exipry time.

### DIFF
--- a/src/modules/rlm_redis_ippool/rlm_redis_ippool.c
+++ b/src/modules/rlm_redis_ippool/rlm_redis_ippool.c
@@ -1327,6 +1327,20 @@ run:
 	return mod_action(inst, request, action);
 }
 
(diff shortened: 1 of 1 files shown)


arr2036:
You're meant to set an explicit action in the section (&control:Pool-Action) so the module does the right thing.  As you saw there's already support for the operation.  The redis ippool code predates the ability to set section specific methods for DHCPv4 by about four years :)

The more complete fix, if you wanted to look into it, is to add generic methods beyond `authorize`, `authenticate`, etc.. and also map protocol specific sections to those methods, and drop support for `Pool-Action` entirely.

It's basically the same set of methods as would be needed for the SQL module.

Users can then call `ippool.update` `ippool.release` if they want to do overrides.

PR #3585, opened 2020-08-20
Implement DHCP-Decline packet handling

ndptech:
Make use of the unused "off" queries to handle a DHCP-Decline packet and
mark addresses as declined.

### DIFF
--- a/raddb/policy.d/dhcp
+++ b/raddb/policy.d/dhcp
@@ -13,3 +13,16 @@ dhcp_sqlippool_release {
 
 }
 
+#  Assign compatibility data to request for sqlippool for DHCP Decline
+dhcp_sqlippool_decline {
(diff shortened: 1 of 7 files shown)


raddb/policy.d/dhcp
```diff
@@ -13,3 +13,16 @@ dhcp_sqlippool_release {
...
+	#  Do a minor hack to the request so that it looks
+	#  like a RADIUS Accounting Off request to the SQL IP Pool module.
+	update request {
+		&Acct-Status-Type = Accounting-Off
+	}
+
+	#  Call the actual module in accounting context
+	dhcp_sqlippool.accounting
+
+}
```
alandekok, line 28: add an extra LF / CR?

PR #3598, opened 2020-08-24
virtual_server_compile_sections() The offset could be 0 too

jpereira:


src/lib/server/virtual_servers.c
```diff
@@ -1301,10 +1301,10 @@ int virtual_server_compile_sections(CONF_SECTION *server, virtual_server_compile
 			 *	Cache the CONF_SECTION which was found.
 			 */
 			if (uctx) {
-				if (list[i].offset > 0) {
+				if (list[i].offset >= 0) {
```
arr2036, line 1304: Useless comparison covers the entire range.

src/lib/server/virtual_servers.c
```diff
@@ -1301,10 +1301,10 @@ int virtual_server_compile_sections(CONF_SECTION *server, virtual_server_compile
...
 					*(CONF_SECTION **) (((uint8_t *) uctx) + list[i].offset) = subcs;
 				}
-				if (list[i].instruction > 0) {
+				if (list[i].instruction >= 0) {
```
arr2036, line 1307: Useless comparison covers the entire range.

arr2036: Trying out my new role as static analysis tool

alandekok:
If `offset` isn't set, it's initialized to zero.  So we can't check for that.  It's better just to pad the input structs.

PR #3611, opened 2020-09-03
mod_perl: Print out the version

jpereira:
It's helpful to know which version was linked.

### DIFF
--- a/src/modules/rlm_perl/rlm_perl.c
+++ b/src/modules/rlm_perl/rlm_perl.c
@@ -1053,7 +1053,11 @@ DIAG_ON(nested-externs)
 
 static int mod_load(void)
 {
+#define LOAD_INFO(_fmt, ...) fr_log(LOG_DST, L_INFO, __FILE__, __LINE__, "rlm_perl - " _fmt,  ## __VA_ARGS__)
 #define LOAD_WARN(_fmt, ...) fr_log_perror(LOG_DST, L_WARN, __FILE__, __LINE__, "rlm_perl - " _fmt,  ## __VA_ARGS__)
+
+	LOAD_INFO("Perl version: %s", PERL_API_VERSION_STRING);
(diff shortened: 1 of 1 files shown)


jpereira:
e.g:

```
...
Thu Sep  3 13:44:00 2020: Debug : rlm_perl loaded, checking if it's valid
Thu Sep  3 13:44:00 2020: Debug : rlm_perl validated.  Handle address 0x60d0000450d0, symbol address 0x5a3d520
Thu Sep  3 13:44:00 2020: Info  : rlm_perl - Perl version: 5.32.0
Thu Sep  3 13:44:00 2020: Info  : Loaded module "rlm_perl"
Thu Sep  3 13:44:00 2020: Debug :     perl {
...
```

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -1053,8 +1053,12 @@ DIAG_ON(nested-externs)
 
 static int mod_load(void)
 {
+#define LOAD_INFO(_fmt, ...) fr_log_perror(LOG_DST, L_INFO, __FILE__, __LINE__, "rlm_perl - " _fmt,  ## __VA_ARGS__)
 #define LOAD_WARN(_fmt, ...) fr_log_perror(LOG_DST, L_WARN, __FILE__, __LINE__, "rlm_perl - " _fmt,  ## __VA_ARGS__)
 
+	LOAD_INFO("Perl version: %s", PERL_API_VERSION_STRING);
```
arr2036, line 1059: Just use INFO, you're not printing an error, and you'll print irrelevant error messages from the error stack if they've not been cleared by previous code.

jpereira:
It might be interesting to have that in v3.0.x. let me know if you agree @arr2036

src/modules/rlm_perl/rlm_perl.c
jpereira: done

PR #3614, opened 2020-09-03
v3.0.x DHCP fixes: Honour hinted IP address; Proper handling of DHCP REQUESTs

terryburton:
Per RFCs for DHCP, we now honours an IP address hint by introducing a new configuration item `req_attribute_name` (request attribute containing the address hint) and `allocate_requested` query (and updating the SPs).

Differentiate REQUESTs from DISCOVERs and fixes a few issues and allows multiple FR instances sharing a database to serve the same set of clients, each receiving the same broadcast DHCP packets from clients:

* Fixes an issue with FR ACKing a different IP address that was requested by the client under some heavily loaded conditions, and when multiple servers sharing a database offer different addresses to the same client.
* Fixes an issue where renewals will not succeed because these REQUESTs will not thave a DHCP-Requested-IP-Address, only a DHCP-Client-IP-Address.
* Implements short offer times and releases all outstanding offers when a particular offer is requested. (As happens in multi-headed configurations.)
* Implements the RFC behaviour where a server should remain quiet when it is not being spoken to, i.e. a mismatching DHCP-Server-Identifier during a REQUEST.

### DIFF
--- a/raddb/mods-available/dhcp_sqlippool
+++ b/raddb/mods-available/dhcp_sqlippool
@@ -32,6 +32,9 @@ sqlippool dhcp_sqlippool {
 	# SQL table to use for ippool range and lease info
 	ippool_table = "dhcpippool"
 
+	# The duration for which a lease is reserved whilst under offer
+	offer_duration = 10
+
 	# IP lease duration. (Leases expire even if no DHCP-Release packet is received)
 	lease_duration = 7200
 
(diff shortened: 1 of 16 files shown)


terryburton:
@alandekok: Tested by @ndptech. Ready for FR core dev review.

ndptech:
@alandekok: I set up FR with each of the 5 database back ends and tested the functionality of the new queries by sending in specifically crafted DHCP packets (e.g. adding a Requested-IP-Address to a DISCOVER) and examining both the FR diagnostics output and the returned packets to check that we have correct behaviour.

PR #3620, opened 2020-09-05
Alter refcount for old storage so it is kept allocated until next SSL storage flush.

lytboris:


lytboris:
This is the last bit for OpenSSL 1.1.1 mentioned in #1831

PR #3625, opened 2020-09-08
Fixed repeated accounting for closed row

TK-Khaw:
Row with period_end that is not null should not be incremented despite its period_start is lesser then acctstoptime of rows in radacct. A check for null for period_end is added before update to prevent such data usage accounting error.

### DIFF
--- a/raddb/mods-config/sql/main/postgresql/process-radacct.sql
+++ b/raddb/mods-config/sql/main/postgresql/process-radacct.sql
@@ -108,7 +108,9 @@ BEGIN
         SET
             acctinputoctets = data_usage_by_period.acctinputoctets + EXCLUDED.acctinputoctets,
             acctoutputoctets = data_usage_by_period.acctoutputoctets + EXCLUDED.acctoutputoctets,
-            period_end = v_end;
+            period_end = v_end
+        WHERE
+            data_usage_by_period.period_end IS NULL;
 
     --
     -- Create an open-ended "next period" for all ongoing sessions and carry a


terryburton:
When there are no ongoing sessions there are no rows in the the table that represents the current, open-ended period. The existing method for picking the start of the current period is erroneously picking the start of the previous closed period.

Picking the start of the current period correctly seems to solve the problem:

```
diff --git a/raddb/mods-config/sql/main/postgresql/process-radacct.sql b/raddb/mods-config/sql/main/postgresql/process-radacct.sql
index 5b486170e3..19c7957832 100644
--- a/raddb/mods-config/sql/main/postgresql/process-radacct.sql
+++ b/raddb/mods-config/sql/main/postgresql/process-radacct.sql
@@ -51,6 +51,7 @@ CREATE TABLE data_usage_by_period (
     acctoutputoctets bigint
 );
 ALTER TABLE data_usage_by_period ADD CONSTRAINT data_usage_by_period_pkey PRIMARY KEY (username, period_start);
+CREATE INDEX data_usage_by_period_pkey_period_end ON data_usage_by_period(period_end);
 
 
 --
@@ -78,7 +79,7 @@ DECLARE v_start timestamp;
 DECLARE v_end timestamp;
 BEGIN
 
-    SELECT COALESCE(MAX(period_start), TO_TIMESTAMP(0)) INTO v_start FROM data_usage_by_period;
+    SELECT COALESCE(MAX(period_end) + INTERVAL '1 SECOND', TO_TIMESTAMP(0)) INTO v_start FROM data_usage_by_period;
     SELECT DATE_TRUNC('second',CURRENT_TIMESTAMP) INTO v_end;
 
     --
```

Please try the above and let me know whether this works for you, in which case I shall create patches.

terryburton:
#3634 and #3635 should be merged instead of this.

TK-Khaw:
Hi glad to know that it had been fixed. May I ask if there's any possible problems that I may had overlooked with my proposed fix? Would be happy if you can enlighten me on the issue thanks.

terryburton:
@TK-Khaw Does you patch not fail to record a new period for the consumed data in the icase where your filter is triggered? It seems to generate the negated carry but not the data for the current period.

TK-Khaw:
@terryburton I see like you said my patch will only record a new period for those rows whose period_end is NULL upon function called. For those whose period_end is not NULL a new row with record of 0 consumed data do not appear as a results of my method. 

I get why this can an issue. Thanks for clarifying it for me  : )

PR #3629, opened 2020-09-09
Some fixes and tests for VMPS

jpereira:


src/tests/unit/protocols/vmps/base.txt
```diff
@@ -0,0 +1,27 @@
...
+#
+proto vmps
+proto-dictionary vmps
+
+#
+#	decode Join-Request
+#
+decode-proto 01 01 00 06 00 00 12 34 00 00 0c 01 00 04 c0 a8 45 01 00 00 0c 02 00 04 65 74 68 30 00 00 0c 03 00 05 6b 61 6c 6f 73 00 00 0c 04 00 0b 74 61 70 69 6f 63 61 2e 6c 61 6e 00 00 0c 07 00 01 00 00 00 0c 06 00 06 ca fe ca fe ca fe
+match Packet-Type = Join-Request, Error-Code = No-Error, Sequence-Number = 4660, Client-IPv4-Address = 192.168.69.1, Port-Name = "eth0", VLAN-Name = "kalos", Domain-Name = "tapioca.lan", Unknown = 0x00, MAC-Address = ca:fe:ca:fe:ca:fe
+
```
alandekok, line 13: we should also check that we can encode it, too

```
encode-proto -
match ...
```

using data from the previous `decode-proto`

jpereira: Done!

PR #3632, opened 2020-09-09
Change internal encoding to use markers in anticipation of extensible…

jejones3141:
… dbuffs


src/protocols/internal/encode.c
```diff
@@ -244,32 +221,24 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
...
-		}
-		memcpy(len_field, buff, flen);
-		enc_field[0] |= ((flen - 1) << 2);
+	/*
+	 *	Ugh, it's a long one, need to memmove the
+	 *	data.
+	 */
+	if (flen > 1) {
+		FR_DBUFF_ADVANCE_RETURN(&work_dbuff, flen - 1);
+		memmove(fr_dbuff_marker_current(&value_field) + flen - 1, fr_dbuff_marker_current(&value_field), slen);
```
arr2036, line 234: This then leaves the current pointer in work_dbuff set incorrectly, which isn't a problem now, but could well be if we ever tried to refactor.  You need to add a dbuff function that operates on a marker or sbuff, and updates all pointers within the segment being moved. The function would also need to check that no markers or `p` are moved past the end of the buffer.  If you don't feel confident in writing such a function then I can take a look.

jejones3141:
Thanks; I'll take a look and ask for help if needed. Will let you know one
way or another tonight.

jejones3141:
I've pushed a change that isn't a new dbuff operation, but which I think
corrects the issue of work_dbuff's position pointer; in the loop over
children of TLVs, I overlooked a needed change and passed
FR_DBUFF_RESERVE() dbuff rather than &work_dbuff, so that in that case,
work_dbuff's position pointer wasn't getting advanced. This needs to happen
whatever is done about the memmove().

PR #3645, opened 2020-09-11
Add counter to DHCP ippool schema

ndptech:
By updating a counter every time a REQUEST is responded to successfully
we ensure that records are changed, thus all database engines return an
affected row count of 1 for successful matches.  This safeguards the new
DHCP site config which uses the affected row count to determine between
ACK and NAK responses.

### DIFF
--- a/raddb/mods-config/sql/ippool-dhcp/mssql/queries.conf
+++ b/raddb/mods-config/sql/ippool-dhcp/mssql/queries.conf
@@ -201,7 +201,9 @@ start_update = "\
 
 #
 #  This query extends an existing lease (or offer) when a DHCP REQUEST packet
-#  arrives
+#  arrives.  This query must update a row when a lease is succesfully requested
+#  - queries that update no rows will result in a NAK reply.  In this example
+#  incrementing "counter" is used to achieve this.
 #
 alive_update = "\
 	UPDATE ${ippool_table} \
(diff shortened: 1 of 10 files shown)


terryburton:
LGTM.

terryburton:
It would be useful to articulate the requirement for the query to return a non-zero number of rows updates for the module to consider the lease extension to have succeeded and therefore return `ok` rather than `notfound`. Perhaps as a comment in the queries so that users who customise their schema are certain to find it.

alandekok:
Is there another field we could use instead of a counter?  i.e. track some additional metric which may be useful for the admin?

A counter ensures that the DB returns the correct number of updated rows.  But the counter field in the DB isn't useful to the admin.  If it has value `5` versus `2112`, that doesn't mean anything to the admin.

Maybe this could even be a timestamp, with sub-second resolution?  i.e. a "lease updated time" would probably still work, and be more informative.

terryburton:
The allocation counter aligns with what is done in `rlm_redis_ippools` in v4, but I concur that if there is a more useful metric then we should use this (and maybe update redis to match?)

ndptech:
We could make it a "last requested" time stamp instead of a counter.  Sub-second resolution on that **should** make it do what's needed, though not totally guaranteed in the case where a number of `REQUEST` packets have got queued up, de-duplication is not running (or the client has changed the xid for each transmission - allowed within the RFCs) and they happen to get handled at the same time.
The counter is guaranteed to achieve the desired result, and is what's done in `rlm_redis_ippools`

PR #3652, opened 2020-09-16
Make marker pointer advances past parent pointer advance parent point…

jejones3141:
…er too

Why? fr_[sd]buff_moves() with marker source/destination will advance the
marker pointer... but if that moves it past the parent's pointer, then
bytes are used that aren't reflected in fr_[sd]buff_used(). For dbuffs,
at least, this breaks the convention for encoding functions.

This change causes setting a marker pointers to check the parent's pointer
and, if and only if the set moves the marker pointer past the parent
pointer, advance the parent pointer to match the marker pointer.

### DIFF
--- a/src/lib/util/dbuff.h
+++ b/src/lib/util/dbuff.h
@@ -75,6 +75,8 @@ struct fr_dbuff_s {
 						///< buffer changes.
 };
 
+static inline void fr_dbuff_set_to_marker(fr_dbuff_marker_t *m);
+
 /** @name utility macros
  * @{
  */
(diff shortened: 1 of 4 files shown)


arr2036:
Can you uninvert the order of the functions in the sbuff code, it makes it really hard to figure out what changed

jejones3141:
Indeed I can and will.

jejones3141:
Change made and pushed.

jejones3141:
Agreed, it's unexpected and surprising behavior. Closing.

PR #3662, opened 2020-09-23
Update Mikrotik dictionary

jpereira:
Based on https://wiki.mikrotik.com/wiki/Manual:RADIUS_Client/reference_dictionary


share/dictionary.mikrotik
```diff
@@ -45,9 +42,9 @@ ATTRIBUTE	Mikrotik-Wireless-Comment		21	string
 ATTRIBUTE	Mikrotik-Delegated-IPv6-Pool		22	string
 ATTRIBUTE	Mikrotik-DHCP-Option-Set		23	string
 ATTRIBUTE	Mikrotik-DHCP-Option-Param-STR1		24	string
-ATTRIBUTE	Mikrotik-DHCP-Option-ParamSTR2		25	string
+ATTRIBUTE	Mikrotik-DHCP-Option-Param-STR2		25	string
```
alandekok, line 45: New attribute names can be added.  We shouldn't delete old names, as it will break existing configurations.

arr2036:
Yeah we only break names across major versions

PR #3666, opened 2020-09-25
v3.0.x: DHCP default policy improvements

terryburton:
Reviewed-by @ndptech up to fa0fff2.

Mainly: DHCP: Policy support for shared-network: Layer 2 with multiple subnets

    Many networks support multiple co-existing subnets without VLAN separation, i.e. we have 
    one pool containing equally valid IP addresses from each subnet.

    To support this configuration we differentiate selecting a pool based on the default or
    overridden DHCP-Network-IP-Address (indicating the Layer 2 network originating the
    request) versus setting the reply options for the subnet to which the allocated IP 
    address belongs, in particular the DHCP-Router-Address.

Also:

  * Allow group membership to be used to guard access to pools (akin to "allow members of")
  * s/DHCP-Network-IP-Address/DHCP-Network-Subnet/g
  * Some additional minor attribute name fixes in the examples
  * Lots of improvements to the text.

### DIFF
--- a/raddb/mods-available/dhcp_files
+++ b/raddb/mods-available/dhcp_files
@@ -5,13 +5,23 @@
 # Instances of the "files" module for managing DHCP options
 #
 files dhcp_network {
-	#  The file containing DHCP options mapping
+	#  The file containing network-specific DHCP options mapping
 	filename = ${modconfdir}/files/dhcp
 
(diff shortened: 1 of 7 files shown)


ndptech:
Looks good to me - gives a clear structure for DHCP configuration that covers a wide range of possible installation environments.

PR #3670, opened 2020-10-01
v4: Use data type of "key" in rlm_files to allow subnet matching for IP addresses

ndptech:
Amend the `PAIR_LIST` structure to use `fr_value_box_t` for the key - enabling data types to be stored as value box and thereby allowing searching to be based on properties of the real data type.

Specifically if the data type is an IP address or prefix, `rlm_files` now stores the key lookup in a patricia trie enabling closest enclosing subnet matching.


src/modules/rlm_attr_filter/rlm_attr_filter.c
```diff
@@ -133,8 +133,11 @@ static int attr_filter_getfile(TALLOC_CTX *ctx, rlm_attr_filter_t *inst, char co
 		     * then bitch about it, giving a good warning message.
 		     */
 		     if (fr_dict_attr_is_top_level(vp->da) && (vp->da->attr > 1000)) {
-			WARN("[%s]:%d Check item \"%s\"\n\tfound in filter list for realm \"%s\".\n",
-			       filename, entry->lineno, vp->da->name, entry->name);
+			 char *str;
+			 fr_value_box_aprint(NULL, &str, entry->name, NULL);
```
alandekok, line 137: You should be able to just pass in the `fr_value_box_t` directly, which is simpler:

```
 WARN("[%s]:%d Check item \"%s\"\n\tfound in filter list for realm \"%pV\".\n",
			      filename, entry->lineno, vp->da->name, entry->name);
```

src/modules/rlm_attr_filter/rlm_attr_filter.c
```diff
@@ -203,20 +206,21 @@ static rlm_rcode_t CC_HINT(nonnull(1,2)) attr_filter_common(void const *instance
...
+		char *str;
+		fr_value_box_aprint(NULL, &str, pl->name, NULL);
 		/*
 		 *  If the current entry is NOT a default,
 		 *  AND the realm does NOT match the current entry,
 		 *  then skip to the next entry.
 		 */
-		if ((strcmp(pl->name, "DEFAULT") != 0) &&
-		    (strcmp(keyname, pl->name) != 0))  {
+		if ((strcmp(str, "DEFAULT") != 0) &&
```
alandekok, line 216: I'd simplify this to something like:

```
if (pl->name->vb_type != FR_TYPE_STRING) continue;
if ((strcmp(pl->name->vb_strvalue, "DEFAULT") !=0) &&
     (strcmp(keyname, pl->name->vb_strvalue) != 0)) {
    continue;
}
```

code profiling shows that `malloc / free` are huge overheads, so we try to avoid them where possible.

src/modules/rlm_attr_filter/rlm_attr_filter.c
```diff
@@ -203,20 +206,21 @@ static rlm_rcode_t CC_HINT(nonnull(1,2)) attr_filter_common(void const *instance
...
+		    (strcmp(keyname, str) != 0))  {
 		    continue;
 		}
 
-		RDEBUG2("Matched entry %s at line %d", pl->name, pl->lineno);
+		RDEBUG2("Matched entry %s at line %d", str, pl->lineno);
```
alandekok, line 221: `%pV` should work here, too

src/modules/rlm_files/rlm_files.c
```diff
@@ -212,10 +220,13 @@ static int getusersfile(TALLOC_CTX *ctx, char const *filename, rbtree_t **ptree)
...
 		entry->next = NULL;
 
+		char *str;
+		fr_value_box_aprint(NULL, &str, entry->name, NULL);
+
 		/*
 		 *	DEFAULT entries get their own list.
 		 */
-		if (strcmp(entry->name, "DEFAULT") == 0) {
+		if (strcmp(str, "DEFAULT") == 0) {
```
alandekok, line 264: you can check for `entry->name->type  == FR_TYPE_STRING` and then just do the `strcmp()`

src/modules/rlm_files/rlm_files.c
```diff
@@ -355,8 +403,12 @@ static rlm_rcode_t file_common(rlm_files_t const *inst, REQUEST *request, char c
 		}
 
 		if (paircmp(request, packet->vps, check_tmp, &reply->vps) == 0) {
-			RDEBUG2("Found match \"%s\" one line %d of %s", pl->name, pl->lineno, filename);
+			char *str;
+			fr_value_box_aprint(NULL, &str, pl->name, NULL);
+
+			RDEBUG2("Found match \"%s\" on line %d of %s", str, pl->lineno, filename);
```
alandekok, line 478: `%pV` should work here, too

src/modules/rlm_attr_filter/rlm_attr_filter.c
ndptech: That gives a compiler error:
src/modules/rlm_attr_filter/rlm_attr_filter.c:27:20: warning: format ‘%s’ expects argument of type ‘char *’, but argument 10 has type ‘fr_value_box_t *’ {aka ‘struct value_box_s *’} [-Wformat=]

arr2036: Fixup the format arg where you're passing the value box to be a `%pV`

src/modules/rlm_attr_filter/rlm_attr_filter.c
```diff
@@ -133,11 +133,8 @@ static int attr_filter_getfile(TALLOC_CTX *ctx, rlm_attr_filter_t *inst, char co
 		     * then bitch about it, giving a good warning message.
 		     */
 		     if (fr_dict_attr_is_top_level(vp->da) && (vp->da->attr > 1000)) {
-			 char *str;
-			 fr_value_box_aprint(NULL, &str, entry->name, NULL);
-			 WARN("[%s]:%d Check item \"%s\"\n\tfound in filter list for realm \"%s\".\n",
-			      filename, entry->lineno, vp->da->name, str);
-			 talloc_free(str);
+			 WARN("[%s]:%d Check item \"%s\"\n\tfound in filter list for realm \"%pV\".\n",
```
arr2036, line 136: If you have time after, could you fix up these log messages.  Embedded newlines won't be printed.  If the log message should be displayed on multiple lines, it needs multiple calls to the logging function.  Trailing newlines aren't needed either.

src/modules/rlm_files/rlm_files.c
```diff
@@ -38,23 +38,27 @@ typedef struct {
 
 	char const 	*filename;
 	rbtree_t 	*common;
+	fr_trie_t	*common_trie;
```
arr2036, line 41:
```
typedef struct {
    char const *filename;
    union {
        rbtree_t    *whole;
        fr_trie_t    *prefix;
    } tree;
} rlm_file_t;
```

Don't really care about the type/field names, just the duplication.  There's three reasons to use a union here:

- It shows someone reading the code that the two trees are used exclusively.
- The code will (likely) break instead of continuing silently if we try and access the wrong type of tree.
- It uses slightly less memory (really the least compelling reason here).

You probably need an enum with the tree types here as well unless it's easy to infer which one is in use.

src/modules/rlm_files/rlm_files.c
```diff
@@ -104,17 +108,37 @@ static int pairlist_cmp(void const *a, void const *b)
 	return fr_value_box_cmp(((PAIR_LIST const *)a)->name, ((PAIR_LIST const *)b)->name);
 }
 
-static int getusersfile(TALLOC_CTX *ctx, char const *filename, rbtree_t **ptree)
+static PAIR_LIST *find_entry(rlm_files_t const *inst, fr_value_box_t *key, rbtree_t *ptree, fr_trie_t const *ptrie)
+{
+	PAIR_LIST my_e;
+
+	switch (inst->key_data_type) {
+	case FR_TYPE_IPV4_ADDR:
```
arr2036, line 116: Can you just add an enum which describes which tree is in use, and then set the value in instantiate.

- Makes it easier to add new lookup types in future, or mix up the logic which determines which tree is in use.
- The same four types aren't copy/pasted around multiple areas of the code.
- It's faster.

alandekok:
Mostly merged as of commit e9d0474bb0209a445.  I still haven't done `attr_filter`.

alandekok:
I've pulled over everything but the `attr_filter` changes.  Since that module has limited scope, I'm happy to leave it alone for now.  If we care, we can have a new PR which fixes just that module.

PR #3673, opened 2020-10-02
v4: Add prepend operator for inserting attributes at the beginning of lists

ndptech:
Implement PR 3667 for FreeRADIUS v4

For DHCP, the sequence of a repeated attribute can be significant.

Adding a prepend operator ^= allows for policy to specifically add attributes at the beginning of a list to specify the sequence of a repeated attribute when returned to the client.

### DIFF
--- a/src/lib/server/map_async.c
+++ b/src/lib/server/map_async.c
@@ -1084,6 +1084,19 @@ int map_list_mod_apply(request_t *request, vp_list_mod_t const *vlm)
 		}
 			goto finish;
 
+		case T_OP_PREPEND:
+		{
+			fr_pair_list_t	vp_from;
+
+			fr_pair_list_init(&vp_from);
+			map_list_mod_to_vps(parent, &vp_from, vlm);
+			fr_assert(!fr_pair_list_empty(&vp_from));
+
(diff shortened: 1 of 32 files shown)


arr2036:
Going to need some test cases for this too

src/lib/server/map_async.c
```diff
@@ -1168,6 +1197,35 @@ int map_list_mod_apply(REQUEST *request, vp_list_mod_t const *vlm)
...
+		fr_cursor_t	to, from, from_reverse;
+		VALUE_PAIR	*vp_from, *vp, *vp_from_reverse = NULL;
+
+		vp_from = map_list_mod_to_vps(parent, vlm);
+		fr_assert(vp_from);
+
+		/*
+		 * First reverse the "from" list so they are in the
+		 * correct sequence to prepend to the "to" list
+		 */
```
alandekok, line 1211: wouldn't it also be possible to prepend the first VP, and then just insert the subsequent VPs?  i.e.

prepend from_head, to_head
for the rest of from
   insert from into to
  next to

arr2036:
@ndptech please rehabilitate

PR #3681, opened 2020-10-08
v3: Add a "substring" xlat keyword

ndptech:
Add an xlat `substring` keyword with the syntax
    "%{substring:<attribute | string> <start> <len>}" 

Follows normal substring behaviour where `start` and `len` can be negative to count from the end of the string.

Particularly useful in conditional work for DHCP - e.g. extracting a vendor prefix from a hardware address.


alandekok:
we already have `unpack`, see `rlm_unpack`.  So this should only be for *printable* strings I think.

arr2036:
We've traditionally done substring extraction using regexes so we haven't needed an xlat.  There's no real difference between unpacking an octets string and a printable string other than maybe you want to perform UTF8 validation, so maybe just add that to unpack?

PR #3685, opened 2020-10-12
DHCP: Dedicated rlm_sql instance to support option lookup

terryburton:
New dhcp_sql instance of rlm_sql for use with DHCP:

  - Separate queries.conf and schema.sql, etc. for DHCP vs RADIUS.
  - Group lookup is disabled by default and there are no check queries provided
    to reduce query load for typical use cases.
  - No accounting-related configuration items and queries.

The schema provides dhcpreply, dhcpgroupreply and dhcpgroup (née "usergroup")
tables as per corresponding RADIUS authorize schema, except:

  - Rename column Username -> Identifier since the lookup key varies based on
    "contexts". There will typically be multiple option merges from different
    context to set network-specific options, device-specific options, classes,
    etc.
    - Set Identifier for lookup using the DHCP-SQL-Option-Identifier internal
      attribute (which gets escaped into SQL-User-Name, as usual).
  - Add a Context column ("network", "subnet", "group", "class-abc", etc.) to
    support lookups using the same tables but with the options data partitioned
    between contexts.
    - Set current Context using the DHCP-SQL-Option-Context internal attribute.
      This is not SQL escaped, however it is not set from tainted data.

Example policy for performing SQL options lookup from SQL provided in
policy.d/dhcp.

### DIFF
--- a/raddb/mods-available/dhcp_sql
+++ b/raddb/mods-available/dhcp_sql
@@ -1,0 +1,92 @@
+# -*- text -*-
+##
+## mods-available/sql -- SQL modules
+##
+##	$Id$
+
+######################################################################
+#
+#  Configuration for the DHCP-specific instance of the SQL module
+#
+#  The database schemas and queries are located in subdirectories:
+#
+#	sql/dhcp/<DB>/schema.sql	Schema
+#	sql/dhcp/<DB>/queries.conf	Reply options lookup queries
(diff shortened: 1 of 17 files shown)


terryburton:
@ndptech FYI.

arr2036:
Not from me

PR #3689, opened 2020-10-15
v4: change uses of vp->next to use cursors

ndptech:


arr2036:
Please don't change the tmpl_eval.c functions.  They're cursor callbacks so they should be accessing the fields directly, and should be rewritten when we change to the new list type.

arr2036:
Can you remove commit 
6e00d54

PR #3694, opened 2020-10-19
Take advantage of FR_DBUFF_EXTEND_LOWAT_OR_RETURN()

jejones3141:
...where appropriate. The mistaken uses of fr_dbuff_extend_lowat() include some in routines willing to just move as many bytes as there are room for, rather than decreeing it an error if not everything can be moved.


jejones3141:
Looks like it's already been taken care of.

Issue #3693, opened 2020-10-19
unable to compile with -frecord-gcc-switches


ZeroChaos-:
- Defect - Unexpected behaviour (obvious or verified by project member).

# Defect
## How to reproduce the issue

build with CFLAGS="-frecord-gcc-switches" and try to compile rlm_python3
 
I can't explain this based on reading configure or configure.ac in src/modules/rlm_python3, but this is the output in src/modules/rlm_python3/config.log
```
configure:2778: python3.7-config's cflags were "-I/usr/include/python3.7m -I/usr/include/python3.7m  -Wno-unused-result -Wsign-compare -Os -march=native -mtune=native -pipe -frecord-gcc-switches -fwrapv -DNDEBUG "
configure:2790: Sanitized cflags were "-isystem/usr/include/python3.7m -isystem/usr/include/python3.7m    -march=native -mtune=native -pipe -frecord -fwrapv "
```

As you can see, somehow -frecord-gcc-switches gets sanitized to -frecord.  I was able to find the sanitize sed lines in configure and configure.ac, but I couldn't reproduce the results...  Hopefully this is more obvious to someone else.

jpereira:
@ZeroChaos- could you please share the output of `python3-config --cflags --ldflags`?

ZeroChaos-:
zero@theprophet ~ % python3-config --cflags --ldflags
-I/usr/include/python3.7m -I/usr/include/python3.7m  -Wno-unused-result -Wsign-compare -Os -march=native -mtune=native -pipe -frecord-gcc-switches -fwrapv -DNDEBUG 
-L/usr/lib64 -lpython3.7m -lcrypt -lpthread -ldl  -lutil -lm 


https://github.com/FreeRADIUS/freeradius-server/issues/3693 looks like it would fix the issue to me.  Although that would fix the issue, I'm more curious to fix how half a cflag got stripped out instead of the whole cflag.....  this fix is papering over the issue caused by something else.

jpereira:
@ZeroChaos- you mean that changes in https://github.com/jpereira/freeradius-server/commits/v3/python3 ?

ZeroChaos-:
Correct, yes.

From what I can see, you are pulling in the cflags used to compile python, and then filtering out some things you don't like.  Something is going horribly wrong and part of a cflag is being removed causing breakage.  Finding out what is actually causing this breakage will prevent the next user from hitting a similar bug with a slightly different cflag.

[...]

PR #3699, opened 2020-10-20
rlm_python3: Strip out -frecord-gcc-switches #3693

jpereira:
We decide if enable that yes or no.


arr2036:
@jpereira Could you put in a similar patch for master, thanks!

jpereira:
> @jpereira Could you put in a similar patch for master, thanks!

for sure. https://github.com/FreeRADIUS/freeradius-server/pull/3702

PR #3703, opened 2020-10-22
Changed the vector member of fr_radius_ctx_t from pointer to array

jejones3141:
This avoids possible issues with pointers changing out from
underneath the context if a dbuff version of, for example,
fr_radius_encode(), decides to use extensible dbuffs.


alandekok:
I've pushed a fix based on your patches.

Some changes:

* leave `original` as the original packet, instead of passing in `original + 4`.  We're not using the rest of the packet for much right now.  But the APIs are usually "encode (reply, request)".  i.e. we pass in the original packet, not sub-fields of it
* remove `nullvector` and replace references to it with `memset`.  now that `packet_ctx.vector` is an array, there's no need for a static `nullvector`
* remove duplicate code for `FR_CODE_ACCOUNTING_REQUEST`.  There's no need to have a separate `case` statement for it.

jejones3141:
OK, thanks.

PR #3705, opened 2020-10-23
Shut up clang scan

jpereira:


jpereira:
Mute the below warning.

```
src/bin/fuzzer.c:58:19: warning: Dereference of null pointer (loaded from variable 'argv')
                proto = strrchr((*argv)[0], '_');
                                ^~~~~~~```

PR #3707, opened 2020-10-27
Correct the fr_net_to_int*() functions

jejones3141:
Sign extension on the less significant half could make the more significant half all 1s.

### DIFF
--- a/src/lib/util/net.h
+++ b/src/lib/util/net.h
@@ -269,35 +269,13 @@ static inline uint64_t fr_net_to_uint64(uint8_t const data[static sizeof(uint64_
 	return ((uint64_t)fr_net_to_uint32(data) << 32) | fr_net_to_uint32(data + sizeof(uint32_t));
 }
 
-/** Read an signed 16bit integer from wire format (big endian)
- *
- * @param[in] data	To convert to a 16bit signed integer of native endianness.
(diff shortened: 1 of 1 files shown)


src/lib/util/net.h
```diff
@@ -269,35 +269,10 @@ static inline uint64_t fr_net_to_uint64(uint8_t const data[static sizeof(uint64_
...
-/** Read an signed 16bit integer from wire format (big endian)
- *
- * @param[in] data	To convert to a 16bit signed integer of native endianness.
- * @return a 16 bit signed integer of native endianness.
- */
-static inline int16_t fr_net_to_int16(int8_t const data[static sizeof(int16_t)])
-{
-	return (((int16_t)data[0]) << 8) | data[1];
-}
+#define fr_net_to_int16(data)	(int16_t) fr_net_to_uint16(data)
```
alandekok, line 272: These need to have brackets around them, so that the "function" call is treated as one object

i.e.

```
#define fr_net_to_int16(_x)	((int16_t) fr_net_to_uint16(_x))
```

We typically also use different names for the arguments to macros, to make it clear that the names don't matter to the macro.

jejones3141: Corrected and pushed; thanks.

PR #3715, opened 2020-10-30
:gear: arp: Update fr_arp_encode() to use dbuff

jpereira:


src/protocols/arp/base.c
```diff
@@ -141,13 +141,16 @@ int fr_arp_entry_add(UNUSED int fd, UNUSED char const *interface,
...
+ssize_t fr_arp_encode(fr_dbuff_t *dbuff, uint8_t const *original, fr_pair_t *vps)
 {
 	ssize_t			slen;
 	fr_pair_t		*vp;
 	fr_arp_packet_t		*arp;
 	fr_cursor_t		cursor;
 	fr_da_stack_t		da_stack;
+	fr_dbuff_t		work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
+
+	FR_DBUFF_CHECK_REMAINING_RETURN(&work_dbuff, FR_ARP_PACKET_SIZE);
```
arr2036, line 153: Remove the explicit check as it'll interfere with variable length buffers.

PR #3720, opened 2020-11-04
Add dbuff read types

jejones3141:
Extend fr_dbuff_out() to support one-byte integers, float, and double.


jejones3141:
Closing; I'll first make the existing _out() routines work with markers as the _in() functions now do.

PR #3721, opened 2020-11-04
Dedicated rlm_sql instance to support DHCP option lookup

terryburton:
Forward port of https://github.com/FreeRADIUS/freeradius-server/pull/3685 and a related typo fix.


arr2036:
Is this level of duplication really required? Can't we just change the attribute used for group lookups?

PR #3724, opened 2020-11-06
v4: Swap all uses of pair_cursor.c to cursor.c

ndptech:
Phase 1 of converting singly linked lists to doubly linked lists

Remove remaining uses of old pair_cursor api and implement missing functionality to use newer cursor api


src/lib/util/pair.c
```diff
@@ -2377,3 +2377,20 @@ fr_pair_t *fr_pair_list_afrom_box(TALLOC_CTX *ctx, fr_dict_t const *dict, fr_val
...
+ * @param item	pointer to a fr_pair_t
+ * @param uctx	da to match
+ *
+ * @return true if the pair matches the da
+ */
+bool pair_matches_da(void *item, void *uctx)
+{
+	fr_pair_t	*vp = item;
+	fr_dict_attr_t	*da = uctx;
+	printf("%s %s\n", vp->da->name, da->name);
```
alandekok, line 2393: Likely not in a production build.  :)

src/lib/util/pair.c
```diff
@@ -2394,3 +2394,28 @@ bool pair_matches_da(void *item, void *uctx)
...
+ * To be used where the da being filtered by is declared as a const
+ *
+ * @param[in] cursor pointing to the list being searched
+ * @param[in] da to match
+ *
+ * @return
+ *	- next item in the list matching the given da
+ *	- NULL if no match found
+ */
+void *fr_cursor_filter_next_by_da(fr_cursor_t *cursor, const fr_dict_attr_t *uctx)
```
alandekok, line 2408: this could be `... fr_cursor_filter_next_by_da(fr_cursor_t *cursor, fr_dict_attr_t const *da)`

i.e. we put `const` after the type, and there's no need to use `uctx` here.  Just use `da`.

The reason to keep with `uctx` is if the function signature is required to take `void*uctx`  We then cast that to the correct type inside of the function.

src/lib/util/pair.c
arr2036: Should have the `fr_` prefix if this is in the util library.

src/lib/util/pair.c
arr2036: unconsting da is wrong here.  The fr_cursor_filter_next code should really take a const and do the unconsting.

arr2036: Reason for `const` after the type is so that it consistently applies to the thing to the left of it.

arr2036:
This conflicts with some work I'm doing, so I'll merge it in now to avoid you needing to rebase.

PR #3729, opened 2020-11-10
Add TFTP support

jpereira:


arr2036:
Still needs the decoder moving to dbuffs.  I was checking out the dbuff functions for decoding and I think we have enough of the API there now.

arr2036:
The resolution in the meeting was to leave this as a PR until we've figured out how to do large file transfers in the server core.

alandekok:
This should be separated based on #4051 .  so this PR should just be the new `src/process/tftp/` and `src/listen/tftp` code.

We also need to figure out a "read file" API which can be used by multiple modules.

jpereira:
> This should be separated based on #4051 . so this PR should just be the new `src/process/tftp/` and `src/listen/tftp` code.
> 
> We also need to figure out a "read file" API which can be used by multiple modules.

@alandekok Alright. As soon as we merge the https://github.com/FreeRADIUS/freeradius-server/pull/4051, I will rebase that branch against the `HEAD`.

PR #3733, opened 2020-11-10
Stop using obsolete fr_dbuff_marker_{current, advance}

jejones3141:
This also eliminates conflicting function declarations left in
src/lib/eap_aka_sim/encode.c


arr2036:
Already fixed, thanks

PR #3738, opened 2020-11-11
Add fr_value_box_from_network_dbuff()

jejones3141:
With this change, fr_value_box_from_network() is made a wrapper
around fr_value_box_from_network_dbuff()

### DIFF
--- a/src/lib/util/value.c
+++ b/src/lib/util/value.c
@@ -1457,9 +1457,16 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
 				  uint8_t const *src, size_t len,
 				  bool tainted)
 {
-	uint8_t	const *p = src;
-	uint8_t const *end = p + len;
-	size_t	min, max;
+	return fr_value_box_from_network_dbuff(ctx, dst, type, enumv, &FR_DBUFF_TMP(src, len), len, tainted);
+}
+
+ssize_t fr_value_box_from_network_dbuff(TALLOC_CTX *ctx,
+					fr_value_box_t *dst, fr_type_t type, fr_dict_attr_t const *enumv,
+					fr_dbuff_t *dbuff, size_t len,
+					bool tainted)
+{
+	size_t		min, max;
+	fr_dbuff_t	work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
 
 	min = fr_value_box_network_sizes[type][0];
 	max = fr_value_box_network_sizes[type][1];
@@ -1481,14 +1488,16 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
 
 	switch (type) {
 	case FR_TYPE_STRING:
-		if (fr_value_box_bstrndup(ctx, dst, enumv, (char const *)src, len, tainted) < 0) {
+		if (fr_value_box_bstrndup_dbuff(ctx, dst, enumv, &work_dbuff, len, tainted) < 0) {
 			return FR_VALUE_BOX_NET_OOM;
 		}
-		return len;
+		return fr_dbuff_set(dbuff, &work_dbuff);
 
 	case FR_TYPE_OCTETS:
-		if (fr_value_box_memdup(ctx, dst, enumv, src, len, tainted) < 0) return FR_VALUE_BOX_NET_OOM;
-		return len;
+		if (fr_value_box_memdup_dbuff(ctx, dst, enumv, &work_dbuff, len, tainted) < 0) {
+			return FR_VALUE_BOX_NET_OOM;
+		}
+		return fr_dbuff_set(dbuff, &work_dbuff);
 	default:
 		break;
 	}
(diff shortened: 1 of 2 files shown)


src/lib/util/value.c
```diff
@@ -1478,12 +1478,20 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
...
 {
-	uint8_t	const *p = src;
-	uint8_t const *end = p + len;
-	size_t	min, max;
+	return fr_value_box_from_network_dbuff(ctx, dst, type, enumv, &FR_DBUFF_TMP(src, len), tainted);
+}
+
+ssize_t fr_value_box_from_network_dbuff(TALLOC_CTX *ctx,
+					fr_value_box_t *dst, fr_type_t type, fr_dict_attr_t const *enumv,
+					fr_dbuff_t *dbuff,
```
arr2036, line 1486: This should still take an explicit length parameter.  The length will usually come from the attribute's length field.  The reason why we'd want to it like this as it means we don't have to write boilerplate code for doing length checks.  We can pass the dbuff representing the rest of the data on the wire, and the length from the attribute header, and fr_value_box_from_network_dbuff should error if len exceeds the available data in the dbuff.

src/lib/util/value.c
```diff
@@ -1502,14 +1510,20 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
 
 	switch (type) {
 	case FR_TYPE_STRING:
-		if (fr_value_box_bstrndup(ctx, dst, enumv, (char const *)src, len, tainted) < 0) {
+		if (fr_value_box_bstrndup(ctx, dst, enumv,
```
arr2036, line 1513: You'll need to add a dbuff variant of `fr_value_box_bstrndup`

src/lib/util/value.c
```diff
@@ -1502,14 +1510,20 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
...
-		return len;
+		fr_dbuff_advance(&work_dbuff, len);
+		return fr_dbuff_set(dbuff, &work_dbuff);
 
 	case FR_TYPE_OCTETS:
-		if (fr_value_box_memdup(ctx, dst, enumv, src, len, tainted) < 0) return FR_VALUE_BOX_NET_OOM;
-		return len;
+		if (fr_value_box_memdup(ctx, dst, enumv,
+			fr_dbuff_current(&work_dbuff), len, tainted) < 0) {
+			return FR_VALUE_BOX_NET_OOM;
```
arr2036, line 1523: And a dbuff variant of `fr_value_box_memdup`

src/lib/util/value.c
```diff
@@ -1478,12 +1478,20 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
...
+					fr_value_box_t *dst, fr_type_t type, fr_dict_attr_t const *enumv,
+					fr_dbuff_t *dbuff,
+					bool tainted)
+{
+	size_t		min, max, len;
+	fr_dbuff_t	work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
 
 	min = fr_value_box_network_sizes[type][0];
 	max = fr_value_box_network_sizes[type][1];
+	len = fr_dbuff_extend_lowat(NULL, &work_dbuff, min);
```
arr2036, line 1494: You should check the requested len against min, max.

src/lib/util/value.c
```diff
@@ -1527,65 +1541,101 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
...
 		break;
 
 	case FR_TYPE_IPV4_PREFIX:
 		dst->vb_ip = (fr_ipaddr_t){
 			.af = AF_INET,
-			.prefix = *p++,
 		};
-		memcpy(&dst->vb_ip.addr.v4, p, (end - p));
+		fr_dbuff_out(&dst->vb_ip.prefix, &work_dbuff);
+		fr_dbuff_out_memcpy((uint8_t *)&dst->vb_ip.addr.v4, &work_dbuff, len - 1);
```
arr2036, line 1552: All these should use the RETURN variant

src/lib/util/value.c
jejones3141: Makes sense given the separate length; thanks.

src/lib/util/value.c
jejones3141: Agreed.

src/lib/util/value.c
jejones3141: Perhaps I don't understand. Isn't that length (more accurately the minimum and maximum length) retrieved from the attribute's type, which is already a parameter?

jejones3141: I didn't understand. It should indeed stay.

jejones3141: That change is now made and pushed.

src/lib/util/value.c
jejones3141: Done and pushed; at least for now, `fr_value_box_bstrndup()` is the original and `fr_value_box_bstrndup_dbuff()` is the wrapper, because there are calls that pass NULL, and `FR_DBUFF_TMP(NULL, 0)` causes complaints from the address sanitizer.

src/lib/util/value.c
jejones3141: Corrected and pushed.

jejones3141: Remaining fr_dbuff_out* converted to the _RETURN versions and pushed.

jejones3141:
Explicit dbuff versions of `fr_value_box_{bstrndup, memdup}()` are now in place.

src/lib/util/value.c
```diff
@@ -1481,14 +1488,18 @@ ssize_t fr_value_box_from_network(TALLOC_CTX *ctx,
 
 	switch (type) {
 	case FR_TYPE_STRING:
-		if (fr_value_box_bstrndup(ctx, dst, enumv, (char const *)src, len, tainted) < 0) {
+		FR_DBUFF_EXTEND_LOWAT_OR_RETURN(&work_dbuff, len);
```
arr2036, line 1491: I'm pretty sure this isn't needed. I mean `fr_value_box_bstrndup_dbuff `should return a negative integer if there's not enough space, so it's a duplicate check?

PR #3747, opened 2020-11-16
CI: Add GitHub Super-Linter workflow

terryburton:
This adds a workflow that runs Super-Linter, which runs a battery of linters over the codebase: https://github.com/github/super-linter

The output of some of the linters (e.g. ShellCheck) is somewhat vexatious but there is some stuff in there that should be cleaned up, so I haven't silenced any of the checks for now.

To keep the noise down it is configured to ordinarily perform a lint of only changed files. However, all files will be linted on push to the "linter" branch so that we can see cleanup progress.

### DIFF
--- a/.github/workflows/super-linter.yml
+++ b/.github/workflows/super-linter.yml
@@ -1,0 +1,27 @@
+name: Super-Linter
+
+on:
+  push:
+    branches-ignore:
+      - coverity_scan
+  pull_request:
+
+jobs:
+
+  lint:
+    name: Super-Linter. ${{ fromJSON('["Changes-only", "Full"]')[github.ref == 'refs/heads/linter'] }}
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


arr2036:
If you look at the output it doesn't seem to actually be checking for C files?

arr2036:
Yeah... doesn't support C. Only C #

terryburton:
ACK.

terryburton:
> Yeah... doesn't support C. Only C #

Probably covers the "everything else" bit which should help catch issues with things such as scripts, SQL, Markdown, Dockerfile anti-patterns...

Example full run: https://github.com/terryburton/freeradius-server/runs/1408381484?check_suite_focus=true

A couple of things that I can see it detected from a brief skim of the output:

  * Bug with passwords containing space in the certificate scripts
  * Typos and uninitialised variables in the init scripts and other scripts
  * Failed expansions: echo 'graph_vlabel Exchanged / ${graph_period}'

PR #3751, opened 2020-11-17
RPM: Set SELinux tunable radius_use_jit=1 after installation

terryburton:
Tested in CentOS 7 and CentOS 8.

Adds a dependency on policycoreutils which installs cleanly on non SELinux environments (tested with Docker).


arr2036:
thanks! any chance of something similar for debian?

mcnewton:
Debian is sane and doesn't use SElinux :-)

arr2036:
Oh I didn't realise that.  I just always assumed it was disabled by default :)

PR #3752, opened 2020-11-18
Add extensible dbuffs that read from a stream.

jejones3141:


src/lib/util/dbuff.h
```diff
@@ -1479,9 +1523,16 @@ static inline ssize_t _fr_dbuff_out_memcpy_dbuff(uint8_t **out_p, fr_dbuff_t *ou
 static inline ssize_t _fr_dbuff_out_##_type(_type##_t *out, uint8_t **pos_p, fr_dbuff_t *in) \
 { \
 	fr_assert(out); \
-	FR_DBUFF_EXTEND_LOWAT_OR_RETURN(in, sizeof(_type##_t)); \
-	*out = fr_net_to_##_type((*pos_p)); \
-	return _fr_dbuff_set(pos_p, in, (*pos_p) + sizeof(_type##_t)); \
+	if (fr_dbuff_extend_lowat(NULL, in, sizeof(_type##_t)) >= sizeof(_type##_t)) { \
```
arr2036, line 1526: This isn't needed because we assume that an Nbit integer will be represented by Nbits on the wire.

arr2036:
This commit does too many things.  The code that deals with extending dbuffs is fine, but I don't agree with the whitepsace changes or the changes to integer parsing.  Can you resubmit with just the code changes related to extending the dbuffs.

jejones3141:
OK. I will resubmit with only those changes; I can use
fr_dbuff_memcpy_out() to show that it can shift/extend more than once to
satisfy a request.

PR #3754, opened 2020-11-18
v4: Replace fr_pair_t * with fr_pair_list_t in structures

ndptech:
Where fr_pair_t * is used for a list in data structures, replace it with fr_pair_list_t and add explicit initialisation calls for the new list headers.

### DIFF
--- a/src/modules/proto_bfd/proto_bfd.c
+++ b/src/modules/proto_bfd/proto_bfd.c
@@ -387,6 +387,8 @@ static const char *bfd_state[] = {
 static void bfd_request(bfd_state_t *session, request_t *request, fr_radius_packet_t *packet)
 {
 	memset(request, 0, sizeof(*request));
+	fr_pair_list_init(&request->control);
+	fr_pair_list_init(&request->state);
 	memset(packet, 0, sizeof(*packet));
 
 	request->packet = packet;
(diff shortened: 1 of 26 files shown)


src/modules/proto_bfd/proto_bfd.c
```diff
@@ -387,7 +387,10 @@ static const char *bfd_state[] = {
 static void bfd_request(bfd_state_t *session, request_t *request, fr_radius_packet_t *packet)
 {
 	memset(request, 0, sizeof(*request));
+	fr_pair_list_init(&request->control);
+	fr_pair_list_init(&request->state);
```
arr2036, line 391: These two would already be initialised by the request init function?

src/modules/proto_bfd/proto_bfd.c
```diff
@@ -387,7 +387,10 @@ static const char *bfd_state[] = {
...
 	memset(packet, 0, sizeof(*packet));
+	fr_pair_list_init(&packet->vps);
```
arr2036, line 393: And this should really be initialised by the packet allocation or init functions

arr2036: Didn't see the `packet->vps` changes in here, but maybe I missed them?

PR #3761, opened 2020-11-23
Add "extensible" dbuffs that read data from a file and relevant tests.

jejones3141:


src/lib/util/dbuff.h
```diff
@@ -1417,26 +1471,28 @@ size_t _fr_dbuff_move_dbuff_marker_to_dbuff_marker(fr_dbuff_marker_t *out, fr_db
...
+	size_t	ext_len, to_copy, remaining;
+
+	for (remaining = outlen; remaining > 0; remaining -= to_copy) {
+		to_copy = remaining;
+		ext_len = _fr_dbuff_extend_lowat(NULL, in, fr_dbuff_end(in) - (*pos_p), 1);
+		if (ext_len == 0) return -remaining;
+		if (ext_len < to_copy) to_copy = ext_len;
+		out += _fr_dbuff_set(pos_p, in,
+				     (*pos_p) + _fr_dbuff_safecpy(out, out + to_copy, (*pos_p), (*pos_p) + to_copy));
+	}
```
arr2036, line 1483: So we're just ignoring outlen now?

arr2036:
Other than the use of "file" instead of "fd" which is used everywhere else in the server, this looks fine.  Can you fixup the names, and remove the eof code and I'll merge this in.

jejones3141:
Ouch; good point. Will do.

jejones3141:
Functions and macros renamed to use fd rather than file, and eof code
removed; the result has been pushed. Thanks.

PR #3767, opened 2020-11-24
Add tests for pairs

jpereira:


src/lib/util/pair_legacy.c
```diff
@@ -283,7 +283,10 @@ static ssize_t fr_pair_list_afrom_substr(TALLOC_CTX *ctx, fr_dict_attr_t const *
 	char const	*p, *next;
 	fr_token_t	last_token = T_INVALID;
 	fr_pair_t_RAW	raw;
-	fr_dict_attr_t const *internal = fr_dict_root(fr_dict_internal());
+	fr_dict_t const *dict_internal = fr_dict_internal();
```
arr2036, line 286: Or initialise the internal dict?

src/lib/util/pair_tests.c
```diff
@@ -0,0 +1,1205 @@
...
+{
+	fr_pair_t   *vp;
+
+	TEST_CASE("Allocation using fr_pair_afrom_da");
+	TEST_CHECK((vp = fr_pair_afrom_da(autofree, attr_test_integer)) != NULL);
+
+	TEST_CASE("Validating VP_VERIFY()");
+	VP_VERIFY(vp);
+
+	vp->vp_uint32 = 12345;
```
arr2036, line 211: The compiler would have to be pretty broken for this test to fail?  Assigning a string here would be a better test.

src/lib/util/pair_tests.c
```diff
@@ -0,0 +1,1205 @@
...
+	TEST_CASE("Validating VP_VERIFY()");
+	VP_VERIFY(vp);
+
+	copy = fr_pair_copy(autofree, vp);
+
+	TEST_CASE("Validating VP_VERIFY()");
+	VP_VERIFY(copy);
+
+	TEST_CHECK(copy && copy->da && copy->da->type == FR_TYPE_UINT32);
+	TEST_MSG("Expected (copy->type == FR_TYPE_UINT32)");
```
arr2036, line 252: There's no check that the values are the same?

arr2036: Check to see if there's a pair_cmp function you could use.

src/lib/util/pair_tests.c
```diff
@@ -0,0 +1,1205 @@
...
+	TALLOC_CTX 	*ctx = talloc_null_ctx();
+
+	TEST_CASE("Allocate a new attribute fr_pair_afrom_da");
+	TEST_CHECK((vp = fr_pair_afrom_da(ctx, attr_test_integer)) != NULL);
+
+	TEST_CASE("Validating VP_VERIFY()");
+	VP_VERIFY(vp);
+
+	TEST_CASE("Stealing 'vp' pair using fr_pair_steal()");
+	fr_pair_steal(autofree, vp); /* It should exit without memory-leaks */
```
arr2036, line 270: Yes, but you need to test the steal was performed using talloc_parent() to check the parent...

src/lib/util/pair_tests.c
```diff
@@ -0,0 +1,1205 @@
...
+	fr_pair_t    *vp;
+	unsigned int attr = FR_MODULE_RETURN_CODE;
+
+	TEST_CASE("Allocation using fr_pair_afrom_child_num");
+	TEST_CHECK((vp = fr_pair_afrom_child_num(autofree, fr_dict_root(dict_test), attr)) != NULL);
+
+	TEST_CASE("Validating VP_VERIFY()");
+	VP_VERIFY(vp);
+
+	TEST_CHECK(vp && vp->da && vp->da->attr == FR_MODULE_RETURN_CODE);
```
arr2036, line 230: Expressions like this in test cases are really annoying because you have no idea which term evaluated to false.  each term should be wrapped in its own `TEST_CHECK` macro call.  Then the test framework tells you which one isn't true.... Though you don't really need to check vp and vp->da because `VP_VERIFY` already checks that.

src/lib/util/pair_tests.c
jpereira: Good catch. Fixed.

PR #3771, opened 2020-11-27
Ensure group attributes are added in the dictionary sequence

ndptech:
For IPv6 DHCP, IPv6 address needs to be added before the expiry time to
match the sequence in the dictionary.

Also make ensure Request packets are handled as updates rather than
allocations.


alandekok:
These changes shouldn't be necessary after commit  d895b5d633fc.  The structs now can take grouped attributes:

`My-Struct = { foo = bar, baz = 1}`

The question now is whether the redis ippool code can update nested structures like that.

arr2036:
I'll add that functionality in today.  I was just thinking of creating groups anywhere there was any kind of filter specified.

PR #3776, opened 2020-12-02
v3.0.x CI tests for build and install of DEBs and RPMs

terryburton:
Build RPM/DEB packages -> Reset the environment -> Install them -> Run `radiusd -XC`

Might eventually want to consider scheduling this rather than running it on push.

### DIFF
--- a/.github/workflows/ci-deb.yml
+++ b/.github/workflows/ci-deb.yml
@@ -1,0 +1,83 @@
+name: CI DEB
+
+on:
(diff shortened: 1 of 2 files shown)


terryburton:
@mcnewton This should prevent you from having to clean up after everyone that adds new files to the tree.

terryburton:
Rebased on top of current v3.0.x so that the CI is clean.

PR #3777, opened 2020-12-02
CI tests for build and install of DEBs and RPMs

terryburton:
Build RPM/DEB packages -> Reset the environment -> Install them -> Run radiusd -XC

NOTE: RPM and DEB packages are broken post-install so don't pull this now if we want to keep the CI clean.

Might eventually want to consider scheduling this rather than running it on push.


arr2036:
Looks like the install tests are currently failing?

PR #3779, opened 2020-12-03
rlm_sql: Fix 'nas_type' client set

jpereira:
Fixes when the 'type' is filled with an invalid value when loading
from DB.


jpereira:
It validate the `nas_type` avoiding be filled with invalid content. e.g: Stored in the DB with '' showing `nas_type = `

```
radmin> show clients list verbose
client {
	ipaddr = 192.168.0.1/32
	proto = *
	secret = testing123
	longname = 127.0.0.1
	shortname = localfromdb
	nas_type =
	number = 2
	virtual_server = tapioca1
	dynamic = yes
	lifetime = 0
	global = yes
}
radmin>
```

jpereira:
It is no longer necessary.

PR #3783, opened 2020-12-04
fixed variable to use the "&" prefix

jpereira:
removal of yellow warning when running with this enabled


arr2036:
wasn't required in v3.0.x, but sure....

PR #3785, opened 2020-12-07
Require cURL >= 7.56.0 for rlm_smtp

terryburton:
Uses rlm_mime_init() el al.

### DIFF
--- a/src/modules/rlm_smtp/all.mk
+++ b/src/modules/rlm_smtp/all.mk
@@ -1,3 +1,19 @@
+#
+#  Functions for comparing versions, taken from:
+#
+#    https://stackoverflow.com/a/15637871/2568535
+#
+S		:=
+SP		:= $S $S
+str.eq		= $(if $(subst $1,,$2),$S,T)
+str.le		= $(call str.eq,$(word 1,$(sort $1 $2)),$1)
+mklist		= $(eval __tmp := $1)$(foreach i,0 1 2 3 4 5 6 7 8 9,$(eval __tmp := $$(subst $$i,$$i ,$(__tmp))))$(__tmp)
+shift		= $(wordlist 2, $(words $1), $1)
+num.le		= $(eval __tmp1 := $(call mklist,$1))$(eval __tmp2 := $(call mklist,$2))$(if $(call str.eq,$(words $(__tmp1)),$(words $(__tmp2))),$(call str.le,$1,$2),$(call str.le,$(words $(__tmp1)),$(words $(__tmp2))))
+list.strip	= $(eval __flag := 1)$(foreach d,$1,$(if $(__flag),$(if $(subst 0,,$d),$(eval __flag :=)$d,$S),$d))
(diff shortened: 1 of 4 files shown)


src/modules/rlm_smtp/all.mk
```diff
@@ -1,12 +1,22 @@
...
+VERSION		:=
 -include $(top_builddir)/src/lib/curl/all.mk
 TARGET		:=
 
 ifneq "$(TARGETNAME)" ""
+
+#  Require cURL >= 7.56.0
+CURL_X		:= $(word 1,$(subst ., ,$(VERSION)))
+CURL_Y		:= $(word 2,$(subst ., ,$(VERSION)))
+CURL_GOOD	:= $(shell test $(CURL_X) -gt 7 -o \( $(CURL_X) -eq 7 -a $(CURL_Y) -ge 56 \) && echo true)
```
arr2036, line 13: The issue with this is it forks for every invocation of make, no matter what the target, it'd be better if you could do it with builtins.

src/modules/rlm_smtp/all.mk
```diff
@@ -1,3 +1,19 @@
...
+SP		:= $S $S
+str.eq		= $(if $(subst $1,,$2),$S,T)
+str.le		= $(call str.eq,$(word 1,$(sort $1 $2)),$1)
+mklist		= $(eval __tmp := $1)$(foreach i,0 1 2 3 4 5 6 7 8 9,$(eval __tmp := $$(subst $$i,$$i ,$(__tmp))))$(__tmp)
+shift		= $(wordlist 2, $(words $1), $1)
+num.le		= $(eval __tmp1 := $(call mklist,$1))$(eval __tmp2 := $(call mklist,$2))$(if $(call str.eq,$(words $(__tmp1)),$(words $(__tmp2))),$(call str.le,$1,$2),$(call str.le,$(words $(__tmp1)),$(words $(__tmp2))))
+list.strip	= $(eval __flag := 1)$(foreach d,$1,$(if $(__flag),$(if $(subst 0,,$d),$(eval __flag :=)$d,$S),$d))
+gen.cmpstr	= $(eval __Tmp1 := $(subst ., ,$1))$(eval __Tmp2 := $(subst ., ,$2))$(foreach i,$(__Tmp1),$(eval j := $(word 1,$(__Tmp2)))$(if $j,$(if $(call str.eq,$i,$j),0,$(if $(call num.le,$i,$j),L,G)),G)$(eval __Tmp2 := $$(call shift,$(__Tmp2))))$(if $(__Tmp2), L)
+ver.lt 		= $(call str.eq,$(word 1,$(call list.strip,$(call gen.cmpstr,$1,$2))),L)
+
```
terryburton, line 16: All of the ways I've looked at doing version comparisons in GNU Make are quite opaque.

The above works. If it's acceptable we could move this into a common include. Where's the best place for this?

Otherwise we may be better off using the autoconf route in the original PR, or doing it the noddy way with a decision tree stepping through a list of unsupported versions.

alandekok: Don't bother using GNU Make macros.  They're horrific.  Instead, run a shell program.  but only once:

something like this:

```
-include $(BUILD_DIR)/make/rlm_smtp.mk

.PHONY: $(BUILD_DIR)/make/rlm_smtp.mk
$(BUILD_DIR)/make/rlm_smtp.mk:
   test $(CURL_X) -gt 7 -o \( $(CURL_X) -eq 7 -a $(CURL_Y) -ge 56 \) && echo "CURL_GOOD=1" > $@
```

Once that's run one, the `rlm_smtp.mk` file contains the `CURL_GOOD` definition, which then is `include`d into the normal build process.

PR #3786, opened 2020-12-08
Enforce sbuff file max

jejones3141:
Don't allow `fr_sbuff_extend_file()` to acquire more than the maximum specified in the context. (That's not quite the same as bytes read given possible buffering; also, in accordance with comments below, it no longer keeps a count of bytes read, instead calculating it.)

### DIFF
--- a/src/lib/util/sbuff.c
+++ b/src/lib/util/sbuff.c
@@ -232,7 +232,7 @@ size_t fr_sbuff_shift(fr_sbuff_t *sbuff, size_t shift)
  */
 size_t fr_sbuff_extend_file(fr_sbuff_t *sbuff, size_t extension)
 {
-	size_t			read, available;
+	size_t			read, available, total_read;
 	fr_sbuff_uctx_file_t	*fctx;
 
 	CHECK_SBUFF_INIT(sbuff);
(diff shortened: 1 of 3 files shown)


src/lib/util/sbuff.h
```diff
@@ -113,6 +113,7 @@ typedef struct {
 	FILE			*file;			//!< FILE * we're reading from.
 	char			*buff_end;		//!< The true end of the buffer.
 	size_t			max;			//!< Maximum number of bytes to read.
+	size_t			read;			//!< Number of bytes read.
```
arr2036, line 116: Isn't this equivalent to `available + shifted`?

jejones3141: I do believe you're correct. I'll make the change, test, and push. Thanks.

jejones3141: Turns out that the equivalent is shifted + (end - buff); will change, retest, and push.

alandekok: maybe something like `shifted + fr_sbuff_used(self)` ?

arr2036: Nah @jejones3141 is correct if we're trying to limit the total amount read from a file, and not limit the total amount we've read from the buffer.

src/lib/util/sbuff.c
```diff
@@ -253,6 +259,7 @@ size_t fr_sbuff_extend_file(fr_sbuff_t *sbuff, size_t extension)
 	}
 
 	available = fctx->buff_end - sbuff->end;
+	if (available > fctx->max - total_read) available = fctx->max - total_read;
```
arr2036, line 262: brackets please!

jejones3141: Brackets added and pushed.

PR #3789, opened 2020-12-09
CI: Package tests including reuse of eapol_test tests in post-install context

terryburton:
RPMs: https://github.com/terryburton/freeradius-server/actions/runs/421435552
DEBs: https://github.com/terryburton/freeradius-server/actions/runs/421435549

### DIFF
--- a/.github/workflows/ci-deb.yml
+++ b/.github/workflows/ci-deb.yml
@@ -38,6 +38,10 @@ jobs:
         echo force-unsafe-io > /etc/dpkg/dpkg.cfg.d/02speedup
         echo 'man-db man-db/auto-update boolean false' | debconf-set-selections
 
+    #
+    #  Required so that the checkout action uses git protocol rather than the GitHub REST API.
+    #  make rpm requires the FR directory to be a git repository.
+    #
     - name: Install recent git
       run: |
         apt-get update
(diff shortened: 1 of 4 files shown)


arr2036:
Is there any reason not to use the existing EAP test config and eapol_test configuration files?

terryburton:
Maybe, I haven't looked. I'm out of time for a day or two and there are some build system issues that I want to address first...

My goal for this is to have something that will flag up any glaring run time issues with OpenSSL, the dynamic linker, etc. which this first iteration should achieve.

It's a shame we can't detect SELinux issues with this approach, however we hopefully can do this in future when the CI system switches to a modern kernel.

alandekok:
as of commit 89ee420077201f6e you should be able to do `make eapol_test`, and it will build _only_ the `eapol_test` binary.

terryburton:
@alandekok That's nice. Thanks.

Rebased and updated workflow for DEB to always use it. Also doing RPM now...

terryburton:
> Rebased and updated workflow for DEB to always use it. Also doing RPM now...

Done. Package build is clean.

terryburton:
Ignore for now: Pushed to wrong branch. Sorry for the noise.

terryburton:
This now reuses the eapol_test CI test but in post-install context without requiring any additional run-time dependencies (other than any non-ancient version of GNU Make).

arr2036:
Hm do we still need those config files embedded in the workflow?

terryburton:
> Hm do we still need those config files embedded in the workflow?

Their benefit is only marginal. I've culled them.

PR #3794, opened 2020-12-10
In fr_sbuff_extend_file(), don't NUL terminate if the read fills the …

jejones3141:
…buffer.

### DIFF
--- a/src/lib/util/sbuff.c
+++ b/src/lib/util/sbuff.c
@@ -252,7 +252,7 @@ size_t fr_sbuff_extend_file(fr_sbuff_t *sbuff, size_t extension)
 
 	read = fread(sbuff->end, 1, available, fctx->file);
 	sbuff->end += read;	/* Advance end, which increases fr_sbuff_remaining() */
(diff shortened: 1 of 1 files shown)


src/lib/util/sbuff.c
```diff
@@ -252,7 +252,7 @@ size_t fr_sbuff_extend_file(fr_sbuff_t *sbuff, size_t extension)
 
 	read = fread(sbuff->end, 1, available, fctx->file);
 	sbuff->end += read;	/* Advance end, which increases fr_sbuff_remaining() */
-	*sbuff->end = '\0';	/* Terminate sbuff */
+	if (sbuff->end < fctx->buff_end) *sbuff->end = '\0';	/* Terminate sbuff */
```
arr2036, line 255: I'm not really sure why we're nul terminating at all?  We're reading from the buffer and all the code is aware of the end of the buffer, so there's not really a good reason for terminating.

jejones3141: Then it probably means I should reconsider my test for what should happen when you hit the maximum bytes read, and the line should just go away. Closing this PR.

PR #3795, opened 2020-12-10
:gear: rlm_python: Fix handling of attribute lists

jpereira:
It fixes the way of how the .py scripts receives the 'dict' within these
major lists. e.g: `p["request"]`, `p["control"]`, `p["reply"]` and `p["session-state"]`

Also, allowing the capability to update them returning a tuple with such a dictionary.

### DIFF
--- a/src/modules/rlm_python/rlm_python.c
+++ b/src/modules/rlm_python/rlm_python.c
@@ -268,98 +268,72 @@ failed:
 	Py_XDECREF(p_traceback);
 }
 
-static void mod_vptuple(TALLOC_CTX *ctx, rlm_python_t const *inst, request_t *request,
-			fr_pair_list_t *vps, PyObject *p_value, char const *funcname, char const *list_name)
+static int tuple_to_list(TALLOC_CTX *ctx, request_t *request, const char *list_name, fr_pair_list_t *list, PyObject *tuple)
 {
-	int		i;
-	Py_ssize_t	tuple_len;
-	tmpl_t	*dst;
-	fr_pair_t	*vp;
-	request_t		*current = request;
+	int         i;
+	Py_ssize_t  tuple_len;
+	tmpl_t      *dst;
+	fr_pair_t   *vp;
+	request_t   *current = request;
 
 	/*
 	 *	If the Python function gave us None for the tuple,
 	 *	then just return.
 	 */
-	if (!p_value || p_value == Py_None) return;
+	if (!tuple || tuple == Py_None) return 0;
 
-	if (!PyTuple_CheckExact(p_value)) {
-		ERROR("%s - non-tuple passed to %s", funcname, list_name);
-		return;
+	if (!PyTuple_CheckExact(tuple)) {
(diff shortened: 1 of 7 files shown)


src/modules/rlm_python/rlm_python.c
```diff
@@ -488,95 +488,147 @@ static int mod_populate_vptuple(rlm_python_t const *inst, request_t *request, Py
...
+	 * If some list is not available, NONE is used instead
+	 */
+	p_arg = PyTuple_New(4);
+	if (!p_arg) {
+		ERROR("%s - Memory cannot be allocated for PyTyple_New", funcname);
+		rcode = RLM_MODULE_FAIL;
+		goto finish;
+	}
+
+	/* If there is a request, fill in the first 4 attribute lists */
```
alandekok, line 568: there should always be a request?  If the function is being called from `recv foo` or xlat, there's always a request.

src/modules/rlm_python/rlm_python.c
```diff
@@ -488,95 +488,147 @@ static int mod_populate_vptuple(rlm_python_t const *inst, request_t *request, Py
...
+			rcode = RLM_MODULE_FAIL;
+			goto finish;
 		}
 	}
 
 	/* Call Python function. */
-	p_ret = PyObject_CallFunctionObjArgs(p_func, p_arg, NULL);
+	p_dict = PyDict_New();
+	if (!p_dict ||
+	    PyDict_SetItemString(p_dict, "request", PyTuple_GET_ITEM(p_arg, 0)) ||
```
alandekok, line 590: Just move this name setting to `mod_populate_vps()`, and pass the name to that function.  less code == better code

src/modules/rlm_python/rlm_python.c
```diff
@@ -587,19 +639,33 @@ static unlang_action_t do_python_single(rlm_rcode_t *p_result,
...
+
+		/* process updateDict */
+		update_dict = PyTuple_GET_ITEM(p_ret, 1);
+		if (!PyDict_CheckExact(update_dict)) {
+			ERROR("%s - updateDict is not dictionary", funcname);
+			rcode = RLM_MODULE_FAIL;
+			goto finish;
+		}
+
+		mod_vptuple(request->packet, inst, request, &request->request_pairs,
```
alandekok, line 654: TBH macros here could shorten this code.  Repetition is opportunity for mistakes

src/tests/modules/python/auth_with_shared_storage.unlang
```diff
@@ -1,13 +1,13 @@
 # Should return noop (no shared variable set)
-pmod_shared_storage
+pmod7_shared_storage
```
alandekok, line 2: Why this change?  I don't understand...

jpereira: Just to keep the same as the previous tests with numbers. e.g: `pmod${number}_$name`

PR #3801, opened 2020-12-14
rlm_wimax: added support for EPS-AKA authentication (LTE/3gpp)

bt4:
Hi, please review my code which implements the LTE style authentication for use with the latest WiMAX specifications. Some of the documents which I referred to:

WMF-T32-001-R022v05_Network-Stage2-Base.pdf page 294
33401-g30.doc page 22 - 24, and 78 (kasme)
33102-g00.doc page 22 and 63 (sqn)
WMF-T33-001-R022v05_Network-Stage3-Base.pdf page 1310 (attributes)

These mods have been tested with a TelRad EPC.

### DIFF
--- a/raddb/mods-available/wimax
+++ b/raddb/mods-available/wimax
@@ -1,6 +1,8 @@
 #
 #	The WiMAX module currently takes no configuration.
 #
+#### INSTRUCTIONS FOR V1 and V2.0 WIMAX #########################
+#
 #	It should be listed in the "authorize" and "preacct" sections.
 #	This enables the module to fix the horrible binary version
 #	of Calling-Station-Id to the normal format, as specified in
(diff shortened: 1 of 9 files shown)


alandekok:
This looks pretty good.  I have minor comments on formatting, etc.  but that's fine.

alandekok:
I've done some rearrangement.  Please double-check that I haven't broken anything. :)

bt4:
Hi Alan. Thanks for merging and cleaning up my code. I did some testing and found one problem with the new code. On line:

https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/modules/rlm_wimax/rlm_wimax.c#L605

Please see my proposed fix here: #3802

However, I am not sure what exactly is the correct way to deal with RAND after a SQN resync. The thing is that normally we have to create RAND in generate the authentication vector. However, in the case of a resync there is already a RAND in the request and I couldn't get a clear idea from the docs which this same RAND should be used or a new one generated in the reply. I have tested both variation and both work with the TelRad EPC which I have access to.

With this fix we by default reuse the same RAND but we also have the option to delete it via unlang before calling the module in post_auth, in which case the module will generate a new RAND.

The idea was to save a copy of RAND in control:WiMAX-SIM-RAND

PR #3810, opened 2020-12-18
Rest raw value

mschneid:
Add config item do disable enumerated attribute values and add new JSON property with the numeric attribute type.


raddb/mods-available/rest
```diff
@@ -19,22 +19,22 @@ rest {
 #		ca_path = ${certdir}
 
 #		certificate_file        = /path/to/radius.crt
-#		private_key_file	= /path/to/radius.key
-#		private_key_password	= "supersecret"
-#		random_file		= /dev/urandom
+#		private_key_file        = /path/to/radius.key
```
alandekok, line 22: We would really prefer patches which don't randomly change whitespace.

arr2036: I think it's ok, there some messed up tab/space stuff that this fixes.

raddb/mods-available/rest
```diff
@@ -98,16 +98,24 @@ rest {
 	#  [outer.][<list>:]<attribute0>=<value0>&[outer.][<list>:]<attributeN>=<valueN>
 	#
 	#  JSON - All attributes and values are escaped according to the JSON specification
+	#  - attribute  Name of the attribute.
+	#  - attr_type  Numerical representation of the attribute type.
```
arr2036, line 104: can you rename to `attr_num` and place inside a `json { }` subsection within the method subsection.

`attr_type` is RADIUS specific terminology, and FreeRADIUS isn't a RADIUS server.  It just happens to have a RADIUS frontend :)

nward:
@mschneid I'm wondering if you have any thoughts on #4130 - can we pass `section` around in request ctx rather than as a parameter on `rest_encode_*` ?

mschneid:
> @mschneid I'm wondering if you have any thoughts on #4130 - can we pass `section` around in request ctx rather than as a parameter on `rest_encode_*` ?

For me, that's fine.

src/lib/print.c
```diff
@@ -495,33 +495,26 @@ char *vp_aprints_type(TALLOC_CTX *ctx, PW_TYPE type)
...
  * @param vp to print.
+ * @param raw_value if true, the raw value is printed and not the enumerated attribute value
  * @return the length of data written to out, or a value >= outlen on truncation.
  */
-size_t vp_prints_value_json(char *out, size_t outlen, VALUE_PAIR const *vp)
+size_t vp_prints_value_json(char *out, size_t outlen, VALUE_PAIR const *vp, bool raw_value)
 {
 	char const	*q;
 	size_t		len, freespace = outlen;
+	bool		raw = raw_value || (!vp->da->flags.has_tag && !vp->da->flags.has_value);
```
nward, line 505: I don't entirely understand this line - why are we checking for has_value here? We previously were checking for has_tag only.

I think this is causing my environment to always have `raw` as true.

I assumed FreeRADIUS only ever sent the raw value until coming across this PR in my travels on the chunked stuff - so am trying to figure out why I don't ever see enumerated values.

arr2036: So that the integer value is printed where the attribute doesn't have a set of enumeration values I guess...

nward: Yeah - that makes sense. I'm confused because in my environment I don't see enumeration values ever being printed, only the integer. I.e. `Acct-Status-Type=Start` going in comes out as `{"Acct-Status-Type":{"type":"integer","value":[1]}}`

I've tested on 3.0.19 through 3.0.23 with very basic config and see the same behaviour though, so, I guess it's not related to this change, and I have some more digging to do.

Thanks!

PR #3811, opened 2020-12-18
Add braces to SBUFF_OUT_TALLOC_FUNC_NO_LEN_DEF() definition

jejones3141:


arr2036:
This causes issues with doxygen.  All uses of this macro supply their own braces.

PR #3814, opened 2020-12-20
Update Aruba Dictionnary. ref #3812

jpereira:


share/dictionary.aruba
```diff
@@ -67,6 +67,10 @@ ATTRIBUTE	Aruba-QoS-Trust-Mode			52	integer
 ATTRIBUTE	Aruba-UBT-Gateway-Role			53	string
 ATTRIBUTE	Aruba-Gateway-Zone			54	string
 
+ATTRIBUTE	STP-Admin-Edge-Port			58	integer
```
arr2036, line 70: Needs the Aruba- prefix if you want it in v3.0.x

jpereira: Done

PR #3822, opened 2020-12-29
rlm_rest - Add option to set http version

jpereira:
this adds enough moving parts to allow direct Okta integration.

### DIFF
--- a/src/modules/rlm_rest/rest.c
+++ b/src/modules/rlm_rest/rest.c
@@ -1569,6 +1569,7 @@ static size_t rest_response_header(void *in, size_t size, size_t nmemb, void *us
 {
 	rlm_rest_response_t *ctx = userdata;
 	REQUEST *request = ctx->request; /* Used by RDEBUG */
+	VALUE_PAIR *vp;
 
 	char const *p = in, *q;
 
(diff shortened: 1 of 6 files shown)


src/modules/rlm_rest/rest.c
```diff
@@ -1647,6 +1648,14 @@ static size_t rest_response_header(void *in, size_t size, size_t nmemb, void *us
 
 		ctx->code = atoi(p);
 
+		/*
+		 *  Save the HTTP return status code.
+		 */
+		vp = pair_make_reply("REST-HTTP-Status-Code", NULL, T_OP_ADD);
```
arr2036, line 1654: Shouldn't this be T_OP_SET?

src/modules/rlm_rest/rest.c
```diff
@@ -221,6 +221,34 @@ const FR_NAME_NUMBER http_content_type_table[] = {
...
+ *
+ * Values we expect to use in curl_easy_setopt()
+ *
+ * @see fr_str2int
+ * @see fr_int2str
+ */
+const FR_NAME_NUMBER http_version_table[] = {
+	{ "1.0",	CURL_HTTP_VERSION_1_0		},
+	{ "1.1",	CURL_HTTP_VERSION_1_1		},
+#ifdef CURL_HTTP_VERSION_2_0	/* Added in 7.33.0 */
```
arr2036, line 236: All these are wrong.  @jpereira you didn't think it was strange we didn't use `#ifdef` in the master branch, and instead used version check macros? It's because these are enum values https://github.com/curl/curl/blob/a0962a6417f17f27fb197ae237cc65811627d7c5/include/curl/curl.h#L2123 not macros (as are all CURLOPT values).  You just broke the build with any curl version earlier than 7.33. You also invented your own values instead of using the ones from master branch, so broke forwards compatibility, and used a different value scheme for HTTP 3... "1.0", "2.0", "3", seriously?  The lack of attention to detail is becoming frustrating.  The comments I make on other people's PRs are usually technical in nature, with a few formatting issues, but for virtually every PR you raise I finding issues like this.  You seriously need to up the quality of your contributions if you want them to continue being merged.

PR #3829, opened 2021-01-04
logrotate config file: don't change global logrotate options

mgerstner:
Global options in files in /etc/logrotate.d/ potentially influence all
other configurations parsed after the "radiusd" configuration file.
Especially the `su radiusd radiusd` line can break things for other
packages, but also the other settings may lead to surprising behaviour,
depending on the context.

Therefore move global options into the logfile specific blocks, grouping
together equal settings to avoid a high degree of redundance.

### DIFF
--- a/scripts/logrotate/freeradius
+++ b/scripts/logrotate/freeradius
@@ -5,44 +5,31 @@
 #
 
 #
-#    Global options for all files
-#
-daily
-rotate 14
-missingok
-compress
-delaycompress
-notifempty
-su radiusd radiusd
-
-#
 #  The main server log
 #
 /var/log/radius/radius.log {
 	copytruncate
+	daily
+	rotate 14
+	missingok
+	compress
+	delaycompress
+	notifempty
+	su radiusd radiusd
 }
 
 #
 #  Session monitoring utilities
 #
-/var/log/radius/checkrad.log /var/log/radius/radwatch.log {
-	nocreate
-}
-
-#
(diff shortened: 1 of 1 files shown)


scripts/logrotate/freeradius
```diff
@@ -4,45 +4,32 @@
...
+/var/log/radius/checkrad.log /var/log/radius/radwatch.log
 #
-#  Session database modules
+#  SQL log files
 #
-/var/log/radius/radutmp /var/log/radius/radwtmp {
-	nocreate
-}
-
+/var/log/radius/sqllog.sql
```
alandekok, line 28: Why were these options deleted?  There's no need for that as part fo a commit which cleans up the global config file.

mgerstner: The new config file should be equivalent to the old one. I just wanted to avoid redundant configuration as far as possible. For example:

```
/var/log/radius/sqllog.sql                                                        
#
# some comments                                                                                 
#
/var/log/radius/radacct/*/detail {                                                
        create                                                                    
        daily                                                                     
        rotate 14                                                                 
        missingok                                                                 
        compress                                                                  
        delaycompress                                                             
        notifempty                                                                
        su radiusd radiusd                                                        
}                              
```

This means that the options block `{ create ... su radiusd radiusd }` applies to both, `/var/log/radius/sqllog.sql` and `/var/log/radius/radacct/*/detail`.

PR #3830, opened 2021-01-04
Simplify internal_encode().

jejones3141:
With a child dbuff to encode data into,
1. We needn't repeat ourselves with FR_DBUFF_RESERVE().
2. The length of the data when it's encoded is just the
   used portion of the child dbuff.

We also switched to fr_dbuff_in_memcpy() to get the data length
into the dbuff; it gets rid of the clutter of FR_DBUFF_TMP().

### DIFF
--- a/src/protocols/internal/encode.c
+++ b/src/protocols/internal/encode.c
@@ -56,11 +56,13 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
 			       fr_cursor_t *cursor, void *encoder_ctx)
 {
 	fr_dbuff_t		work_dbuff = FR_DBUFF_NO_ADVANCE(dbuff);
-	fr_dbuff_marker_t	enc_field, value_field, len_field, value_dest;
+	fr_dbuff_marker_t	enc_field, len_field, value_field;
+	fr_dbuff_t		value_dbuff;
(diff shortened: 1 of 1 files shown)


src/protocols/internal/encode.c
```diff
@@ -103,18 +105,23 @@ static ssize_t internal_encode(fr_dbuff_t *dbuff,
...
+	value_dbuff = FR_DBUFF_RESERVE(&work_dbuff, sizeof(uint64_t) - 1);
+	fr_dbuff_marker(&value_field, &value_dbuff);
 
 	switch (da->type) {
 	case FR_TYPE_VALUE:
-		slen = fr_value_box_to_network(&FR_DBUFF_RESERVE(&work_dbuff, sizeof(uint64_t) - 1), &vp->data);
+		slen = fr_value_box_to_network(&value_dbuff, &vp->data);
 		if (slen < 0) return PAIR_ENCODE_FATAL_ERROR;
-		FR_PROTO_HEX_DUMP(fr_dbuff_current(&value_field), slen, "value %s",
+		FR_PROTO_HEX_DUMP(fr_dbuff_start(&value_dbuff), slen, "value %s",
```
arr2036, line 124: Mmm, should that be start? Probably actually need a marker before calling `fr_value_box_to_network` i.e. what was there before :)

PR #3832, opened 2021-01-05
rlm_rest - Add option to set http version

jpereira:
This adds enough moving parts to allow direct Okta integration.

1. Option to set the HTTP negotiation version.
2. Get the response status code.

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -169,6 +169,9 @@ rest {
 	#  2xx    successful    yes           ok/updated
 	#  5xx    server error  no            fail
 	#  xxx    -             no            invalid
+	#
+	#  The status code is held in %{reply:REST-HTTP-Status-Code}.
+	#
 	authorize {
 		uri = "${..connect_uri}/user/%{User-Name}/mac/%{Called-Station-ID}?action=authorize"
 		method = 'get'
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -54,6 +54,12 @@ rest {
 #	connect_timeout = 4.0
 
 	#
+	# Specify HTTP protocol version to use. one of '1.0', '1.1', '2.0', '2.0+auto',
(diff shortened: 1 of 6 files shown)


src/modules/rlm_rest/rest.c
```diff
@@ -1619,6 +1620,14 @@ static size_t rest_response_header(void *in, size_t size, size_t nmemb, void *us
 
 		ctx->code = atoi(p);
 
+		/*
+		 *  Save the HTTP return status code.
+		 */
+		vp = pair_make_reply("REST-HTTP-Status-Code", NULL, T_OP_SET);
+		vp->vp_integer = ctx->code;
+
+		RDEBUG2("Adding reply:REST-HTTP-Status-Code += \"%d\"", vp->vp_integer);
```
arr2036, line 1672: Better to handle this in `rest_request_perform`.  You can get the ctx from the curl handle.

raddb/mods-available/rest
```diff
@@ -53,6 +53,12 @@ rest {
 	#
 #	connect_timeout = 4.0
 
+	#
+	# Specify HTTP protocol version to use. one of '1.0', '1.1', '2.0', '2.0+auto',
+	# '2.0+tls' or 'default'. (libcurl option CURLOPT_HTTP_VERSION)
+	#
+#	http_negotiation = 1.1
```
arr2036, line 60: This isn't enough to support HTTP2, there's also some header parser changes required.

https://github.com/FreeRADIUS/freeradius-server/commit/6286520698a3cc4053b4d49eb0a61d9ba77632aa#diff-0cca2197649953eab611e29cf27fdcc98624c3b46bed73725b4294674c3fcb44

Some of the other changes in the commit series may also be required, you need to _test_ the code against a HTTP2.0 server to be sure.

arr2036:
Did you test against a HTTP2 implementation?

raddb/mods-available/rest
jpereira: yep, the OKTA service supports HTTP 1.0, 1.1 and 2.0 as well.

PR #3833, opened 2021-01-05
add HW-Ext-Specific RADIUS attribute for bounce port using CoA

nqb:
Hello,

I would like to use `HW-Ext-Specific` RADIUS attribute as described in [Huawei's documentation](https://support.huawei.com/hedex/hdx.do?lib=EDOC1100101074AEI08211&docid=EDOC1100101074&lang=en&v=05&tocLib=EDOC1100101074AEI08211&tocV=05&id=EN-US_CLIREF_0176366782&tocURL=resources%2525252Fdc%2525252Fradius-server_authorization_hw-ext-specific_command_bounce-port_disable.html&p=t&fe=1&ui=3&keyword=radius-server%25252Bauthorization%25252Bhw-ext-specific%25252Bcommand%25252Bbounce-port%25252Bdisable) in order to bounce a port using a CoA message.

According to documentation, RADIUS attribute identifier is `HW-Ext-Specific` and not `Huawei-Ext-Specific`. I've noticed a changed between FreeRADIUS master and v3.0 branch where vendor name is not added anymore before each RADIUS attribute in dictionnary. I'm not sure my PR is correct and how to handle a vendor which has two aliases: 
- Huawei
- HW

Thanks in advance

PR #3835, opened 2021-01-06
add HW-Ext-Specific RADIUS attribute for bounce port using CoA

nqb:
Next part of #3833 for v3.0.x branch.


arr2036:
This should have the `Hauwei-` prefix like the rest of the attributes in that dictionary.

nqb:
Thanks for you feedback, I did change requested.

PR #3841, opened 2021-01-12
rlm_python: The detach() function in example.py expects parameter

jpereira:


jpereira:
It fixes:

```
Tue Jan 12 11:10:11 2021: rlm_python (python) - <class 'TypeError'> (detach() takes 0 positional arguments but 1 was given)
```

PR #3848, opened 2021-01-17
CI: Expand FreeBSD build and tests

terryburton:
Update `exec` tests to use /bin/sh rather than /bin/bash, since this path is more stable between OSs.

Expand build for FreeBSD to enable all modules whose dependencies can be satisfied from standard Ports packages.

Perform `make test` for FreeBSD build, for the time being disabling tests that currently fail.

CI results based on last successful build:
https://github.com/terryburton/freeradius-server/actions/runs/490932807
https://github.com/terryburton/freeradius-server/actions/runs/490932806

### DIFF
--- a/.github/workflows/ci-freebsd.yml
+++ b/.github/workflows/ci-freebsd.yml
@@ -16,18 +16,68 @@ jobs:
 
     steps:
 
+    # Checkout, but defer pulling LFS objects until we've restored the cache
     - uses: actions/checkout@v2
       with:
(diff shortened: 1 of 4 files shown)


alandekok:
It's hard to see how the build failures are a result of these changes.  I think we should just merge it in, and separately figure out the build failures.

terryburton:
Yes, I re-based on top of the last successful build prior to this PR and it was clean, as per the links above.

PR #3853, opened 2021-01-18
Modernize and clean up dhcpv4 encoding code to make better use of dbu…

jejones3141:
…ffs and markers

### DIFF
--- a/src/protocols/dhcpv4/encode.c
+++ b/src/protocols/dhcpv4/encode.c
@@ -87,28 +87,52 @@ static ssize_t encode_value(fr_dbuff_t *dbuff,
 /** Extend an encoded option in-place.
  *
  * @param[in] dbuff	buffer containing the option
- * @param[in] hdr	where the option starts
+ * @param[in] hdr	marker (with dbuff as parent) set to where the option starts
  * @param[in] len	length of the data being written
  * @return
- *	- NULL if we can't extend the option
- *	- !NULL which is the start of the next option
+ *	- false if we can't extend the option
+ *	- true  if we can, with hdr set to where the next option should start
+ * @note	The option starts with a two-byte (type, length) header, where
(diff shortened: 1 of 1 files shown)


src/protocols/dhcpv4/encode.c
```diff
@@ -235,20 +265,25 @@ static ssize_t encode_rfc_hdr(fr_dbuff_t *dbuff,
...
 		FR_PROTO_HEX_DUMP(dbuff->p, fr_dbuff_used(&work_dbuff), NULL);
 
-		if ((hdr[1] + len) <= 255) {
-			hdr[1] += len;
-			FR_PROTO_TRACE("%u byte(s) available in option", 255 - hdr[1]);
+		if ((option_len + len) <= 255) {
+			option_len += len;
+			fr_dbuff_advance(&hdr, 1);
+			fr_dbuff_in(&hdr, option_len);
+			fr_dbuff_advance(&hdr, -2);
```
arr2036, line 272: Hm I know it works but it's weird behaviour.  Maybe add a `fr_dbuff_retreat` as that's the atonym...

arr2036: Usually you'd use markers for this kind of thing though because you can be sure the data hasn't been shifted out.  Consider an extension function which writes out data to a socket and shifts it to the left when the buffer is full.

arr2036: Remember you can use fr_sbuff_set on markers now....

jejones3141: Good point. I could add another marker so hdr can stay as is. It would only be used for the length, so it needn't be retreated..

jejones3141: Pushed change to leave hdr alone-no more retreating. I did rename markers, but it seemed worth it for consistency and indicating the markers' purpose.

PR #3857, opened 2021-01-19
Update dictionary.juniper

lyubomirtraykov:
Add Juniper-CWA-Redirect to dictionary.juniper

### DIFF
--- a/share/dictionary/radius/dictionary.juniper
+++ b/share/dictionary/radius/dictionary.juniper
@@ -48,6 +48,7 @@ ATTRIBUTE	Local-Group-Name			46	string
 ATTRIBUTE	Local-Interface				47	string
 ATTRIBUTE	Switching-Filter			48	string
 ATTRIBUTE	VoIP-Vlan				49	string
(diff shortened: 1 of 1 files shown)


share/dictionary/radius/dictionary.juniper
```diff
@@ -48,6 +48,7 @@ ATTRIBUTE	Local-Group-Name			46	string
 ATTRIBUTE	Local-Interface				47	string
 ATTRIBUTE	Switching-Filter			48	string
 ATTRIBUTE	VoIP-Vlan				49	string
+ATTRIBUTE	Juniper-CWA-Redirect			50	string
```
arr2036, line 51: Please remove the "Juniper-" prefix.  Vendor-specific attributes exist in their own name/number space in master. For example if you were referencing `VoIP-VLAN` in RADIUS, you'd use `Vendor-Specific.Juniper-VoIP-VLAN`

lyubomirtraykov: Sorry. Done

PR #3859, opened 2021-01-19
Integrate fuzzer run into CI (push); cleanup matrix conditionals

terryburton:
Fuzzer's reproducer outputs are uploaded on failure.

Might want to tweak the fuzzer runtime for each protocol to give more weighting to active areas of code change. Aim is to not unnecessarily delay the overall workflow result.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,8 @@ env:
 #  REDIS_IPPOOL_TEST_SERVER: 127.0.0.1
   ANALYZE_C_DUMP: 1
   FR_GLOBAL_POOL: 4M
+  TEST_CERTS: yes
+  DO_BUILD: yes
   CI: 1
   GH_ACTIONS: 1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -383,9 +384,32 @@ jobs:
...
+    # Fuzz each protocol for a short time, continuing on failures
+    # Aim to keep to overall runtime of this job in line with other jobs in the CI workflow
+    - name: Run fuzzer tests
+      if: ${{ matrix.env.TEST_TYPE == 'fuzzing' }}
+      run: |
+        # For fuzzing we won't be needing eapol_test
+        mkdir -p build/tests/eapol_test
+        : > build/tests/eapol_test/eapol_test.mk
+        CPUS=`nproc`
+        mkdir -p fuzzer_artifacts/radius && make -j $CPUS test.fuzzer.radius FUZZER_RUNTIME=240 FUZZER_ARTIFACTS=fuzzer_artifacts/radius/ FUZZER_JOBS=$CPUS FUZZER_WORKERS=$CPUS || : ; mv -f fuzz-*.log fuzzer_artifacts/radius/
```
arr2036, line 403: I think this belongs more in the build system.  This should really be calling a single target.

.github/workflows/ci.yml
```diff
@@ -310,18 +314,17 @@ jobs:
...
+    - name: Configure
       run: |
         if $CC -v 2>&1 | grep clang > /dev/null; then
             echo "Enabling llvm sanitizers"
             enable_llvm_sanitizers="--enable-llvm-address-sanitizer --enable-llvm-undefined-behaviour-sanitizer"
             if [ "`uname`" != "Darwin" ]; then
                 enable_llvm_sanitizers="$enable_llvm_sanitizers --enable-llvm-leak-sanitizer"
             fi
+            if [ "$TEST_TYPE" = "fuzzing" ]; then
+                enable_llvm_sanitizers="$enable_llvm_sanitizers --enable-llvm-fuzzer-sanitizer"
```
arr2036, line 326: now `--enable-llvm-fuzzer`

.github/workflows/ci.yml
terryburton: @arr2036 Do you mean something like `make ci-fuzzer` to execute the fuzzing for each protocol?

alandekok: it already supports `make test.fuzzer`  which runs the fuzzer tests for all protocols, for 10 seconds.  You can just do `make -j 4 test.fuzzer`, and it should run all 4 jobs in parallel.

The build system should put the outputs into `artifacts/$(PROTOCOL)/`

I'm also not sure that setting `FUZZER_RUNTIME=240` here will do anything.  The `src/bin/all.mk` file has `FUZZER_RUNTIME := 0`, which _always_ sets it to zero, no matter what the environment variables say.

alandekok:
I've pushed some changes in commit 2594728e73ba8 which unfortunately conflict, but also address some comments.

I think your commit c6a85f9 is fine, tho.

In general, build / output artifacts should go into `$(BUILD_DIR)/something`, which makes it a bit easier to control.

terryburton:
Reworked to use 2594728e73ba8fe0785ea9886b62b3d64b1b663b.

For the moment we run the fuzzing of the protocols in parallel. We might want to be more nuanced in the allocate of CPU time to the protocols in future.

PR #3865, opened 2021-01-21
Move lists of fr_pair_t to doubly linked lists

ndptech:


src/bin/radclient.c
```diff
@@ -462,10 +461,12 @@ static int radclient_init(TALLOC_CTX *ctx, rc_file_pair_t *files)
 				}
 
 				if (vp->da == attr_packet_type) {
-					vp = fr_cursor_remove(&cursor);	/* so we don't break the filter */
+					fr_pair_t *next;
+					next = fr_pair_list_next(&request->filter, vp);	/* so we don't break the filter */
 					request->filter_code = vp->vp_uint32;
+					fr_pair_remove(&request->filter, vp);
```
arr2036, line 467: Removing and freeing are going to be very common patterns.  Could we get an `fr_pair_free` function which takes a list head and the pair to free.  We could actually do the unlinking in the pair destructor and just use `talloc_free` but that has the feeling of a decision we'll regret later.

src/lib/eap_aka_sim/state_machine.c
```diff
@@ -169,15 +168,17 @@ static bool identity_req_set_by_user(request_t *request, eap_aka_sim_session_t *
...
-	     vp = fr_cursor_next(&cursor)) {
+	     vp = fr_pair_list_next(&request->reply_pairs, vp)) {
 		if (vp->da == attr_eap_aka_sim_permanent_id_req) {
 			eap_aka_sim_session->id_req = AKA_SIM_PERMANENT_ID_REQ;
 		found:
 			set_by_user = true;
 			RDEBUG2("Previous section added &reply.%pP, will request additional identity", vp);
-			fr_cursor_free_item(&cursor);
+			prev = fr_pair_list_prev(&request->reply_pairs, vp);
+			fr_pair_delete(&request->reply_pairs, vp);
```
arr2036, line 180: `fr_pair_delete` vs `fr_pair_remove`? Does one free the pair? Maybe have fr_pair_delete return the previous pair to make iteration easier

src/lib/ldap/map.c
```diff
@@ -40,12 +40,11 @@ int fr_ldap_map_getvalue(TALLOC_CTX *ctx, fr_pair_list_t *out, request_t *reques
 {
 	fr_ldap_result_t	*self = uctx;
 	fr_pair_list_t		head;
+	fr_pair_list_t		tmp_list;
```
arr2036, line 43: not initialised?

src/lib/eap/base.c
```diff
@@ -115,15 +115,14 @@ fr_dict_attr_autoload_t eap_base_dict_attr[] = {
...
 {
 	int		total, size;
 	uint8_t const *ptr;
-	fr_pair_list_t	head;
+	fr_pair_list_t	*head;
 	fr_pair_t	*vp;
-	fr_cursor_t	out;
 
-	fr_pair_list_init(&head);
+	head = fr_pair_list_alloc(packet);
```
arr2036, line 125: Yeah this seems like a bad pattern.  This function should really take a list and a ctx to do the allocations in.

src/bin/radclient.c
ndptech: This should have been fr_pair_delete() which does both.

src/lib/eap_aka_sim/state_machine.c
ndptech: fr_pair_remove() just removes the pair from the list (mainly used where pairs are being moved from one list to another), fr_pair_delete() also frees.  I'll add the return of the previous pair.

src/lib/ldap/map.c
ndptech: Oops

src/lib/eap/base.h
```diff
@@ -68,7 +68,7 @@ typedef struct {
 /*
  *	interfaces in eapcommon.c
  */
-fr_pair_t		*eap_packet_to_vp(fr_radius_packet_t *packet, eap_packet_raw_t const *reply);
+fr_pair_list_t		*eap_packet_to_vp(fr_radius_packet_t *packet, eap_packet_raw_t const *reply);
```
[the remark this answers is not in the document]
ndptech: Amended - though actually it would appear not to be used anywhere...

src/lib/server/state.c
```diff
@@ -687,9 +687,9 @@ int fr_request_to_state(fr_state_tree_t *state, request_t *request)
 	request_data_list_init(&data);
 	request_data_by_persistance(&data, request, true);
 
-	if (!request->session_state_pairs && fr_dlist_empty(&data)) return 0;
+	if (fr_pair_list_empty(&request->session_state_pairs) && fr_dlist_empty(&data)) return 0;
 
-	if (request->session_state_pairs) {
+	if (!fr_pair_list_empty(&request->session_state_pairs)) {
```
arr2036, line 692: This doesn't seem right.  `request->session_state_pairs` is a pointer to a list head, shouldn't need `&`?

src/lib/server/trigger.c
```diff
@@ -406,7 +406,7 @@ int trigger_exec(request_t *request, CONF_SECTION const *cs, char const *name, b
 	ctx->name = talloc_strdup(ctx, value);
 
 	if (request) {
-		if (request->request_pairs) {
+		if (!fr_pair_list_empty(&request->request_pairs)) {
```
arr2036, line 409: Hm same here?

src/lib/eap/base.c
```diff
@@ -115,39 +115,33 @@ fr_dict_attr_autoload_t eap_base_dict_attr[] = {
 	{ NULL }
 };
 
-fr_pair_t *eap_packet_to_vp(fr_radius_packet_t *packet, eap_packet_raw_t const *eap)
+void eap_packet_to_vp(fr_radius_packet_t *packet, eap_packet_raw_t const *eap, fr_pair_list_t *list)
```
arr2036, line 118: https://wiki.freeradius.org/contributing/coding-standards#argument-order

Should be `int eap_packet_to_vp(TALLOC_CTX *ctx, fr_pair_list_t *list,  eap_packet_raw_t const *eap)`

Ctx should be `request->request_ctx` this must have been one of the ones I missed.

src/lib/server/trigger.c
```diff
@@ -461,15 +461,15 @@ int trigger_exec(request_t *request, CONF_SECTION const *cs, char const *name, b
  *	- NULL on failure, or if triggers are not enabled.
  *	- list containing Pool-Server and Pool-Port
  */
-fr_pair_t *trigger_args_afrom_server(TALLOC_CTX *ctx, char const *server, uint16_t port)
+fr_pair_list_t *trigger_args_afrom_server(TALLOC_CTX *ctx, char const *server, uint16_t port)
```
arr2036, line 464: Another one where we should really be passing in the list and adding the pairs to it

src/lib/util/dpair.c
```diff
@@ -0,0 +1,2672 @@
...
+#ifdef WITH_VERIFY_PTR
+	fr_pair_t *i;
+#endif
+
+	if (!add) return;
+
+	VP_VERIFY(add);
+
+#ifdef WITH_VERIFY_PTR
+	for (i = fr_pair_list_head(list); i; i = fr_pair_list_next(list, i)) {
```
arr2036, line 668: I believe there's a verify list now

src/protocols/dhcpv6/base.c
```diff
@@ -752,23 +752,21 @@ ssize_t	fr_dhcpv6_decode(TALLOC_CTX *ctx, uint8_t const *packet, size_t packet_l
...
+void *fr_dhcpv6_next_encodable(fr_dlist_head_t *list, void *to_eval, void *uctx)
 {
-	fr_pair_t	*c, *p;
+	fr_pair_t	*c;
 	fr_dict_t	*dict = talloc_get_type_abort(uctx, fr_dict_t);
 
 	if (!to_eval) return NULL;
 
-	for (p = *prev, c = to_eval; c; p = c, c = c->next) {
+	for (c = to_eval; c; c = fr_dlist_next(list, c)) {
```
arr2036, line 762: Hm this should really be operating on a `fr_pair_list_t`

src/protocols/radius/encode.c
```diff
@@ -1481,7 +1481,7 @@ static ssize_t fr_radius_encode_proto(UNUSED TALLOC_CTX *ctx, fr_pair_list_t *vp
 /*
  *	No one else should be using this.
  */
-extern void *fr_radius_next_encodable(void **prev, void *to_eval, void *uctx);
+extern void *fr_radius_next_encodable(fr_dlist_head_t *list, void *to_eval, void *uctx);
```
arr2036, line 1484: Here too, seems odd..

src/lib/server/state.c
ndptech: `session_state_pairs` is defined as `pair_list.state->children`, making `request->session_state_pairs` `request->pair_list.state->children` which is an `fr_pair_list_t` (not a pointer)

arr2036: Yes I realised this later.  Sorry I thought i'd removed these.

src/protocols/dhcpv6/base.c
ndptech: This is a `dcursor` iterator function, and since dcursors are able to operate on any type of dlist, we have to pass in the `fr_dlist_head` rather than a `fr_pair_list_t`

src/protocols/radius/encode.c
ndptech: This is a `dcursor` iterator function, and since dcursors are able to operate on any type of dlist, we have to pass in the `fr_dlist_head` rather than a `fr_pair_list_t`

PR #3866, opened 2021-01-21
Add support for timezone on alpine

x-jokay:
It would be great to have support for timezone on the alpine using environment variable `TZ`, e.g. `TZ=Europe/Zurich`.

```yml
version: '3.8'

services:
  freeradius:
    image: docker.io/freeradius/freeradius-server:3.0.21-alpine
    volumes:
      - ./raddb:/etc/raddb
    ports:
      - 1812:1812
      - 1813:1813
    environment:
      - TZ=Europe/Zurich
    networks:
      - default

networks:
  default:
```

Which will result in

```sh
[user@docker freeradius]$ docker exec freeradius date
Thu Jan 21 14:13:56 CET 2021
```

instead of

```sh
[user@docker freeradius]$ docker exec freeradius date
Thu Jan 21 12:13:56 UTC 2021
```

### DIFF
--- a/scripts/docker/alpine/Dockerfile
+++ b/scripts/docker/alpine/Dockerfile
@@ -57,7 +57,7 @@ COPY --from=build /opt /opt
 # These are needed for the server to start
 #
 RUN apk update \
-    && apk add talloc libressl pcre libwbclient \
+    && apk add talloc libressl pcre libwbclient tzdata \
     \
 #
 #  Libraries that are needed dependent on which modules are used


x-jokay:
@arr2036 ready for review.

mcnewton:
Is this needed for any of the other distributions, or is it just alpine?

Though the theory is that the alpine image is as tiny as possible, and if you want extra stuff then you add it yourself in a local Dockerfile, or use the Ubuntu image.

x-jokay:
Only for alpine, to have the same supported ENV variable which is already supported on the Ubuntu image.

I don't think this will have a remarkable impact on the size of the image.

Would be great to have logs with the "correct" time.

x-jokay:
compare

mcnewton:
Looks fine, thanks for checking.

arr2036:
@mcnewton could you pull this into master too.

mcnewton:
There's no alpine in master at the moment, the docker images are different.

PR #3872, opened 2021-01-24
CI: Rework package installation for on-push CI and scheduled fuzzing

terryburton:
Install packages from debian/control file rather than a list in the workflow file.

Install extra CI-specific packages from new scripts/ci/extra-packages.debian.control file

Defer fixture package installation until after build.

Disable rebuild of initramfs to optimise runtime.


terryburton:
This touches the same file as https://github.com/FreeRADIUS/freeradius-server/pull/3874 but different parts so should hopefully apply cleanly (or at least any conflict will be simple to resolve.)...

arr2036:
Looks good, thanks!

PR #3877, opened 2021-01-25
Rename dpair and dpair_legacy files to pair and pair_legacy

ndptech:


arr2036:
Looks like there's a problem with configure?

PR #3878, opened 2021-01-25
Remove CC_HINT(warn_unused_result) from fr_pair_steal()

jejones3141:
It "returns" void, and thus has no result to not use.


jejones3141:
Already corrected.

PR #3881, opened 2021-01-26
Update license

x-jokay:
Updates the license using the correct template from [licenses /
license-templates](https://github.com/licenses/license-templates/blob/master/templates/gpl2.txt) so it gets correctly recognized by GitHub too 😉 

And this is what it will look like:

license

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,12 +1,12 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 2, June 1991
+		    GNU GENERAL PUBLIC LICENSE
+		       Version 2, June 1991
 
  Copyright (C) 1989, 1991 Free Software Foundation, Inc.,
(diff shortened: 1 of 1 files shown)


arr2036:
Fine with this in principle.  Looks to be mostly formatting changes.  Only bit that stands out is changing `Lesser` to `Library` in the final paragraph. @alandekok are you ok with this?

x-jokay:
You are right, given the fact [this](https://github.com/ome/bioformats/blob/develop/LICENSE.txt) gets recognized as well, you may try this which results in lesser changes.

x-jokay:
Strange, it seems it didn't work 🧐

x-jokay:
Could be due to multiple `LICENSE` files, e.g. [here](https://github.com/FreeRADIUS/freeradius-server/blob/master/doc/LICENSE).

PR #3907, opened 2021-02-05
Add missing ruckus AVPs

joary:
Updated AVPs based on the latest SmartZone 5.1.1 RADIUS data dictionary, see https://support.ruckuswireless.com/articles/000005220

Mirror changes from wireshark: https://gitlab.com/wireshark/wireshark/-/merge_requests/1894


arr2036:
Could you remove the prefixes for both the top level attributes and the TLVs. Thanks.

alandekok:
I've pulled this in manually.

* v4 doesn't use the `Ruckus-` prefix
* we avoid `BEGIN-TLV` in v4, it's not necessary
* use consistent formatting

In short, make these attributes look like the other attributes in this dictionary, which look exactly like the rest of the dictionaries in the server.

Having random formatting is just confusing.

PR #3909, opened 2021-02-05
Update Arista dictionary for v3.0.x

hzhongmj:
This is to update the Arista dictionary to catch up what's in master.


hzhongmj:
This has conflicts, abandon and re-submit.

PR #3913, opened 2021-02-09
Convert lists of map_t to doubly linked lists

ndptech:


src/lib/server/map.c
```diff
@@ -799,69 +803,123 @@ static void map_sort_split(map_t *source, map_t **front, map_t **back)
...
+	*map = map_sort_merge(list, &a, &b, cmp);
+}
+
+
+/** Sort a doubly linked list of #map_t using merge sort
+ *
+ * @param[in,out] list of #map_t to sort.
+ * @param[in] cmp to sort with
+ */
+void map_sort(fr_map_list_t *list, fr_cmp_t cmp)
```
arr2036, line 882: @ndptech I have a feeling we'll probably want to sort different structures built on dlists.  Maybe add merge sort functions to dlist.h and use them here?  Shouldn't be much work seeing as you have most of the code :)

ndptech: It's now in dlist.h, and I've got that being used by fr_pair_list_sort() as well.

src/lib/ldap/base.h
```diff
@@ -319,7 +319,7 @@ typedef struct {
  * Used to store the array of attributes we'll be querying for.
  */
 typedef struct {
-	map_t const *maps;				//!< Head of list of maps we expanded the RHS of.
+	fr_map_list_t	const *maps;			//!< Head of list of maps we expanded the RHS of.
```
arr2036, line 322: Not sure this should be a pointer?

src/lib/server/map.c
```diff
@@ -58,7 +58,10 @@ static void map_dump(request_t *request, map_t const *map)
 
 static inline map_t *map_alloc(TALLOC_CTX *ctx)
 {
-	return talloc_zero(ctx, map_t);
+	map_t *map;
+	map = talloc_zero(ctx, map_t);
```
arr2036, line 62: `MEM(map = talloc_zero(ctx, map_t));`

src/lib/unlang/map_priv.h
```diff
@@ -31,7 +31,7 @@ extern "C" {
 typedef struct {
 	unlang_group_t		group;
 	tmpl_t			*vpt;
-	map_t		*map;		//!< Head of the map list
+	fr_map_list_t		*map;		//!< Head of the map list
```
arr2036, line 34: Should this be a pointer?

src/modules/rlm_radius/rlm_radius.c
```diff
@@ -326,9 +326,10 @@ static int status_check_update_parse(TALLOC_CTX *ctx, void *out, UNUSED void *pa
 	int			rcode;
 	CONF_SECTION		*cs;
 	char const		*name2;
-	map_t		*head = NULL;
+	fr_map_list_t		*head = talloc_zero(ctx, fr_map_list_t);
```
arr2036, line 329: hm?

src/modules/rlm_radius/rlm_radius.h
```diff
@@ -70,7 +70,7 @@ struct rlm_radius_s {
 	uint32_t		proxy_state;  		//!< Unique ID (mostly) of this module.
 	uint32_t		*types;			//!< array of allowed packet types
 	uint32_t		status_check;  		//!< code of status-check type
-	map_t		*status_check_map;	//!< attributes for the status-server checks
+	fr_map_list_t		*status_check_map;	//!< attributes for the status-server checks
```
arr2036, line 73: Should this be a pointer?

src/lib/ldap/base.h
ndptech: It's assigned in fr_ldap_map_expand() - which is passed a pointer - so yes.

arr2036: Ah you're right.  I'll double check this isn't the case with the rest.

PR #3914, opened 2021-02-09
Clean up vmps dbuff usage

jejones3141:


src/protocols/vmps/vmps.c
```diff
@@ -297,20 +299,22 @@ ssize_t fr_vmps_encode(fr_dbuff_t *dbuff, uint8_t const *original,
 		size_t len;
 
 		if (vp->da == attr_packet_type) {
-			fr_dbuff_current(&hdr)[1] = (uint8_t)vp->vp_uint32;
+			fr_dbuff_set(&io, fr_dbuff_current(&hdr) + 1);
```
arr2036, line 302: Need to check the return code of the `fr_dbuff_set` and the `fr_dbuff_in` calls

jejones3141:
Good point.Will do.

jejones3141:
OK. All writes that extend the dbuff are now of the FR_DBUFF_OUT*_RETURN()
flavor.

PR #3928, opened 2021-02-10
v3.0.x: Activate tmate on failure when pushing to the ci-debug branch

terryburton:


terryburton:
This now spawns tmate sessions for the main CI jobs as well as any failing package build and post-install test jobs.

PR #3932, opened 2021-02-12
rlm_redis: Add new 'logfile' option for rlm_redis. Closes #1114

### LINKED ISSUE #1114: The rlm_redis could provide a log_file option
Hi,

The rlm_redis could provide a log_file option like found in the rlm_sql. this could be
great to help the depuration.

jpereira:
It allows the user to redirect all REDIS queries to a logfile.


arr2036:
I really don't see the point in doing this for v3.  The module v3 is super basic, it's just used for xlat and doing simultaneous use checks.  Also v3 is against a single redis server, so you can monitor the commands from the redis side.  In the original issue i'd said v3.1.x and that was 5-6 years ago...  For master I honestly don't think this functionality is useful.  You can get the redis commands issued if you use a high enough debug level, and you can get them on the redis server itself... and it's not like SQL where the queries you log could be played back.  In most cases if you're doing something like accounting you'll be issuing multiple commands are they're likely to get interleaved with commands from other requests (at least in the current code), because you'll be issuing them from a series xlat expansions.

PR #3934, opened 2021-02-12
v4: Amend lists in dl.c to be doubly linked lists

ndptech:
Cursors were only used for walking the list and basic inserting and removal.  Replaced by direct pointer manipulation to remove the overhead of cursor initialisation and allowing the removal of cursors.


arr2036:
Can you just use dlists for this instead.  Then we get all the use after free checks (for the talloc chunks) as part of using that API.  Same for any other singly linked lists you're converting away from cursors, just make it a doubly linked list.

PR #3938, opened 2021-02-16
Only set TEST_SERVER variables for feature tests

ndptech:
Prevents builds without the feature test servers from attempting to test
against servers that don't exist.


terryburton:
LGTM

PR #3945, opened 2021-02-18
Add pre-order and post-order rbtree iterators.

jejones3141:
The test we have is based on static test data with the values
appropriate for pre and post iterators determined by hand. We
can generate additional test cases as desired.

### DIFF
--- a/src/lib/util/rbtree_tests.c
+++ b/src/lib/util/rbtree_tests.c
@@ -40,7 +40,7 @@ static int fr_rb_test_cmp(void const *one, void const *two)
 	return CMP(a->num, b->num);
 }
 
-static void test_rbtree_iter(void)
+static void test_rbtree_iter_inorder(void)
 {
 	rbtree_t 		*t;
 	fr_rb_test_node_t	sorted[MAXSIZE];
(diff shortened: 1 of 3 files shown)


src/lib/util/rbtree_tests.c
```diff
@@ -78,6 +87,42 @@ static void test_rbtree_iter(void)
 	pthread_mutex_unlock(&t->mutex);
 
 	talloc_free(t);
+
+	/*
+	 *	Build a tree for pre- and post-order iterator tests
+	 */
+	t = rbtree_alloc(NULL, fr_rb_test_node_t, node, fr_rb_test_cmp, NULL, RBTREE_FLAG_LOCK);
```
arr2036, line 94: Could you break pre and post order tests out into separate functions and add them to the test array at the bottom (maybe rename the existing test too). Other than that everything looks good!

jejones3141: Will do, thanks!

jejones3141: Changes made and pushed.

PR #3949, opened 2021-02-19
Add rbtree_iter_delete() and corresponding test.

jejones3141:
rbtree_iter_delete() deletes the current node in an iterator
and remembers the next one; rbtree_iter_next_inorder() notices
when that's happened and uses the saved node, otherwise it
behaves as usual.

### DIFF
--- a/src/lib/util/rbtree.c
+++ b/src/lib/util/rbtree.c
@@ -916,6 +916,17 @@ void *rbtree_iter_next_inorder(fr_rb_tree_iter_t *iter)
 	 */
 	if (unlikely(iter->node == NIL)) return NULL;
 
+	/*
+	 *	rbtree_iter_delete() has already deleted this node,
+	 *	and saved the next one for us. (We check for NULL;
+	 *	NIL just means we're at the end.)
+	 */
+	if (!iter->node) {
+		iter->node = iter->next;
+		iter->next = NULL;
+		return iter->node->data;
+	}
+
 	if (x->right != NIL) {
 		x = x->right;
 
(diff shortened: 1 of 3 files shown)


src/lib/util/rbtree.c
```diff
@@ -1122,6 +1133,23 @@ void *rbtree_iter_next_postorder(fr_rb_tree_iter_t *iter)
...
+ * @note Only makes sense for in-order traversals.
+ *
+ * @param[in] iter	previously initialised with #rbtree_iter_init
+ */
+void rbtree_iter_delete(fr_rb_tree_iter_t *iter)
+{
+	fr_rb_node_t *x = iter->node;
+
+	if (unlikely(x == NIL)) return;
+	(void) rbtree_iter_next_inorder(iter);
```
arr2036, line 1147: Could you fixup the iterator initialisers to insert a copy of the iterator function pointer in the iterator struct and call that instead.  Then this code should work for all iterator types.

src/lib/util/rbtree.c
```diff
@@ -916,6 +916,17 @@ void *rbtree_iter_next_inorder(fr_rb_tree_iter_t *iter)
...
+	/*
+	 *	rbtree_iter_delete() has already deleted this node,
+	 *	and saved the next one for us. (We check for NULL;
+	 *	NIL just means we're at the end.)
+	 */
+	if (!iter->node) {
+		iter->node = iter->next;
+		iter->next = NULL;
+		return iter->node->data;
+	}
```
arr2036, line 928: Could you add similar checks to the other iterators

src/lib/util/rbtree.c
jejones3141: Will do (with the analogous change to the next functions you mention in the other comment), and definitely try the same test for the pre and post with delete.

PR #3956, opened 2021-02-23
Bring back the rbtree_iter_done() function, and add an inorder flavor

jejones3141:
rbtree_t is only complete within rbtree.c, so a macro in rbtree.h
doesn't have the information to do what the rbtree_iter_done()
function in rbtree.c could do. So we'll bring back that function
and add rbtree_iter_inorder_done() that of course takes an
fr_rb_tree_iter_inorder_t *.


jejones3141:
Better solution in place; closing.

PR #3960, opened 2021-02-26
Get rid of the last use (and user) of rbtree_walk(), rbtree_flatten().

jejones3141:
That done, we also get rid of rbtree_walk() and the walk_*_order() functions.
The one use of rbtree_flatten() is now more clearly done with an iterator.


src/lib/io/network.c
```diff
@@ -1404,22 +1404,18 @@ static void fr_network_post_event(UNUSED fr_event_list_t *el, UNUSED fr_time_t n
...
-
-		len = rbtree_flatten(nr, (void ***)&sockets, nr->sockets, RBTREE_IN_ORDER);
-
-		for (i = 0; i < len; i++) talloc_free(sockets[i]);
-
-		talloc_free(sockets);
+	for (socket = rbtree_iter_init_inorder(&iter, nr->sockets);
+	     socket;
+	     socket = rbtree_iter_next_inorder(&iter)) {
+		talloc_free(socket);
```
arr2036, line 1418: talloc_free will call `_network_socket_free` which will result in a delete operation which will invalidate the cursor.

arr2036: the simple fix would be to delete the socket from `nr->sockets` then call talloc_free

arr2036: If you pull the latest code you'll likely now see the server deadlocks on exit

jejones3141: Ah. Thanks; I will correct it.

PR #3983, opened 2021-03-15
v4: Initial version of xlat arg parser plus first two xlats to use the parser

ndptech:


src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+		/*
+		 *	xlat takes all input as a single vb.
+		 */
+		vb = fr_dlist_head(in);
+		if (vb) {
+			if (args->concat) {
+				/* Concat all vbs to one, casting to the required type. */
+				if (fr_value_box_list_concat(ctx, vb, in, args->type, true) < 0) {
+concat_error:
+					RERROR("Failed concatenating input");
```
arr2036, line 266: Wants to be `RPERROR` to print the library error

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+					return XLAT_ACTION_FAIL;
+				}
+			} else {
+				if ((vb->type == args->type) && (vb->type != FR_TYPE_VALUE_BOX)) return XLAT_ACTION_DONE;
+
+				/* Just cast the first vb to the required type */
+				if (fr_value_box_cast_in_place(ctx, vb, args->type, NULL) < 0) {
+cast_error:
+					RERROR("Failed converting input to %s",
```
arr2036, line 275: `RPERROR` here too.

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+		 *	xlat needs no input processing just return.
+		 */
+		return XLAT_ACTION_DONE;
+
+	case XLAT_INPUT_MONO:
+		/*
+		 *	xlat takes all input as a single vb.
+		 */
+		vb = fr_dlist_head(in);
+		if (vb) {
```
arr2036, line 261: A lot of nesting here.  Maybe

```c
if (!vb && args->required) {
missing:
    RERROR("Missing required argument");
    return XLAT_ACTION_FAIL;
}
```

makes it a bit easier to follow the code too.

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+		 *	xlat consumes a sequence of arguments.
+		 */
+		arg = args;
+		vb = fr_dlist_head(in);
+		while (arg->type != FR_TYPE_INVALID) {
+			/*
+			 *	pre-advance, in case the vb is replaced during processing
+			 */
+			next = fr_dlist_next(in, vb);
+
```
arr2036, line 301:
```
if (!vb) {
    if (arg->required) goto missing;
    break;
}
```

src/lib/unlang/xlat_eval.c
ndptech: Isn't the the same level of nesting?

arr2036: It lets you unindent the code between lines 260-277

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+					fr_dlist_insert_after(in, vb, child_vb);
+					if ((child_vb->type != arg->type) && (arg->type != FR_TYPE_VALUE_BOX)) {
+						if (fr_value_box_cast_in_place(ctx, child_vb, arg->type, NULL) < 0)
+							goto cast_error;
+					}
+				}
+
+				fr_dlist_remove(in, vb);
+				talloc_free(vb);
+			} else if ((vb->type != arg->type) && (arg->type != FR_TYPE_VALUE_BOX)) {
```
arr2036, line 330: Hmm, boxed values can't contain boxed values.  Don't worry about ```arg->type != FR_TYPE_VALUE_BOX``` the cast function will catch the error.

src/lib/unlang/xlat_eval.c
arr2036: https://wiki.freeradius.org/contributing/coding-standards#code-style_return-early :)

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,121 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+			// *todo* make this escape tainted value boxes
+			if ((arg->concat) && (vb->type == FR_TYPE_GROUP)) {
+				fr_value_box_t *new_vb = fr_value_box_alloc(ctx, arg->type, NULL, false);
+				if (fr_value_box_list_concat(ctx, new_vb, &vb->vb_group, arg->type, true) < 0) {
+					talloc_free(new_vb);
+					goto concat_error;
+				}
+				fr_dlist_replace(in, vb, new_vb);
+				talloc_free(vb);
+			} else if ((vb->type == FR_TYPE_GROUP) && (arg->type != FR_TYPE_GROUP)) {
```
arr2036, line 315: I don't think moving up a level is correct here as it'd make it impossible for the xlat to determine the position of arguments after this one. Most xlats that take multiple boxes for an argument want to iterate over them as multiple values for that particular argument.  Casting in place is definitely correct though.

arr2036: There should probably be a validation pass when registering argument arrays which checks that the type passed in is in the set `FR_TYPE_VALUE`.

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,154 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+				talloc_free(vb);
+				goto arg_parsed;
+			}
+
+			/*
+			 *	We're neither concatenating nor do we only expect a single value,
+			 *	cast all child values to the required type.
+			 */
+
+			do {
```
arr2036, line 350: Should be a condition around the loop checking that `arg->type != FR_TYPE_VOID`, i.e. only cast children if a non void type is specified.

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,154 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+
+		if (args->concat) {
+			/* Concat all vbs to one, casting to the required type. */
+			if (fr_value_box_list_concat(ctx, vb, in, args->type, true) < 0) {
+		concat_error:
+				RPERROR("Failed concatenating input");
+				return XLAT_ACTION_FAIL;
+			}
+		} else {
+			if ((vb->type == args->type) && (vb->type != FR_TYPE_VALUE_BOX)) return XLAT_ACTION_DONE;
```
arr2036, line 278: should remove subcondition check for FR_TYPE_VALUE_BOX

src/lib/unlang/xlat_eval.c
```diff
@@ -222,6 +222,154 @@ static inline void xlat_debug_log_result(request_t *request, fr_value_box_t cons
...
+						goto cast_error;
+					}
+				}
+			 } while ((child_vb = fr_dlist_next(&vb->vb_group, child_vb)));
+
+		arg_parsed:
+			if (arg->variadic == false) {
+				arg++;
+			} else {
+				variadic_parsed = true;
```
arr2036, line 362: `if (arg->variadic && !fr_dlist_peek_next(in)) break` would mean you could remove `variadic_parsed`

PR #3985, opened 2021-03-16
🔧 tmpl: Fix tmpl_find_or_add_vp()

jpereira:
It should add the vp into the correct list.


arr2036:
Does that work for nested attributes within groups, where the groups need creating?

PR #3987, opened 2021-03-16
v4: 5 more xlats with argument definitions

ndptech:


src/lib/unlang/xlat_builtin.c
```diff
@@ -3414,7 +3410,7 @@ do { \
 	XLAT_REGISTER_MONO("md5", xlat_func_md5, xlat_func_md5_arg);
 	xlat_register(NULL, "module", xlat_func_module, false);
 	XLAT_REGISTER_MONO("pack", xlat_func_pack, xlat_func_pack_arg);
-	xlat_register(NULL, "pairs", xlat_func_pairs, false);
+	XLAT_REGISTER_MONO("pairs", xlat_func_pairs, xlat_func_pairs_arg);
```
arr2036, line 3397: This this actually an args xlat.  It's an xlat function with exactly one argument, which is the attribute reference.  The one attribute reference should be single valued.

src/lib/unlang/xlat_eval.c
```diff
@@ -923,9 +923,13 @@ xlat_action_t xlat_frame_eval_resume(TALLOC_CTX *ctx, fr_dcursor_t *out,
 	 *	to debug problems if they free or change elements
 	 *	and don't remove them from the list.
 	 */
-	if (!fr_dlist_empty(result)) { fr_dlist_verify(result); }
+#ifndef TALLOC_GET_TYPE_ABORT_NOOP
```
arr2036, line 926: Not needed

PR #3989, opened 2021-03-16
CI: Cache hostapd build directory to avoid unnecessary rebuild

terryburton:
Saves bandwidth and some build time in CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,8 @@ env:
   FR_GLOBAL_POOL: 4M
   TEST_CERTS: yes
   DO_BUILD: yes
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -101,6 +103,14 @@ jobs:
     - name: Git LFS pull
       run: git lfs pull
 
+    - name: Restore eapol_test build directory from cache
+      uses: actions/cache@v2
+      id: hostapd-cache
+      with:
+        path: ${{ env.HOSTAPD_BUILD_DIR }}
+        key: hostapd-${{ runner.os }}-${{ env.HOSTAPD_GIT_TAG }}-v3
```
terryburton, line 111: -v3 is because I botched stuff in my CI and needed to bump the cache epoch.

PR #3995, opened 2021-03-18
v4: define arguments for %(concat: ) xlat

ndptech:


arr2036:
Can you split concat into two keywords, concat and join, where concat takes a multi-valued argument as input, and an optional separator, and concat joins multiple variadic arguments into a single list.

PR #4000, opened 2021-03-18
v4: Make a generic pad function and use for both lpad and rpad xlats

ndptech:


arr2036:
Pushed an alternative version which just uses sbuffs instead of a mix of APIs.

PR #4013, opened 2021-03-22
Various linter fixes

terryburton:
~50% of the way towards an appeased Super Linter...


terryburton:
Ignore the linter complaints from this run for the changed Perl and Bash files. They're part of a future set of scripts to fix.

PR #4014, opened 2021-03-23
v4: Convert %(integer: ) to new xlat api

ndptech:


src/lib/unlang/xlat_builtin.c
```diff
@@ -1117,97 +1117,98 @@ static xlat_action_t xlat_func_explode(TALLOC_CTX *ctx, fr_dcursor_t *out,
...
-		if (vp->vp_length > 8) {
-			break;
-		}
+		if (in_vb->vb_length > 8) goto error;
 
-		if (vp->vp_length > 4) {
-			memcpy(&int64, vp->vp_octets, vp->vp_length);
-			return snprintf(*out, outlen, "%" PRIu64, htonll(int64));
+		if (in_vb->vb_length > 4) {
+			fr_value_box_cast_in_place(ctx, in_vb, FR_TYPE_INT64, NULL);
```
arr2036, line 1178: That should be UINT64

src/lib/unlang/xlat_builtin.c
```diff
@@ -1117,97 +1117,98 @@ static xlat_action_t xlat_func_explode(TALLOC_CTX *ctx, fr_dcursor_t *out,
...
 
 	uint64_t	int64 = 0;	/* Needs to be initialised to zero */
-	uint32_t	int32 = 0;	/* Needs to be initialised to zero */
-
-	fr_skip_whitespace(fmt);
+	char		buff[40];	/* Temporary string representation special conversions */
 
-	if ((xlat_fmt_get_vp(&vp, request, fmt) < 0) || !vp) return 0;
+	switch (in_vb->type) {
+	case FR_TYPE_UINT8:
```
arr2036, line 1147: case FR_TYPE_INTEGER:

src/lib/unlang/xlat_builtin.c
```diff
@@ -1117,97 +1117,98 @@ static xlat_action_t xlat_func_explode(TALLOC_CTX *ctx, fr_dcursor_t *out,
...
+	case FR_TYPE_INT32:
+	case FR_TYPE_INT64:
+	case FR_TYPE_TIME_DELTA:
+	case FR_TYPE_SIZE:
+	case FR_TYPE_BOOL:
+		/*
+		 *	Ensure enumeration is NULL so that the integer
+		 *	version of a box is returned
+		 */
+		fr_value_box_cast_in_place(ctx, in_vb, in_vb->type, NULL);
```
arr2036, line 1162: You just need to set the field in the box to NULL.  If there's issues manipulating the boxes directly add a helper function to clear the enum field in value.c

src/lib/unlang/xlat_builtin.c
```diff
@@ -1117,97 +1117,98 @@ static xlat_action_t xlat_func_explode(TALLOC_CTX *ctx, fr_dcursor_t *out,
...
-
-	case FR_TYPE_UINT32:
-		return snprintf(*out, outlen, "%u", vp->vp_uint32);
-
-	case FR_TYPE_UINT8:
-		return snprintf(*out, outlen, "%u", (unsigned int) vp->vp_uint8);
-
-	case FR_TYPE_UINT16:
-		return snprintf(*out, outlen, "%u", (unsigned int) vp->vp_uint16);
+		snprintf(buff, 40, "%lu", (long unsigned int)ntohl(in_vb->vb_ip.addr.v4.s_addr));
```
arr2036, line 1191: You can just use the normal cast function here

src/lib/unlang/xlat_builtin.c
```diff
@@ -1117,97 +1117,98 @@ static xlat_action_t xlat_func_explode(TALLOC_CTX *ctx, fr_dcursor_t *out,
...
+		goto output_buff;
 
 	/*
 	 *	Ethernet is weird... It's network related, so it
 	 *	should be bigendian.
 	 */
 	case FR_TYPE_ETHERNET:
-		int64 = vp->vp_ether[0];
+		int64 = in_vb->vb_ether[0];
```
arr2036, line 1199: I think just cast to a UINT64 and ignore the comment.  If there's not that capability in the box casting function already then add it.

src/lib/unlang/xlat_builtin.c
ndptech: FR_TYPE_INTEGER includes FR_TYPE_DATE, which needed different handling to pass tests

PR #4029, opened 2021-03-25
v4: Define argument for %{rest: } xlat

ndptech:


arr2036:
This is one of those annoying ones which need escaping and you can't use the argument framework for.

You need to ingest each of the value boxes in turn, tracking where you are in the URI.  For tainted boxes you need to apply escaping consistent with where in the URI you are, or error out if the input is invalid (for example, tainted boxes for schemes should not be allowed).

A very similar scheme will be needed for the LDAP URIs but they're more complicated.

arr2036:
closing for now, so it's not merged accidentally

PR #4033, opened 2021-03-26
v4: convert %(config :) to new xlat api

ndptech:


src/lib/unlang/xlat.h
```diff
@@ -98,14 +98,15 @@ extern size_t xlat_action_table_len;
...
+ * @param[in] ctx		to allocate any additional buffers in.
  * @param[in] request		being processed.  Used mostly for debugging.
  * @param[in,out] vb		to escape
  * @param[in] uctx		a "context" for the escaping
  * @return
  *	- 0 on success.
  *	- -1 on failure.
  */
-typedef int (*xlat_escape_func_t)(request_t *request, fr_value_box_t *vb, void *uctx);
+typedef int (*xlat_escape_func_t)(TALLOC_CTX *ctx, request_t *request, fr_value_box_t *vb, void *uctx);
```
arr2036, line 109: I don't think so... They'd allocate buffers with the vb as the ctx?

PR #4041, opened 2021-03-30
v4: Convert %(xlat: ) to new xlat api

ndptech:


src/lib/unlang/xlat_builtin.c
```diff
@@ -1317,47 +1317,54 @@ static xlat_action_t xlat_func_next_time(TALLOC_CTX *ctx, fr_dcursor_t *out, req
...
 
-	fr_skip_whitespace(fmt);
-
-	if (outlen < 3) {
-	nothing:
-		return 0;
-	}
-
-	if ((xlat_fmt_get_vp(&vp, request, fmt) < 0) || !vp) goto nothing;
+	if ((xlat_fmt_get_vp(&vp, request, in_head->vb_strvalue) < 0) || !vp) return XLAT_ACTION_FAIL;
```
arr2036, line 1333: It should just take a string, there's no reason to lookup an attribute.

src/lib/unlang/xlat_builtin.c
```diff
@@ -1317,47 +1317,54 @@ static xlat_action_t xlat_func_next_time(TALLOC_CTX *ctx, fr_dcursor_t *out, req
...
 
+	MEM(vb = fr_value_box_alloc_null(ctx));
 	/*
 	 *	If it's a string, expand it again
 	 */
 	if (vp->vp_type == FR_TYPE_STRING) {
-		slen = xlat_eval(*out, outlen, request, vp->vp_strvalue, NULL, NULL);
-		if (slen <= 0) return slen;
+		char	buff[1024];
+		slen = xlat_eval(buff, 1024, request, vp->vp_strvalue, NULL, NULL);
```
arr2036, line 1344: Not the correct function to use here.  You need to use the async capable call.  You might just be able to get away with parenting the expansion off the ctx returned by `interpreter_frame_talloc_ctx()`, and pushing the new xlat onto the interpreter stack for evaluation.

src/lib/unlang/xlat_builtin.c
arr2036: This should really be a mono xlat too.

PR #4046, opened 2021-04-02
Fix EAP tests

jpereira:


alandekok:
Pulled over manually because I hate squashing multiple commits into one.

PR #4050, opened 2021-04-08
Updated all SQL schema adding a new 'Class' column.

jpereira:
It helps the admin to locate accounting packets.


alandekok:
Merged in as commit 287feed2c6

PR #4055, opened 2021-04-13
v4: Various fixes which allow sqlippools to work correctly, especially for DHCP

ndptech:


alandekok:
I've pulled over most of these manually.  I'm less sure about adding DHCP options to the main `sqlippool` module.  I'd prefer to not have multiple things in the same example configuration.

Perhaps instead just add a `mods-available/dhcp_sqlippool` sample configuration.

Tho if we have lots of these kind of things, it may be worth investigating how to make these even more protocol agnostic.   Tho that may take a deeper investigation.

For now, duplicating the module and having a clear "dhcp" config seems better than have one "sqlippool" config with a lot of "for RADIUS do X, for DHCP do Y".

PR #4059, opened 2021-04-16
bootstrap: check existence of openssl's commands

kkang-wr:
It calls openssl's commands 'dhparam' and 'pkcs12' in script bootstrap                                                                                                       
which are configurable based on openssl configure options 'no-dh' and 
'no-des'. So check existence of these commands. If not exist, abort from
running of the script.

1. https://github.com/openssl/openssl/blob/master/apps/build.info#L37
2. https://github.com/openssl/openssl/blob/master/apps/build.info#L22

Signed-off-by: Kai Kang <kai.kang@windriver.com>

### DIFF
--- a/raddb/certs/bootstrap
+++ b/raddb/certs/bootstrap
@@ -14,6 +14,14 @@ umask 027
 cd "$(dirname "$0")" || exit 1
 
 
+#  check commands of openssl exist
+for cmd in dhparam pkcs12; do
+	if ! openssl ${cmd} -help >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


alandekok:
This just aborts the entire script if those commands don't exist. Why not check for the existence of the commands, and skip them if they don't exist?

i.e.

```
if openssl supports FOO
   run command foo
```

kkang-wr:
I suppose these openssl commands are essential to create certificates in this scripts. Without these commands, script bootstrap should be unfunctional.

alandekok:
You can always run the scripts on another system.  So these aren't critical fixes.

PR #4062, opened 2021-04-16
v4: Convert %(soh: ) to new xlat api

ndptech:


mcnewton:
Is it worth updating SoH, or should it just be scrapped? IIRC Windows 10 dropped it a few years ago, and aside from w7/Vista/8 nothing else ever did.

arr2036:
yeah should probably be scrapped

arr2036:
but... lets merge this in anyway until someone has time to do it :)

PR #4071, opened 2021-04-29
Add new 'Class' column into post-auth / SQL schemas

jpereira:


raddb/mods-config/sql/main/oracle/queries.conf
```diff
@@ -419,10 +419,11 @@ post-auth {
...
 		INSERT INTO ${..postauth_table} \
-			(username, pass, reply, authdate) \
+			(username, pass, reply, authdate, class) \
 		VALUES (\
 			'%{User-Name}', \
 			'%{%{User-Password}:-%{Chap-Password}}', \
 			'%{reply.Packet-Type}', \
-			TO_TIMESTAMP('%S.%M','YYYY-MM-DDHH24:MI:SS.FF'))"
+			TO_TIMESTAMP('%S.%M','YYYY-MM-DDHH24:MI:SS.FF'), \
+			'%{Class}')"
```
arr2036, line 428: Wrong list

raddb/mods-config/sql/main/postgresql/queries.conf
```diff
@@ -469,10 +469,11 @@ post-auth {
...
 		INSERT INTO ${..postauth_table} \
-			(username, pass, reply, authdate) \
+			(username, pass, reply, authdate, class) \
 		VALUES(\
 			'%{User-Name}', \
 			'%{%{User-Password}:-Chap-Password}', \
 			'%{reply.Packet-Type}', \
-			'%S.%M')"
+			'%S.%M', \
+			'%{Class}')"
```
arr2036, line 478: Wrong list

arr2036:
Class attribute is in the reply list `%{reply.Class}` for authentication, because we're providing the value for the NAS.

raddb/policy.d/accounting
```diff
@@ -107,3 +107,47 @@ acct_counters64.preacct {
...
+#  window of opportunity is much smaller.  i.e. milliseconds, instead
+#  of seconds.
+#
+#  This policy can also be used to "bootstrap" accounting sessions.
+#  If there is data which is only available in the Access-Request,
+#  it can be placed in the accounting table.  Then, when accounting
+#  packets are received, they will update the row which contains
+#  the session information.
+#
+sql_session_start.post-auth {
```
arr2036, line 150: This should be generic.  We can create a fake accounting packet in v4, so we no longer need an "SQL" specific policy.  You should create a subrequest, then use the call keyword to loop it back around to the current virtual server.  The policy should fail if we don't have a `Acct-Session-Id`.  Any other comments @alandekok ?

alandekok:
We should probably also keep commit messages to ASCII.

PR #4079, opened 2021-05-17
Fix Debian package path

jpereira:
Just completing the previous commit https://github.com/FreeRADIUS/freeradius-server/commit/2238543273092adb981242badea1b92884528d6c


jpereira:
Just completing the previous commit https://github.com/FreeRADIUS/freeradius-server/commit/2238543273092adb981242badea1b92884528d6c

PR #4083, opened 2021-05-19
v4: Convert %(idn: ) to new xlat API

ndptech:


arr2036:
I think this one would be a mono function as it's only accepting the domain to internationalise

PR #4086, opened 2021-05-19
Updating Aruba dict with VSA used in Aruba C2C product

antonimilton:
Adding two new aruba VSA for Aruba's campus-to-cloud (C2C) environment.


alandekok:
Pulled in via commit f3baa777f31e

PR #4087, opened 2021-05-20
v4: Convert %(unpack: ) to new xlat API

ndptech:


src/modules/rlm_unpack/rlm_unpack.c
```diff
@@ -59,154 +64,93 @@ fr_dict_attr_autoload_t rlm_unpack_dict_attr[] = {
...
-			GOTO_ERROR;
+			REDEBUG("Zero length hex string in '%s'", data_vb->vb_strvalue);
+			return XLAT_ACTION_FAIL;
 		}
+	} else if (data_vb->type == FR_TYPE_STRING) {
+		input = (uint8_t const *)data_vb->vb_strvalue;
+		input_len = data_vb->length;
 	} else {
-		GOTO_ERROR;
+		input = data_vb->vb_octets;
```
alandekok, line 109: this is removing the requirement above that `unpack requires the input attribute to be 'string' or 'octets'"`

i.e. it should be

```
} else if (data_vb->type == FR_TYPE_OCTETS) {
   ... do octets stuff
} else {
  complain and return an error
}

alandekok: If the data isn't "octets" form, it might be worth just casting it to octets?  That would allow for things like unpacking 24-bit vendor IDs from ethernet addresses.

I suppose the user could manually add a cast to the xlat, but it might be nice to just do it for them.

ndptech: At the top of the routine there is a test for whether the data is octets or string - anything else throws an error and returns.

I could add a cast for other data types.

alandekok: OK, for now let's just leave it.

alandekok:
Merged manually.  I really dislike the GitHub "squash and merge" for complex commits.

PR #4088, opened 2021-05-20
v4: Convert %{escape: } and %{unescape: } to new xlat API

ndptech:


src/modules/rlm_escape/rlm_escape.c
```diff
@@ -55,68 +63,86 @@ static char const hextab[] = "0123456789abcdef";
...
+	fr_value_box_t		*arg = fr_dlist_head(in);
+	char const		*p = arg->vb_strvalue;
+	char			*out_p;
+	size_t			reqlen = 0;
+	size_t			len;
+	fr_value_box_t		*vb;
+	fr_sbuff_t		sbuff;
+
+	memcpy(&instance, xlat_inst, sizeof(instance));
+	inst = talloc_get_type_abort(instance, rlm_escape_t);
```
arr2036, line 81: Instance data can be const too?

arr2036: Also all the complexity around utf8 chars can be removed.  That was just to stop truncation occurring in the middle of a char.  Now we have dynamically sized buffers it's not an issue

src/modules/rlm_escape/rlm_escape.c
```diff
@@ -55,68 +63,86 @@ static char const hextab[] = "0123456789abcdef";
...
+		if (fr_utf8_strchr(&chr_len, inst->allowed_chars, len, p) == NULL) {
+			reqlen += chr_len * 3;	/* '=' + ([hex]{2}) */
+		} else {
+			reqlen += chr_len;
+		}
+		p += chr_len;
+	}
+
+	MEM(vb = fr_value_box_alloc_null(ctx));
+	if (fr_value_box_bstr_alloc(ctx, &out_p, vb, NULL, reqlen, false) < 0) {
```
arr2036, line 98: you can create a talloc sbuff using the box as the ctx, with an initial size of double the input string, then trim it at the end, and use the shallow copy functions to assign it to the box.  Probably faster and much less code.

PR #4092, opened 2021-05-24
v4: Convert %(perl: ) to new xlat API

ndptech:


src/modules/rlm_perl/rlm_perl.c
```diff
@@ -424,10 +421,14 @@ static ssize_t perl_xlat(UNUSED TALLOC_CTX *ctx, char **out, size_t outlen,
 			(void)POPs;
 		} else if (count > 0) {
 			tmp = POPp;
-			strlcpy(*out, tmp, outlen);
-			ret = strlen(*out);
-
-			RDEBUG2("Len is %zu , out is %s freespace is %zu", ret, *out, outlen);
+			MEM(vb = fr_value_box_alloc_null(ctx));
+			if (fr_value_box_strdup(ctx, vb, NULL, tmp, true) < 0) {
```
arr2036, line 425: should use bstrndup

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -351,6 +351,49 @@ static void xs_init(pTHX)
...
+			ssize_t	slen;
+
+			slen = fr_value_box_print_quoted(&FR_SBUFF_OUT(buffer, sizeof(buffer)), vb, T_BARE_WORD);
+			if (slen < 0) return -1;
+			sv = newSVpvn(buffer, (size_t)slen);
+		}
+			break;
+		}
+		if (!sv) return -1;
+		SvTAINT(sv);
```
arr2036, line 393: Probably want to reflect the taint status of the individual boxes?

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -394,6 +394,123 @@ static int perl_vblist_to_av(AV *av, fr_value_box_list_t *head) {
...
+	HV		*hv;
+	I32		sv_len, i;
+	int		type;
+
+	type = SvTYPE(sv);
+
+	switch (type) {
+	case SVt_IV:
+	/*	Integer or Reference */
+		DEBUG3("Integer or Reference returned");
```
arr2036, line 427: References are special, we should probably deref if it's an array and convert that.  Blessed references can also contain package names, so we could in theory support certain cpan libraries and their underlying types... maybe that's future work though.

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -394,6 +394,123 @@ static int perl_vblist_to_av(AV *av, fr_value_box_list_t *head) {
...
+		if (SvROK(sv)) {
+			if (perl_sv_to_vblist(ctx, list, request, SvRV(sv)) < 0) return -1;
+			break;
+		}
+		MEM(vb = fr_value_box_alloc(ctx, FR_TYPE_INT32, NULL, false));
```
arr2036, line 432: DEBUG3("Integer")

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -394,6 +394,123 @@ static int perl_vblist_to_av(AV *av, fr_value_box_list_t *head) {
...
+	int		type;
+
+	type = SvTYPE(sv);
+
+	switch (type) {
+	case SVt_IV:
+	/*	Integer or Reference */
+		DEBUG3("Integer or Reference returned");
+		if (SvROK(sv)) {
+			if (perl_sv_to_vblist(ctx, list, request, SvRV(sv)) < 0) return -1;
```
arr2036, line 429: DEBUG3("Reference returned")

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -394,6 +394,123 @@ static int perl_vblist_to_av(AV *av, fr_value_box_list_t *head) {
...
+	/*	Float */
+		DEBUG3("Float returned");
+		MEM(vb = fr_value_box_alloc(ctx, FR_TYPE_FLOAT64, NULL, false));
+		vb->vb_float64 = SvNV(sv);
+		break;
+
+	case SVt_PV:
+	/*	String */
+		DEBUG3("String returned");
+		tmp = SvPVbyte(sv, len);
```
arr2036, line 446: `SvPVutf8`

src/tests/modules/perl/test.pl
```diff
@@ -67,14 +67,14 @@ sub authenticate {
 	if ($RAD_REQUEST{'User-Name'} =~ /^baduser/i) {
 		# Reject user and tell him why
 		$RAD_REPLY{'Reply-Message'} = "Denied access by rlm_perl function";
-		return RLM_MODULE_REJECT;
+		return RLM_MODULE_OK;
```
arr2036, line 70: hm?

src/modules/rlm_perl/rlm_perl.c
```diff
@@ -351,6 +351,166 @@ static void xs_init(pTHX)
...
+		switch (vb->type) {
+		case FR_TYPE_STRING:
+			sv = newSVpvn(vb->vb_strvalue, vb->length);
+			break;
+
+		case FR_TYPE_OCTETS:
+			sv = newSVpvn((char const *)vb->vb_octets, vb->vb_length);
+			break;
+
+		default:
```
arr2036, line 381: Handling `FR_TYPE_GROUP` probably wouldn't hurt.

src/modules/rlm_perl/rlm_perl.c
arr2036: I think we can actually get the taint status from perl as well for all of these.

src/tests/modules/perl/test.pl
ndptech: For the purpose of the test this was needed - returning RLM_MODULE_REJECT causes the test to stop immediately.  I'll change it to RLM_MODULE_NOTFOUND and add a comment

PR #4100, opened 2021-06-01
v4: Fixups of the handling of pair lists passed as inputs to rlm_exec calls

ndptech:
Plus add simple tests for calling exec as a module


alandekok:
Merged manually.

PR #4110, opened 2021-06-14
Tentative: Nightly pkgs: Build only on distros that satisfy JSON-C >= 1.13

terryburton:
The current JSON-C version requirement is only satisfied by CentOS >= 8,
Ubuntu >= 20.04 and Debian >= Bullseye (testing).

Building on earlier distros would require that we maintain a local
JSON-C package.

So we remove these builds for now to reduce noise...

### DIFF
--- a/.github/workflows/ci-deb.yml
+++ b/.github/workflows/ci-deb.yml
@@ -31,10 +31,7 @@ jobs:
           M=$(cat <<EOF
           {
             "env": [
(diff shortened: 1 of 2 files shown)


arr2036:
I think @mcnewton is working on custom packages for JSON-C so this shouldn't be required.

mcnewton:
Yes, this is fixed now. Thanks.

PR #4118, opened 2021-06-17
v4: Convert %(unbound: ) to new xlat api

ndptech:


src/modules/rlm_unbound/rlm_unbound.c
```diff
@@ -77,6 +77,15 @@ static const CONF_PARSER module_config[] = {
 	CONF_PARSER_TERMINATOR
 };
 
+/*
+ *	Cleanup events after unbound has either completed or timed out
+ */
+static void unbound_event_cleanup(unbound_xlat_thread_inst_t *xt)
```
arr2036, line 80: You also need to call this as a cancellation callback (it's a function you pass when you yield, along with the resumption callback).  I'm also suspicious of this approach in general.. Won't unbound re-use the same FD for multiple queries? In which case you'll need to track how many outstanding requests you have before deleting the FD.  By removing the write callbacks you also introduce the possibility of exhausting the FD buffer, because you always write, no matter what state the connection is in.

This is why the code in io.c was there.  Unbound would tell us whether it had outstanding requests on a given FD, and we would only allow it to write data when the FD was writable...

PR #4122, opened 2021-06-25
v4: Convert %(unbound: ) to new xlat API

ndptech:


arr2036:
As a general rule, when you have a resource from a 3rd party library that's bound to a talloc chunk it's better to use a destructor to free that resource.  You have the `done` field in the unbound request.  If you could add a destructor that looked at that, and called cancel if the unbound request was being freed but done was `false`, that'd be ideal.

 It's about minimising the potential for bad code paths which leak resources.
 
The parenting of the unbound request is incorrect, in that the lifetime should be bound to the current unlang stack frame and not the request.  Again, this provides more safety, as if the request is cancelled, and partially unwinds the stack, the unbound request is freed, and the cancel function would be called.

The function you want to use to get that ctx is `unlang_interpret_frame_talloc_ctx` which'll return a `TALLOC_CTX`.

PR #4126, opened 2021-07-01
debug: don't set resource hard limit to zero

antoniotorresm:
Setting the resource hard limit to zero is irreversible, meaning if it
is set to zero then there is no way to set it higher. This means
enabling core dump is not possible, since setting a new resource limit
for RLIMIT_CORE would fail. By only setting the soft limit to zero, we
can disable and enable core dumps without failures.

Ticket in RHEL Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1977572

### DIFF
--- a/src/lib/debug.c
+++ b/src/lib/debug.c
@@ -599,7 +599,7 @@ int fr_set_dumpable(bool allow_core_dumps)
 		struct rlimit no_core;
 
 		no_core.rlim_cur = 0;
-		no_core.rlim_max = 0;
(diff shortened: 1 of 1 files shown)


antoniotorresm:
Just for the record, this change is already in the master branch: https://github.com/FreeRADIUS/freeradius-server/blob/2d092d1436dcbe16b815440eb5d15a83e4aef2d1/src/lib/util/debug.c#L847

alandekok:
fixed in commit 1ce4508c

PR #4127, opened 2021-07-02
v4: Define arguments for %(rest: ) xlat and parse / escape URIs

ndptech:


src/lib/unlang/xlat_eval.c
```diff
@@ -2216,3 +2216,78 @@ bool xlat_async_required(xlat_exp_t const *xlat)
...
+		 */
+		if (uri_vb->length == 0) continue;
+
+		/*
+		 *	Look for URI part terminator
+		 */
+		fr_sbuff_init(&sbuff, uri_vb->vb_strvalue, uri_vb->length);
+
+		do {
+			if ((p = strchr(uri_part->term_chars, *fr_sbuff_current(&sbuff)))) {
```
arr2036, line 2277: There's sbuff functions to handle searcing for the next char

arr2036:
```
	fr_sbuff_term_t const 		bareword_terminals =
					FR_SBUFF_TERMS(
						L("\t"),
						L("\n"),
						L(" "),
						L("%"),
						L("&"),
						L(")"),
						L("+"),
						L("-"),
						L("/"),
						L("^"),
						L("|")
					);
```

arr2036:
```size_t	fr_sbuff_adv_until(fr_sbuff_t *sbuff, size_t len, fr_sbuff_term_t const *tt, char escape_chr);
```

arr2036: Then the sbuff will be left at the first character of the termination substring.  The terminal sequences do need to be ordered in the array though.

Issue #4130, opened 2021-07-08
Using a combination of rlm_rest, body = json, and chunk > 0 crashes


nward:
# Issue type
- Defect - Crash or memory corruption.

### How to reproduce the issue

Using a combination of rlm_rest, body = json, and chunk > 0 crashes.

I'm not certain if this is supposed to work - as chunk is undocumented, however chunk seems to work with body = post just fine - for the few hundred or so requests I put through it when testing.

I'm needing to use chunk, as otherwise some of my RADIUS accounting messages are larger than REST_BODY_MAX_LEN's 8192B - I can of course increase that, but noted that chunk exists so trying to get it to play nicely first.

Also interestingly (though not really related to this exactly problem (probably) - if REST_BODY_MAX_LEN is exceeded, radiusd.log says `ERROR: Server returned no data`, but if I log the request afterwards I get:
```
        Module-Failure-Message = "mqrest: Failed creating HTTP body content"
        Module-Failure-Message = "mqrest: Server returned no data"
```
My messages are so long as I'm unpacking ERX-DHCP-Options etc.

Anyway - I'm digging in to this, trying to remember how curl's API works, but posting here in case @arr2036 (who appears to have written most of rlm_rest, at least recently) says "no, don't use chunk, it's broken and we're not going to fix it" :-)

My intent presently is to solve it and document chunk.

### Output of ``[radiusd|freeradius] -X`` showing issue occurring

This was run in gdb.

```text
(0)   accounting {
(0)     if (request:Chargeable-User-Identity) {
(0)     if (request:Chargeable-User-Identity)  -> TRUE
(0)     if (request:Chargeable-User-Identity)  {
(0)       redundant {
rlm_rest (mqrest): Reserved connection (0)
(0) mqrest: Expanding URI components
(0) mqrest: EXPAND http://127.0.0.1:8080
(0) mqrest:    --> http://127.0.0.1:8080
(0) mqrest: EXPAND /accounting
(0) mqrest:    --> /accounting
(0) mqrest: Sending HTTP POST to "http://127.0.0.1:8080/accounting"
(0) mqrest: Encoding attribute "User-Name"

<crash>
```
### Full backtrace from LLDB or GDB

```text
Program received signal SIGSEGV, Segmentation fault.
[...]

PR #4131, opened 2021-07-08
Fix chunked rlm_rest HTTP body

nward:
Re. #4130.

This fixes rlm_rest's somewhat hidden chunked encoding HTTP feature, which stopped working in #3810.

Additionally, this documents the chunk config parameter.

I've tested this fairly well, I think - however raw_value doesn't seem to do anything for me - even before I made my changes, so I can't really validate that this still works.
In my environment, the value param is always the raw integer value - i.e. Acct-Status-Type in the JSON is always a number, and never "Start" or "Stop" or what not. My understanding is that `raw_value = yes` should make this be the case, but it always seems to be for me, so not sure what's going on there. I didn't even realise it should be translating it for me, so didn't know what I was missing until reading what `raw_value` is supposed to do !

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -90,6 +90,11 @@ rest {
 	#    password     - Password to use for authentication, will be expanded.
 	#    require_auth - Require HTTP authentication.
 	#    timeout      - HTTP request timeout in seconds, defaults to 4.0.
+	#    chunk        - Chunk size to use. If set, HTTP chunked encoding is used to
+	#                   send data to the REST server. Make sure that this is large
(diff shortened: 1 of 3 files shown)


nward:
```
E: Failed to fetch http://debian-archive.trafficmanager.net/debian/pool/main/d/dbus/dbus_1.12.20-0+deb10u1_amd64.deb  Undetermined Error [IP: 52.247.76.206 80]
1090
```

Well, that's hardly fair! :-)

arr2036:
Could you provide some more information on how this fixes the issue?

PR #4136, opened 2021-07-10
Removed the ability to pass NULL as data argument to fr_heap_extract().

jejones3141:
It was set up to return fr_heap_pop() if passed NULL, but the only
users of that capability were in heap test code.

### DIFF
--- a/src/lib/util/heap.c
+++ b/src/lib/util/heap.c
@@ -214,30 +214,22 @@ int fr_heap_extract(fr_heap_t *hp, void *data)
 	int32_t parent, child, max;
 
 	/*
-	 *	Extract element.  Default is the first one (pop)
+	 *	Extract element.
 	 */
-	if (!data) {
-		if (unlikely((hp->num_elements == 0) || !hp->p[0])) {
-			fr_strerror_const("Tried to extract element from empty heap");
-			return -1;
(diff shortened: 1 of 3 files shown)


src/lib/util/heap.c
```diff
@@ -214,30 +214,22 @@ int fr_heap_extract(fr_heap_t *hp, void *data)
...
 	 */
-	if (!data) {
-		if (unlikely((hp->num_elements == 0) || !hp->p[0])) {
-			fr_strerror_const("Tried to extract element from empty heap");
-			return -1;
-		}
-		parent = 0;
-	} else {		/* extract from the middle */
-		parent = index_get(hp, data);
+	parent = index_get(hp, data);
```
alandekok, line 219: `index_get()` dereferences `data`.  For safety, `fr_heap_extract()` should check for `data==NULL`, and return an error if that's passed to it.

jejones3141:
Sorry for the late reply. The change also changes heap.h so the
CC_HINT(nonnull) applies to both parameters of fr_heap_extract(). I've seen
the compiler complain about code checking whether parameters so hinted are
NULL, so I didn't put in the check; perhaps the argument is that static
analysis will find code that risks passing a NULL?

PR #4150, opened 2021-07-23
Fix capabilities in service file

jspricke:
As freeradius is not run as root we need to request extra capabilities
wiht AmbientCapabilities instead of limiting the set with
CapabilityBoundingSet.

see: https://bugs.debian.org/985967

### DIFF
--- a/debian/freeradius.service
+++ b/debian/freeradius.service
@@ -42,7 +42,7 @@ RestartSec=5
 NoNewPrivileges=true
 
 # Allow binding to secure ports, broadcast addresses, and raw interfaces.
-CapabilityBoundingSet=CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_NET_BROADCAST CAP_NET_RAW CAP_SETUID CAP_SETGID CAP_CHOWN CAP_DAC_OVERRIDE
+AmbientCapabilities=CAP_NET_ADMIN CAP_NET_BIND_SERVICE CAP_NET_BROADCAST CAP_NET_RAW CAP_SETUID CAP_SETGID CAP_CHOWN CAP_DAC_OVERRIDE
 
 # Private /tmp that isn't shared by other processes
 PrivateTmp=true


arr2036:
I'm not sure this is the right fix.  The systemd documentation isn't great and I may be interpreting it wrong... But from my understanding `CapabilityBoundingSet` was meant to set the list permitted capabilities, and `AmbientCapabilities` was meant to set the list of permitted capabilities and the list of effective capabilities.

If this is correct, then the fix here would be to have radiusd request the capability using the cap API (as it does already with v4.0.x/master).  The rationale here is that it's better to notify the user explicitly that a given capability is not available if the process requires it to perform some task... and If we're going to the trouble of checking if a capability is effective for a process, we may as well request it. 

There's already utility functions in master 

https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/util/cap.c
https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/util/cap.h

Seeing where `fr_cap_enable` is called in master would give an indication of the capabilities we need and where we need them.  @jspricke if you want to pull that code back to the v3.0.x branch and add the required calls it shouldn't be that much work...

`fr_strerror_cont` calls can just be replaced with `fr_strerror_printf`.

Note: You opened this fix against master branch where I don't believe the debian issue applies.  The fix should be opened against v3.0.x.

jspricke:
> I'm not sure this is the right fix. The systemd documentation isn't great and I may be interpreting it wrong... But from my understanding `CapabilityBoundingSet` was meant to set the list permitted capabilities, and `AmbientCapabilities` was meant to set the list of permitted capabilities and the list of effective capabilities.

My understanding is that `CapabilityBoundingSet` can reduce the existing capabilities whereas `AmbientCapabilities` extends them from the default (based on https://unix.stackexchange.com/a/581337). I tested this with the Debian package and `AmbientCapabilities` solved the problem, as reported in the bug as well.

> If this is correct, then the fix here would be to have radiusd request the capability using the cap API (as it does already with v4.0.x/master). The rationale here is that it's better to notify the user explicitly that a given capability is not available if the process requires it to perform some task... and If we're going to the trouble of checking if a capability is effective for a process, we may as well request it.
> 
> There's already utility functions in master
> 
> https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/util/cap.c
> https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/util/cap.h
> 
> Seeing where `fr_cap_enable` is called in master would give an indication of the capabilities we need and where we need them. @jspricke if you want to pull that code back to the v3.0.x branch and add the required calls it shouldn't be that much work...

I only changed the Debian package to make sure it will be in the next stable version, backporting the code from master would probably too big of a change as Debian is frozen for the release, already. Anyway, if there is already a different fix in master, that's fine with me, closing then. Thanks for the feedback!

PR #4165, opened 2021-07-29
v3.0.x: PROXY protocol fixes v2

terryburton:


src/main/tls_listen.c
```diff
@@ -278,14 +278,17 @@ static int proxy_protocol_check(rad_listen_t *listener, REQUEST *request)
...
-		RDEBUG("(TLS) Received PROXY protocol connection from client %s:%s -> %s:%s, via proxy %s:%u -> %s:%u",
-		       argv[0], argv[2], argv[1], argv[3],
-		       inet_ntop(af, &sock->haproxy_src_ipaddr, src_buf, sizeof(src_buf)),
-		       sock->haproxy_src_port,
-		       inet_ntop(af, &sock->haproxy_dst_ipaddr, dst_buf, sizeof(dst_buf)),
-		       sock->haproxy_dst_port);
+		char src_buf[128], dst_buf[128], csrc_buf[128], cdst_buf[128];
+
+		RDEBUG("(TLS) Received PROXY protocol connection from client %s:%d -> %s:%d, via proxy %s:%u -> %s:%u",
+		      inet_ntop(af, &src.ipaddr, csrc_buf, sizeof(csrc_buf)),
```
alandekok, line 284: Ah... argv[0] doesn't work because of the preceding memmove()...

src/main/tls_listen.c
```diff
@@ -375,7 +375,6 @@ static int tls_socket_recv(rad_listen_t *listener)
 	if (listener->proxy_protocol) {
 		rcode = proxy_protocol_check(listener, request);
 		if (rcode < 0) goto do_close;
-		if (rcode == 0) return 1;
```
alandekok, line 373: The `proxy_protocol_check()` function should return:

* -1 on error
* 0 for "need to read more data"
* 1 for "done

I'll push a slightly different fix.

alandekok:
I've pushed some of those patches, and some slightly different fixes.  I hope it works this time. :)

src/main/tls_listen.c
terryburton: That works.

terryburton:
> I've pushed some of those patches, and some slightly different fixes. I hope it works this time. :)

Looks good. Thanks.

PR #4167, opened 2021-08-01
Remove unused has_value

nward:
Since bc06a68ef97c654f54f6947125e548e04c5b2709 (in release_3_0_0_beta1!) has_value hasn't been set anywhere (as far as I can see) but it's still checked by some code:

- `src/lib/print.c` checks this in `vp_prints_value_json` (as of db94f6bb261 / #3810)
- `src/modules/rlm_couchbase/mod.c` checks this in code which is "heavily based on" `vp_prints_value_json`

I note that has_value_alias is checked in dict.c in three places, and is also never set (as far as I can see).

This PR is draft at the moment to share my work and get some early feedback in case I'm going about this the wrong way. I say "as far as I can see" a bit here, as I am not certain it's not set by some bitwise operation on an ATTR_FLAGS, or something along those lines.

Open questions:
- If we're not setting *or* checking has_value (and maybe has_value_alias) can they be removed from ATTR_FLAGS entirely?


nward:
OK - have done some work here.. though looking at v3.1.x and v4 (when tracking down VALUE-ALIAS stuff, removed in v3.1.x) I see that has_value is set with FLAG_SET in both v3.1.x and v4... perhaps rather than do all this, I should just be reverting this line: https://github.com/FreeRADIUS/freeradius-server/commit/bc06a68ef97c654f54f6947125e548e04c5b2709#diff-09f96093b9caf7ed788c5bc0d7002766678f853f63ffce95866a9e435af0a63cL982

I'll test this, and see how it goes.

nward:
Yeah - ok. I think doing this by setting has_value is the best approach here, as it keeps things in line with where v4 is. I hope my spam hasn't been annoying as I've gotten to this point.

I've pushed changes to that effect. I've left my "working" so far, but obviously intend to squash this if there's agreement with the "final" approach.

So:
- Updated src/lib/dict.c to set has_value again using effectively the UNCONST define from v4
- Updated src/lib/print.c to have different logic - previously raw_value could cause a raw value to be printed even if has_tag was set - and I've documented the logic
- Updated src/modules/rlm_couchbase/mod.c - add a raw_value param to retain v3.0.x behaviour by default, and update the logic to attempt to print the raw value if has_value is false, and document the logic
- Set raw_value to default to "yes" in rlm_rest to retain v3.0.x behaviour by default

I have tested rlm_rest and it behaves as expected, but I have no way of testing rlm_couchbase.

alandekok:
I'll take a look.  I think part of the confusion is that a `raw` attribute usually means that the underlying attribute is malformed.  i.e. an `ipaddr` type which comes in an `Access-Accept` as having only 3 bytes of data.

So part of the confusion here may be a mis-use of the word `raw`.

For printing, we have to check:

* if the attribute is `raw`, then the data type should be `octets`
* otherwise if the attribute has a tag, then the tag has to be printed
* if the attribute has a value, then the value name also has to be printed

For parsing, we gave to do:

* if it has a tag, try to parse the tag.  If that fails, then it's raw
* if it has a value, then try to look up the value as a name
* otherwise try to parse it as an integer

arr2036:
@nward can you squash some of these commits together to make review easier.

nward:
Hi - thanks @arr2036, yes you're right re. the usage of `raw` in this context. I think it would be reasonable to call it something else if there's a conflict, but I'm just using what was already there. Sorry @alandekok for not replying to you on that - I had thought I had!

I have rebased the commits to make more sense. I'd originally left them verbatim to show my working - though appreciate that it was probably not that important :-)

PR #4190, opened 2021-08-17
Switch to unsigned type for fr_lst_index_t

jejones3141:
This involves adding one to the true index when storing it in an
item, and subtracting one when reading it for use.


arr2036:
Looks like the error is from this PR

jejones3141:
Saw the error thanks to gcc. It's an issue with an error check. Will make sure it works and push.

PR #4203, opened 2021-08-20
Declare capacity and num_elements to have type fr_lst_index_t

jejones3141:
This lets them track changes to the fr_lst_index_t typedef.

While one could theoretically have an LST with capacity
and size equal to one more than the maximum fr_lst_index_t
value, given the typedef, one would run out of RAM long before
that.


arr2036:
The issue is they're not indexes, and so using that type name is somewhat confusing.  The main reason I was arguing for tree specific index types was so that you could easily tell which type of tree was being inserted into.

PR #4204, opened 2021-08-21
Fix unlang compile error message

jpereira:
Print out a helpful message helping the user to know that such method
"foo" does not have any method.

### DIFF
--- a/src/lib/unlang/compile.c
+++ b/src/lib/unlang/compile.c
@@ -3445,10 +3445,13 @@ static unlang_t *compile_module(unlang_t *parent, unlang_compile_t *unlang_ctx,
 			cf_log_err(ci, "The \"%s\" module does not have a '%s %s' method.",
 				   inst->module->name,
 				   unlang_ctx->section_name1, unlang_ctx->section_name2);
-		} else {
+		} else if (unlang_ctx->section_name1) {
 			cf_log_err(ci, "The \"%s\" module does not have a '%s' method.",
 				   inst->module->name,
 				   unlang_ctx->section_name1);
+		} else {
+			cf_log_err(ci, "The \"%s\" module does not have any method. e.g: %s.method",
+				   inst->module->name, inst->module->name);
 		}
 
 		return NULL;


alandekok:
I guess looking at it some more, when does the `unlang_ctx` have no `section_name1`?  I don't think that should ever happen.

arr2036:
Groups could theoretically have that in the future, there's no real need to name them, but I don't know about anything currently.

jpereira:
If we do something like:

```
recv Accounting-Request { sql }
```

We got `unlang_ctx->section_name1 == NULL` and `unlang_ctx->section_name2 == NULL` printing out a `(null)` message.

alandekok:
Except that the SQL module can handle `recv accounting-request`.  and the `unlang_ctx->section_name1` should not be NULL

so why is it NULL?

jpereira:
@alandekok I've tested using `stats` like:

```
recv Accounting-Request {
	stats
}
```

 and we got the same `null`

e.g:

```
Tue Aug 24 19:40:38 2021: Compiling policies in - recv Accounting-Request {...}
Tue Aug 24 19:40:38 2021: ...../set/v4/etc/raddb//sites-enabled/default[1293]: The "stats" module does not have a '(null)' method.
Tue Aug 24 19:40:38 2021: ...../set/v4/etc/raddb//sites-enabled/default[1293]: Invalid keyword "stats".
```

jpereira:
@alandekok I noticed that for example:

```
recv Accounting-Request {
	files
}
```

is working fine because the `Accounting-Request` expects the `methods[MOD_PREACCT](...)` as can be seen in https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_files/rlm_files.c#L679 and we are facing `(null)` when we use _sql_ because it is not being set in https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_sql/rlm_sql.c#L1636 

Otherwise, I am not sure if it is expected or we are missing it.

alandekok:
I've pushed a different fix which prints out which method it's trying to compile.  That error is a little more descriptive.

PR #4209, opened 2021-08-25
Fix Typo with wrong Bracket

Janfred:
Just a small typo.


alandekok:
I've pushed a fix.  If the commit isn't PGP signed, then github won't merge it.

PR #4217, opened 2021-08-31
rlm_sqlcounter Add MySQL query sample based on specific day

jpereira:


jpereira:
good point @ndptech , I've updated the PR.

PR #4224, opened 2021-09-10
v4: Add a status to trunks and watchers on state changes

ndptech:
The initial use case for this is handling LDAP referrals where multiple possible servers can be contacted to chase a referral.  If no existing trunks are open to any of the possible destinations, then multiple trunks can be opened and the first one to become active will trigger the watcher which will send the query.

### DIFF
--- a/src/lib/server/trunk.c
+++ b/src/lib/server/trunk.c
@@ -171,6 +171,18 @@ struct fr_trunk_connection_s {
   	/** @} */
 };
 
+/** An entry in a trunk watch function list
+ *
+ */
+typedef struct fr_trunk_watch_entry_s {
+	fr_dlist_t		entry;			//!< List entry.
+	fr_trunk_watch_t	func;			//!< Function to call when a trunk enters
+							///< the state this list belongs to
+	bool			oneshot;		//!< Remove the function after it's called once.
+	bool			enabled;		//!< Whether the watch entry is enabled.
+	void			*uctx;			//!< User data to pass to the function.
+} fr_trunk_watch_entry_t;
+
 /** Main trunk management handle
  *
  */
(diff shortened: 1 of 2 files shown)


src/lib/server/trunk.c
```diff
@@ -361,6 +361,13 @@ static fr_table_num_indexed_bit_pos_t const fr_trunk_conn_trigger_names[] = {
 };
 static size_t fr_trunk_conn_trigger_names_len = NUM_ELEMENTS(fr_trunk_conn_trigger_names);
 
+static fr_table_num_ordered_t const fr_trunk_states[] = {
```
arr2036, line 380: You used the ordered table type but the states aren't lexicographically ordered.

src/lib/server/trunk.c
```diff
@@ -3882,6 +3891,27 @@ static void trunk_manage(fr_trunk_t *trunk, fr_time_t now)
 		while ((tconn = fr_dlist_head(&trunk->to_free))) talloc_free(fr_dlist_remove(&trunk->to_free, tconn));
 	}
 
+	/*
+	 *	Update the state of the trunk
+	 */
+	if (fr_heap_num_elements(trunk->active)) {
+		new_state = FR_TRUNK_STATE_ACTIVE;
+	} else {
+		new_state = ((fr_dlist_num_elements(&trunk->init) +
```
arr2036, line 4016: There's functions to get this already see `fr_trunk_connection_count_by_state`

src/lib/server/trunk.c
```diff
@@ -3882,6 +3891,27 @@ static void trunk_manage(fr_trunk_t *trunk, fr_time_t now)
 		while ((tconn = fr_dlist_head(&trunk->to_free))) talloc_free(fr_dlist_remove(&trunk->to_free, tconn));
 	}
 
+	/*
+	 *	Update the state of the trunk
+	 */
+	if (fr_heap_num_elements(trunk->active)) {
+		new_state = FR_TRUNK_STATE_ACTIVE;
```
arr2036, line 4014: Best to use `fr_trunk_connection_count_by_state` here too

PR #4225, opened 2021-09-11
add dictionary for mellanox

aaronknister:
add dictionary for mellanox including the vendor-specific local user mapping attribute used by MLNX-OS


jpereira:
@aaronknister Please, add reference for the new  " share/dictionary/radius/dictionary.mellanox" inside the "share/dictionary/radius/dictionary"

PR #4243, opened 2021-09-24
New dictionary for Cambium

terryburton:
Compiled from various online sources and tidied up. Untested, but at least the IANA enterprise number is valid...


arr2036:
Can you do the same PR for v4 (with names corrected)

PR #4248, opened 2021-10-06
v4: Ground work for moving LDAP connections to per thread trunks

ndptech:
This series of commits defines the first round of structures and functions required to move LDAP connections to use trunks rather than the connection pool.

Each trunk will be specific to a host / bind dn combination.

By default, trunks will be created for the configured host / bind dn in the module configuration.  Additional trunks will be created as required either to satisfy LDAP xlats or to chase referrals.

None of this series of commits changes the current behaviour - it is just setting the stage for the next phase of changes.


src/lib/ldap/base.h
```diff
@@ -421,6 +421,21 @@ static inline void fr_ldap_berval_to_value_shallow(fr_value_box_t *value, struct
...
+ * @param[in] one	first connection to compare.
+ * @param[in] two	second connection to compare.
+ * @return CMP(one, two)
+ */
+static inline int8_t fr_ldap_trunk_cmp(void const *one, void const *two)
+{
+	fr_ldap_thread_trunk_t const	*a = one, *b = two;
+	int uricmp = strcmp(a->uri, b->uri);
+
+	if (uricmp == 0) return (strcmp(a->bind_dn, b->bind_dn));
```
arr2036, line 529: strcmp can overflow the return value.  You should `CMP(strcmp(a->bind_db, b->bind_dn), 0);`

arr2036: It's a slightly cleaner flow if you just return if `uricmp != 0`.

src/lib/ldap/base.h
```diff
@@ -389,6 +389,50 @@ typedef struct fr_ldap_thread_trunk_s {
...
+
+	char const		*dn;		//!< Base DN for searches, DN for modifications.
+	char const		**attrs;	//!< Attributes being requested in a search.
+	int			scope;		//!< Search scope.
+	char const		*filter;	//!< Filter for search.
+
+	LDAPControl		*serverctrls[LDAP_MAX_CONTROLS];	//!< Server controls specific to this query.
+	LDAPControl		*clientctrls[LDAP_MAX_CONTROLS];	//!< Client controls specific to this query.
+
+	LDAPMod			**mods;		//!< Changes to be applied if this query is a modification.
```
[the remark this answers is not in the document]
arr2036: Or if it's a pain to change the field references, just make it an anonymous struct.

src/lib/ldap/base.c
```diff
@@ -884,6 +930,50 @@ int fr_ldap_global_config(int debug_level, char const *tls_random_file)
...
+	if (query->referral_urls) ldap_memvfree((void **)query->referral_urls);
+
+	if (query->result) ldap_msgfree(query->result);
+
+	if (query->ldap_url) {
+		/*
+		 *	Queries originating from parsed URLs may have parsed
+		 *	server controls as well
+		 */
+		if ((query->ldap_url->lud_exts) && (query->serverctrls[0])) ldap_controls_free(query->serverctrls);
```
arr2036, line 950: We probably always want to free the client and server controls, not just when query->ldap_url->lud_exts is set?

src/lib/ldap/connection.c
```diff
@@ -549,3 +566,470 @@ int fr_ldap_connection_timeout_reset(fr_ldap_connection_t const *c)
...
+			ROPTIONAL(REDEBUG, ERROR, "Maximum LDAP referral depth (%d) exceeded", t->t->config->referral_depth);
+			query->ret = LDAP_RESULT_EXCESS_REFERRALS;
+			break;
+		}
+
+		/*
+		 *	If we've come here as the result of an existing referral
+		 *	clear the previous list of URLs before getting the next list.
+		 */
+		if (query->referral_urls) ldap_memvfree((void **)query->referral_urls);
```
arr2036, line 893: Probably want to set this to NULL after freeing.

src/lib/ldap/connection.c
```diff
@@ -549,3 +566,470 @@ int fr_ldap_connection_timeout_reset(fr_ldap_connection_t const *c)
...
+static void ldap_request_cancel_mux(fr_trunk_connection_t *tconn, fr_connection_t *conn, UNUSED void *uctx)
+{
+	fr_trunk_request_t	*treq;
+	fr_ldap_connection_t	*ldap_conn = talloc_get_type_abort(conn->h, fr_ldap_connection_t);
+	fr_ldap_query_t		*query;
+
+	while ((fr_trunk_connection_pop_cancellation(&treq, tconn)) == 0) {
+		query = treq->preq;
+		ldap_abandon_ext(ldap_conn->handle, query->msgid, NULL, NULL);
+		fr_rb_remove(ldap_conn->queries, query);
```
arr2036, line 588: Don't we need to free this as well?

src/lib/ldap/connection.c
```diff
@@ -549,3 +566,470 @@ int fr_ldap_connection_timeout_reset(fr_ldap_connection_t const *c)
...
+
+	default:
+		break;
+	}
+
+	find.msgid = ldap_msgid(result);
+	query = fr_rb_find(ldap_conn->queries, &find);
+
+	if (!query) {
+		WARN("Ignoring msgid %i - doesn't match any outstanding queries", find.msgid);
```
arr2036, line 856: `WARN("Ignoring msgid %i - doesn't match any outstanding queries (it may have been cancelled)", find.msgid);`

src/lib/ldap/base.c
ndptech: Sometimes the server controls are set from static data and sometimes parsed from URLs - maybe the solution is a flag for each to indicate they need freeing ..?

arr2036: There's an equivalent function `fr_ldap_control_clear(fr_ldap_connection_t *conn)` for connections.

PR #4308, opened 2021-11-09
Make TEST_CHECK_*() macros display stringified _got

jejones3141:
Makes it clearer what's being compared with the expected value,
though it might take an explicit #define of TEST_MSG_MAXSIZE for
a sufficiently convoluted expression.


jejones3141:
Closing pending independent correction of ALLOC_REQ() conditional compilation to make Mac compiler happy.

Issue #4317, opened 2021-11-21
[defect]: please do not use locate in configure


stintel:
### What type of defect/bug is this?

Unexpected behaviour (obvious or verified by project member)

### How can the issue be reproduced?

Try to build freeradius on a box which contains a backup or NFS export of a root fs of another Linux box. It breaks horribly due to the use of locate in the freeradius configure script. One example: it picks up libexecinfo from the backup, which is not compatible, breaks compilation.

This snippet from configure output made several people go down a wtf rabbit hole for several hours:
```
checking for execinfo.h... no
checking for execinfo.h in /usr/local/include/... no
checking for execinfo.h in /opt/include/... no
checking for execinfo.h in /srv/archive/archive/quagmire/usr/include... no
checking for execinfo.h in /srv/backup/daffy/root/usr/include... yes
checking for backtrace_symbols in -lexecinfo... no
checking for backtrace_symbols in -lexecinfo in /srv/backup/daffy/root/usr/portage/dev-libs... no
checking for backtrace_symbols in -lexecinfo in /srv/backup/daffy/root/usr/portage/dev-libs/libexecinfo... no
checking for backtrace_symbols in -lexecinfo in /srv/backup/daffy/root/usr/portage/dev-libs/libexecinfo/files... no
checking for backtrace_symbols in -lexecinfo in /srv/backup/daffy/root/usr/portage/metadata/md5-cache/dev-libs... no
checking for backtrace_symbols in -lexecinfo in /srv/nfs/rpi3/usr/portage/dev-libs... no
checking for backtrace_symbols in -lexecinfo in /srv/nfs/rpi3/usr/portage/dev-libs/libexecinfo... no
checking for backtrace_symbols in -lexecinfo in /srv/nfs/rpi3/usr/portage/dev-libs/libexecinfo/files... no
checking for backtrace_symbols in -lexecinfo in /srv/nfs/rpi3/usr/portage/metadata/md5-cache/dev-libs... no
checking for backtrace_symbols in -lexecinfo in /usr/local/lib... no
checking for backtrace_symbols in -lexecinfo in /opt/lib... no
```

This later causes a compile failure:
```
stijn@ministore /tmp/freeradius-server-3.0.21 $ make 
CC jlibtool.c
HEADER src/include/attributes.h
HEADER src/include/features.h
HEADER src/include/missing.h
HEADER src/include/tls.h
HEADER src/include/rfc2865.h
HEADER src/include/rfc2866.h
[...]

PR #4318, opened 2021-11-21
Remove FR_LOCATE_DIR macro

floppym:
Calling "locate" to find include and library paths makes no sense. The
locate database may contain paths comprised of files meant for entirely
different toolchains.

Fixes: https://github.com/FreeRADIUS/freeradius-server/issues/4317

### DIFF
--- a/acinclude.m4
+++ b/acinclude.m4
@@ -38,73 +38,6 @@ fi
 ])
 
 dnl #
-dnl #  Locate the directory in which a particular file is found.
-dnl #
-dnl #  Usage: FR_LOCATE_DIR(MYSQLLIB_DIR, libmysqlclient.a)
-dnl #
-dnl #    Defines the variable MYSQLLIB_DIR to be the directory(s) in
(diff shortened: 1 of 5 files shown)


thesamesam:
Note that ``locate`` may also find files which do not exist on the system (removed since the last DB update, usually done daily or weekly) or fail to find new files. It's really intended for users to find files rather than within build systems or automated processes which expect to find them in standard-ish locations.

We already have pkg-config which is being used + some additional logic in the build system within FreeRADIUS to go looking for libs and headers, so ``locate`` should not be needed at all.

stintel:
Tested-by: Stijn Tintel <stijn@linux-ipv6.be>

floppym:
@alandekok The code still exists in the master branch. Is there some reason for that?

alandekok:
@floppym We haven't gotten to it yet.

Plus, the "master" branch has not been released as v4, and is not yet supported.

floppym:
Thanks, just wanted to make sure it wasn't missed by mistake.

PR #4321, opened 2021-11-25
v4: Convert %{xlat: } to new xlat api

ndptech:


arr2036:
Didn't we come up with a fix for needing to mess with the dlist headers?

arr2036:
I think you can just call the xlat tokenizer directly, and calling tmpl_aexpand breaks async evaluation.  You need to tokenize the inputs as an ephemeral xlat, then push the result onto the unlang stack for evaluation

arr2036:
Calling xlat_frame_eval isn't enough because it doesn't allow yielding

PR #4322, opened 2021-12-01
Revert "Scheduled fuzzing: Merge and push pack the corpus (#4313)"

terryburton:
Pushback of the corpus fails due to a protection rule on the master branch that
limits commits to a particular group within the FreeRADIUS organisation. (The
GH Actions "bot" cannot be added to this group.)

So we revert to reinstate scheduled fuzzing pending some alternative approach.

This reverts commit 4512200851b4d0d7131ff7314259f8d9394764db.


arr2036:
I've removed the branch protection that explicitly lists a team.  I suspect this'll still fail because the commits aren't signed.

arr2036:
We have a nwkrad bot account, do you want to try and use that?

terryburton:
Commits pushed from GH Actions using the GH API are treated as verified, so without the team restriction it should now work. (The team restriction was the only part I didn't model when testing.)

The current approach is nice because the update_file API just tacks a commit (containing a single file modification) on to the head of the branch, which is what we need in this case.

The more protracted approach (that I haven't tested) is to create GPG keys for a bot account and store a signing key as a repo secret. After the fuzzer updates the corpus, perform a pull -ff on the checked out repo (to collect recent commits), add the updated corpus, and create a new signed commit using the stored key material.

PR #4326, opened 2021-12-03
v4: Initial sweep to remove legacy xlat code

ndptech:


arr2036:
Arg, I have pending changes for this already.  The difficulty comes in actually removing the legacy interpreter.  Still working on the fixes for that.

PR #4328, opened 2021-12-07
Scheduled fuzzing: Cache merged corpus; push back only after 30 days

terryburton:
This reduces the corpus push back from nightly to monthly. To maintain nightly continuity each merged corpus is instead stored in the GH Actions "cache".

A limitation is that a cache entry must be hit within 3 days in order to be kept. This shouldn't be a problem since the cache is downloaded early in the run so that it is accessed each night even when the builds / unit tests fail.

### DIFF
--- a/.github/workflows/ci-scheduled-fuzzing.yml
+++ b/.github/workflows/ci-scheduled-fuzzing.yml
@@ -151,6 +151,29 @@ jobs:
     - name: Git LFS pull
       run: git lfs pull
 
+    #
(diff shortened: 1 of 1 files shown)


arr2036:
Do we still need all the LFS cache steps now?

terryburton:
> Do we still need all the LFS cache steps now?

Update removes the initial lfs checkout since this has only ever pulled in the radsniff test pcap artefact (from cache), which is unused by this fuzzer build.

PR #4330, opened 2021-12-09
Don't hide other homeservers when one is removed

terryburton:
With dynamic_homeservers the "bynumber" lookup becomes gappy due to deletions,
so we should not stop at such gaps when listing all homeservers.


mcnewton:
Merged manually to 3.0 & 3.2

PR #4333, opened 2021-12-12
FreeBSD CI is now fast enough to run on push/pull...

terryburton:
... but whether it's reliable enough remains to be seen!

(Previously rsync could occasionally time out.)


terryburton:
Re-based so that the build will succeed.

PR #4342, opened 2021-12-18
ssl_mysql_arg use int type instead bool

czy-1:
I used freeradius 3.0.25 with mysql-community-devel 8.0.27.
I set the tls_required = yes but it not work.
I see the mysql.h MYSQL_OPT_SSL_MODE should be a enum.


alandekok:
fixed in commit f8083917.  Which also fixes it to use the correct enum name, both where the variable is declared, and later in the code.

PR #4346, opened 2021-12-22
fix compile error

czy-1:
I see the commit [f808391](https://github.com/FreeRADIUS/freeradius-server/commit/f8083917aec91b17b9d72e46dbd5883114270cc5).
but in mysql.h, the enum name should be SSL_MODE_PREFERRED not the SL_MODE_PREFERRED, I think this is a spelling mistake.
But in different versions of mysql-devel, some versions don't have enum mysql_ssl_mode,  such as mariadb-devel.x86_64 1:5.5.68, does it cause compilation failure in these versions and do we need to care about this.


alandekok:
I've pushed fixes, thanks.

PR #4347, opened 2021-12-22
CI: Build and use OpenSSL 3.0.1 for LIBS_ALT build

terryburton:
To stop the build from failing until the port to the OpenSSL 3 API is complete, we include an easy to revert commit that passes in `-Wno-error=deprecated-declarations`.

However, the build is still failing at this time due to the HMAC tests failing. This failure is identical to the MacOS build failure which seems to have began when MacOS switched from OpenSSL 3.0.0 to 3.0.1.


alandekok:
Merged manually, thanks.

PR #4355, opened 2022-01-07
udpfromto: Use plain sendto if our source address is 0.0.0.0 or ::/0

terryburton:


alandekok:
pulled manually, thanks.

PR #4363, opened 2022-01-19
Fix eapol_test against OpenSSL 3.x

jpereira:


alandekok:
Merged as individual commits.

PR #4378, opened 2022-01-26
tls: close opened dir to avoid resource leak

antoniotorresm:


alandekok:
Thanks.  I've pushed a fix where it closes on normal exit, too.

We require that the commits be signed, so I can't use the GitHub merge process.

PR #4384, opened 2022-01-28
Fix SQL/db2 build

jpereira:


arr2036:
I believe those are actually pointers? Why are you swapping `NULL` for `0`?

jpereira:
The `stmt` are `SQLHANDLE`. then, If you take a look at `/usr/local/db2/include/sqlcli.h`, the `SQLHANDLE` is `typedef  SQLINTEGER SQLHANDLE;` (At least, building without any extra -DFOO definition) so, that patch fixes the errors/warnings during the build.

PR #4387, opened 2022-02-01
Split DHCPv4 dictionaries to pair attributes/values with their RFCs

jejones3141:


share/dictionary/dhcpv4/dictionary.rfc5986
```diff
@@ -0,0 +1,13 @@
...
+# Version $Id$
+##############################################################################
+#
+#  RFC5986 - Discovering the Local Location Information Server (LIS)
+#
+#  $Id$
+#
+##############################################################################
+
+ATTRIBUTE	Access-Network-Domain-Name		213	string # really one dns_label
```
arr2036, line 13: I think we can do DNS labels now?

arr2036:
Is the intent to do an organisation pass and a separate fixup pass?

share/dictionary/dhcpv4/dictionary.rfc5986
alandekok: The DHCPv4 encoder / decoder doesn't yet support `dns_label`.  The relevant code has to be back-ported from DHCPv6.

PR #4388, opened 2022-02-01
If the SSL session isn't complete after handshake, and there aren't more writes to do, then there is more data on the socket to be read

aren:


alandekok:
I'll pull it over manually, as we require commits to be signed.

Thanks.

PR #4391, opened 2022-02-02
ci-rpm: As Centos8 is EOL, let's use the Rocky

jpereira:


alandekok:
Isn't this already in commit 1e0293473 ?

jpereira:
@alandekok no, it completes the commit `1e02934`.

jpereira:
Merged manually https://github.com/FreeRADIUS/freeradius-server/commit/09ca2c700b42d8fde4b1861e66a479ddcdd1cf39

PR #4394, opened 2022-02-03
ci-rpm: As Centos8 is EOL, let's use the Rocky

jpereira:


mcnewton:
Not sure we want to get rid of CentOS 8 - plenty of people will still be using it, so it needs to be tested. But we should add Rocky 8 so there's testing for RHEL8 equivalent.

PR #4402, opened 2022-02-23
rlm_mschap: Avoid memory leak

jpereira:


src/modules/rlm_mschap/rlm_mschap.c
```diff
@@ -1007,23 +1007,25 @@ static int CC_HINT(nonnull (1, 2, 4, 5)) do_mschap_cpw(rlm_mschap_t *inst,
 
 			ctx = EVP_CIPHER_CTX_new();
 			if (!ctx) {
+			error:
 				REDEBUG("Failed getting RC4 from OpenSSL");
+				if (ctx) EVP_CIPHER_CTX_free(ctx);
```
arr2036, line 1012: So now it prints a spurious error line for every other error?  Why isn't the error goto under EVP_CIPHER_CTX_set_key_length?

alandekok:
I’ll push a simple fix tomorrow.

PR #4404, opened 2022-03-03
Don't refer to errno when the pthread library fails

miztake:
When the pthread library fails, errno is referenced even though errno is not set.
Fix to refer to the return code of the pthread library instead of errno.


alandekok:
Can you sign the commits?  Otherwise it will be difficult for us to merge them.

arr2036:
I wish there was a GitHub option to require signed commits before allowing people to open a PR *sigh*

arr2036:
Closed via https://github.com/FreeRADIUS/freeradius-server/commit/6d443117d3f0799613ca4e8965c222424aa9bc85

PR #4411, opened 2022-03-10
test: Remove listen{} detail

jpereira:
It only pollutes the logs. We don't need it.

### DIFF
--- a/src/tests/config/test.conf
+++ b/src/tests/config/test.conf
@@ -45,12 +45,6 @@ policy {
 #
 server test {
 	listen {
-		type = detail
-		filename = ${radacctdir}/detail
-		load_factor = 10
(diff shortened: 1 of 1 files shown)


jpereira:
It's boring to troubleshoot on debug mode with tons of 

``` 
Wed Mar  9 21:13:00 2022 : Debug: detail (/Users/jpereira/Devel/FreeRADIUS/freeradius-server-v3.0.x.git/src/tests/detail): Polling for detail file
Wed Mar  9 21:13:00 2022 : Debug: detail (/Users/jpereira/Devel/FreeRADIUS/freeradius-server-v3.0.x.git/src/tests/detail): Detail listener state unopened waiting 1.082667 sec
```

PR #4414, opened 2022-03-14
ci-rpm: Fix packages.endpointdev.com URL

jpereira:
It was renamed as described in https://packages.endpointdev.com/

```
(Note that this service previously operated at packages.endpoint.com and moved to packages.endpointdev.com on 2021-11-30.)
```


jpereira:
That patch fixes such error found in https://github.com/FreeRADIUS/freeradius-server/runs/5529552227?check_suite_focus=true#step:4:8

PR #4419, opened 2022-03-17
Add attributes for Ruckus SmartCell Insight

TigerHuang:
Ref: https://docs.commscope.com/bundle/sci-5.5-user-guide/page/GUID-58C0D969-59B8-4B3A-BF39-B095171B48CE.html


jpereira:
Hi @TigerHuang 

it looks we have a typo. the FreeRADIUS expects `string`, not `strin`.

PR #4428, opened 2022-03-21
v4: Rework talloc_array_null_terminate()

ndptech:
talloc_realloc_fn() amends a talloc'd array in such a way that talloc_array_length() is then unable to return the length of the array.

This rework gets round that by creating a new array.

### DIFF
--- a/src/lib/util/talloc.c
+++ b/src/lib/util/talloc.c
@@ -717,13 +717,15 @@ void **talloc_array_null_terminate(void **array)
 	size_t		len;
 	TALLOC_CTX	*ctx;
 	void		**new;
+	size_t		size;
 
 	if (!array) return NULL;
 
 	len = talloc_array_length(array);
 	ctx = talloc_parent(array);
+	size = talloc_get_size(array) / talloc_array_length(array);
 
-	new = talloc_realloc_fn(ctx, array, len + 1);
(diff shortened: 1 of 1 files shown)


src/lib/util/talloc.c
```diff
@@ -723,9 +723,17 @@ void **talloc_array_null_terminate(void **array)
 	len = talloc_array_length(array);
 	ctx = talloc_parent(array);
 
-	new = talloc_realloc_fn(ctx, array, len + 1);
+	new = talloc_array(ctx, void *, len + 1);
```
arr2036, line 726: Can you try this with `void *_talloc_realloc_array(const void *ctx, void *ptr, size_t el_size, unsigned count, talloc_get_name(ptr));`

src/lib/util/talloc.c
```diff
@@ -723,7 +724,7 @@ void **talloc_array_null_terminate(void **array)
 	len = talloc_array_length(array);
 	ctx = talloc_parent(array);
 
-	new = talloc_realloc_fn(ctx, array, len + 1);
+	new = _talloc_realloc_array(ctx, array, size, len + 1, talloc_get_name(array));
```
arr2036, line 727: size = `talloc_get_size() / talloc_array_length()`

PR #4436, opened 2022-03-22
eap: Fix OpenSSL3 integration

jpereira:
The latest OpenSSL 3 always send the 'SSL3_RT_HEADER' and
'SSL3_RT_INNER_CONTENT_TYPE' over the cbtls_msg().
As we don't use it for anything, it's fine just ignore it.


alandekok:
This is essentially the same patch which broke Windows 11. So we'll have to try something else.

PR #4439, opened 2022-03-23
v4: Remove ifdefs for longstanding OpenLDAP features

ndptech:
Since only OpenLDAP's libldap is supported, the `ifdef`s for features which have been in libldap for years can be removed.


arr2036:
Session tracking checks need to stay

PR #4440, opened 2022-03-23
Change Client-FQDN to be a struct

jejones3141:
It is split out into its own dictionary and has tests.


jejones3141:
Struct and tests updated.

share/dictionary/dhcpv4/dictionary.rfc4072
```diff
@@ -0,0 +1,19 @@
+# -*- text -*-
+# Copyright (C) 2022 The FreeRADIUS Server project and contributors
+# This work is licensed under CC-BY version 4.0 https://creativecommons.org/licenses/by/4.0
+# Version $Id$
+##############################################################################
+#
+#  RFC4072 - The Dynamic Host Configuration Protocol (DHCP) Client
```
alandekok, line 7: It's RFC 4702.  RFC 4072 is "Diameter extensions for EAP"

jejones3141:
Corrections to files (both names and content) made and pushed to the branch.

alandekok:
I merged the changes manually.  We'd prefer to do this as one commit instead of a bunch of smaller ones.  It keeps the git history a bit more readable.

PR #4444, opened 2022-03-29
freeradius-snmp "proxy": Various fixes, improvements and cleanup

terryburton:
Accelerate walks with a "next OID" hash.
Enable collection of per-client stats by default.
Populate accounting client table with real data.
Avoid race by hoisting radius_stats_init() out of updater thread.
Have radius_updater thread terminate on SIGINT.
Refresh RADIUS-AUTH-SERVER-MIB and RADIUS-ACC-SERVER-MIBs.
Add IPv6 client support.
Assume default value (0) for non-existent stats attributes.
Whitespace and style.


alandekok:
IPv6 MIBs are defined in https://datatracker.ietf.org/doc/html/rfc4669

terryburton:
> IPv6 MIBs are defined in https://datatracker.ietf.org/doc/html/rfc4669

Thanks.

Force pushed. Now with IPv6 client support.

PR #4455, opened 2022-04-08
v4: Implement LDAP persistent search handling

ndptech:
Allows a virtual server to receive requests as entries change in an LDAP directory which could be used to trigger CoA or disconnect packets.

Three variants of persistent search implementation are supported:
- RFC 4533
- Active Directory
- https://tools.ietf.org/id/draft-ietf-ldapext-psearch-03.txt


src/lib/ldap/sync.h
```diff
@@ -0,0 +1,45 @@
...
+ * @file lib/ldap/sync.h
+ *
+ * @brief Common definitions required by both network and worker for LDAP sync
+ *
+ * @copyright 2022 Network RADIUS SARL (legal@networkradius.com)
+ */
+
+/** Types of the internal packets for processing LDAP sync messages
+ */
+typedef enum {
```
alandekok, line 29: These should be available automatically (not as an enum, tho) as `src/include/protocol/ldap/freeradius.internal.h`

there are some admittedly horrific Perl scripts which convert the dictionaries to header files.

src/listen/ldap_sync/active_directory.c
```diff
@@ -27,6 +27,297 @@
...
+ * @param[in] ctx	to allocate nodes in.
+ * @param[out] root	of tree of parsed nodes.
+ * @param[in] filter	to parse.
+ * @param[in] attrs	specified in the sync config.  Used to warn user if the filter
+ *			references attributes not being requested.
+ * @return
+ *	- 0 on success
+ *	< 0 on failure
+ */
+static int ldap_filter_parse(TALLOC_CTX *ctx, ldap_filter_t **root, char const *filter, char const **attrs)
```
alandekok, line 295: If this is generic to LDAP and not to AD, it might be better to put this code into `src/lib/ldap`, so that other things can use it.

But that's a minor nit.

src/lib/ldap/sync.h
arr2036: Yeah best solution is to use the attribute macros as the values then they're in sync but we still have an enum for debugging.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+	ldap_filter_t		*child_node;
+
+	/*
+	 *	Firstly, look for the characters which indicate the start of a group of filters
+	 *	to be combined with a logical operator.
+	 */
+	switch(fr_sbuff_current(sbuff)[0]) {
+	case '&':
+		node->logic_op = LDAP_FILTER_LOGIC_AND;
+		node->orig = talloc_asprintf(node, "&");
```
arr2036, line 96: `talloc_typed_strdup`

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+		goto sync_filter_new_group;
+
+	case '|':
+		node->logic_op = LDAP_FILTER_LOGIC_OR;
+		node->orig = talloc_asprintf(node, "|");
```
arr2036, line 101: `talloc_typed_strdup`

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+		goto sync_filter_new_group;
+
+	case '!':
+		node->logic_op = LDAP_FILTER_LOGIC_NOT;
+		node->orig = talloc_asprintf(node, "!");
```
arr2036, line 106: `talloc_typed_strdup`

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+	sync_filter_new_group:
+		if ((fr_sbuff_remaining(sbuff) < 2) || (fr_sbuff_current(sbuff)[1] != '(')) {
```
arr2036, line 108: Use `fr_sbuff_extend` instead of remaining.   Add an `fr_sbuff_peek(_sbuff, offset)` which returns '\0' if the access is out of bounds if you need an offset.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+static int ldap_filter_parse_node(ldap_filter_t *node, fr_sbuff_t *sbuff, int depth, char const **attrs)
+{
+	fr_sbuff_marker_t	filter_marker, attr_s_marker, attr_e_marker, value_marker;
+	ldap_filter_t		*child_node;
+
+	/*
+	 *	Firstly, look for the characters which indicate the start of a group of filters
+	 *	to be combined with a logical operator.
+	 */
+	switch(fr_sbuff_current(sbuff)[0]) {
```
arr2036, line 93: Just deref with `*`, `[0]` encourages people reading the code to access offsets of current which is unsafe.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+			return -1;
+		}
+
+		DEBUG3("%*sCreating LDAP filter group %s", depth, "", node->orig);
+		node->filter_type = LDAP_FILTER_GROUP;
+		child_node = node->child = talloc_zero(node, ldap_filter_t);
+		depth += 2;
+		fr_sbuff_advance(sbuff, 2);
+		if (ldap_filter_parse_node(child_node, sbuff, depth, attrs) < 0) return -1;
+		while ((fr_sbuff_current(sbuff)[0] == '(') && (fr_sbuff_remaining(sbuff) > 1)) {
```
arr2036, line 119: Just use `fr_sbuff_extend` or `fr_sbuff_extend_lowat` in all places where `fr_sbuff_remaining` is used.  I might remove `fr_sbuff_remaining`, it's too easy to use it instead of the extend functions...

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+			case LDAP_FILTER_OP_PRESENT:
+			case LDAP_FILTER_OP_SUBSTR:
+			case LDAP_FILTER_OP_UNSET:
+				fr_assert(0);
+				break;
+			}
+			/*
+			 *	Take a copy of the original filter for debug output
+			 */
+			node->orig = talloc_zero_array(node, char, fr_sbuff_diff(sbuff, &filter_marker) + 1);
```
arr2036, line 243: `MEM()`

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+
+			fr_sbuff_marker(&attr_e_marker, sbuff);
+			fr_sbuff_advance(sbuff, 1);
+			fr_sbuff_marker(&ext_marker, sbuff);
+			if ((fr_sbuff_adv_until(sbuff, SIZE_MAX, &filter_terms, '\0') == 0) ||
+			    (fr_sbuff_current(sbuff)[0] != ':')) {
+				fr_strerror_const("Extensible match filters need two \":\" surrounding the rule OID");
+				return -1;
+			}
+			if (strncmp(fr_sbuff_current(&ext_marker), LDAP_MATCHING_RULE_BIT_AND,
```
arr2036, line 152: `fr_sbuff_adv_past_str_literal`?

arr2036: Or if you really need the length constraint `fr_sbuff_adv_past_str`.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+				    fr_sbuff_diff(sbuff, &ext_marker)) == 0) {
+				node->op = LDAP_FILTER_OP_BIT_AND;
+				break;
+			}
+			if (strncmp(fr_sbuff_current(&ext_marker), LDAP_MATCHING_RULE_BIT_OR,
```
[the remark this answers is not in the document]
arr2036: Or if you really need the length constraint `fr_sbuff_adv_past_str`.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+				fr_strerror_const("Missing value after \"=\"");
+				return -1;
+			}
+
+			/*
+			 *	If we've not already set where the end of the attribute is, set it now
+			 */
+			if (fr_sbuff_diff(&attr_s_marker, &attr_e_marker) == 0) fr_sbuff_marker(&attr_e_marker, sbuff);
+
+			node->attr = talloc_zero_array(node, char, fr_sbuff_diff(&attr_e_marker, &attr_s_marker) + 1);
```
arr2036, line 177: MEM()

src/listen/ldap_sync/active_directory.c
ndptech: Isn't `fr_sbuff_extend` about extending buffers you are writing to?  I'm after seeing how close we are to the end of a buffer I'm reading from

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+
+			/*
+			 *	Since we don't have the LDAP schema, we make an assumption that <=, >= and
+			 *	bitwise operators are going to be used with numeric attributes
+			 */
+			case LDAP_FILTER_OP_GE:
+			case LDAP_FILTER_OP_LE:
+			case LDAP_FILTER_OP_BIT_AND:
+			case LDAP_FILTER_OP_BIT_OR:
+				fr_value_box_from_str(node, node->value, FR_TYPE_UINT32, NULL,
```
arr2036, line 226: Check error condition here.  There's tons of reasons the parsing operations can fail and we definitely want to report them.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+			    (node->op == LDAP_FILTER_OP_EQ)) {
+				node->op = LDAP_FILTER_OP_PRESENT;
+				break;
+			}
+
+			node->value = fr_value_box_alloc_null(node);
+
+			switch (node->op) {
+			case LDAP_FILTER_OP_EQ:
+				fr_value_box_bstrndup(node, node->value, NULL, fr_sbuff_current(&value_marker),
```
arr2036, line 209: Technically return code should be checked here

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+ * @param[in,out] node	to populate with parsed filter
+ * @param[in] sbuff	pointing to filter to parse
+ * @param[in] depth	to indent debug output, indicating nesting of groups
+ *
+ * @return
+ *  - 0 on success
+ *  - < 0 on error
+ *  - 1 if the final advance of the sbuff moved forward.
+ */
+static int ldap_filter_parse_node(ldap_filter_t *node, fr_sbuff_t *sbuff, int depth, char const **attrs)
```
arr2036, line 84: This should be returning `fr_slen_t` and providing positive and negative parse offsets so that we can prettyprint an error marker for the user.

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+		node->orig = talloc_asprintf(node, "!");
+	sync_filter_new_group:
+		if ((fr_sbuff_remaining(sbuff) < 2) || (fr_sbuff_current(sbuff)[1] != '(')) {
+			fr_strerror_const("Logical operators must be followed by '('");
+			return -1;
+		}
+
+		DEBUG3("%*sCreating LDAP filter group %s", depth, "", node->orig);
+		node->filter_type = LDAP_FILTER_GROUP;
+		child_node = node->child = talloc_zero(node, ldap_filter_t);
```
arr2036, line 115: MEM()

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+ *	< 0 on failure
+ */
+static int ldap_filter_parse(TALLOC_CTX *ctx, ldap_filter_t **root, char const *filter, char const **attrs)
+{
+	fr_sbuff_t	sbuff;
+
+	/*
+	 *	Filters are always wrapped in ()
+	 */
+	if ((talloc_array_length(filter) < 4) || (filter[0] != '(') ||
```
arr2036, line 302: Just try to parse the filter an error out where it no longer matches the expected grammar

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,684 @@
...
+		/*
+		 *	LDAP filters only use one wildcard character '*' for zero or more
+		 *	character matches.
+		 */
+		case LDAP_FILTER_OP_SUBSTR:
+		{
+			char const	*v, *t, *v_end, *t_end;
+			bool		skip;
+
+			t_end = node->value->vb_strvalue + node->value->length - 1;
```
arr2036, line 517: Why is one being subtracted here?  Usually in parsing end is set to the char after the last char in the buffer, and we check for `p < end`

src/listen/ldap_sync/persistent_search.c
```diff
@@ -0,0 +1,289 @@
...
+	ber = ber_init(&ctrls[i]->ldctl_value);
+	if (!ber) {
+		ERROR("Failed allocating ber to handle syncStateValue control");
+
+	free_ber:
+		if (ber) ber_free(ber, 1);
+		goto error;
+	}
+
+	if (ber_scanf(ber, "{e", &change_type) == LBER_ERROR) {
```
arr2036, line 220: Can you add some inline docs for the format of the control

src/listen/ldap_sync/proto_ldap_sync.c
```diff
@@ -0,0 +1,359 @@
...
+		talloc_free(dl_mod_inst);
+		return -1;
+	}
+	*((dl_module_inst_t **)out) = dl_mod_inst;
+	return 0;
+}
+
+/** Decode an internal LDAP sync packet
+ */
+static int mod_decode (UNUSED void const *instance, request_t *request, uint8_t *const data, size_t data_len)
```
arr2036, line 111: whitespace

src/listen/ldap_sync/proto_ldap_sync.c
```diff
@@ -0,0 +1,359 @@
...
+		if (request->async->packet_ctx) {
+			sync_packet = talloc_get_type_abort(request->async->packet_ctx, sync_entry_packet_t);
+		}
+		break;
+
+	case FR_LDAP_SYNC_CODE_COOKIE_LOAD:
+	case FR_LDAP_SYNC_CODE_COOKIE_STORE:
+		break;
+	default:
+		return -1;
```
arr2036, line 143: Could this be an assert?

src/listen/ldap_sync/proto_ldap_sync.c
```diff
@@ -0,0 +1,359 @@
...
+	 */
+	pair_append_request(&vp, attr_ldap_sync_packet_id);
+	fr_value_box_from_network(vp, &vp->data, vp->da->type, NULL, &dbuff, sizeof(uint32_t), false);
+	request->packet->id = vp->data.vb_uint32;
+	request->reply->id = vp->data.vb_uint32;
+
+	/*
+	 *	Extract attributes from the passed data
+	 */
+	while (fr_dbuff_remaining(&dbuff)) {
```
arr2036, line 157: Use extend here too

src/listen/ldap_sync/proto_ldap_sync.c
```diff
@@ -0,0 +1,359 @@
...
+
+static int mod_open(void *instance, fr_schedule_t *sc, UNUSED CONF_SECTION *conf)
+{
+	proto_ldap_sync_t	*inst = talloc_get_type_abort(instance, proto_ldap_sync_t);
+	fr_listen_t		*li;
+
+	/*
+	 *	Build the #fr_listen_t.
+	 */
+	li = talloc_zero(inst, fr_listen_t);
```
arr2036, line 233: MEM

src/listen/ldap_sync/active_directory.c
arr2036: It works for both depending on what the underlying sbuff is.  The `fr_sbuff_t` can be initialised with different extend functions.  Where we're using it for printing the extend function can realloc the underlying buffer for more space, or potentially write data in the buffer to a socket, and shift everything to the left.  When we're using it for parsing, we're usually shifting parsed stuff to the left, and copying in more data at the end of the buffer from a `FILE *` or another data source.

The point of markers isn't really to delimit sections, you use recursive function calls for that.  Markers are used if you might need to refer back to text earlier in the buffer, or write to a position earlier in the buffer.  They're used to stop the data you're referring to being shifted out of the buffer.

src/listen/ldap_sync/active_directory.c
arr2036: For explanation, the `asprintf` functions in general are horribly slow, I'm also not sure they set the type to `char`, which is what it should be for a string buffer in talloc.   I think in general the talloc string manipulation functions don't set the type correctly so we have `_typed_` wrapper functions internally which do.

raddb/sites-available/ldap_sync
```diff
@@ -27,194 +102,391 @@
...
-#			realm = 'example.org'
-		}
+				#
+				#  Specify the list of attributes to return in entries,
+				#  one config item per LDAP attribute.
+				#
+				#  If no attr config items are specified, then all attributes
+				#  will be returned.
+				#
+				attrs = 'cn'
```
arr2036, line 180: These should probably be singular not plural, don't know who did that initially.

raddb/sites-available/ldap_sync
```diff
@@ -27,194 +102,391 @@
...
+#				attrs = 'isDeleted'
+#				attrs = 'sAMAccountName'
+#				attrs = 'userAccountControl'
+#				attrs = 'lastKnownParent'
+#
+#				update {
+#					&User-Name := 'sAMAccountName'
+#					&Tmp-Integer-0 := 'userAccountControl'
+#					&Tmp-String-0 := 'lastKnownParent'
+#				}
```
arr2036, line 294: Is there a reason to have the attributes manually specified, or could we just take them from the RHS of the update section?

raddb/sites-available/ldap_sync
```diff
@@ -27,194 +102,391 @@
...
+	#  For other directories, this will be called prior to the search query being
+	#  sent to the server, so could be used for any initial setup of cache datastores.
+	#
 	#  A request will be generated with the following attributes:
 	#
-	#  - &request.LDAP-Sync-DN		the base_dn of the sync.
-	#  - &request.LDAP-Sync-Filter		the filter of the sync (optional).
-	#  - &request.LDAP-Sync-Scope		the scope of the sync (optional).
-	#  - &request.LDAP-Sync-attr		the attributes returned by the sync (optional).
+	#  - &request.LDAP-Sync.DN		the base_dn of the sync.
```
arr2036, line 314: As these attributes creates in a TLV, or was this a search and replace when moving the LDAP attributes into a protocol specific dictionary?

arr2036: Looks like a TLV.  That's fine.

src/lib/ldap/base.h
```diff
@@ -483,6 +487,57 @@ typedef struct {
...
+			ldap_filter_logic_t	logic_op;	//!< Logical operator for this group
+			ldap_filter_t		*child;		//!< First child of this group
+		};
+		struct {
+			char			*attr;		//!< Attribute for the filter node
+			ldap_filter_op_t	op;		//!< Operator to be used for comparison
+			fr_value_box_t		*value;		//!< Value to compare with
+		};
+	};
+	ldap_filter_t		*next;			//!< Next node in a group
```
arr2036, line 537: I know this is a pain, but we're trying to move everything to dlists.  Could you setup a type wrapper for dlists and use those instead?

src/lib/ldap/filter.c
```diff
@@ -0,0 +1,565 @@
...
+
+	MEM(*root = talloc_zero(ctx, ldap_filter_t));
+
+	ret = ldap_filter_parse_node(*root, &sbuff, 0, attrs);
+	if (ret < 0) {
+		char	*spaces, *text;
+
+		fr_canonicalize_error(ctx, &spaces, &text, ret + 1, filter);
+		ERROR("Invalid LDAP filter: %s", text);
+		ERROR("                     %s^ - %s", spaces, fr_strerror());
```
arr2036, line 346: Use `EMARKER`

arr2036: Generally for parsing it ends up being more flexible to have the caller pass in the sbuff, and return an `fr_slen_t`, then the caller can use the appropriate logging function.  In this case it's probably better to use one of the cf_log_err functions so we get info about the config item containing the error.

src/listen/ldap_sync/proto_ldap_sync.c
arr2036: between mod_decode and parens

src/lib/ldap/filter.c
```diff
@@ -0,0 +1,579 @@
...
+	/*
+	 *	If we're at the very top level we should be at the end
+	 *	of the buffer
+	 */
+	if ((depth == 0) && (fr_sbuff_extend(sbuff))) {
+		fr_strerror_const("Extra characters at the end of LDAP filter");
+		return fr_sbuff_error(sbuff);
+	}
+
+	return 0;
```
arr2036, line 323: This should return the number of bytes parsed

src/lib/ldap/filter.c
```diff
@@ -0,0 +1,579 @@
...
+	fr_dlist_insert_head(*root, node);
+
+	ret = ldap_filter_parse_node(node, filter, 0, attr_check, uctx);
+	if (ret < 0) {
+		talloc_free(*root);
+		*root = NULL;
+		return ret;
+	}
+
+	return 0;
```
arr2036, line 356: Should return number of bytes parsed

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+
+	return;
+}
+
+/** Dummy mod_open for child listener
+ *
+ * Openning of the real conneciton is done before this is called, but a mod_open
+ * is required in order to register the mod_read and mod_write.
+ */
+static int proto_ldap_child_mod_open(UNUSED fr_listen_t *li)
```
arr2036, line 533: Can you fix the I/O code to check for the presence of this callback and remove the dummy code

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+		     fr_table_str_by_value(sync_ldap_msg_table, type, "<invalid>"));
+	free_msg:
+		if (ctrls) ldap_controls_free(ctrls);
+		ldap_msgfree(msg);
+		return ret;
+	}
+
+	find.msgid = msgid;
+
+	sync = fr_rb_find(tree, &find);
```
arr2036, line 625:
```
&(sync_state_t){ .msgid = msgid }
```

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+		 */
+		fr_assert (packet_id <= talloc_array_length(inst->sync_config));
+
+		/*
+		 *	Look for the returned cookie.
+		 */
+		if (fr_dbuff_remaining(&dbuff)) {
+			FR_DBUFF_OUT_RETURN(&attr, &dbuff);
+			if (attr == attr_ldap_sync_cookie->attr) {
+
```
arr2036, line 746: WS

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	/*
+	 *	Verify that the LDAP server configuration is valid, either
+	 *	distinct server and port or an LDAP url.
+	 */
+	fr_assert(inst->handle_config.server_str[0]);
+
+	server = inst->handle_config.server_str[0];
+
+	if (ldap_is_ldap_url(server)) {
```
arr2036, line 1133: All this URL check code should be moved to a common library function and removed from rlm_ldap too

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	LDAPMessage			*result;
+
+	size_t				i;
+
+	/*
+	 *	Timeout has to be > 0 msec, otherwise libldap always returns LDAP_TIMEOUT
+	 *	Very short timeouts have also been seen to return LDAP_TIMEOUT even though
+	 *	there is data to read, depending on computer speed.  10ms appears to be safe.
+	 */
+	status = fr_ldap_result(&result, NULL, ldap_conn, dir_ctx->msgid, LDAP_MSG_ALL, NULL, fr_time_delta_from_msec(10));
```
arr2036, line 839: I don't really understand what's happening here.  Why is there a fixed timeout?  Won't open_Read be called on an I/O event?

src/listen/ldap_sync/active_directory.c
```diff
@@ -0,0 +1,204 @@
...
+	ctrl2.ldctl_iscritical = 1;
+	ctrls[1] = &ctrl2;
+	ctrl2.ldctl_iscritical = 1;
+
+	/*
+	 *	Active Directory only takes the filter (objectClass=*) when the
+	 *	notification control is used.  To compensate for this we parse the
+	 *	filter and implement our own basic version of a filter.
+	 */
+	if (strcasecmp(sync->config->filter, "(objectClass=*)") != 0) {
```
arr2036, line 104: A slightly more robust way of doing this might be to use the filter parser and check out the parsed nodes, that'd account for whitespace variations.

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+
+#include "proto_ldap_sync_ldap.h"
+#include "rfc4533.h"
+#include "persistent_search.h"
+#include "active_directory.h"
+
+extern fr_app_io_t proto_ldap_sync_ldap;
+extern fr_app_io_t proto_ldap_sync_child;
+
+static CONF_PARSER sasl_mech_static[] = {
```
arr2036, line 41: The majority of this config stuff is identical to the configuration for rlm_ldap, could you move it into a common library and then have it work on offsets of `fr_ldap_config_t` then strip it out of rlm_ldap as well.

arr2036: This one can just operate on `fr_ldap_sasl_t`

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+
+	{ FR_CONF_OFFSET("scope", FR_TYPE_STRING, sync_config_t, scope_str), .dflt = "sub" },
+	/* For persistent search directories, setting this to "no" will load the whole directory. */
+	{ FR_CONF_OFFSET("changes_only", FR_TYPE_BOOL, sync_config_t, changes_only), .dflt = "yes" },
+
+	CONF_PARSER_TERMINATOR
+};
+
+static CONF_PARSER option_config[] = {
+	{ FR_CONF_OFFSET("idle", FR_TYPE_TIME_DELTA, proto_ldap_sync_ldap_t, handle_config.keepalive_idle), .dflt = "60" },
```
arr2036, line 64: This can operate on fr_ldap_config_t

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	/* timeout on network activity */
+	{ FR_CONF_OFFSET("net_timeout", FR_TYPE_TIME_DELTA, proto_ldap_sync_ldap_t, handle_config.net_timeout), .dflt = "10" },
+	/* timeout for reconnections */
+	{ FR_CONF_OFFSET("reconnection_delay", FR_TYPE_TIME_DELTA, proto_ldap_sync_ldap_t, handle_config.reconnection_delay), .dflt = "5" },
+
+	CONF_PARSER_TERMINATOR
+};
+
+static CONF_PARSER tls_config[] = {
+	{ FR_CONF_OFFSET("ca_file", FR_TYPE_FILE_INPUT, proto_ldap_sync_ldap_t, handle_config.tls_ca_file) },
```
arr2036, line 84: fr_ldap_config_t again

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	{ FR_CONF_OFFSET("ca_path", FR_TYPE_FILE_INPUT, proto_ldap_sync_ldap_t, handle_config.tls_ca_path) },
+	{ FR_CONF_OFFSET("certificate_file", FR_TYPE_FILE_INPUT, proto_ldap_sync_ldap_t, handle_config.tls_certificate_file) },
+	{ FR_CONF_OFFSET("private_key_file", FR_TYPE_FILE_INPUT, proto_ldap_sync_ldap_t, handle_config.tls_private_key_file) },
+	{ FR_CONF_OFFSET("start_tls", FR_TYPE_BOOL, proto_ldap_sync_ldap_t, handle_config.start_tls), .dflt = "no" },
+	{ FR_CONF_OFFSET("require_cert", FR_TYPE_STRING, proto_ldap_sync_ldap_t, handle_config.tls_require_cert_str) },
+
+	CONF_PARSER_TERMINATOR
+};
+
+static const CONF_PARSER global_config[] = {
```
arr2036, line 94: These are global

arr2036: Not really sure the offsets here are right.  It's probably only working because `handle_config` appears first in `proto_ldap_sync_ldap_t`?

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	 *	Areas of the DIT to listen on
+	 */
+	{ FR_CONF_SUBSECTION_ALLOC("sync", FR_TYPE_SUBSECTION | FR_TYPE_MULTI | FR_TYPE_REQUIRED, proto_ldap_sync_ldap_t, sync_config, ldap_sync_search_config) },
+
+	/*
+	 *	Extra configuration items
+	 */
+	{ FR_CONF_POINTER("options", FR_TYPE_SUBSECTION, NULL), .subcs = (void const *) option_config },
+	{ FR_CONF_POINTER("global", FR_TYPE_SUBSECTION, NULL), .subcs = (void const *) global_config },
+	{ FR_CONF_OFFSET("tls", FR_TYPE_SUBSECTION, proto_ldap_sync_ldap_t, handle_config), .subcs = (void const *) tls_config },
```
arr2036, line 121: See our base address is handle_config inside `proto_ldap_sync_ldap_t` but the subcs is looking at offsets from `proto_ldap_sync_ldap_t`

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+};
+
+static CONF_PARSER const proto_ldap_sync_ldap_config[] = {
+	/*
+	 *	LDAP server definition
+	 */
+	{ FR_CONF_OFFSET("server", FR_TYPE_STRING | FR_TYPE_MULTI | FR_TYPE_REQUIRED, proto_ldap_sync_ldap_t, handle_config.server_str) },
+	{ FR_CONF_OFFSET("port", FR_TYPE_UINT16, proto_ldap_sync_ldap_t, handle_config.port) },
+	{ FR_CONF_OFFSET("identity", FR_TYPE_STRING, proto_ldap_sync_ldap_t, handle_config.admin_identity) },
+	{ FR_CONF_OFFSET("password", FR_TYPE_STRING | FR_TYPE_SECRET, proto_ldap_sync_ldap_t, handle_config.admin_password) },
```
arr2036, line 108: These could be moved into common code as a preprocessor macro taking the structure and the field within the structure for the `fr_ldap_config_t`

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+ * This allows individual syncs to be stopped without destroying the underlying connection.
+ *
+ * Removes the sync's msgid from the tree of msgids associated with the connection.
+ *
+ * @param[in] sync to abandon.
+ * @return 0
+ */
+static int sync_state_free(sync_state_t *sync)
+{
+
```
arr2036, line 202: WS

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	DEBUG3("Abandoning sync");
+
+	if (!sync->conn->handle) return 0;	/* Handled already closed? */
+
+	/*
+	 *	Tell the remote server to stop sending results
+	 */
+	if (sync->msgid >= 0) ldap_abandon_ext(sync->conn->handle, sync->msgid, NULL, NULL);
+	fr_rb_delete(tree, &find);
```
arr2036, line 215: `&(sync_state_t){  .msgid = sync->msgid }`

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+
+	return 0;
+}
+
+/** Serialise values from sync config for sending to worker.
+ *
+ * @param[out] dbuff	to append attributes to
+ * @param[in] config	to read values from
+ */
+static int serialise_sync_config_values(fr_dbuff_t *dbuff, sync_config_t const *config)
```
arr2036, line 260: It'd make more sense here to use the internal encoder and decoder, then the pairs can just pop into existence when they're decoded on the other side?

arr2036: It makes it less work for us to maintain because we only need to update the serialization function

arr2036: I envisaged similar issues with the async redis code, which was the reason for writing the internal encoder/decoder

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+		 *	For Cookie Load packets, use the sync array index as the packet ID.
+		 */
+		if (fr_dbuff_in(&dbuff, (uint32_t)i) < 0) goto dbuff_free;
+
+		if (serialise_sync_config_values(&dbuff, config) < 0) goto dbuff_free;
+
+		/*
+		 *	If the root DN has been identified, populate the attribute.
+		 */
+		if (config->root_dn) {
```
arr2036, line 946: Yeah, all of this should just be adding pairs to a temporary pair list, and then calling the internal encoder

arr2036: You can stick them all in thread local structure if you want to avoid the overhead of all the allocs

src/listen/ldap_sync/proto_ldap_sync_ldap.c
```diff
@@ -0,0 +1,1325 @@
...
+	CONF_SECTION		*sync_cs;
+	sync_config_t		*sync_conf;
+	char const		*server;
+	char			*url;
+	size_t			i, len;
+	CONF_SECTION		*map_cs;
+	map_t const		*map;
+	tmpl_rules_t		parse_rules = {
+		.attr = {
+			.allow_foreign = true	/* Because we don't know where we'll be called */
```
arr2036, line 1121: We shouldn't allow foreign attributes as the listener should be bound to a virtual server.  There's functions to access the virtual server dictionary in virtual_server.c, you just need to figure out the CONFIG_SECTION.  I think other code might walk back up the config tree looking for a server section... there's functions for that in cf_util.c

src/listen/ldap_sync/proto_ldap_sync_ldap.h
```diff
@@ -0,0 +1,125 @@
...
+extern fr_table_num_sorted_t const sync_op_table[];
+extern size_t sync_op_table_len;
+
+int8_t sync_state_cmp(void const *one, void const *two);
+
+sync_state_t *sync_state_alloc(TALLOC_CTX *ctx, fr_ldap_connection_t *conn, sync_config_t const *config);
+
+int ldap_sync_cookie_store(sync_state_t *sync, uint8_t const *cookie, bool refresh);
+
+int ldap_sync_entry_send(sync_state_t *sync, uint8_t const uuid[SYNC_UUID_LENGTH], struct berval *orig_dn,
```
arr2036, line 122: `uint8_t const uuid[static SYNC_UUID_LENGTH]` to actually attempt to enforce the requirement

src/listen/ldap_sync/active_directory.c
arr2036: Fair enough 👍🏻

src/listen/ldap_sync/proto_ldap_sync_ldap.h
ndptech: Only RFC4533 servers provide a UUID, so for the others we have to be able to pass NULL in here.

src/listen/ldap_sync/proto_ldap_sync_ldap.c
ndptech: The elements it refers to are in `proto_ldap_sync_ldap_t` - it's the same as in rlm_ldap where the elements exist in `rlm_ldap_t`.  They are not in `handle_config`

src/listen/ldap_sync/proto_ldap_sync_ldap.c
ndptech: I think when this was originally coded I didn't have `res_timeout` being set correctly.  If the timeout that `ldap_result()` is called with is too short then it returns a TIMEOUT even though there is data available - maybe due to the time it is taking to parse what's come in on the wire.
Having got `res_timeout` plumbed in correctly, we can call things with zero there and it's reliable.

src/lib/ldap/conf.h
```diff
@@ -0,0 +1,111 @@
...
+ * $Id$
+ * @file lib/ldap/conf.h
+ * @brief Conf parser sections common to rlm_ldap and ldap sync
+ *
+ * @copyright 2022 The FreeRADIUS Server Project.
+ */
+
+#include <freeradius-devel/ldap/base.h>
+
+static CONF_PARSER fr_ldap_sasl_mech_static[] = {
```
arr2036, line 12: Can you still this all in conf.c and then add `extern` declarations in conf.h, that way we don't have the bytecode being duplicated for every use.

src/lib/ldap/conf.h
```diff
@@ -0,0 +1,111 @@
...
+
+	{ FR_CONF_OFFSET("reconnection_delay", FR_TYPE_TIME_DELTA, fr_ldap_config_t, reconnection_delay), .dflt = "10" },
+
+	CONF_PARSER_TERMINATOR
+};
+
+/*
+ *  Macros for including common LDAP configuration items
+ */
+#define FR_LDAP_GLOBAL_CONF(_conf) static const CONF_PARSER global_config[] = { \
```
arr2036, line 99: Same here... We really need to fix the global stuff so it's only allowed to occur in one place, and outside of the module instance configuration.

PR #4458, opened 2022-04-13
Make POSIXification in exec_pair_to_env() notice lower case first char

jejones3141:
The loop as written always incremented p after the test to deal with
a leading digit, so it wouldn't notice a leading lower case letter.

### DIFF
--- a/src/lib/server/exec.c
+++ b/src/lib/server/exec.c
@@ -154,8 +154,8 @@ static int exec_pair_to_env(char **env_p, size_t env_len, fr_sbuff_t *env_sbuff,
 		 *	for the first char.
 		 */
 		p = fr_sbuff_current(&env_m[i]);
-		if (isdigit((int)*p)) *p = '_';
-		while (p++ < fr_sbuff_current(&sbuff)) {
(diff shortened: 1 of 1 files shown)


src/lib/server/exec.c
```diff
@@ -154,8 +154,8 @@ static int exec_pair_to_env(char **env_p, size_t env_len, fr_sbuff_t *env_sbuff,
 		 *	for the first char.
 		 */
 		p = fr_sbuff_current(&env_m[i]);
-		if (isdigit((int)*p)) *p = '_';
-		while (p++ < fr_sbuff_current(&sbuff)) {
+		if (isdigit((int)*p)) *p++ = '_';
```
alandekok, line 157: I'm also not clear why this line is here.  If we're changing all of the characters, why make the first one special?

This could probably be just:

```
for (p = fr_sbuff_current(&env_m[i]);
        p < fr_sbuff_current(&sbuff);
        p++ {
   ... check the characters ...
}

jejones3141: First one is special because it alone can't be a digit.

PR #4459, opened 2022-04-15
Update dictionary.h3c on public manuals 

rightone:
Update dictionary.h3c on public manuals


mcnewton:
Merged manually, thanks.

PR #4460, opened 2022-04-19
v3.2.x: alpine image does not build

Ejdesgaard:
alpine/Dockerfile: Changed git release var from v3.0.x to 3.2.x
alpine/Dockerfile: Changed alpine tagged to 3.13. It's the last tag that provides libcouchbase-dev-2.x


mcnewton:
Useful, thanks!

PR #4462, opened 2022-04-22
Get rid of superfluous test in loop in command_encode_dns_label(). CID #1455228

jejones3141:
By virtue of CC_HINT(nonnull), in, and hence p, will be non-NULL
the first time through the loop. To take another iteration, next
must be non-NULL, so that p, set to next + 1, will also be non-NULL;
hence the "while (p)" is effectively "while (true)".


jejones3141:
Apparently my reasoning is flawed.

PR #4464, opened 2022-04-23
alpine/Dockerfile: chmod +x docker-entrypoint.sh, or permission denied.

Ejdesgaard:
When chmod -x is fixed, then the next issue is:
standard_init_linux.go:228: exec user process caused: exec format error

Starting the container with sh and manually executing sh ./docker-entrypoint.sh works, but I have not looked into why.


mcnewton:
Thanks.

You've done changes to 3.2.x, but a PR against master, so this is impossible to read as there are so many incorrect commits. Pull Requests need to be made against the correct branch.

I've fixed this up and pushed changes for v3.2.x and v3.0.x.

The entrypoint mode is normally copied from the host where it should be set there correctly, but adding an extra chmod won't hurt just in case. However if you're still seeing other errors after fixing it up on your system then maybe you have other permissions problems. The v3 dockerfiles work elsewhere without problem.

PR #4469, opened 2022-04-25
Fix some potential dead-locks in src/lib/server/pool.c

jpereira:


alandekok:
The `connection_find()` function is supposed to return with the mutex held.  I'll push a fix for `fr_pool_free()`

PR #4490, opened 2022-05-04
Remove short-cut check for FR_TYPE_ in fr_value_box_from_substr() (CID #1504016)

jejones3141:
dst_type == FR_TYPE_NULL is caught and handled  at the start
of the function, so by the switch on dst_type, it will never
equal FR_TYPE_NULL.


arr2036:
This needs a little bit more investigation

arr2036:
The assert needs removing, not this case

arr2036:
Basically we need to be able to print a NULL type to a string, and perform the inverse parse operation.

jejones3141:
OK. Thanks.

PR #4501, opened 2022-05-10
Fix typo in 90fde2339e

jpereira:


alandekok:
I'll revert the commit instead.  We shouldn't be adding features to 3.0.

PR #4503, opened 2022-05-11
Use fr_fast_rand() to select operation in burn-in (CID #1503935, #150…

jejones3141:
…4015)


arr2036:
There's a bunch of new warnings about uninitialised rand ctxs in covertiy

jejones3141:
Yeah. Looking into it now.

jejones3141:
Looks like it's getting faked out by done_init, which as far as I can tell
is pointless, so I'm pulling it so it always sets the fields in rand_ctx.
Testing it now.

PR #4509, opened 2022-05-12
Catch invalid status values in tls_cache_session_ticket_app_data_get …

jejones3141:
…(CID #1503904)

The switch didn't deal with all defined status values. We add them and
leave the default always present to make explicit how they should be
treated and to catch invalid values... and to assure the reader and
coverity that we will never pass a NULL request pointer to
tls_cache_app_data_get().

### DIFF
--- a/src/lib/tls/cache.c
+++ b/src/lib/tls/cache.c
@@ -1265,6 +1265,9 @@ static SSL_TICKET_RETURN tls_cache_session_ticket_app_data_get(SSL *ssl, SSL_SES
 	switch (status) {
 	case SSL_TICKET_EMPTY:
 	case SSL_TICKET_NO_DECRYPT:
+	case SSL_TICKET_FATAL_ERR_MALLOC:
+	case SSL_TICKET_FATAL_ERR_OTHER:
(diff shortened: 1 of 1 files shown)


src/lib/tls/cache.c
```diff
@@ -1265,9 +1265,10 @@ static SSL_TICKET_RETURN tls_cache_session_ticket_app_data_get(SSL *ssl, SSL_SES
 	switch (status) {
 	case SSL_TICKET_EMPTY:
 	case SSL_TICKET_NO_DECRYPT:
-#ifdef __clang_analyzer__
+	case SSL_TICKET_FATAL_ERR_MALLOC:
+	case SSL_TICKET_FATAL_ERR_OTHER:
+	case SSL_TICKET_NONE:
 	default:
```
arr2036, line 1271: The idef should be retained around default, or default should be omitted.  That tells us when new error codes are added that we might need to deal with.

jejones3141: Makes sense; so be it. That is liable to mean having to think of another way to placate coverity, so I will make the change and consider what to do about that.

PR #4511, opened 2022-05-13
Use OpenSSL 3.0 library context to access MD4 and MD5

antoniotorresm:
Enabling the legacy provider in the global context is not enough if the
system is in FIPS mode. Instead, load providers in local context for both MD5 and
MD4.

Remove EVP_MD_CTX_FLAG_NON_FIPS_ALLOW flag since it does not have any
effect anymore.

### DIFF
--- a/src/include/md4.h
+++ b/src/include/md4.h
@@ -40,7 +40,7 @@ extern "C" {
 
 #ifndef HAVE_OPENSSL_MD4_H
 /*
- * The MD5 code used here and in md4.c was originally retrieved from:
+ * The MD4 code used here and in md4.c was originally retrieved from:
  *   http://www.openbsd.org/cgi-bin/cvsweb/src/include/md4.h?rev=1.12
  *
  * This code implements the MD4 message-digest algorithm.
@@ -83,6 +83,7 @@ USES_APPLE_DEPRECATED_API
 #  define fr_md4_destroy(_x)
 #else
 #include <openssl/evp.h>
+#include <openssl/provider.h>
 
 /*
  *	Wrappers for OpenSSL3, so we don't have to butcher the rest of
@@ -90,18 +91,28 @@ USES_APPLE_DEPRECATED_API
  */
 typedef struct FR_MD4_CTX {
 	EVP_MD_CTX	*ctx;
-	EVP_MD const   	*md;
+	EVP_MD const  	*md;
 	unsigned int	len;
+
+	OSSL_PROVIDER	*default_provider;
+	OSSL_PROVIDER	*legacy_provider;
+	OSSL_LIB_CTX	*libctx;
 } FR_MD4_CTX;
 
 static inline void fr_md4_init(FR_MD4_CTX *ctx)
 {
 	ctx->ctx = EVP_MD_CTX_new();
-//	ctx->md = EVP_MD_fetch(NULL, "MD4", "provider=legacy");
-	ctx->md = EVP_md4();
+
+	if (EVP_default_properties_is_fips_enabled(NULL)) {
+		ctx->libctx = OSSL_LIB_CTX_new();
+		ctx->legacy_provider = OSSL_PROVIDER_load(ctx->libctx, "legacy");
+		ctx->md = EVP_MD_fetch(ctx->libctx, "MD4", NULL);
+	} else {
+		ctx->md = EVP_md4();
+	}
+
 	ctx->len = MD4_DIGEST_LENGTH;
 
-	EVP_MD_CTX_set_flags(ctx->ctx, EVP_MD_CTX_FLAG_NON_FIPS_ALLOW);
 	EVP_DigestInit_ex(ctx->ctx, ctx->md, NULL);
 }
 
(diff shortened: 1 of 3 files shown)


alandekok:
The code is called from multiple threads, so it's likely to crash when the providers are loaded / unloaded for each MD4 / MD5 call.

The providers should instead be put into the `FR_MD4_CTX` structures, where they are completely local.

alandekok:
Please try this patch.  It passes my preliminary tests.

I've moved the providers into the `FR_MD4_CTX`, where they can't affect anything else.  I've also made the load / delete of the providers dependent on FIPS mode.

MD4 / MD5 is used in the server a *lot*.  So loading / unloading the providers on every operation has huge performance impacts.  An alternative is just to use the `md4.c` and `md5.c` files included with the server.

[patch.gz](https://github.com/FreeRADIUS/freeradius-server/files/8692941/patch.gz)

antoniotorresm:
Hi Alan, thanks for your patch, it seems to work fine.

I've updated the PR with the changes from your patch, and also:

1) Removed EVP_MD_CTX_FLAG_NON_FIPS_ALLOW usage, since it doesn't work in OpenSSL 3.0: https://github.com/openssl/openssl/blob/openssl-3.0.3/include/openssl/evp.h#L208-L209
2) Instead of loading both default and legacy providers for both MD4 and MD5, only load default provider for MD5 and legacy provider for MD4.

You are right about the performance impact, though. Probably moving the providers to a global state would be better, just as it's done in TLS: https://github.com/FreeRADIUS/freeradius-server/blob/3935d3c985b10f24397ab3614df285724c470673/src/lib/tls/base.c#L62
We can't reuse the providers from TLS since we need to setup a library context to bypass FIPS limitations. Where do you think we could move this global library context?

Can we use the macros at: https://github.com/FreeRADIUS/freeradius-server/blob/06e4a82dacf7ec806d96e094baac9c3946b07622/src/include/threads.h#L45
to make this global library context thread safe?

arr2036:
Is there really a need to load these into a specific library context?  I mean sure, we should be using a library context for everything in FreeRADIUS to separate our configuration of libssl from 3rd party libraries, but that's not a change we're going to be making for v3, so the majority of this code seems superfluous.  Just load the provider into the default context, and be done....

antoniotorresm:
The issue is that when the system is configured to be in FIPS mode, some algorithms are blocked from the global default context (including MD5). To bypass this limitation, we need to use a standalone context.

alandekok:
The new patch makes a bit more sense:

* create a local libctx
* load the provider into that
* load the MD4/MD5 EVP from the libctx, which now can use the provider.

TBH, this is slow as heck.  Plus, we sometimes need to copy the `MD5_CTX`, and I'm not entirely sure that copy will work properly if the `MD5_CTX` has a bunch of other things in it.

The simpler solution would be to just drop / delete the checks for `<openssl/md4.h>`, and use our own code in `src/lib/md4.c`.  The code has been there for 20+ years, and it works.  It also avoids any kind of OpenSSL weirdness.

We won't get any kind of CPU / SSL engine optimizations for MD4/MD5 functions.  But I don't think that matters for the tiny amount of data which we're hashing.

antoniotorresm:
I agree with you, just using the internal MD4 and MD5 is probably the simpler solution. I've prepared a patch that drops all the checks and directly uses the internal implementations.

[internal_md5_md4.zip](https://github.com/FreeRADIUS/freeradius-server/files/8707282/internal_md5_md4.zip)

I've tested and this works fine under a FIPS environment. However, there are two `EVP_md5()`  left at:

https://github.com/FreeRADIUS/freeradius-server/blob/06e4a82dacf7ec806d96e094baac9c3946b07622/src/modules/rlm_eap/libeap/mppe_keys.c#L105

and:

https://github.com/FreeRADIUS/freeradius-server/blob/06e4a82dacf7ec806d96e094baac9c3946b07622/src/modules/rlm_otp/otp_radstate.c#L121

that I'm unsure how to adapt to use internal MD5. If this code is not frequently executed, I guess we can load the providers if adapting to internal MD5 is not an option.

What do you think? Thanks!

alandekok:
I'll take a look at the patch in a bit.

For `src/modules/rlm_eap/libeap/mppe_keys.c`, just load the `EVP_MD` dynamically.  That code is only used for EAP-FAST, and then only if TLS 1.0 or 1.1 is used.  Everything else uses `SSL_export_keying_material()`, which doesn't have this problem.  See `eaptls_gen_mppe_keys()`.

alandekok:
and for `rlm_otp`, just disable it.  We've removed it in 3.2.0.  Because so far as we can tell, no one has used it in ~8 years.

antoniotorresm:
Here's the updated patch:

[internalmd4md5.zip](https://github.com/FreeRADIUS/freeradius-server/files/8728815/internalmd4md5.zip)

I've removed all `EVP_MD_CTX_FLAG_NON_FIPS_ALLOW` left, as well as load MD5 dynamically in `mppe_keys.c` if FIPS mode is enabled. If this version is fine then I will close this PR and open a new one so we can get a CI run.

alandekok:
The supplied patch makes it use the internal MD4 / MD5 functions always, for all systems.  I would strongly prefer that to be configurable.

I've pushed commit 947d5d6b, which adds a flag `configure --enable-fips-workaround`.  When that flag is passed to the `configure` stage, then it chooses the internal MD4/MD5 functions.  If the flag isn't passed, then the OpenSSL ones are used.

This workaround isn't perfect.  If someone does a build without `--enable-fips-workaround`, and then enables FIPS mode, the server won't work.   The solution is to recompile.  :(

The `master` branch has run-time detection of FIPS, and it switches between using the internal functions and the OpenSSL ones "on the fly".  However, I'm also careful of the "stable" nature of v3, and want to do the minimum possible code changes.

I've also pushed commit  bf758fabeed6b, which implements the dynamic loading of the `EVP_MD()` functions.  This is only used for EAP-FAST, so any performance issues are acceptable for now.

The longer-term alternative is to just move to thread-local `OSSL_LIB_CTX` which loads the MD4 / MD5 functions.  But for now let's see if these changes work.

PR #4520, opened 2022-05-18
Check return values in rlm_date.c (CID #1503947, #1503971)

jejones3141:


arr2036:
Just use `MEM()` like everywhere else

jejones3141:
Will do.

PR #4524, opened 2022-05-20
Feature/eap cache

jrouzierinverse:
Add the ability to cache the EAP state using the cache module.


alandekok:
Thanks.  It seems pretty good.  I'll take a look and review it over the next few days.

Is there a way to add unit tests for this functionality?  That may be difficult, tho.

alandekok:
I've done some cleanups and pushed the changes to https://github.com/alandekok/freeradius-server/tree/feature/eap-cache

* we're particular about formatting.  All of the code should look the same
* I've protected the feature via WITH_EAP_CACHE, so that normal builds will work without changes
* the module and virtual server have been renamed.  I don't want to confuse EAP session caching with this caching.

Please let me know if this works.

jrouzierinverse:
This is perfect I will start adding other eap types soon.

alandekok:
Please rebase your changes on top of the branch I pushed to my public repository.  I understand it's easier to just keep committing on top of your local repository.  However, that makes it difficult to merge the changes back in.

I would very much prefer to not have lots of commits which are "do this", followed by a later commit of "whoops, let's do it a different way".  Those are developmental changes which can (and should) be cleaned up before a PR is sent.

jrouzierinverse:
Will do.

alandekok:
In the end, OpenSSL cannot be made to save / restore "in progress" TLS connections.

PR #4525, opened 2022-05-21
rlm_expr: Fix typo with double '&'

jpereira:


jpereira:
Just to avoid double `&`

e.g:

```
Fri May 20 22:49:28 2022: WARN  : (0)          Can't find &&Acct-Delay-Time.  Using 0 as operand value
```

PR #4530, opened 2022-05-23
ssl: Fix minor compile warn about const/discards qualifiers.

jpereira:


jpereira:
It fixes the https://github.com/FreeRADIUS/freeradius-server/runs/6530665115?check_suite_focus=true#step:17:1532

PR #4532, opened 2022-05-24
Deal with unchecked return value Coverity reports

jejones3141:
This deals with the following CIDs: 1503943, 1503949,
1455233, 1503981

### DIFF
--- a/src/lib/eap_aka_sim/state_machine.c
+++ b/src/lib/eap_aka_sim/state_machine.c
@@ -722,7 +722,7 @@ RESUME(store_pseudonym)
 
 			fr_value_box_bstrdup_buffer_shallow(NULL, &vp->data, NULL, identity, false);
 		}
-		pair_update_request(&new, attr_session_id);
+		MEM(pair_update_request(&new, attr_session_id) >= 0);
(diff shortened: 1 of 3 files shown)


src/lib/eap_aka_sim/state_machine.c
```diff
@@ -722,7 +722,7 @@ RESUME(store_pseudonym)
 
 			fr_value_box_bstrdup_buffer_shallow(NULL, &vp->data, NULL, identity, false);
 		}
-		pair_update_request(&new, attr_session_id);
+		MEM(pair_update_request(&new, attr_session_id) >= 0);
```
jpereira, line 725: @arr2036 maybe we could use the _Coccinelle_ to detect the use of `pair_update_request()` without such `MEM()`. what do you think @arr2036 ?

jejones3141: Looking more closely at main_config.c, it ought to return -1 if the `fr_value_box_bstr_realloc()` fails in `xlat_config_escape()`. I will make that change.

arr2036: Yeah for these functions we could add an assert.  We'd need to go through and remove all the MEM() wrappers around the calls though.  If you want to look at it @jpereira go ahead.

PR #4540, opened 2022-05-29
New optional lightweight Acct-On/Off strategy with helper SP and view

terryburton:
Currently supported by MySQL and PostgreSQL.


raddb/mods-config/sql/main/mysql/process-radacct.sql
```diff
@@ -150,3 +150,130 @@ BEGIN
...
+    DECLARE v_last_report DATETIME DEFAULT 0;
+    DECLARE v_last BOOLEAN DEFAULT FALSE;
+    DECLARE v_batch_size INT(12);
+
+    --
+    --  This works for many circumstances
+    --
+    SET v_batch_size = 2500;
+
+    SELECT MIN(radacctid) INTO v_a FROM radacct;
```
ndptech, line 237: Would it be worth this being `SELECT MIN(radacctid) INTO v_a FROM radacct WHERE acctstoptime IS NULL` so that the routine steps over a much smaller range of rows in the case when there is significant amount of history in radacct?

raddb/mods-config/sql/main/sqlite/process-radacct-close-after_reload.pl
```diff
@@ -0,0 +1,100 @@
...
+                    acctterminatecause
+                END
+            FROM nasreload WHERE nasipaddress = radacct.nasipaddress
+        )
+    WHERE
+        radacctid BETWEEN ? AND ?
+        AND acctstoptime IS NULL
+EOF
+
+my $sth = $dbh->prepare('SELECT MIN(radacctid), MAX(radacctid) FROM radacct');
```
ndptech, line 61: Add `WHERE acctstoptime IS NULL` ?

raddb/mods-config/sql/main/mysql/process-radacct.sql
terryburton: Yes.

raddb/mods-config/sql/main/sqlite/process-radacct-close-after_reload.pl
terryburton: Done.

alandekok:
I think it looks good.  Oracle, etc. can be done when someone with an Oracle license supplies a patch. :)

terryburton:
@alandekok Thanks. I've forced-pushed a squashed version for merging - no changes.

PR #4541, opened 2022-05-30
v4: Framework for configuring, initialising and freeing libraries

ndptech:
e.g. libldap where there are global options which should be set once even though the library may be used by more than one FreeRADIUS module

### DIFF
--- a/src/lib/server/global_lib.c
+++ b/src/lib/server/global_lib.c
@@ -1,0 +1,233 @@
+/*
+ *  This program is free software; you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
+ *  the Free Software Foundation; either version 2 of the License, or
+ *  (at your option) any later version.
+ *
+ *  This program is distributed in the hope that it will be useful,
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/*
+ * $Id$
+ *
+ * @file global_lib.c
+ * @brief Handle global configuration, initialisation and freeing for libraries
+ *
+ * @copyright 2022 The FreeRADIUS server project
+ */
+#include <freeradius-devel/server/global_lib.h>
+#include <freeradius-devel/server/main_config.h>
+#include <freeradius-devel/util/atexit.h>
+
+/*
+ *  Terminator for array of global_lib_autoinit_t
+ */
+global_lib_autoinit_t const lib_autoinit_terminator = { .name = NULL };
+
+/*
+ *  Global list of libraries
+ */
+typedef struct {
+	fr_rb_tree_t		libs;
+} global_lib_list_t;
+
+static global_lib_list_t *lib_list;
+
+/** Structure to track use of libraries.
+ *
+ */
+typedef struct {
+	fr_rb_node_t			entry;			//!<  Entry in tree of libraries
+	global_lib_autoinit_t const	*autoinit;		//!<  Autoinit structure used to manage this library
+	uint32_t			instance_count;		//!<  Number of current uses of this library
+	bool				initialised;		//!<  Has the init callback been run for this library
+} global_lib_inst_t;
+
+/** Parse the global config section for a library and call its init function
+ *
+ * @param[in] lib	to configure and initialise
+ * @return
+ *	- 0 on success
+ *	- -1 on failure
+ */
+static int lib_init_call(global_lib_inst_t *lib)
+{
+	CONF_SECTION	*global_cs, *cs;
+
+	/*
+	 *	Find relevant config section
+	 *	If the section does not exist allocate an empty one
+	 *	so default rules can be evaluated.
+	 */
+	global_cs = cf_section_find(cf_root(main_config->root_cs), "global", NULL);
+	if (!global_cs) global_cs = cf_section_alloc(main_config->root_cs, main_config->root_cs, "global", NULL);
+
+	cs = cf_section_find(global_cs, lib->autoinit->name, NULL);
+	if (!cs) cs = cf_section_alloc(global_cs, global_cs, lib->autoinit->name, NULL);
+
+	if ((cf_section_rules_push(cs, lib->autoinit->config)) < 0 ||
+	    (cf_section_parse(lib, lib->autoinit->inst, cs) < 0)) {
+		cf_log_err(cs, "Failed evaluating configuration for libldap");
+		return -1;
+	}
+
+	/*
+	 *  Call the init callback if defined
+	 */
+	if (lib->autoinit->init && (lib->autoinit->init()) < 0) return -1;
+
+	lib->initialised = true;
+
+	return 0;
+}
+
+/** Instantiate a list of libraries
+ *
+ * @param to_init	Array of autoinit structures detailing libraries to initialise
+ * @return
+ *	- 0 on succcess
+ *	- -1 on failure
+ */
+static int lib_auto_instantiate(global_lib_autoinit_t * const *to_init)
+{
+	global_lib_autoinit_t * const *p;
+
+	for (p = to_init; *p != &lib_autoinit_terminator; p++) {
+		global_lib_inst_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(global_lib_inst_t){ .autoinit = *p });
+
+		/*
+		 *  If the library is already initialised, just increase the reference count
+		 */
+		if ((lib) && (lib->initialised)) {
+			lib->instance_count++;
+			return 0;
+		}
+
+		if (!lib) {
+			MEM(lib = talloc_zero(lib_list, global_lib_inst_t));
+			lib->autoinit = *p;
+			fr_rb_insert(&lib_list->libs, lib);
+		}
+		lib->instance_count++;
+
+		/*
+		 *  If the main config parsing is not complete we can't initialise the library yet
+		 */
+		if (!main_config->root_cs) continue;
+
+		DEBUG2("Instantiating %s", lib->autoinit->name);
+		if (lib_init_call(lib) < 0) return -1;
+	}
+
+	return 0;
+}
+
+/** Callback for creation of "lib" symbols
+ *
+ */
+int global_lib_auto_instantiate(UNUSED dl_t const *module, void *symbol, UNUSED void *user_ctx)
+{
+	if (lib_auto_instantiate((global_lib_autoinit_t **)symbol) < 0) return -1;
+
+	return 0;
+}
+
+/** Run free callbacks for external libraries no-longer in use
+ *
+ * @param[in] to_free	Array of autoinit structures detailing libraries to free
+ */
+static void lib_autofree(global_lib_autoinit_t * const *to_free)
+{
+	global_lib_autoinit_t * const *p;
+
+	for (p = to_free; *p != &lib_autoinit_terminator; p++) {
+		global_lib_inst_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(global_lib_inst_t){ .autoinit = *p });
+
+		fr_assert_msg(lib, "Library %s already freed", (*p)->name);
+
+		if (--lib->instance_count > 0) return;
+
+		/*
+		 *  Only run the free callback if the library was successfully initialised
+		 */
+		if (lib->initialised && ((*p)->free)) (*p)->free();
+
+		fr_rb_remove(&lib_list->libs, lib);
+		talloc_free(lib);
+	}
+}
+
+/** Callback for freeing of "lib" symbols
+ *
+ */
+void global_lib_autofree(UNUSED dl_t const *module, void *symbol, UNUSED void *user_ctx)
+{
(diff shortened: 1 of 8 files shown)


src/lib/server/lib.c
```diff
@@ -0,0 +1,229 @@
...
+ */
+static int lib_init_call(fr_lib_t *lib)
+{
+	CONF_SECTION	*global_cs, *cs;
+
+	/*
+	 *	Find relevant config section
+	 */
+	global_cs = cf_section_find(cf_root(main_config->root_cs), "global", NULL);
+	if (!global_cs) {
```
alandekok, line 54: I'd just have this, and the `lib->name` check return 0 if there's no configuration section found.  "Do nothing" should be a perfectly fine configuration.

src/lib/server/lib.c
```diff
@@ -0,0 +1,229 @@
...
+
+	for (; (*p)->name; p++) {
+		fr_lib_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(fr_lib_t){ .name = (*p)->name });
+
+		/*
+		 *  If the library is already initialised, just increase the reference count
+		 */
+		if ((lib) && (lib->initialised)) {
```
alandekok, line 100: what about `if lib && !lib->initialized` ?  Maybe make that an assert?

ndptech: That's a valid thing which could happen.  When libraries are loaded during config parsing, the entry would be created in the tree on the first occasion the library is loaded, however initialisation can't happen until the config parsing is completed as `main_config->root_cs` is not populated until that point.
So if the same library is used by more than one module in FR which causes it to be loaded more than once during config parsing, after the first time it is loaded, subsequent calls will have `lib && !lib-initialised`.  The initialisation will happen after config parsing has been completed by a call to `lib_init()`

src/lib/server/lib.c
arr2036: Agreed, warnings not required here.

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+	global_cs = cf_section_find(cf_root(main_config->root_cs), "global", NULL);
+	if (!global_cs) {
+		WARN("Missing global config section");
+		goto init;
+	}
+
+	cs = cf_section_find(global_cs, lib->name, NULL);
+	if (!cs) {
+		WARN("Missing \"%s\" subsection in global config", lib->name);
+		goto init;
```
arr2036, line 62: You still need to evaluate the default rules.  Probably easiest to alloc the sections if they're not found and just add them in.

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+	for (; (*p)->name; p++) {
+		fr_lib_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(fr_lib_t){ .name = (*p)->name });
+
+		/*
+		 *  If the library is already initialised, just increase the reference count
+		 */
+		if ((lib) && (lib->initialised)) {
+			lib->instance_count ++;
```
arr2036, line 102: WS

src/lib/server/lib.c
arr2036: Agree with @ndptech.  Multiple modules use things like `libfreeradius-curl`, and this is the global config, i.e. there's one config structure for the entire process.

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with this program; if not, write to the Free Software
+ *  Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+/*
+ * $Id$
+ *
+ * @file lib.c
```
arr2036, line 20: This is a bit of a generic name, maybe `global.c` instead?

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+	fr_rb_node_t		entry;			//!<  Entry in tree of libraries
+	void			*data;			//!<  Library configuration data
+	fr_lib_autoinit_t	*autoinit;		//!<  Autoinit structure used to manage this library
+	uint32_t		instance_count;		//!<  Number of current uses of this library
+	bool			initialised;		//!<  Has the init callback been run for this library
+} fr_lib_t;
+
+typedef struct {
+	fr_rb_tree_t		libs;
+} fr_lib_list_t;
```
arr2036, line 65: Move into the .c file

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+ *
+ */
+typedef struct {
+	char const		*name;			//!<  Name of library
+	fr_rb_node_t		entry;			//!<  Entry in tree of libraries
+	void			*data;			//!<  Library configuration data
+	fr_lib_autoinit_t	*autoinit;		//!<  Autoinit structure used to manage this library
+	uint32_t		instance_count;		//!<  Number of current uses of this library
+	bool			initialised;		//!<  Has the init callback been run for this library
+} fr_lib_t;
```
arr2036, line 61: Move into the .c file

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+
+typedef struct {
+	fr_rb_tree_t		libs;
+} fr_lib_list_t;
+
+int fr_dl_lib_autoinit(dl_t const *module, void *symbol, void *user_ctx);
```
arr2036, line 67: Don't need the fr_ prefix that's only for functions in the util library and the protocol libraries, i.e. stuff that needs to be namespaced.  The server library is there so we can call the same code from the test harness binaries and `radiusd`, it's not meant for external consumption in the same way.

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+/** Structure to define how to initialise libraries with global configuration
+ *
+ */
+typedef struct {
+	char const 		*name;			//!<  Name of library and section within global config
+	CONF_PARSER const	*config;		//!<  Config parser for this library's global options
+	void			*inst;			//!<  Module data to parse global config into
+	fr_lib_init_t		init;			//!<  Callback to initialise library
+	fr_lib_free_t		free;			//!<  Callback to free library
+} fr_lib_autoinit_t;
```
arr2036, line 44: `global_lib_autoinit_t`?

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+
+/** Run free callbacks for external libraries no-longer in use
+ *
+ * @param[in] to_free	Array of autoinit structures detailing libraries to free
+ */
+static void lib_autofree(fr_lib_autoinit_t * const *to_free)
+{
+	fr_lib_autoinit_t * const *p = to_free;
+
+	for (; (*p)->name; p++) {
```
arr2036, line 143: Maybe stick the init on the same line.  Just makes it slightly clearer.

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+		fr_lib_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(fr_lib_t){ .name = (*p)->name });
+		if (!lib) {
```
arr2036, line 147: I think this should probably be an assert, as there shouldn't be any way this happens normally.

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+ * @param to_init	Array of autoinit structures detailing libraries to initialise
+ * @return
+ *	- 0 on succcess
+ *	- -1 on failure
+ */
+static int lib_autoinit(fr_lib_autoinit_t * const *to_init)
+{
+	fr_lib_autoinit_t * const *p = to_init;
+
+	for (; (*p)->name; p++) {
```
arr2036, line 93: Or `p != fr_lib_autoinit_terminator`.  Ensures people actually use the terminator :)

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+ *
+ */
+static int8_t _lib_name_cmp(void const *one, void const *two)
+{
+	fr_lib_t const	*a = one;
+	fr_lib_t const	*b = two;
+	int 		ret;
+
+	ret = strcmp(a->name, b->name);
+	return CMP(ret, 0);
```
arr2036, line 182: `return CMP(a->autoinit, b->autoinit)`?

src/lib/server/lib.c
arr2036: Maybe stick the init on the same line. Just makes it slightly clearer.

src/lib/server/lib.c
arr2036: Or p != fr_lib_autoinit_terminator. Ensures people actually use the terminator :)

src/lib/server/lib.c
arr2036: `.autoinit = p`

src/lib/server/lib.c
```diff
@@ -0,0 +1,230 @@
...
+ *	- -1 on failure
+ */
+static int lib_autoinit(fr_lib_autoinit_t * const *to_init)
+{
+	fr_lib_autoinit_t * const *p = to_init;
+
+	for (; (*p)->name; p++) {
+		fr_lib_t	*lib = NULL;
+
+		lib = fr_rb_find(&lib_list->libs, &(fr_lib_t){ .name = (*p)->name });
```
arr2036, line 96: `.autoinit = p`

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+extern const fr_lib_autoinit_t fr_lib_autoinit_terminator;
+
+/** Structure to track use of libraries.
+ *
+ */
+typedef struct {
+	char const		*name;			//!<  Name of library
+	fr_rb_node_t		entry;			//!<  Entry in tree of libraries
+	void			*data;			//!<  Library configuration data
+	fr_lib_autoinit_t	*autoinit;		//!<  Autoinit structure used to manage this library
```
arr2036, line 58: `const`

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+/*
+ *	To be used as terminator in an array of fr_lib_autoinit_t
+ */
+extern const fr_lib_autoinit_t fr_lib_autoinit_terminator;
+
+/** Structure to track use of libraries.
+ *
+ */
+typedef struct {
+	char const		*name;			//!<  Name of library
```
arr2036, line 55: no longer needed

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+
+typedef int (*fr_lib_init_t)(void);
+
+typedef void (*fr_lib_free_t)(void);
+
+/** Structure to define how to initialise libraries with global configuration
+ *
+ */
+typedef struct {
+	char const 		*name;			//!<  Name of library and section within global config
```
arr2036, line 39: no longer needed

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+	CONF_PARSER const	*config;		//!<  Config parser for this library's global options
+	void			*inst;			//!<  Module data to parse global config into
```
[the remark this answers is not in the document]
arr2036: ...wait, what's the data field in `fr_lib_t`?

src/lib/server/lib.h
```diff
@@ -0,0 +1,73 @@
...
+ */
+extern const fr_lib_autoinit_t fr_lib_autoinit_terminator;
+
+/** Structure to track use of libraries.
+ *
+ */
+typedef struct {
+	char const		*name;			//!<  Name of library
+	fr_rb_node_t		entry;			//!<  Entry in tree of libraries
+	void			*data;			//!<  Library configuration data
```
arr2036, line 57: Doesn't look like this is used?

src/lib/server/lib.c
arr2036: Maybe stick the init on the same line. Just makes it slightly clearer.

src/lib/server/global_lib.c
```diff
@@ -0,0 +1,232 @@
...
+ *  Terminator for array of global_lib_autoinit_t
+ */
+global_lib_autoinit_t const lib_autoinit_terminator = { .name = NULL };
+
+/*
+ *  Global list of libraries
+ */
+typedef struct {
+	fr_rb_tree_t		libs;
+} fr_lib_list_t;
```
arr2036, line 39: Take off the `fr_` prefix from anything in the server library.

src/lib/server/global_lib.h
```diff
@@ -0,0 +1,57 @@
...
+	lib_init_t		init;			//!<  Callback to initialise library
+	lib_free_t		free;			//!<  Callback to free library
+} global_lib_autoinit_t;
+
+/*
+ *	To be used as terminator in an array of fr_lib_autoinit_t
+ */
+extern const global_lib_autoinit_t lib_autoinit_terminator;
+
+int dl_lib_autoinit(dl_t const *module, void *symbol, void *user_ctx);
```
arr2036, line 51: global_lib_autoinit

src/lib/server/global_lib.h
```diff
@@ -0,0 +1,57 @@
...
+
+void dl_lib_autofree(dl_t const *module, void *symbol, void *user_ctx);
```
arr2036, line 53: global_lib_autofree

src/lib/server/global_lib.h
```diff
@@ -0,0 +1,57 @@
...
+
+int lib_list_init(void);
+
+int lib_init(void);
```
arr2036, line 57: global_list_init

src/lib/server/global_lib.c
arr2036: `global_list_list_t`

PR #4542, opened 2022-05-31
automatic dynamic server discovery as per RFC7585

restena-sw:
scripts and documentation update to facilitate dynamic server discovery in a more complete fashion (automates DNS lookup and radmin commands) so that enabling dynamic discovery for consortia like eduroam and OpenRoaming is now purely a config change

### DIFF
--- a/doc/configuration/dynamic_home_servers.md
+++ b/doc/configuration/dynamic_home_servers.md
@@ -190,9 +190,9 @@ authorize {
 			case {
 				# no home server exists, ask DNS
 				update control {
-					&My-Local-String := `%{config:prefix}/bin/naptr-eduroam-freeradius.sh %{1} %{config:prefix}`
+					&Temp-Home-Server-String := `%{config:prefix}/bin/naptr-eduroam-freeradius.sh %{1} %{config:prefix}`
 				}
-				if ("%{control:My-Local-String}" == "" ) {
+				if ("%{control:Temp-Home-Server-String}" == "" ) {
 					reject
 				} else {
 					update control {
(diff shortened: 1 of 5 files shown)


alandekok:
I've pushed most of the changes back to the v3.2.x branch.  I've done some merging of commits, so that it looks like everything was correct on the first commit.

The one thing left is installing and running the scripts.  I'm not sure it's best to put them into a global `bin` directory.  They're not really intended for random people (or even administrators) to run.

Perhaps they should go into a new directory, `raddb/bin`  ?  That way it's clear that only the server uses them.  And people who don't use the scripts can simply delete that directory.

restena-sw:
An admin might want to call this by hand to debug DNS issues with their own realm, to see if discovery would work in the server? That's admittedly the only manual-action call for the script though.

alandekok:
@restena-sw Hmm... that's true.

Putting very use-specific scripts into `/usr/bin` just seems weird to me.  If the scripts do go there, they should be named `freeradius-foo`, to make it very clear that nothing else needs them.

alandekok:
I've pushed some cleanups and rearrangements.  Rather than running a script inside of a script, I just pulled the `verify` functions into one script.

I added a `-d raddb` parameter, ala `radtest`.  And created a `raddb/home_servers/tls.conf` file.  That way all of the site-local configuration can be in `tls.conf`, and not in the script.

restena-sw:
How about putting the script into mods-config/suffix/... ? Its only use is in conjunction with that module. It's arguably to have a shell script there, but other mods also have "not really config" in that space, like SQL schemata or perl/example.pl .

And thanks much for cleaning up my mess :-)

alandekok:
OK, moved to `mods-config/realm/freeradius-naptr-to-home-server.sh`.  I updated the docs to match.

PR #4543, opened 2022-05-31
Annotate to assure coverity that buffers are initialized (CID #150404…

jejones3141:
…2, #1504436)


jejones3141:
I came across it here:
https://community.synopsys.com/s/article/how-to-add-code-annotation-I-do-not-know-which-name-should-exist-with-coverity

jejones3141:
Not docs as such, but examples. Looks like it's a comment with
coverity[<abbreviation for defect>] as the body.

PR #4548, opened 2022-06-06
Amend building of test certs to avoid `openssl ca` race condition

ndptech:


raddb/all.mk
```diff
@@ -167,46 +149,20 @@ ${top_srcdir}/raddb/certs/passwords.mk: $(wildcard ${top_srcdir}/raddb/certs/*cn
...
-$(foreach dir,rsa ecc,$(foreach file,ca server client ocsp,$(eval $(call BUILD_CERT,${top_srcdir}/raddb/certs,${dir},${file}))))
+build.raddb: ${top_srcdir}/raddb/certs/ecc/ocsp.pem
 
-${top_srcdir}/raddb/certs/dh: ${top_srcdir}/raddb/certs/passwords.mk
-	${Q}echo CERT-DH $@
-	${Q}$(MAKE) -C ${top_srcdir}/raddb/certs/ $(notdir $@)
-	${Q}touch $@
+${top_srcdir}/raddb/certs/ecc/ocsp.pem:
+	${Q}echo Making certs for $@
+	${Q}$(MAKE) -C ${top_srcdir}/raddb/certs/ dh
```
arr2036, line 159: I think you can just make all?

ndptech: That does a few other bits, which aren't required for CI, and I think also had other parallel run issues

arr2036: It won't have parallel run issues because by default Make only builds one target at a time.

ndptech: The issues were with the .vrfy targets in raddb/certs/Makefile - they did not have all their dependencies listed

PR #4551, opened 2022-06-07
redis cache module

jrouzierinverse:


src/modules/rlm_cache/drivers/rlm_cache_redis/rlm_cache_redis.c
```diff
@@ -0,0 +1,417 @@
...
+		error:
+			talloc_free(c);
+			freeReplyObject(reply);
+			return CACHE_ERROR;
+		}
+		break;
+	case REDIS_REPLY_NIL:
+		talloc_free(c);
+		freeReplyObject(reply);
+		return CACHE_MISS;
```
jpereira, line 234: Maybe a simple `goto error;` here could be helpful.

arr2036: But it's returning notfound... which is is correct

PR #4556, opened 2022-06-10
Placate Coverity for mschap_xlat() (CID #1503902)

jejones3141:


arr2036:
Could you use `MEM()` here?

jejones3141:
Will do.

PR #4561, opened 2022-06-14
Handle coverity false positive in SBUFF_PARSE_FLOAT_DEF() (CID #1504042)

jejones3141:
`fr_sbuff_out_bstrncpy_allowed()` always copies at least the
terminating NUL, so buff is always initialized before the `_func()`
call.


jejones3141:
redundant with uninit

PR #4570, opened 2022-06-16
Revise SBUFF_PARSE_[U]INT_DEF() to avoid always-true checks (CID #150…

jejones3141:
…4045, #1504004)
_Generic() lets us check the type of the output pointer to avoid the always-true checks that Coverity points out.


jejones3141:
Annotating will work. This is definitely overkill, so I will switch to that.

PR #4571, opened 2022-06-20
Set *err to -2 when the list isn't found in tmp_dcursor_init()

jejones3141:
This change honors the comments describing how err is handled, and
also means, given tmpl_find_vp()'s use of tmpl_dcursor_init(),
means callers of tmpl_find_vp() can check its return value
rather than having to look at the written vp (e.g. sometimes_return();
see CID #1503995).

### DIFF
--- a/src/lib/server/tmpl_dcursor.c
+++ b/src/lib/server/tmpl_dcursor.c
@@ -373,6 +373,7 @@ fr_pair_t *tmpl_dcursor_init(int *err, TALLOC_CTX *ctx, tmpl_dcursor_ctx_t *cc,
 		if (!list_head) {
 			fr_strerror_printf("List \"%s\" not available in this context",
 					   fr_table_str_by_value(pair_list_table, tmpl_list(vpt), "<INVALID>"));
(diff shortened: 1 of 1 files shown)


src/lib/server/tmpl_dcursor.c
```diff
@@ -371,8 +371,11 @@ fr_pair_t *tmpl_dcursor_init(int *err, TALLOC_CTX *ctx, tmpl_dcursor_ctx_t *cc,
 	if (!vpt->rules.attr.list_as_attr) {
 		list_head = tmpl_list_head(request, tmpl_list(vpt));
 		if (!list_head) {
-			fr_strerror_printf("List \"%s\" not available in this context",
-					   fr_table_str_by_value(pair_list_table, tmpl_list(vpt), "<INVALID>"));
+			if (err) {
+				*err = -2;
+				fr_strerror_printf("List \"%s\" not available in this context",
```
arr2036, line 376: This should be outside the condition

jejones3141: So be it.

PR #4574, opened 2022-06-22
Create relative symlinks so that the source tree can be shared across several hosts to be accessed remotely

mkhon:


alandekok:
I'll merge this, but the build system needs many more changes in order to have multiple builds from one source tree.

arr2036:
What's the use case?

mkhon:
> What's the use case?

vagrant or multipass (source code is in a vagrant or multipass VM, code editor on laptop)

arr2036:
Ah fair enough.

PR #4576, opened 2022-06-22
Fix fr_sbuff_extend_file()

mkhon:
The fix allows to parse policy files longer than 8k (file I/O buffer size of `pairlist_read()`), when policy strings span I/O buffer boundaries.

However, looks like more complete (and intrusive) fix is to use bound child sbufs (obtained via `FR_SBUFF_BIND_CURRENT()`) at least in the following functions:
- cond_tokenize_operand()
- cond_tokenize()
- fr_cond_tokenize()
- map_afrom_substr()
- tmpl_afrom_value_substr()
- tmpl_afrom_octets_substr()
- tmpl_afrom_substr()
- tokenize_expression()
- xlat_tokenize_ephemeral()
- xlat_tokenize_argv()
- xlat_tokenize()
- dict_attr_search()
- dict_attr_search_qualified()
- fr_sbuff_out_bstrncpy_exact()
- fr_value_box_from_substr()

The pattern used in these functions

```
our_in = FR_SBUFF(in);
// read potentially unbounded objects (strings, octets) from our_in
return fr_sbuff_set(in, &our_in);
```

is not suitable when input buffer can be smaller than the parsed object - when input is streamed `fr_sbuff_set()` will fail if the buffer was shifted, or moreover, buffer shifting will fail because parent sbuff (`in`) was not advanced while reading from `our_in`, or there are sbuff markers bound to `in` (up the calling stack) or `our_in` (in the current function context).


arr2036:
The reason the functions create their own sbuffs is to allow backtracking.  If this really becomes an issue we can dynamically resize the buffers.  There's code to do that already for aggregation sbuffs, where we're printing data of an unknown length.

I agree with your fix in `fr_sbuff_extend_file`, but I'm not sure why `fr_sbuff_out_bstrncpy` and others need to change... They all record the amount data has been shifted during the lifetime of the sbuff, and add that to `(start - p)` to get the total amount of data consumed...

arr2036:
...and yes, you could say there's no point in allowing backtracking for something like `fr_cond_tokenize`, but the other issue is we pretty print positional error markers, indicating where in an expression an error occurred, and for that we need the original string, so we can't shift it out of the temporary buffer.

mkhon:
> I agree with your fix in `fr_sbuff_extend_file`, but I'm not sure why `fr_sbuff_out_bstrncpy` and others need to change... They all record the amount data has been shifted during the lifetime of the sbuff, and add that to `(start - p)` to get the total amount of data consumed.
> 
> The reason the functions create their own sbuffs (that are not bound to their parents) is to allow backtracking. If this really becomes an issue we can dynamically resize the buffers. There's code to do that already for aggregation sbuffs where we're printing data of an unknown length.

Returning `fr_sbuff_used_total(our_in)` is not correct when the `our_in` was shifted. For the buffer size of 8K you return 8K once the buffer was first shifted and parsing breaks.

arr2036:
> I am not sure I get your comment about dynamic buffer resizing - input buffers are of fixed size and I think they are supposed to be fixed.

There's no requirement for the underlying char buffer to be stack allocated.  It's fine to allocate them on the heap.  It just requires explicit cleanup after we're done with the `fr_sbuff_t`.

> Returning fr_sbuff_used_total(our_in) is not correct when our_in was shifted. You return at least our_in->shifted when at EOF position (and should return 0) once the buffer was shifted and parsing breaks because of that. See the tests that I added that demonstrate this issue.

Really `shifted` should not have been initialised to the value from the parent sbuff.  I've now fixed that (it now gets initialised to 0).  Could you rebase, and try your tests again without the changes to `fr_sbuff_out_bstrncpy` and `fr_sbuff_out_bstrncpy_allowed`.

arr2036:
No problem.  Thanks for the other fixes and the test cases!

PR #4584, opened 2022-06-25
Sync "git for-each-ref" output parsing with other Dockerfile.deps

mkhon:


arr2036:
Could you apply the same fix to the other distros, thanks!

mkhon:
> Could you apply the same fix to the other distros, thanks!

I think all others are already fixed - this is from where I obtained the patch

PR #4587, opened 2022-06-27
Define top_srcdir and BUILD_DIR when making "rpm" so that tarball creation succeeds

mkhon:


Makefile
```diff
@@ -132,29 +132,35 @@ define n
...
-endif
-endif
-endif
+endif # crossbuild
+else # rpm
+#
+#  We're building rpm
+#
+top_srcdir=	$(realpath $(dir $(lastword $(MAKEFILE_LIST))))
+BUILD_DIR=	build
```
mcnewton, line 161: This would be a bit clearer if it were in a separate `ifeq "$(MAKECMDGOALS)" "rpm"` section after the endifs, rather than in an else. This if section is rather large already.

mkhon: BUILD_DIR and top_srcdir is defined for the other case (libfreeradius-make) in the same block

PR #4589, opened 2022-06-28
Fix source files in debuginfo rpm

mkhon:
Without the fix I get the following when trying to build rpm with -g (CFLAGS=-g make rpm) on CentOS 7:

```
cpio: freeradius-4.0.0-1.el7.x86_64/src/bin/dhcpclient.c: Cannot stat: No such file or directory
cpio: freeradius-4.0.0-1.el7.x86_64/src/bin/radclient.c: Cannot stat: No such file or directory
cpio: freeradius-4.0.0-1.el7.x86_64/src/bin/radclient.h: Cannot stat: No such file or directory
...
```

And no source files appear in the debuginfo package.

Note that this reproduces only when freeradius is compiled with CFLAGS=-g - makes sense to always use -g so that correct debuginfo packages are generated.

### DIFF
--- a/redhat/freeradius.spec
+++ b/redhat/freeradius.spec
@@ -506,29 +506,6 @@ export CFLAGS="$CFLAGS -g3 -fpic"
 export CXXFLAGS="$CFLAGS"
 %endif
 
-# The build system seems to feed the compiler relative paths for the source files.  This is usually fine
-# but it means the source paths in the ELF headers of the binaries are also relative.
-#
-# As a post install step a helper script (find-debuginfo.sh) is called to create the debug info files and
-# perform stripping on the original binaries.
-#
-# find-debuginfo.sh calls another utility (debugedit) to rewrite the source paths in the ELF headers of
-# the binaries so they'll match where the source files will actually be installed.
-#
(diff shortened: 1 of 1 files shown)


arr2036:
This would break local builds for us.  It'd be better to figure out the prefix map for your situation.

mkhon:
> This would break local builds for us. It'd be better to figure out the prefix map for your situation.

How do you run your local builds? Is there -g in CFLAGS? By default it is not there.

Which distro do you use to build RPM - the problem can easily be reproduced on CentOS 7.

If you don't have -g in CFLAGS you won't see the issue. If you run "CFLAGS=-g make
rpm" you will see a lot of error messages that .h file is not found when debuginfo is being extracted.

Additionally, I think we should append -g to CFLAGs by default when building RPM, otherwise debuginfo packages are incomplete and can not be used for source-level debugging

arr2036:
We build on Centos 7 and Rocky 8.  The prefix map is needed there to produce a useable debuginfo package, otherwise lldb/gdb fail to locate the source files installed with the package.  We're building with CFLAGS `-g3`.

mkhon:
Can you please re-check on latest CentOS 7? Because this is 100% reproducible on my CentOS 7 build machines

I will try to test on Rocky 8

PR #4590, opened 2022-06-28
map_afrom_cp: unescape double-quoted and back-quoted RHS prior to parsing it

mkhon:
This fixes the following regression: the following unlang code could be properly parsed by 3.2.x:

```
update control {
    &DHCP-Redis-Pre-Script := "%(redis:EVAL \"some lua script\" 1 %{control.IP-Pool.Name}"
}
```

In 4.x this generates the following error:

```
/etc/raddb/sites-enabled/dhcp[276]: %(redis:EVAL \"some lua script...
/etc/raddb/sites-enabled/dhcp[276]:         ^: Unexpected text after argument
```

The reason is that `xlat_tokenize_function_args` (and `xlat_tokenize_expansion`) do not obey `fr_sbuff_parse_rules_t` passed to `xlat_tokenize_string` so unescaping RHS does not happen. Moreover - they seem to expect unescaped text and apply their own unescape rules.

### DIFF
--- a/src/lib/server/map.c
+++ b/src/lib/server/map.c
@@ -99,6 +99,7 @@ int map_afrom_cp(TALLOC_CTX *ctx, map_t **out, map_t *parent, CONF_PAIR *cp,
 {
 	map_t		*map;
 	char const	*attr, *value, *marker_subject;
+	fr_sbuff_parse_rules_t const *p_rules;
 	ssize_t		slen;
 	fr_token_t	type;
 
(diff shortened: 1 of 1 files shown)


alandekok:
The intent of the `"%(redis:EVAL \"some lua script" 1 %{control.IP-Pool.Name}"` call is that it gets parsed:

1. As a double-quoted string
2. Then the `"` parser sees the `%{...}`, and parses it as an xlat
3. The xlat parser sees `redis:...` and ensures that the `redis` function exists
4. sees the `redis_args` from `src/modules/rlm_redis/rlm_redis.c`
5. ensures that the `redis_xlat()` function is called with the appropriate arguments

Note that by the time we get to step (2), the intent is that the double-quoted string has already been un-escaped.  So the arguments to the `redis` function are just a string: `EVAL "some lua script" 1 %{control.IP-Pool.Name}`.

Those arguments should be parsed as whatever they are.  i.e. one string with spaces, etc.  Because the `redis_args` definition has `.variadic = true`.

I think that the correct fix is to update `xlat_tokenize_argv()` to get passed the `xlat_arg_parser_t` array, so that it knows how to parse the arguments.  In this case, we have `.variadic = true`, so the parser should just keep parsing anything until it's done.

mkhon:
> Note that by the time we get to step (2), the intent is that the double-quoted string has already been un-escaped.

Yes. But the double-quoted string is not unescaped when xlat parser is called from `map_afrom_cp` - and this is something I tried to fix in my PR

mkhon:
> This API is deprecated and will be removed in the next couple of months in favour of one that supports async I/O.

Which API do you mean?

PR #4603, opened 2022-07-07
Introduce sigalgs_list option to control signature algorithms used for TLS negotiaiton.

lytboris:
Available for OpenSSL 1.1.1+.

This allows to work-around broken TPMs with PSS algorithm broken in a weird way.

See also: https://lists.freeradius.org/pipermail/freeradius-users/2021-January/099364.html

p/s This change could be merged into v3.0.x as-is, let me know if anyone willing to do so, will make an another PR (or just merge it :)


src/main/tls.c
```diff
@@ -4233,6 +4237,26 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client, char const *chain_
 		}
 	}
 
+#if OPENSSL_VERSION_NUMBER >= 0x10101000L
+	/*
+	 * SSL_CTX_set1_sigalgs_list() macro is missing const modifier for string argument
+	 * thus we have to make a copy of const char *sigalgs_list prior calling the macro.
+	 */
+	if (conf->sigalgs_list) {
+		char *sigalgs_list_copy = talloc_strdup(conf, conf->sigalgs_list);
```
alandekok, line 4246: You could just use `memcpy` from one pointer to another.  :)  See `/* const issues */` all over the server code.

lytboris:
It looks like CI/CD is broken. Const issues also? :)

alandekok:
I've manually merged this in commit 28a7e84c3ad3e

alandekok:
We'll leave v3.0.x alone for now.  That branch is frozen, and only bug fixes && security fixes are applied to it.

PR #4606, opened 2022-07-10
CentOS 7 has glibc 2.17 which does not support %Z in strptime(3)

mkhon:
%Z support was added in glibc 2.19: https://sourceware.org/bugzilla/show_bug.cgi?id=14876

### DIFF
--- a/src/tests/keywords/date
+++ b/src/tests/keywords/date
@@ -13,7 +13,7 @@ update request {
 }
 
 # Some systems report GMT some UTC...
-if (&Tmp-String-0 != "Fri 22 Sep 17:25:00 GMT 2017") && (&Tmp-String-0 != "Fri 22 Sep 17:25:00 UTC 2017") {
+if (&Tmp-String-0 != "Fri 22 Sep 17:25:00 2017") {
 	test_fail
 }
 
--- a/src/tests/keywords/radius.conf
+++ b/src/tests/keywords/radius.conf
@@ -38,7 +38,7 @@ modules {
 	}
 
 	date {
-		format = "%a %d %b %H:%M:%S %Z %Y"
+		format = "%a %d %b %H:%M:%S %Y"
 		utc = yes
 	}
 


alandekok:
The times needs to be printed with time zone.  It's not good to remove`%Z` for *all* platforms, simply because *one* platform doesn't support it.

Plus, CentOS 7 is from 2014.   Version 4 generally news much more up-to-date kernels, libc, and compilers

mkhon:
@alandekok I haven't changed default format - this change only fixes tests

CentOS 7 is the last available CentOS (sic!) after CentOS 8 died and it is still used in production

alandekok:
https://rockylinux.org/

mkhon:
@alandekok it's not my choice to use CentOS 7

alandekok:
We have decided what we're going to support.  We only have limited resources.

If you want a RADIUS server on CentOS7, use 3.2.  It's supported there.  If you want the features of v4, use a modern distribution.  It's your choice.

mkhon:
@alandekok I understand about the limited resources.

I am currently on contract for a rather large organization to make freeradius v4 to work (thus my recent PRs) - I am not in a position to decide about infrastructure (which is CentOS 7) - and changing base OS will require months of work of their infrastructure team to update their custom Docker images, CI pipelines and so on.

I've just thought that making all the tests working on most architectures and OSes is a nice thing which improves freeradius stability.

In my change default date format of rlm_date is not changed - I only changed a config of a single "keyword date" test

alandekok:
Removing functionality means that it's untested.

git repositories are free.  It's possible to maintain your own fork, including any custom patches for distributions which are almost 10 years out of date.

arr2036:
@mkhon adding a shim to support `%Z` with glibc versions which don't support timezones would be the preferred way of fixing this, not removing tests.

mkhon:
@arr2036 I haven't removed any tests.

Adding a shim for function like getptime(3) will mean adding several hundred lines of code for 10-year-old OS, and also some autoconf glue for this - does not make sense to me.

I am perfectly ok with this patch living in my local repo.

PR #4610, opened 2022-07-11
rlm_files: allow to specify proto (default "radius")

mkhon:


arr2036:
You should already be able to qualify attribute names with protocols.  If that's not working then that needs to be addressed.  There's very little reason so manually load protocols like this because the demand for them should already be there from proto and process modules.

mkhon:
> You should already be able to qualify attribute names with protocols. If that's not working then that needs to be addressed. There's very little reason so manually load protocols like this because the demand for them should already be there from proto and process modules.

@arr2036 can you please clarify how this is expected to work? May be an example of rlm_files configuration file

alandekok:
It's not clear why you need to force a protocol in the `files` module.  The tests use RADIUS, and that works.

If you want to refer to protocol-specific attributes, you can do `&radiusd.User-Name` or `&dhcpv4.Client-Identifier`

A deeper issue is that every module can be used in multiple virtual servers.  It would be a lot of work to manually specify a `namespace = ...` in each module, when that information is already in the virtual server.

A longer term solution would be to automatically inform each module (during instantiation) about the namespace, so that everything just works, with no additional intervention from the administrator.

mkhon:
@alandekok it is actually dhcpv4.Client-Identifier (without &). I will close this PR. Thanks!

Issue #4598, opened 2022-07-05
[defect]: ASSERT FAILED src/lib/server/trunk.c[1051]: 0: Trunk request 0 invalid transition INIT -> UNASSIGNED


mkhon:
### What type of defect/bug is this?

Crash or memory corruption (segv, abort, etc...)

### How can the issue be reproduced?

Use rlm_ldap

### Log output from the FreeRADIUS daemon

```shell
The problem can not be reproduced with "radiusd -X" (only when running with multiple threads):

Mon Jul  4 13:29:48 2022: Info  : Network 0 - Starting
Mon Jul  4 13:29:48 2022: Info  : Worker 0 - Starting
Mon Jul  4 13:29:48 2022: Info  : Worker 1 - Starting
Mon Jul  4 13:29:48 2022: Error : Failed setting connection option new TLS context: Unknown error
Mon Jul  4 13:29:48 2022: Error : ldap - [2] - Connection initialisation failed
Mon Jul  4 13:29:53 2022: Info  : rlm_ldap - [3] Trunk connection changed state INIT -> CONNECTING
Mon Jul  4 13:29:53 2022: Info  : rlm_ldap (ldap) - Directory type: OpenLDAP
Mon Jul  4 13:29:53 2022: Info  : rlm_ldap (ldap) - Directory supports RFC 4533
Mon Jul  4 13:29:53 2022: Info  : Scheduler created successfully with 1 networks and 2 workers
Mon Jul  4 13:29:53 2022: Info  : post-suid-down capabilities: = cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,35,36+ep
Mon Jul  4 13:29:53 2022: Info  : Ready to process requests
[...]

PR #4628, opened 2022-07-22
Attempt to fix LDAP query cancellation issues

mkhon:
See https://github.com/FreeRADIUS/freeradius-server/issues/4598#issuecomment-1192615161

1) fr_ldap_trunk_search(): revert rev. 2a47fc01ba326431171eb1e53d64d0c009facf44
which causes fr_trunk_request_signal_cancel() to be called twice for some queries
(from _ldap_search_sync_timeout() and ldap_trunk_query_cancel()

2) ldap_trunk_query_cancel(): ignore query->treq == NULL:
query->treq can be NULL for completed queries

3) fr_ldap_trunk_search(), fr_ldap_trunk_modify(): use ttrunk->trunk talloc
context for LDAP queries - the same lifetime as for fr_trunk_request_t,
otherwise queries are free'd when stack frames are freed in frame_cleanup()
and ldap_request_cancel_mux() later on gets treq with free'd treq->preq

4) fr_trunk_request_signal_cancel_complete(): handle treq in
FR_TRUNK_REQUEST_STATE_CANCEL_PARTIAL and FR_TRUNK_REQUEST_STATE_CANCEL,
otherwise ldap_request_cancel_mux() loops on the same treq if this treq
was previously cancelled by fr_trunk_request_signal_cancel() and entered
FR_TRUNK_REQUEST_STATE_CANCEL state

### DIFF
--- a/src/lib/ldap/base.c
+++ b/src/lib/ldap/base.c
@@ -619,6 +619,7 @@ static void ldap_trunk_query_cancel(UNUSED request_t *request, fr_state_signal_t
 	if (!query->treq) return;
 
 	fr_trunk_request_signal_cancel(query->treq);
+	query->treq = NULL;
 
 }
 
@@ -648,9 +649,7 @@ static unlang_action_t ldap_trunk_query_start(UNUSED rlm_rcode_t *p_result, UNUS
  */
 static void _ldap_search_sync_timeout(UNUSED fr_event_list_t *el, UNUSED fr_time_t now, void *uctx)
 {
-	fr_trunk_request_t *treq = talloc_get_type_abort(uctx, fr_trunk_request_t);
-
-	fr_trunk_request_signal_cancel(treq);
+	ldap_trunk_query_cancel(NULL, FR_SIGNAL_CANCEL, uctx);
 }
 
 /** Run an async or sync search LDAP query on a trunk connection
@@ -722,7 +721,7 @@ unlang_action_t fr_ldap_trunk_search(rlm_rcode_t *p_result,
 		 */
 		if (fr_time_delta_ispos(timeout)) {
 			if (fr_event_timer_in(ctx, unlang_interpret_event_list(request), &ev, timeout,
-					      _ldap_search_sync_timeout, query->treq) < 0) goto error;
+					      _ldap_search_sync_timeout, query) < 0) goto error;
 		}
 
 		*p_result = unlang_interpret_synchronous(unlang_interpret_event_list(request), request);
--- a/src/lib/server/trunk.c
+++ b/src/lib/server/trunk.c
@@ -1434,7 +1434,6 @@ static void trunk_request_enter_cancel_complete(fr_trunk_request_t *treq)
 	switch (treq->pub.state) {
 	case FR_TRUNK_REQUEST_STATE_CANCEL_SENT:
 	case FR_TRUNK_REQUEST_STATE_CANCEL:
-		REQUEST_EXTRACT_CANCEL_SENT(treq);
 		break;
 
 	default:


mkhon:
I will check unit test failures later. Please treat this PR as draft to discuss fixes for #4598

arr2036:
Reverting https://github.com/FreeRADIUS/freeradius-server/commit/2a47fc01ba326431171eb1e53d64d0c009facf44 isn't an option.  That event timer stops the server from deadlocking when we don't receive a response from LDAP.  If there's no timer event, then the call to `unlang_interpret_synchronous` will never return in some situations.

mkhon:
1) can you please explain why `worker_max_request_timer()` will not help in this case? I see it being fired in my scenario (but may be not in all cases)

I think I can make a change that both timeout handlers are set but would like to understand what is happening here better

4) I will make the proposed changes, however if we have a request in `trunk->partial` we may never exit `ldap_request_cancel_mux()` loop. May be we should add an assertion that the state is not `CANCEL_PARTIAL`

arr2036:
1.  It was clear when I made the change but I don't recall.  I'll dig into it tomorrow.  Might be something to do with the synchronous event loop not processing timer events from the main event loop for the worker.  In which case there may be a simple fix...

4. You can add an assert if you like.  But the LDAP code would need to explicitly signal that a `treq` should transition to the `CANCEL_PARTIAL` state, nothing else is going to do that.

ndptech:
@arr2036 

3. Parenting the `fr_ldap_query_t` off the `treq` won't work as things stand - `ldap_trunk_request_demux()` calls `fr_trunk_request_signal_complete()`which frees the children of the `treq` and therefore the `fr_ldap_query_t` would be freed.  Then we'd have use after free issues with `ldap_xlat_resume()`.  I think your suggestion of a common structure to hold the msgid and replacing the `treq->preq` with that when cancelling a query should be a robust solution.

arr2036:
It looks like both event loops are using the same event list so the request timer should eventually kick in and cancel the request (I'm not sure why it's not).  The other event is still need for now as there's no event loop integration between libldap and FreeRADIUS, so we can't rely on any timeouts set for libldap regarding client side search or connection timeouts.

arr2036:
> @arr2036
> 
> 3. Parenting the `fr_ldap_query_t` off the `treq` won't work as things stand - `ldap_trunk_request_demux()` calls `fr_trunk_request_signal_complete()`which frees the children of the `treq` and therefore the `fr_ldap_query_t` would be freed.  Then we'd have use after free issues with `ldap_xlat_resume()`.  I think your suggestion of a common structure to hold the msgid and replacing the `treq->preq` with that when cancelling a query should be a robust solution.

@ndptech Bah, yeah you're right that's what I was afraid of.  So we need to alloc a new structure to hold the msgid, and bind that to the treq.  Then we're guaranteed to have it cleaned up once the `treq` is done being cancelled.  I saw you put the destructors in there to clean up extensions and such, so we can just free the existing `fr_ldap_query_t` on cancellation and swap out the pointer.

mkhon:
> > @arr2036
> > 
> > 3. Parenting the `fr_ldap_query_t` off the `treq` won't work as things stand - `ldap_trunk_request_demux()` calls `fr_trunk_request_signal_complete()`which frees the children of the `treq` and therefore the `fr_ldap_query_t` would be freed.  Then we'd have use after free issues with `ldap_xlat_resume()`.  I think your suggestion of a common structure to hold the msgid and replacing the `treq->preq` with that when cancelling a query should be a robust solution.
> 
> @ndptech Bah, yeah you're right that's what I was afraid of. So we need to alloc a new structure to hold the msgid, and bind that to the treq. Then we're guaranteed to have it cleaned up once the `treq` is done being cancelled. I saw you put the destructors in there to clean up extensions and such, so we can just free the existing `fr_ldap_query_t` on cancellation and swap out the pointer.

@arr2036 are you going to implement this?

mkhon:
> .4. You can add an assert if you like. But the LDAP code would need to explicitly signal that a `treq` should transition to the `CANCEL_PARTIAL` state, nothing else is going to do that.

@arr2036 Can you please elaborate about `CANCEL_PARTIAL` transition? In which function?

mkhon:
> It looks like both event loops are using the same event list so the request timer should eventually kick in and cancel the request (I'm not sure why it's not). The other event is still need for now as there's no event loop integration between libldap and FreeRADIUS, so we can't rely on any timeouts set for libldap regarding client side search or connection timeouts, at least I think that's right @ndptech? For background, I wrote the original trunk code, and @ndptech did most of the async LDAP work.

@arr2036 can you please elaborate about "the other event"?

Also, if both events are in the same event list - we still don't need the second one?

mkhon:
@arr2036 can you review the updated PR?

1) I removed re-parenting for LDAP queries (should be implemented differently, like described by you)

2) I moved `treq` state changes to `fr_trunk_request_signal_cancel_complete()`

arr2036:
I pushed a couple of commits to deal with some of the none contentious issues.

> @arr2036 are you or @ndptech going to implement this?

Yeah I'll have a go at making the change.

> Also, if both events are in the same event list - we still don't need the second one (_ldap_search_sync_timeout())?

No, because the request timeout period will likely exceed the ldap specific timeout period.  It's a temporary hack.  We want to implement timeouts outside of the modules eventually, likely with sections in unlang which specify the total time for a given part of the policy language to complete.  That's not there today, so we need the extra event to implement LDAP module specific timeouts.

> @arr2036 Can you please elaborate about CANCEL_PARTIAL transition? In which function?

In `fr_trunk_request_signal_cancel_complete` a transition from `FR_TRUNK_REQUEST_STATE_CANCEL_PARTIAL` to `FR_TRUNK_REQUEST_STATE_CANCEL_COMPLETE` is invalid.  That transition would indicate that the cancellation muxer, wrote out half a cancellation request, then gave up and just marked it as done.  That's not allowed.  `FR_TRUNK_REQUEST_STATE_CANCEL_PARTIAL` is a state that's only used for stream protocols, and only when the send buffer is full, and we wrote out part of a cancellation request.  It isn't used at all by the ldap module.

mkhon:
> I pushed a couple of commits to deal with some of the none contentious issues.

@arr2036 I think in ac75bcd993150f3a734c81f0966927322528c0cd calling `trunk_request_enter_cancel_complete()` in `FR_TRUNK_REQUEST_STATE_CANCEL` will cause `REQUEST_BAD_STATE_TRANSITION`

mkhon:
> > Also, if both events are in the same event list - we still don't need the second one (_ldap_search_sync_timeout())?
> 
> No, because the request timeout period will likely exceed the ldap specific timeout period. It's a temporary hack. We want to implement timeouts outside of the modules eventually, likely with sections in unlang which specify the total time for a given part of the policy language to complete. That's not there today, so we need the extra event to implement LDAP module specific timeouts.

In this case there should be protection for calling timeout handlers for the same treq twice

mkhon:
> > I pushed a couple of commits to deal with some of the none contentious issues.
> 
> @arr2036 I think in [ac75bcd](https://github.com/FreeRADIUS/freeradius-server/commit/ac75bcd993150f3a734c81f0966927322528c0cd) calling `trunk_request_enter_cancel_complete()` in `FR_TRUNK_REQUEST_STATE_CANCEL` will cause `REQUEST_BAD_STATE_TRANSITION`

@arr2036 I think there is still an issue with ffc3f85a795b2999653fddaf26645d5b97041152: doing `REQUEST_EXTRACT_CANCEL_SENT()` in `FR_TRUNK_REQUEST_STATE_CANCEL` is not correct. I think we should just remove `REQUEST_EXTRACT_CANCEL_SENT()` as in my PR because it will be done by `trunk_request_remove_from_conn()` a few lines below

mkhon:
@arr2036 can you please review the updated PR?

arr2036:
All those fixes looked OK.  So I guess the outstanding issue is swapping out the preq on cancellation which I can handle.

PR #4629, opened 2022-07-25
Fix tests for 'update{}' section

jpereira:


src/tests/keywords/base64
```diff
@@ -71,12 +63,10 @@ if (&Tmp-String-9[0] != 'Og==') {
 	test_fail
 }
 
-&request <= {
```
alandekok, line 74: That's the new syntax tho... no `update` section.  It should be left alone.

src/tests/keywords/call
```diff
@@ -1,8 +1,8 @@
 #
-#  PRE: if
+#  PRE: update
```
alandekok, line 2: If we're removing `update`, then none of the tests should depend on `update`

src/tests/keywords/call-empty
```diff
@@ -14,6 +15,6 @@ if (&reply.Reply-Message[0] != "call second") {
 	test_fail
 }
 
-&reply -= &Reply-Message
+&reply.Reply-Message !* ANY
```
arr2036, line 18: Not equivalent?

src/tests/keywords/cast-integer
```diff
@@ -8,9 +8,9 @@ if (<integer>&Class[0] == 257) {
 }
 
 if (<integer>&Class[0] < 256) {
-	test_fail
+	&reply.Filter-Id += "wrong"
 }
 
 if (<integer>&Class[0] > 257) {
-	test_fail
+	&reply.Filter-Id += "wrong"
```
arr2036, line 15: Because why?

PR #4635, opened 2022-07-27
missing_lock, not missing_unlock (CID #1414431)

jejones3141:
Typo in coverity annotation.
This occurs in several files, but the defect only occurs in
src/lib/server/pool.c, so ultimately it would be worth pulling
the others; no need to have unnecessary annotations.


arr2036:
Could you remove the unnecessary ones, thanks!

jejones3141:
Will do.

PR #4657, opened 2022-08-08
Add callback to tmpl_dcursors to build out missing pairs

ndptech:


src/lib/server/tmpl_dcursor.c
```diff
@@ -206,6 +206,21 @@ fr_pair_t *_tmpl_cursor_eval(fr_dlist_head_t *list_head, fr_pair_t *curr, tmpl_d
...
 	} else goto all_inst;	/* Used for TMPL_TYPE_LIST */
 
+	/*
+	 *	If no pair was found and there is a fill
+	 *	callback, call that, depending on the suffix
+	 */
+	if ((!vp) && (cc->build) && (ar)) switch (ar->ar_num) {
+	case NUM_UNSPEC:
+	case NUM_LAST:
+	case 0:
```
alandekok, line 216: This could perhaps check the `ar->ar_num` against the iteration?  So if there's 2 vps, and the caller says `foo[2] := ...`, it could add the third one.

Tho that is a bit more of an esoteric use-case.

We may also want to create the vp for `NUM_ALL`?  Tho I'd probably start with the 3 items here, and then write test cases to see how we want it to behave.

src/lib/server/tmpl_dcursor.c
```diff
@@ -466,6 +466,23 @@ void tmpl_dcursor_clear(tmpl_dcursor_ctx_t *cc)
...
+ * @param[in] uctx		unused.
+ * @return
+ *	- newly allocated #fr_pair_t.
+ *	- NULL on error.
+ */
+fr_pair_t *tmpl_dcursor_pair_build(fr_pair_t *parent, fr_dcursor_t *cursor, fr_dict_attr_t const *da, UNUSED void *uctx)
+{
+	fr_pair_t *vp;
+	vp = fr_pair_afrom_da(parent, da);
+	if (vp) fr_dcursor_append(cursor, vp);
```
alandekok, line 483: which always appends to the tail, because no matching vp was found.

src/lib/server/tmpl_dcursor.c
arr2036: `NUM_ALL` can be added to the above cases.  It's up to the build callback to figure out what it wants to do with it.
I think `case 0` is wrong, it should be checking for `ar->ar_num ==  fr_pair_list_num_elements(&parent->vp_group)`.  Otherwise we'd never be able to add more than one child of a given type to the parent?

src/lib/server/tmpl_dcursor.c
```diff
@@ -206,6 +206,21 @@ fr_pair_t *_tmpl_cursor_eval(fr_dlist_head_t *list_head, fr_pair_t *curr, tmpl_d
 		break;
 	} else goto all_inst;	/* Used for TMPL_TYPE_LIST */
 
+	/*
+	 *	If no pair was found and there is a fill
+	 *	callback, call that, depending on the suffix
+	 */
+	if ((!vp) && (cc->build) && (ar)) switch (ar->ar_num) {
```
arr2036, line 213: Um and a little bit of overbracketing here? :)

src/lib/server/tmpl_dcursor.c
arr2036: Is there really much value in having the callback build the pair, seeing as in all cases it's just building a `fr_pair_t` from the `tmpl_attr_t` and appending it to the current list of children?  Seems like we'd just end up with the same boilerplate in every build function.

arr2036: I'd also like to keep this separate from the normal eval function, which is likely to get significantly larger in future when we add proper filter syntax.

src/lib/server/tmpl_dcursor.c
```diff
@@ -466,6 +466,23 @@ void tmpl_dcursor_clear(tmpl_dcursor_ctx_t *cc)
...
+ * @param[in] da		of new pair.
+ * @param[in] uctx		unused.
+ * @return
+ *	- newly allocated #fr_pair_t.
+ *	- NULL on error.
+ */
+fr_pair_t *tmpl_dcursor_pair_build(fr_pair_t *parent, fr_dcursor_t *cursor, fr_dict_attr_t const *da, UNUSED void *uctx)
+{
+	fr_pair_t *vp;
+	vp = fr_pair_afrom_da(parent, da);
```
arr2036, line 486: `MEM()` as this is server code, so we abort on memory allocation errors.

src/lib/server/tmpl_dcursor.c
ndptech: The `case 0` was partly a work round for tmpl parsing currently not handling TLVs / structs without a suffix.  If we want to add a specific number suffix if all the ones up to that already exist, then it's comparing the iterator count with `ar->ar_num` rather than the number of elements in the parent as those could be any attribute.

arr2036: ah good point

PR #4658, opened 2022-08-08
Annotate false positive in use of fr_base16_encode() result (CID #150…

jejones3141:
…3918)

fr_base16_encode() always puts something in the buffer, if only a NUL
terminator, but coverity doesn't notice, hence the annotation.


jejones3141:
Will submit PR that gathers up annotations where coverity doesn't notice sbuff operations that are guaranteed to put data in the buffer.

PR #4666, opened 2022-08-11
v4: Handle nested attributes and multiple protocols with the internal encoder

ndptech:


alandekok:
I think this looks good.  I have no comments.

PR #4667, opened 2022-08-11
v4: Define fr_pair_append_by_parent_da() for automatic creation of nested pairs

ndptech:
plus macros for appending and populating pairs

### DIFF
--- a/src/lib/util/pair.c
+++ b/src/lib/util/pair.c
@@ -1268,6 +1268,75 @@ int fr_pair_prepend_by_da(TALLOC_CTX *ctx, fr_pair_t **out, fr_pair_list_t *list
 	return 0;
 }
 
+/** Alloc a new fr_pair_t, adding the parent attributes if required
+ *
+ * A child pair will be added to the first available matching parent
+ * found.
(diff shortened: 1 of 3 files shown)


src/lib/util/pair.c
```diff
@@ -1232,6 +1232,60 @@ int fr_pair_prepend_by_da(TALLOC_CTX *ctx, fr_pair_t **out, fr_pair_list_t *list
...
+ *	- -1 on failure.
+ */
+int fr_pair_append_by_da_parent(TALLOC_CTX *ctx, fr_pair_t **out, fr_pair_list_t *list, fr_dict_attr_t const *da)
+{
+	fr_pair_t	*vp, *container = NULL;
+
+	/*
+	 *  Walk up the dictionary finding / adding parent attributes first
+	 */
+	if (!fr_dict_attr_is_top_level(da)) {
```
alandekok, line 1256: Perhaps this could be made iterative, like ` fr_pair_find_by_da_nested()` using a da_stack?

That would likely be faster and use less memory.

PR #4668, opened 2022-08-11
v4: Add fr_pair_find_by_da_nested() to find pairs by their da structure

ndptech:


src/lib/util/pair.c
```diff
@@ -716,6 +716,33 @@ fr_pair_t *fr_pair_find_by_da_idx(fr_pair_list_t const *list, fr_dict_attr_t con
...
+
+	if (!fr_dict_attr_is_top_level(da)) {
+		vp = fr_pair_find_by_da_nested(list, da->parent);
+		if (!vp) return NULL;
+
+		fr_assert(fr_type_is_structural(vp->da->type));
+		parent = &vp->vp_group;
+	}
+
+	vp = fr_pair_find_by_da(parent, NULL, da);
```
alandekok, line 742: minor nit: just `return fr_pair_find_by_da(parent, NULL, da)`

arr2036: That's an interesting way of doing the search, but there's a couple of issues.  The first is that it doesn't consider alternative branches, so if you have Leaf-Foo nested in TLV-Bar, but only in the second instance, of TLV-Bar, then this function will fail to find Leaf-Foo.  The second issue is it's quite inefficient to set up the entire stack before starting the search.  You probably didn't spot this (it's not super obvious), but most of the DAs have a pre-built stack containing pointers to their parents.  There are also functions to generate DA stacks dynamically for the ones that don't (i.e. the deeply nested ones).  Using a cached stack or dynamically building one should be transparent to the caller.

Once you have the DA stack, you can start at the top, looking for attributes that match in the list, if you don't find one, you're done, you return NULL.  If you do find one, you then recurse with the child list of that attribute, and an offset da stack, keep doing that until you get to the leaf.  If at any level the attribute doesn't exist, in the next frame back up the stack, you need to loop looking for the next instance of the da at that level, and thus evaluating the entire tree, not just the first path you find through the tree.

arr2036: I guess just as a general thing, do we really want `fr_pair_find_by_da` and `fr_pair_find_by_da_nested`, or do we just want the nested behaviour for `fr_pair_find_by_da`?

ndptech: `fr_pair_find_by_da` has the ability to set a starting entry for the search, which I found used a couple of times in existing code - not sure how that would work with a nested search.

arr2036: It should be equivalent to finding the n'th instance in a flattened version of the tree.

alandekok: I think the most common use of this kind of API is just "find the first one".  In which case this function is fine.

we also need "find the next one by da", with the caveat that the list it's in changes depending on if it has a flat or nested hierarchy.

The good news is that every vp now points to it's parent list, so it's less critical to pass the parent list around everywhere.  It's still useful, but it means we can do fixups without changing everything

src/lib/util/pair.c
```diff
@@ -879,26 +879,112 @@ fr_pair_t *fr_pair_find_by_da_idx(fr_pair_list_t const *list, fr_dict_attr_t con
...
 
-		fr_assert(fr_type_is_structural(vp->da->type));
-		parent = &vp->vp_group;
+	/*
+	 *	If we've been given a value for prev, build out
+	 *	the pairs at each level
+	 */
+	while (prev && (prev->da->depth > 0)) {
+		prev_stack[prev->da->depth] = UNCONST(fr_pair_t *, prev);
+		prev = fr_pair_parent(prev);
```
alandekok, line 917: you can call `fr_proto_da_stack_build(&da_stack, prev->da)` here, too?

alandekok:
this looks good.  Just push them to `master` once the tests are finished.

PR #4673, opened 2022-08-15
Attempt to quiet coverity on fr_value_box_t assignments (CIDs follow)

jejones3141:
CIDs: 1508477, 1508479, 1508480, 1508481, 1508482, 1508483,
      1508484, 1508486

Some members of an fr_value_box_t are const qualified, hence the
complaints about attempts to assign to them or to members of the
union of possible values, some of which are const qualified
(though coverity doesn't appear to complain consistently, e.g. the
FR_TYPE_UINT8 and FR_TYPE_UINT16 cases in the switch in
sim_decode_pair_value() in src/lib/eap_aka_sim/decode.c). We're
trying not to use annotations.

### DIFF
--- a/src/lib/eap_aka_sim/decode.c
+++ b/src/lib/eap_aka_sim/decode.c
@@ -802,7 +802,7 @@ static ssize_t sim_decode_pair_value(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_di
 		break;
 
 	case FR_TYPE_UINT16:
-		memcpy(&vp->vp_uint16, p, sizeof(vp->vp_uint16));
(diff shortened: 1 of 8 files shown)


alandekok:
I think the underlying code is mostly hacks to get around the `const` parts of a `fr_value_box_t`.  It might be better to define a macro `fr_value_box_copy_unsafe()`, and then use that instead of the various `memcpy` things.  That could make it a little clearer.

The underlying `fr_value_box_copy()` function works around the `const` issues through using some macro magic.  But it's fairly heavy-weight, and likely not what we want to use here.

PR #4674, opened 2022-08-15
Annotate use of truncated time for seed (CID #1508485)

jejones3141:
Here the time's just used for a random number seed. 2038 isn't a
problem. From CID #1508485 we know that an explicit cast doesn't
suffice. coverity will still consider it a defect, so we annotate.


alandekok:
I'll push a slightly different fix.

PR #4676, opened 2022-08-16
More "update {....}" to edit against policy.d/*

jpereira:


raddb/policy.d/eap
```diff
@@ -8,14 +8,11 @@ Xeap.authorize {
...
-				&Cache-TTL := 0
-			}
+			&control.Cache-TTL := 0
+
 			cache_eap
 
-			update control {
-				&State !* ANY
-			}
+			&control.State !* ANY
```
alandekok, line 15: This won't work.  It's likely causing CI to fail, too.  You have to do:

```
&control -= &State[*]
```

raddb/policy.d/eap
```diff
@@ -74,9 +71,7 @@ permit_only_eap {
 #
 remove_reply_message_if_eap {
 	if (&reply.EAP-Message && &reply.Reply-Message) {
-		update reply {
-			&Reply-Message !* ANY
-		}
+		&reply.Reply-Message !* ANY
```
alandekok, line 74: Same here.  `!*` is no longer allowed.  If it's not an error in the current parser, then it should be an error.

raddb/policy.d/eap
```diff
@@ -109,16 +102,14 @@ use_tunneled_reply {
...
-		User-Name !* ANY
-		Message-Authenticator !* ANY
-		EAP-Message !* ANY
-		Proxy-State !* ANY
-		MS-CHAP-NT-Enc-PW !* ANY
-		MS-MPPE-Encryption-Types !* ANY
-		MS-MPPE-Send-Key !* ANY
-		MS-MPPE-Recv-Key !* ANY
-	}
+	&reply.User-Name !* ANY
```
alandekok, line 105: These are all `&reply -= &User-Name[*]`, etc.

raddb/policy.d/accounting
```diff
@@ -64,23 +62,17 @@ acct_unique {
...
-		update request {
-			&Acct-Unique-Session-Id := "%{hex:%{md5:%{string:%{User-Name},%{Acct-Multi-Session-ID},%{Acct-Session-ID},%{%{NAS-IPv6-Address}:-%{NAS-IP-Address}},%{NAS-Identifier},%{NAS-Port-ID},%{NAS-Port}}}}"
-		 }
+		&request.Acct-Unique-Session-Id := %{hex:%{md5:%{string:%{User-Name},%{Acct-Multi-Session-ID},%{Acct-Session-ID},%{%{NAS-IPv6-Address}:-%{NAS-IP-Address}},%{NAS-Identifier},%{NAS-Port-ID},%{NAS-Port}}}}
 	}
 
-	update request {
-		&Tmp-String-9 !* ANY
-	}
+	&request.Tmp-String-9 !* ANY
```
alandekok, line 68: This should also be `&request -= &Tmp-String-9[*]`

You can take a pass looking for `!*` and fix them.

alandekok:
Merged manually with some fixups.

PR #4680, opened 2022-08-17
Clean up if statements that don't affect control flow (CID #1504005)

jejones3141:
Since one does a "goto error" no matter what, unlang_function_clear()
only matters for its side effects.


jejones3141:
OK... I should've looked at the history. Will annotate to quiet coverity.,

PR #4689, opened 2022-08-24
fix libressl build

ffontaine:
Fix the following build failure with libressl:

```
src/modules/rlm_eap/types/rlm_eap_fast/rlm_eap_fast.c: In function '_session_secret':
src/modules/rlm_eap/types/rlm_eap_fast/rlm_eap_fast.c:228:47: error: invalid use of incomplete typedef 'SSL' {aka 'struct ssl_st'}
  228 |         eap_fast_session_ticket(tls_session, s->s3->client_random, s->s3->server_random, secret, secret_len);
      |                                               ^~
```

Fixes:
 - http://autobuild.buildroot.org/results/c8df444f4c39f83e254dbb642a5852a1c956f7bb

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/src/main/cb.c
+++ b/src/main/cb.c
@@ -47,7 +47,7 @@ void cbtls_info(SSL const *s, int where, int ret)
 		if (RDEBUG_ENABLED3) {
 			char const *abbrv = SSL_state_string(s);
 			size_t len;
-#if OPENSSL_VERSION_NUMBER >= 0x10100000L
+#if OPENSSL_VERSION_NUMBER >= 0x10100000L && !defined(LIBRESSL_VERSION_NUMBER)
 			STACK_OF(SSL_CIPHER) *client_ciphers;
 			STACK_OF(SSL_CIPHER) *server_ciphers;
 #endif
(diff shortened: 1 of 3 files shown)


arr2036:
@mcnewton @alandekok my feelings about libressl in v4.0.x are if it's going to be done, it needs to be done with feature flag macros... but this is 3.2 which I care very little about, so it is up to you guys :)

alandekok:
I really hate the `ifdef` soup that LibreSSL requires.  There's no easy way to do compatibility flags, so it's all just patching things at each individual call site.  As the patches show, LibreSSL is randomly compatible with various versions of OpenSSL.

There's no simple clean way to do this. :(

alandekok:
These patches are also available in the OpenBSD CVSweb.  http://cvsweb.openbsd.org/cgi-bin/cvsweb/ports/net/freeradius/patches/

ffontaine:
I'm a little bit puzzled.
The PR was closed and the libressl patch doesn't seem to be available in OpenBSB CVSweb (or I didn't found it).
So, should I understand that building with libressl is not supported?
In that case, I'll update buildroot.

alandekok:
The link I posted includes patches for FreeRADIUS to build with libressl.  Please go check.

Building with libressl will be done once that can be done in a clean and maintainable way.  Putting masses of logic in `ifdef` is fragile and hard to maintain.

PR #4706, opened 2022-08-30
enable parallel builds for Debian

jimdigriz:
Enable support for parallel builds for Debian.

Though `compat=9` supports parallel (with an argument) `compat=10` and later automatically enable it when `dpkg-buildpackage` is called with `-j/--jobs`. Debian 'jessie' 8 is still in Extended LTS but fortunately comes with `debhelper` version 10.x, and we only support Debian `stretch` 9 or later anyway so just bump `compat` to 10.

Only fly in the ointment is there is some race condition when running `install-sh`, so we paste over the problem with a `mkdir -p` and pretend everything is peachy.

**N.B.** this also applies cleanly to v3.0.x so can be cherry picked.


alandekok:
Merged manually/

PR #4710, opened 2022-09-01
More convert from 'update {....}' to 'edit' against src/tests/keywords

jpereira:


src/tests/keywords/smash
```diff
@@ -3,6 +3,6 @@
 #
 #  This gives the game away.
 #
-update { &control.User-Name := 'bob', &request.Reply-Message := "foo" }
+&control.User-Name := 'bob', &request.Reply-Message := "foo"
```
alandekok, line 6: This test can be left alone as a test for `update`

src/tests/keywords/update-all
```diff
@@ -3,8 +3,6 @@
 #
```
alandekok, line 3: All of the `update-foo` tests should be left alone, as they're testing the `update` functionality.

Once `update` is gone, these tests can be deleted.

PR #4713, opened 2022-09-02
Annotate false positive tainted_data (CID #1503893)

jejones3141:
In fr_dhcv4_raw_packet_recv() (is there a reason for that
spelling?), coverity claims the downcast of packet->data
in the fr_dhcpv4_packet_get_option() call taints the contents
of packet->data, but it's cast to, and the called function
takes, a const-qualified pointer, so in what sense can it be
tainted?


jejones3141:
Note also the preceding comment.

PR #4724, opened 2022-09-14
More convert 'update' to 'edit'

jpereira:


alandekok (approved):

src/tests/auth/chap
```diff
@@ -5,6 +5,4 @@
 #
 #  Set CHAP-Password
 #
-update request {
-	&CHAP-Password := "%(chap_password:%{request.CHAP-Password})"
-}
+&request.CHAP-Password := "%(chap_password:%{request.CHAP-Password})"
```
alandekok, line 8: `request` is the default, so we don't need to add it all of the time.

src/tests/auth/md5_password
```diff
@@ -1,7 +1,5 @@
 #
 #  over-ride password set in radiusd.conf
 #
-update control {
-	&Password.Cleartext -= ANY
-	&Password.With-Header := '{md5}5d41402abc4b2a76b9719d911017c592'
-}
+&control.Password.Cleartext -= ANY
```
alandekok, line 4: that doesn't work.  See my previous comments in other commits.  You must use

`&control -= &Password.Cleartext[*]`

src/tests/performance/proxy.conf
```diff
@@ -59,9 +57,7 @@ server default {
 
 	recv Accounting-Request {
 		if (!&Event-Timestamp) {
```
alandekok, line 59: this could likely be just

`&Event-Timestamp = "%l"  # only sets it if there's no Event-Timestamp`

PR #4726, opened 2022-09-15
redis_ippool: Implement static leases

mkhon:
Static leases can be created in any pool using the following commands:

SET {POOL}:device:OWNER IP
HSET {POOL}:ip:IP is_static true
ZREM {POOL}:pool IP

### DIFF
--- a/src/modules/rlm_redis_ippool/rlm_redis_ippool.c
+++ b/src/modules/rlm_redis_ippool/rlm_redis_ippool.c
@@ -211,41 +211,58 @@ static char lua_alloc_cmd[] =
 	 */
 	"exists = redis.call('GET', owner_key);" EOL							/* 8 */
 	"if exists then" EOL										/* 9 */
-	"  local expires_in = tonumber(redis.call('ZSCORE', pool_key, exists) - ARGV[1])" EOL		/* 10 */
-	"  if expires_in > 0 then" EOL									/* 11 */
-	"    ip = redis.call('HMGET', '{' .. KEYS[1] .. '}:"IPPOOL_ADDRESS_KEY":' .. exists, 'device', 'range', 'counter')" EOL	/* 12 */
-	"    if ip and (ip[1] == ARGV[3]) then" EOL							/* 13 */
-	"      if expires_in < tonumber(ARGV[2]) then" EOL						/* 14 */
-	"        redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2], exists)" EOL			/* 15 */
-	"        expires_in = tonumber(ARGV[2])" EOL							/* 16 */
-	"      end" EOL											/* 17 */
-	"      return {" STRINGIFY(_IPPOOL_RCODE_SUCCESS) ", exists, ip[2], expires_in, ip[3] }" EOL	/* 18 */
-	"    end" EOL											/* 19 */
-	"  end" EOL											/* 20 */
-	"end" EOL											/* 21 */
+	"  address_key = '{' .. KEYS[1] .. '}:"IPPOOL_ADDRESS_KEY":' .. exists" EOL 			/* 10 */
+	"  ip = redis.call('HMGET', address_key, 'device', 'range', 'counter', 'is_static', 'expires_at')" EOL	/* 11 */
+	"  if ip then" EOL										/* 12 */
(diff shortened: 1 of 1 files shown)


arr2036:
>The real use-case for static (or sticky) leases is to create a static binding between OWNER and IP
So that when IP is allocated you need to find static leases by OWNER - thus modifying only ZSET is not enough.
Iterating over HSET in order to find the IP address by OWNER is not an option here.

But there's already an explicit association between the owner and IP address?  I don't really see what the issue is here.

- Resolve owner to existing address: https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_redis_ippool/rlm_redis_ippool.c#L212
- Lookup IP in ZSET: https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_redis_ippool/rlm_redis_ippool.c#L214
- If IP has not expired, retrieve metadata: https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_redis_ippool/rlm_redis_ippool.c#L216

For update we're actually cheating a bit here:
- https://github.com/FreeRADIUS/freeradius-server/blob/master/src/modules/rlm_redis_ippool/rlm_redis_ippool.c#L287 we should really be checking the _current_ expiry value, and if we do that, and find it's +inf we could short circuit the ZADD and 'expire' operation on the owner key.  I guess adding a flag to the ip address metadata for "is_static" would allow us to continue cheating, and not have the performance penalty of a ZSCORE for renews.  If you've got a way of benchmarking easily it might be worth examining both approaches to see how big of a performance penalty ZSCORE adds.

mkhon:
Another set of changes is that we don't want to break renewal logic for the clients, so they should get sane lease time values. Also, as the IP for static leases is removed from ZSET, we can not use the ZSET score as a lease time and keep it in HSET instead.

Given this, my patch does exactly what you described - IP address is removed from ZSET, and is kept in SET on release.

arr2036:
@mkhon Infinity is a sane lease time... Not only is it sane, it's explicitly supported by[ DHCPv4](https://www.rfc-editor.org/rfc/rfc2131.html#section-3.3) and [DHCPv6](https://www.rfc-editor.org/rfc/rfc8415.html#page-2) and the correct value to be handing back here.  Anything else is a site specific customisation that you should be handling in policy.  It should NOT be hardcoded into the module.

> Also, as the IP for static leases is removed from ZSET, we can not use the ZSET score as a lease time and keep it in HSET instead.
I'm not using your patch set as a basis for my statements.  I'm talking about the minimum changes to the existing code in order to get static IPs functional.

mkhon:
> @mkhon Infinity is a sane lease time... Not only is it sane, it's explicitly supported by[ DHCPv4](https://www.rfc-editor.org/rfc/rfc2131.html#section-3.3) and [DHCPv6](https://www.rfc-editor.org/rfc/rfc8415.html#page-2) and the correct value to be handing back here. Anything else is a site-specific customisation that you should be handling in policy. It should not be the default behaviour of the ippool module.
 
The default behavior is to return Lease-Time as configured by `lease_time` option and we needed this behavior. Of course we can always override this if static leases will be implemented as you mentioned.

I should note that `sql_ippool` module does not have this "feature" in the provided examples - both static and dynamic allocations use the same `lease_duration` parameter as configured in the `sqlippool` configuration files.

In general, lease duration should not depend on whether the lease static or dynamic - it should be set as configured (`lease_durtation` or `lease_time`). This is the feature that I wanted to keep when implementing static leases and that seem to be lost with your suggestion.

Additionally, I do not see that your changes are less intrusive (judging by the number of lines changed). I am ok with them as long as some support for static leases is in and do not mind which implementation is better.

But we will have to fix the behavior of `lease_time` elsewhere and it will not match the behavior of sqlippool module.

mkhon:
> This is what I'm proposing
> [arr2036@9f42360](https://github.com/arr2036/freeradius-server/commit/9f423600f4cadde3e4d2251eb7332e5ed9713cc3)

I think you're missing skipping the results of `ZREVRANGE` which have INF lease time.

mkhon:
@arr2036 can we get back to this?

arr2036:
Alternative scheme implemented in: https://github.com/FreeRADIUS/freeradius-server/commit/df48890e9ccbb499dce2caa14039dcc6d31a1ebd and others

PR #4735, opened 2022-09-20
More "update {....}" to edit against src/tests/modules/*

jpereira:


src/tests/modules/eap_sim/sim_xlat_id_aka.unlang
```diff
@@ -98,12 +86,8 @@ if (&control.Tmp-String-1 != "0%{User-Name[1]}") {
...
 #  2.1 - Encrypt the permanent ID for fastauth
 #
-update control {
-	&Tmp-String-0 := '1420032219455259'
-}
-update control {
-	&User-Name[1] := "%(3gpp_temporary_id_encrypt:%{User-Name[1]} %{control.Tmp-String-0} 6 fastauth)"
-}
+&control.Tmp-String-0 := '1420032219455259'
+&control.User-Name[1] = "%(3gpp_temporary_id_encrypt:%{User-Name[1]} %{control.Tmp-String-0} 6 fastauth)"
```
alandekok, line 90: That should be fixed to append the `User-Name` to the `control` list.  As discussed in chat, assigning *new* values to index `[1]` is almost always wrong.

src/tests/modules/exec/sync.unlang
```diff
@@ -1,37 +1,33 @@
...
-if (&Tmp-String-0 != "") {
+if (&Tmp-String-0) {
 	test_fail
 }
 
-update request {
-	&Tmp-String-0 := "%(exec_sync:/bin/sh $ENV{MODULE_TEST_DIR}/fail.sh)"
-}
+&request -= &Tmp-String-0
+&Tmp-String-0 := "%(exec_sync:/bin/sh $ENV{MODULE_TEST_DIR}/fail.sh)"
```
alandekok, line 24: You don't need to delete it, and then assign it using `:=`.  You can just use `:=`.  This is documented in the `edit.adoc` file.

PR #4736, opened 2022-09-21
Initial round of moving rlm_ldap to be fully async

ndptech:
- Use existing async bind in mod_authenticate
- Define functions for async retrieval of user DN
- Use async retrieval of user DN in mod_authenticate

### DIFF
--- a/src/modules/rlm_ldap/rlm_ldap.c
+++ b/src/modules/rlm_ldap/rlm_ldap.c
@@ -176,6 +176,17 @@ global_lib_autoinst_t const *rlm_ldap_lib[] = {
 	GLOBAL_LIB_TERMINATOR
 };
 
+/** Holds state of in progress async authentication
+ *
+ */
+typedef struct {
+	char const		*dn;
+	char const		*password;
+	rlm_ldap_t const	*inst;
+	fr_ldap_thread_t	*thread;
+	fr_ldap_thread_trunk_t	*ttrunk;
+} ldap_auth_ctx;
+
 
 static xlat_arg_parser_t const ldap_escape_xlat_arg = { .required = true, .concat = true, .type = FR_TYPE_STRING };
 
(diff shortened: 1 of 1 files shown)


src/modules/rlm_ldap/user.c
```diff
@@ -35,6 +35,138 @@ USES_APPLE_DEPRECATED_API
...
+		if (tmpl_aexpand(ctx, &user_ctx->filter, request, inst->userobj_filter,
+				 fr_ldap_escape_func, NULL) < 0) {
+			REDEBUG("Unable to create filter");
+			return -1;
+		}
+	}
+
+	if (tmpl_aexpand(ctx, &user_ctx->base_dn, request, inst->userobj_base_dn, fr_ldap_escape_func, NULL) < 0) {
+		REDEBUG("Unable to create base_dn");
+		return -1;
```
jpereira, line 163: Maybe we need to call `TALLOC_FREE(user_ctx->filter)` here, due to being allocated previously by `tmpl_aexpand()`

ndptech: I'll re-work the allocation hierarchy so we can `TALLOC_FREE(user_ctx)` and it will clear up properly

src/modules/rlm_ldap/rlm_ldap.h
```diff
@@ -150,6 +150,18 @@ extern HIDDEN fr_dict_attr_t const *attr_user_name;
 /*
  *	user.c - User lookup functions
  */
+#define rlm_ldap_find_user_cached(_dn, _request) \
```
arr2036, line 153: just static inline it...  Macros are fine if you want to manipulate other variables in scope directly, or implement foreach loops, but they have no advantages over static inlines, and are harder to debug in a compiler.

src/modules/rlm_ldap/user.c
```diff
@@ -35,6 +35,136 @@ USES_APPLE_DEPRECATED_API
...
+				REDEBUG("%s", dn);
+				ldap_memfree(dn);
+			}
+			REXDENT();
+			RETURN_MODULE_INVALID;
+		}
+	}
+
+	entry = ldap_first_entry(query->ldap_conn->handle, query->result);
+	if (!entry) {
```
arr2036, line 80: Just to make sure... There's a separate code path which deals with users that weren't found, right?  Else this should be `RETURN_MODULE_NOTFOUND`

src/modules/rlm_ldap/user.c
```diff
@@ -35,6 +35,136 @@ USES_APPLE_DEPRECATED_API
...
+	fr_ldap_util_normalise_dn(dn, dn);
+
+	RDEBUG2("User object found at DN \"%s\"", dn);
+
+	MEM(pair_update_control(&vp, attr_ldap_userdn) >= 0);
+	fr_pair_value_strdup(vp, dn, false);
+	ldap_memfree(dn);
+	if (user_ctx->out) *user_ctx->out = user_ctx->query;
+
+	return UNLANG_ACTION_CALCULATE_RESULT;
```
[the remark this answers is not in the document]
arr2036: I think/hope the interpreter records the last rcode in `request->rcode` so we can get at it easily, otherwise there's going to need to be some way of explicitly passing a result back in the uctx.

src/modules/rlm_ldap/user.c
```diff
@@ -35,6 +35,136 @@ USES_APPLE_DEPRECATED_API
...
+	user_ctx = talloc_zero(ctx, ldap_user_find_ctx);
+	*user_ctx = (ldap_user_find_ctx) {
+		.inst = inst,
+		.ttrunk = ttrunk,
+		.attrs = attrs,
+		.out = query_out
+	};
+
+	if (inst->userobj_filter) {
+		if (tmpl_aexpand(user_ctx, &user_ctx->filter, request, inst->userobj_filter,
```
arr2036, line 145: Can you switch these out to the async variant if you're writing async code :)  I think it's somewhere in src/lib/unlang... It might be worth spending a little time figuring out how to flesh out that code for general consumption once this initial patch set is in.

arr2036: For yielding in between expansions it might be worth pushing an unlang function onto the stack, and then changing its resume function to represent the different expansion states.

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -1025,16 +1077,21 @@ static unlang_action_t CC_HINT(nonnull) mod_authenticate(rlm_rcode_t *p_result,
...
+		.inst = inst,
+		.ttrunk = ttrunk
+	};
 
 	/*
-	 *	Attempt a bind using the thread specific connection for bind auths
+	 *	Check for a cahed copy of the DN
 	 */
-	if (fr_ldap_bind_auth_async(request, thread, dn, password->vp_strvalue, true) < 0) RETURN_MODULE_FAIL;
+	rlm_ldap_find_user_cached(auth_ctx->dn, request);
```
arr2036, line 1091: Do we need to do something different here based on the return code?

arr2036, line 1091: `rlm_find_user_dn_cached`... that confused me for a moment... if we're looking up a cached user why aren't we checking the return code :)

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -1025,16 +1077,21 @@ static unlang_action_t CC_HINT(nonnull) mod_authenticate(rlm_rcode_t *p_result,
 
 	RDEBUG2("Login attempt by \"%pV\"", &username->data);
 
-	/*
-	 *	Get the DN by doing a search.
-	 */
-	dn = rlm_ldap_find_user(inst, request, ttrunk, NULL, false, NULL, NULL, &rcode);
-	if (!dn) RETURN_MODULE_RCODE(rcode);
+	auth_ctx = talloc_zero(unlang_interpret_frame_talloc_ctx(request), ldap_auth_ctx);
```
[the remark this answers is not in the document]
arr2036: Yeah. just checking through the code. I made some changes a few months ago to remove all the old code that handles module methods by section identifier.  You can just add a couple of extra fields to `module_method_name_t` like `size_t rctx_size`  and `char const *rctx_type`, then check for them the first time a module call is being made, and allocate the appropriate rctx in the ctx supplied by `unlang_interpret_frame_talloc_ctx`.  This wasn't possible before as many of the module methods were stuck in an array indexed by section code, so we couldn't have extra fields.

PR #4737, opened 2022-09-21
_Generic for SBUFF_PARSE_{U}INT_DEF(), v2.0. (CID #1504004, #1504045)

jejones3141:
It doesn't look like coverity will fix the annotation in macro definitions issue any time soon, so that leaves _Generic() as the only way to not generate known false comparisons for long long or unsigned long long which in turn make coverity complain about operands not affecting results.


jejones3141:
There's a better equivalent formulation of the underflow/overflow check. Bye, _Generic()...

PR #4742, opened 2022-09-23
Correct error check on # of bits to shift (CIDs #1515536, #1515537)

jejones3141:


alandekok:
Pushed a fix already.

PR #4759, opened 2022-10-08
Deal with remaining declarations after statements.

jejones3141:
These cases are less trivial than double semicolons between declarations. Some just require moving declarations to the front of the block, others require otherwise gratuitous blocks.


jejones3141:
Merging with config changes to detect declaration after statement.

PR #4764, opened 2022-10-11
Explicitly ignore the curl_mime_type() return (CID #1504071)

jejones3141:
This includes the curl_mime_type() call that gives rise to the
    coverity defect, but also checks the mime and mime part allocations
    that must work for body_init() to work.

### DIFF
--- a/src/modules/rlm_smtp/rlm_smtp.c
+++ b/src/modules/rlm_smtp/rlm_smtp.c
@@ -674,7 +674,7 @@ static int body_init(fr_mail_ctx_t *uctx, curl_mime *mime)
 	int 		body_elements = 0;
 
 	/* Initialize a second mime to apply special conditions to the body elements */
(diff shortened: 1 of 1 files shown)


src/modules/rlm_smtp/rlm_smtp.c
```diff
@@ -700,7 +700,7 @@ static int body_init(fr_mail_ctx_t *uctx, curl_mime *mime)
 	 */
 	part = curl_mime_addpart(mime);
 	curl_mime_subparts(part, mime_body);
-	curl_mime_type(part, "multipart/mixed" );
+	(void) curl_mime_type(part, "multipart/mixed" );
```
arr2036, line 703: Can you wrap this in MEM then :)

jejones3141: Will do.

PR #4771, opened 2022-10-15
Fallback to `uname -n` on missing `hostname` in radtest

eworm-de:
This should work without inetutils being installed, so add a fallback.

Signed-off-by: Christian Hesse <mail@eworm.de>


eworm-de:
Thanks! I would like to see this in `v3.2.x` branch. Will it be cherry-picked automatically or should I send another request?

PR #4774, opened 2022-10-17
Check fr_event_timer_in() return value (CID #1503926)

jejones3141:


jejones3141:
The majority of uses of `fr_event_timer_in()`, 25 of them, do check the result. At least one of the uses that don't check have some hard-to-unwind things done before the call; I will look again to see if there's a way around it.

alandekok:
The larger question is when the result is checked, is there anything that can be done?  i.e. are `fr_event_timer_in()` failures _fatal_ errors or not?

If the errors are transient and recoverable, then they should be caught and handled.  If the errors are fatal (OOM, internal sanity check failure), then it's better to just abort in `fr_event_timer_in()`

PR #4777, opened 2022-10-18
Port closefrom() calls in exec.c to OpenBSD

guijan:
OpenBSD has a unique version of the closefrom() function which returns int and can be interrupted. Every other OS follows the original Solaris API which returns void and has no error condition. It's possible to be compatible with both.

### DIFF
--- a/src/lib/server/exec.c
+++ b/src/lib/server/exec.c
@@ -309,7 +309,15 @@ static NEVER_RETURNS void exec_child(request_t *request, char **argv, char **env
 	 *	want to leave dangling FD's for the child process
 	 *	to play funky games with, so we close them.
 	 */
-	closefrom(STDERR_FILENO + 1);
+	do {
+		/*
+		 *	Solaris' closefrom returns void.
+		 *	OpenBSD's closefrom returns int and can be EINTR'd.
+		 *	Calling it like this is compatible with both APIs.
+		 */
+		errno = 0;
+		closefrom(STDERR_FILENO + 1);
+	} while (errno == EINTR);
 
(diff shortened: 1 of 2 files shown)


alandekok:
I've pushed a different fix.  We prefer to do OS-specific things like this in a macro (once), and not throughout the code.

guijan:
I'm adding that function with the OpenBSD API to my OpenBSD compatibility library at [guijan/libobsd](https://github.com/guijan/libobsd). Could you make it so the macro is used unconditionally?

alandekok:
There is no reason to have all platforms use code which works around a problem on only one platform.

If you want to do OS-specific magic, you have the source code, and you can do it.

But we have to make FreeRADIUS work on many different operating systems.  If we put OS-specific fixes into every platform, then the project would soon become unmaintainable.

guijan:
That's true, but this trick in particular works on either version of the API, that's why I didn't use #ifdef. It's like calling malloc(1) instead of malloc(0) to be compatible with systems that return NULL and also systems that return a unique pointer.

I plan to make FreeRADIUS work with both Freedesktop's libbsd and my libobsd, hopefully this means some compat code can be dropped in the future.

alandekok:
The "trick" is _useless_ and _irrelevant_ on every platform other than OpenBSD.  The correct way to fix things for one broken platform is to have fixes _only_ for that platform.

Your libraries are nice, but we _cannot_ and _will not_ require that everyone use them in order to run FreeRADIUS.  FreeRADIUS **MUST** build and work on all platforms, with minimal dependencies.

PR #4783, opened 2022-10-22
Fix comment on status line length in rlm_rest

Natureshadow:
I stumbled upon this minor miscalculation (actually, a serious of miscalculations – you only need to count wrong often enough to make it work out in the end ;)) when debugging HTTP/2 issues on an old version of FreeRADIUS.

There's nothing wrong here (my FreeRADIUS was simply too old), it's only a minor issue in the comment.


arr2036:
Haha, wow I don't know how I screwed that up so badly.  At least the code's ok (hopefully).

PR #4784, opened 2022-10-24
v4: Add functionality to rate limit LDAP sync store Cookie

ndptech:
Two additional configuration items define when to store cookies
 - cookie_interval - maximum time between storing cookies
 - cookie_changes - maximum number of changes to be processed between storing cookies

Received cookies are stored in the list of packets being processed, in sequence, and are passed to a worker to run the `store Cookie` section if all preceding changes have been processed and either the time interval has passed or sufficient changes have been processed.

### DIFF
--- a/src/listen/ldap_sync/persistent_search.c
+++ b/src/listen/ldap_sync/persistent_search.c
@@ -46,10 +46,11 @@
  * an ldap_abandon message to the server to tell it to cancel the search.
  *
  * @param[in] conn 		Connection to issue the search request on.
- * @param[in] config		containing callbacks and search parameters.
+ * @param[in] sync_no		number of the sync in the array of configs.
+ * @param[in] inst		instance of ldap_sync this query relates to.
  * @param[in] cookie		not applicable to persistent search LDAP servers.
  */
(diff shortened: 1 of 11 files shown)


arr2036:
It's a bit more complicated than this.  Here's the conditions for doing a cookie store, for a given entry.

- All entries before the candidate entry have completed processing.  e.g. if we have entries A, B, C, then C can only become a checkpoint if A and B have finished processing.  We can't have "holes".
- A specific number of entries have been processed between the previous checkpoint and this checkpoint.
OR
- A specific amount of time has passed between the previous checkpoint and this checkpoint.

src/listen/ldap_sync/persistent_search.c
```diff
@@ -129,6 +129,13 @@ int persistent_sync_state_init(fr_ldap_connection_t *conn, size_t sync_no, proto
 
 	DEBUG3("Sync created with msgid %i", sync->msgid);
 
+	/*
+	 *	Register event to store cookies at a regular interval
+	 *	Whilst persistent search LDAP servers don't provide cookies as such
+	 *	we treat change numbers, if provided, as cookies.
+	 */
+	fr_event_timer_in(sync, conn->conn->el, &sync->cookie_ev, inst->cookie_interval, ldap_sync_cookie_event, sync);
```
arr2036, line 137: Probably a good idea to delete this when the connection is closed

PR #4789, opened 2022-10-30
Fix the use of the ctype classification and mapping functions.

he32:
Even though the argument type for these functions is `int`, there is only a restricted subset of values which are actually valid. The values are either EOF (which is negative) or the values representable as an `unsigned char`, says the standard.

On NetBSD with pkgsrc, where we typically build with -Wchar-subscripts, the existing use (without casts of arguments) typically gets us
   warning: array subscript has type 'char'

Someone has apparently been in here before to silence warnings and added casts to int, which will invoke undefined behaviour for half of the values of a `signed char` argument, as a negative `signed char` will be sign-extended to a negative `int` by the cast, and most probably not to the EOF value.

I could not offhand see any use which would deal with EOF, so added casts to `u_char` to make the behaviour well-defined, and this should both properly get rid of the warnings and the undefined behaviour.

### DIFF
--- a/scripts/build/dlopen.c
+++ b/scripts/build/dlopen.c
@@ -597,7 +597,7 @@ static void ad_have_feature(char const *symbol)
 
 	for (p = def->name + 5; *p != '\0'; p++) {
 		if (islower((int) *p)) {
-			*p = toupper((int) *p);
+			*p = toupper((u_char) *p);
 
 		} else if ((*p == '/') || (*p == '.')) {
 			*p = '_';
@@ -667,7 +667,7 @@ static void ad_update_variable(char const *name, char *value)
 
 	p = strstr(old, value);
 	if (p) {
-		if (!p[value_len] || isspace((int) p[value_len])) {
+		if (!p[value_len] || isspace((u_char) p[value_len])) {
 			gmk_free(old);
 			return;
 		}
@@ -1007,7 +1007,7 @@ static char *make_ad_dump_defines(__attribute__((unused)) char const *nm, unsign
 	FILE *fp;
 	unsigned int i;
 
-	if ((argc == 0) || !*argv[0] || isspace((int) *argv[0])) {
+	if ((argc == 0) || !*argv[0] || isspace((u_char) *argv[0])) {
 		/*
 		 *	Print Makefile rules to redefine the variables we've created.
 		 */
--- a/scripts/jlibtool.c
+++ b/scripts/jlibtool.c
@@ -1062,7 +1062,7 @@ static int parse_long_opt(char const *arg, command_t *cmd)
 		/*
 		 *	Smash the target to lower case
 		 */
-		for (i = 0; i < len; i++) value[i] = tolower(value[i]);
+		for (i = 0; i < len; i++) value[i] = tolower((u_char)value[i]);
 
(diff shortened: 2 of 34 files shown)


alandekok:
The casts to `u_char` should likely be to `uint8_t`, as that's what we use everywhere else.

There's a larger problem, however.  The code in the `master` branch is _not_ part of any released version.  If it works, great.  If not, we suggest sending patches.

Since it isn't part of an official release, it isn't officially supported.  And it has no business being part of `pkgsrc`.  The code may or may not build on a daily basis.  The configuration file format may change without warning.

It really should not be shipped to thousands of unsuspecting people who install it from a package, and then are surprised that it doesn't do what they want.

he32:
I admit to walking in here without knowing the specific conventions for this project, and apparently mistakenly thought the code in the `master` branch would be where the code is maintained. (The fact that `./configure; make` didn't work ought to have been a clue...)

Secondly, I think the ctype(3) man page specifically talks about `unsigned char`, and even though it's pretty far-fetched to have a system where `unsigned char` and `uint8_t` has a different base type, is that guaranteed to be the case?

The pkgsrc package is currently based on version 3.0.25.  I can of course also submit a set of patches to the `v3.0.x` branch.

alandekok:
For data types, the issue is consistency.  Nothing else in the code uses `u_char`, so it should be avoided on general principle.  We use `uint8_t` for "unsigned 8-bit value".

arr2036:
Agreed.  The entire codebase uses `uint8_t *` to mean a binary buffer, I don't see any reason to change that... If there's a differing number of bits in `u_char` to a `uint8_t` then all kinds of assumptions break.

arr2036:
Woo look at those build failures fly...

he32:
Well.  At least all didn't fail.  Or ... Actually, the FreeBSD build gets a green checkmark, but still failed -- that should have been a red checkmark, no?  It looks like I have been using a NetBSDism, `u_char` is apparently only defined under `_NETBSD_SOURCE` in our `<sys/types.h>`.  I'll take your guidance and use `uint8_t` in preference to `unsigned char` (which is longish) in my pull request for the v3.0.x branch, and hopefully that goes smoother.  (I need to delete my fork and re-do it to include all the branches from the original, by the looks of it, unless I get some help how to replicate them "after the fact".)

alandekok:
The simplest way to do 99% of this work is something like this:

```
perl -p -i -e 's/(tolower|toupper|isdigit|isalpha|isspace|isxdigit)\(\(int\)/${1}((uint8_t)/g' $(find . -name "*.[ch]" -print)
```

That catches most things automatically.

alandekok:
See commit d80731fd1f in the v3.0.x branch.  This should fix almost all of the issues.

he32:
> The simplest way to do 99% of this work is something like this:
> 
> ```
> perl -p -i -e 's/(tolower|toupper|isdigit|isalpha|isspace|isxdigit)\(\(int\)/${1}((uint8_t)/g' $(find . -name "*.[ch]" -print)
> ```
> 
> That catches most things automatically.

Not quite.  I think that in my review, about a third of the invocations of those functions had the `(int)` cast already, and the rest had no cast, and the data type passed was typically a `char`, leading to possibly undefined behaviour.

alandekok:
@he32 Please read the actual commit message, and check the commit.  The changes I pushed should catch pretty much everything.

PR #4799, opened 2022-11-11
Deal with dl_open_by_name() toctou (CIS #1400053)

jejones3141:
dl_open_by_name() gives you the option of looking though a list of directories for a library, using the dlopen() function. If a dlopen() succeeds, dl_open_by_name() succeeds and all is well. If it fails, though, the question is why? If it doesn't exist in that directory, you want to try the next on the list. If it does, permissions are likely to be the problem, so report the issue and return failure.

The previous code uses access() on the path if the dlopen() fails, hence toctou.

Unfortunately, there's not a good way to deal with it.

There is no dlopenat(), so one can't avoid toctou with dlopenat() followed by accessat(). The only indication of why dlopen() failed is in the string returned by dlerror()... which only guarantees that the string is human-readable, NUL-terminated, and doesn't end with a newline.

Looking at the result of dlerror() when the dlopen() for the directory list fails suggests that it has the following format:

path COLON SPACE message COLON SPACE strerror_output

In the particular cases in the tests, that last part was always "No such file or directory", what strerror() returns for the ENOENT dl_open_by_name() is checking for.

As the least bad alternative, the code looks at the dlerror() output to determine whether to continue or get out of the loop.

### DIFF
--- a/src/lib/util/dl.c
+++ b/src/lib/util/dl.c
@@ -528,7 +528,7 @@ dl_t *dl_by_name(dl_loader_t *dl_loader, char const *name, void *uctx, bool uctx
 
 		ctx = paths = talloc_typed_strdup(NULL, search_path);
 		while ((path = strsep(&paths, ":")) != NULL) {
-			int access_mode = R_OK | X_OK;
+			char *dlerror_txt;
 
 			/*
 			 *	Trim the trailing slash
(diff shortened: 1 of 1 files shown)


arr2036:
I wouldn't compare with a static string, maybe to the output of `fr_strerror()` with an appropriate errno... I'd also fail back to `access` if the error doesn't match (and just shut up coverity manually)... I don't think is going to be reliable across multiple platforms... and it's not really a security issue.

jejones3141:
It's now doing `strstr()` against `fr_syserror(ENOENT)`, which leaves the fallback to ``access`. Which platforms are verified, and how can I determine that at compile time?

jejones3141:
OK. It will now use `access()` if __linux__ isn't defined.

jejones3141:
Apologies; forgot to clean up history. That's done now.

PR #4802, opened 2022-11-14
Add needed socket closure to avoid resource leak (CID #1516897)

jejones3141:


alandekok:
Thanks.  I've got a fix in my local repo which will get pushed as soon as I iron out other issues.

jejones3141:
Okay, thanks.

PR #4810, opened 2022-11-22
v4: Add limit for number of outstanding ldap sync packets

ndptech:
To push back pressure to the LDAP server when processing is lagging


alandekok:
Seems fine to me.

arr2036:
Yeah the sanitizer failures are @alandekok's code

PR #4813, opened 2022-11-23
rlm_redis_ippool: Fix lua_alloc_cmd

mkhon:
This fixes the following error:

```
(0)      redis_ippool - ERROR: (0) error   : ERR Error running script (call to f_555f9a10a670669b1725edfc4b558c4c5eb3f9ad): @user_script:14: user_script:14: attempt to compare number with string
```

which appeared after 587fcafdcd3350cba4319cbc3c5f2ce9fc8ec2f4
Looks like rlm_redis_ippool.c change in this commit was accidental (and thus not tested)


mkhon:
@arr2036 please review

PR #4814, opened 2022-11-23
Fix RPM build: pass $(PKG_VERSION) to the rpm-build

mkhon:


mcnewton:
Need a bit more than this, I've been looking at it and will push a more complete fix.

PR #4815, opened 2022-11-23
Unshallow git clone so that 'git describe' works properly during the build

mkhon:


mcnewton:
Why `unshallow` in this Dockerfile, rather than just do a full clone in `Dockerfile.deps`? But in either case, I'm not sure either is the correct solution, we should probably just pass the version in from the `dockerbuild` script, or otherwise use the version from the `VERSION` file.

`Dockerfile.deps` probably needs to die now though, it's left over from past now obsolete CI usage and likely can be merged into the main Dockerfiles now, like v3.

mkhon:
> Why `unshallow` in this Dockerfile, rather than just do a full clone in `Dockerfile.deps`? But in either case, I'm not sure either is the correct solution, we should probably just pass the version in from the `dockerbuild` script, or otherwise use the version from the `VERSION` file.
> 
> `Dockerfile.deps` probably needs to die now though, it's left over from past now obsolete CI usage and likely can be merged into the main Dockerfiles now, like v3.

Which version do you mean to get from VERSION file? Dockerfile.deps does shallow clone to get the dependencies

arr2036:
Any more thoughts on this @mcnewton

arr2036:
I'm fine merging this, but the unshallow checkout needs to be optional and off by default.  Feel free to re-submit the PR with those changes.

mcnewton:
I'm currently in the process of redoing all the v4 Dockerfiles from templates anyway (as done recently in v3), so any suggested changes should hold off until it's been updated.

PR #4818, opened 2022-11-25
Add UBNT dictionary

zhangyoufu:
# Submission checklist master branch

- [x] Vendor prefix NOT added for vendor attributes
- [x] Parent prefix NOT added for tlv attributes
- [ ] Dictionary tested by starting the server with the new dictionary loaded (radiusd -C).
- [x] Dictionary run through [format.pl](https://github.com/FreeRADIUS/freeradius-server/blob/master/scripts/dict/format.pl) (scripts/dict/format.pl <path to dictionary>).
- [x] New dictionaries only - Dictionary added to the include list in the top level dictionary file of the protocol dictionary.

### DIFF
--- a/share/dictionary/radius/dictionary
+++ b/share/dictionary/radius/dictionary
@@ -309,6 +309,7 @@ $INCLUDE dictionary.trapeze
 $INCLUDE dictionary.travelping
 $INCLUDE dictionary.tripplite
 $INCLUDE dictionary.tropos
+$INCLUDE dictionary.ubnt
 $INCLUDE dictionary.ukerna
 $INCLUDE dictionary.unisphere
 $INCLUDE dictionary.unix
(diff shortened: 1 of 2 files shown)


zhangyoufu:
I have no idea how to enforce separate `Tunnel-Password-Entry` in `Access-Accept` reply.

Something like

```
update reply {
    &UBNT-Tunnel-Password += "12345678"
    &UBNT-Tunnel-Password-Keyid += ""
    &UBNT-Tunnel-Password += "87654321"
    &UBNT-Tunnel-Password-Keyid += "87654321"
    &UBNT-Tunnel-Password += "deadbeef"
    &UBNT-Tunnel-Password-Keyid += ""
    &UBNT-Tunnel-Password += "badc0ffe"
    &UBNT-Tunnel-Password-Keyid += "badc0ffe"
}
```

does not work for me on 3.0.X branch.

alandekok:
The issue of v3 if a bit more complex.  It does support nested attributes, but it's more complex than with v4.

PR #4819, opened 2022-11-25
radict: Add new options

jpereira:
Adding the new options:
-c print out in CSV format
-H print out the Headers

It might be helpful to lookup the id from the attributes.

### DIFF
--- a/src/bin/radict.c
+++ b/src/bin/radict.c
@@ -40,8 +40,15 @@ RCSID("$Id$")
 #endif
 #include <assert.h>
 
+typedef enum {
+	RADICT_OUT_FANCY = 1,
+	RADICT_OUT_CSV
+} radict_out_t;
+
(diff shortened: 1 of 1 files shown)


jpereira:
Example:

```
 [jpereira@Sugarloaf:freeradius-server.git]$ ./build/bin/local/radict  -D share/dictionary/ -p radius -H User-Password
 Dictionary    OID    Attribute    ID    Type    Flags
 RADIUS    User-Password    User-Password    2    string    encrypt=1
 [jpereira@Sugarloaf:freeradius-server.git]$ ./build/bin/local/radict  -D share/dictionary/ -p radius -H -c User-Password
 Dictionary,OID,Attribute,ID,Type,Flags
 RADIUS,User-Password,User-Password,2,string,encrypt=1
 [jpereira@Sugarloaf:freeradius-server.git]$
```

PR #4821, opened 2022-11-30
Annotate coverity missing_unlock issues (CIDs below)

jejones3141:
NOTE: It's unclear whether an annotation behind a macro
will have the desired effect.

1233595,
1206499,
1233595: coverity doesn't notice that all error returns
         from exfile_open() leave the mutex unlocked, and
         that exfile_close() unlocks the mutex
1414423: connection_find() intentionally leaves the mutex
         locked if it finds a connection
1414434: connection_spawn() intentionally leaves the mutex
         locked if the unlock parameter is false

### DIFF
--- a/src/lib/server/pool.c
+++ b/src/lib/server/pool.c
@@ -295,6 +295,7 @@ static fr_pool_connection_t *connection_find(fr_pool_t *pool, void *conn)
 #endif
 
 			fr_assert(this->in_use == true);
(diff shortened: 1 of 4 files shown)


arr2036:
This is really the wrong way to achieve this.  You should submit a model which describes the behaviour of these functions as locking/unlocking.

jejones3141:
Will use model instead; closing this PR.

PR #4822, opened 2022-12-05
close files when coa_init() succeeds (CIDs #1517228, 1517229)

jejones3141:


alandekok:
Thanks, I've got some other cleanups pending which will fix this.

PR #4827, opened 2022-12-15
opendir may fail

lifeiyu2022:


alandekok:
That patch returns `1` for "success", and checks the `quiet` flag before printing a message.

It should be an `ERROR`, and a fatal one.  I've pushed a fix.

PR #4828, opened 2022-12-19
Disable developer build explicitly when building without _with_developer as it is implicitly enabled when building from git

mkhon:
debian build does not have this problem as it uses --disable-developer explicitly

### DIFF
--- a/redhat/freeradius.spec
+++ b/redhat/freeradius.spec
@@ -596,6 +596,8 @@ export RADIUSD_VERSION_RELEASE="%{release}"
 %if %{?_with_developer:1}%{!?_with_developer:0}
         --enable-developer=yes \
         --with-gperftools \
+%else
+	--disable-developer \
 %endif
 %if %{?_with_address_sanitizer:1}%{!?_with_address_sanitizer:0}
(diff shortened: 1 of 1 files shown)


arr2036:
and what's the problem...?

mkhon:
> and what's the problem...?

The problem is that docker builds are built from git and have developer mode enabled with all the expensive checks enabled

arr2036:
We need more information.  By default a source tarball is created, and that's copied into the build directory.  The autoconf scripts will only enable a developer build by default, if ./configure is running in a directory with a .git folder... So it's confusing why developer builds would need to be explicitly disabled.

mkhon:
"make rpm" runs rpmbuild as

```
rpmbuild --define "_topdir $$cwd/rpmbuild"
```

So rpmbuild/ directory is under current directory (which can be a git clone).

So "git status" which is run during configure to determine if we are building from git checkout finds git repo under rpmbuild/ parent and enables developer mode implicitly.

Additionally, .deb build uses --disable-developer explicitly.

This _topdir stuff should be left intact because it enables proper versioning when "make rpm" is run from a git clone.

PR #4839, opened 2023-01-06
Handle corner case in using fr_sbuff_is...() functions on markers

ndptech:
The previous implementation of the `fr_sbuff_is...()` functions had the following issue.  If the sbuff was advanced off the end of the buffer, then the functions would return `false` regardless of whether the character pointed to by the marker matched the test criteria or not.

This was found while exploring some anomalies in the parsing of the delay interval in rlm_delay, and also lead to the fix in parsing strings into fr_time_delta_t.


arr2036:
Can you prefix or undef the `IS_*` macros, other than that, this looks OK.

PR #4840, opened 2023-01-07
rlm_rest: Allow a Content-Type to be set when literal data is given in an xlat

terryburton:


arr2036:
You can already do that with the header attributes?

terryburton:
Superseded by #4841.

PR #4849, opened 2023-01-17
pyfr: Add Python binding for libfreeradius API.

jpereira:
first version of our new Python "pyfr" module
exporting some API behaviors.


src/language/python/Makefile
```diff
@@ -0,0 +1,35 @@
+-include ../../../Make.inc
```
alandekok, line 1: Can this be integrated into the main FreeRADIUS build?  I.e. the boilermake framework?

Why does this have to be a separate `Makefile`?

The default should be to build everything, so long as the correct headers / libraries are on the system.

src/language/python/bootstrap.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
alandekok, line 1: I don't think this file is necessary.  It's targeted specifically to homebrew, which isn't good.  I think it should just be deleted, and the various build / test rules placed in a `make` file.

src/language/python/Makefile
arr2036: Python specific build system

jpereira: I converted to _boiler.mk_ framework as can be seen in `src/language/python/all.mk`

arr2036:
We'll keep the code to use as a reference (exists in origin as v4/pyfr), but the abstractions for this are still wrong after multiple rounds of feedback, and I don't trust that it's any way correct.

PR #4851, opened 2023-01-18
Changes to reflect ClusterFuzz patches, so that patches can be remove…

mrw42:
…d from ClusterFuzz builds

### DIFF
--- a/src/bin/fuzzer.c
+++ b/src/bin/fuzzer.c
@@ -38,12 +38,12 @@ RCSID("$Id$")
  *	./build/make/jlibtool --mode=execute ./build/bin/local/fuzzer_radius -D share/dictionary /path/to/corpus/directory/
  */
 
-static bool			init = false;
-static fr_test_point_proto_decode_t *tp	= NULL;
-static dl_t			*dl = NULL;
-static dl_loader_t		*dl_loader;
+static bool				init	= false;
+static fr_test_point_proto_decode_t	*tp	= NULL;
+static dl_t				*dl	= NULL;
+static dl_loader_t			*dl_loader;
 
-static fr_dict_t		*dict = NULL;
+static fr_dict_t			*dict	= NULL;
 
 int LLVMFuzzerInitialize(int *argc, char ***argv);
 int LLVMFuzzerTestOneInput(const uint8_t *buf, size_t len);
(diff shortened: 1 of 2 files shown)


mrw42:
Changes include mapping the Clusterfuzz patches (found in oss-fuzz/projects/freeradius/patches.diff)  into configure.ac and src/bin/fuzzer.c.  

The configure.ac changes allow keeping the original CFLAGS while adding -g3, instead of setting the CFLAGS to -g3.  Part of this is commenting out the link at the top that exports CFLAGS, LIBS, LDFLAGS and CPPFLAGS.  I'd appreciate thoughts on whether it is better to comment out the line, or just remove CFLAGS from it.

The changes to fuzzer.c change this line:
if (!dict_dir) dict_dir = DICTDIR;

Into code that actually allocates space for the dict_dir and frees it later.  I don't know if this is completely necessary, but presumably because we shouldn't be setting the value of a const char *?

src/bin/fuzzer.c
```diff
@@ -137,8 +137,24 @@ int LLVMFuzzerInitialize(int *argc, char ***argv)
 		}
 	}
 
-	if (!dict_dir) dict_dir = DICTDIR;
-
+	/*	if (!dict_dir) dict_dir = int; -- removed to match oss-fuzz patch    */
+
+        DICTDIR free_dict = 0;
```
jpereira, line 142: Maybe is it `int free_dict = 0;`?

mrw42: That looks like a desirable change.  Thank you.

alandekok:
See commit f1f751393564a2 and commit 9f8880b373.  I think this should work:

* check for `LIB_FUZZING_ENGINE` in the environment.  It's defined for OSS-FUZZ and not really anything else. 
 https://google.github.io/oss-fuzz/advanced-topics/ideal-integration/
* wrap the OSS-FUZZ specific code in an `#ifdef`, so that it doesn't affect normal builds.  We don't want to append the `/dict` and `/lib` directories for normal builds.  They're only relevant for OSS-FUZZ
* use `talloc_asprintf()` to format the strings.  In the previous patch, if the binary name just happened to be less than 6 characters, the final `snprintf()` would overflow the allocated buffer.
* we don't need `to_free` flags, we can just set a variable `to_free`, and then unconditionally call `talloc_free(to_free)`.  The free function can be passed `NULL`, so this is less visible code for us.

I think this should be enough to get clusterfuzz working again.  With these changes, I hope we don't need any patches specific to clusterfuzz

mrw42:
Can I just close this pull request, without it ever being pulled?  There is one changes still needed in configure.ac to fix some bash syntax that snuck in there at some point, but I think it would be better to make that (one line) change cleanly over Alan's changes.  I think that Alan's changes to fuzzer.c have removed the need for my changes.

PR #4856, opened 2023-01-23
Prevent overrun of fr_tacacs_packet_codes[] (CID #1520416)

jejones3141:
In mod_process(), request->packet->code is constrained to [1, FR_TACACS_PACKET_TYPE_MAX)... but FR_TACACS_PACKET_TYPE_MAX - 1 is equal to FR_TACACS_CODE_MAX, and hence if the code is not allowed, it will fall off the end of fr_tacacs_packet_codes[] in the REDEBUG() invocation.

We therefore add an element to fr_tacacs_packet_codes[].

### DIFF
--- a/src/protocols/tacacs/base.c
+++ b/src/protocols/tacacs/base.c
@@ -99,7 +99,7 @@ fr_dict_attr_autoload_t libfreeradius_tacacs_dict_attr[] = {
 	{ NULL }
 };
 
-char const *fr_tacacs_packet_codes[FR_TACACS_CODE_MAX] = {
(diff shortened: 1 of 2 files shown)


alandekok:
I think the solution here is to regularize the use of the various `MAX_CODE` macros.  In this case, adding another entry to the array isn't the right solution.

* The CODE_MAX definition should be one more than the last valid packet code.

* The DO_NOT_RESPOND value should be a value which will never show up in a real packet.

* The PACKET_CODE_VALID macro should be used instead of relying on manual checks of CODE_MAX

jejones3141:
Makes sense.

PR #4860, opened 2023-01-25
Fix actual missing unlock (CID #1414423)

jejones3141:


jejones3141:
Leaving it locked is intentional.

PR #4861, opened 2023-01-25
v3.2.x: Docker builds for ubuntu22, and consider switching to it for v3.2.2

terryburton:
Reviewed diff of build output from 20.04 and 22.04.
Basic testing with EAP-TLS, rlm_json and rlm_rest.


mcnewton:
Merged, thanks.

PR #4862, opened 2023-01-25
Get rid of missing unlock (CID #1518998)

jejones3141:
Attempt to get rid of duplicate unlock left a gap for a missing unlock, namely if _fr_atexit_global_once_funcs() succeeded.

### DIFF
--- a/src/lib/util/atexit.h
+++ b/src/lib/util/atexit.h
@@ -132,14 +132,10 @@ static inline int fr_atexit_talloc_free(void *to_free)
 	if (unlikely(!atomic_load(&_init_done))) { \
 		pthread_mutex_lock(&_init_mutex); \
 		if (!atomic_load(&_init_done)) { \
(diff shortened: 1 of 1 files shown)


src/lib/util/atexit.h
```diff
@@ -132,14 +132,10 @@ static inline int fr_atexit_talloc_free(void *to_free)
 	if (unlikely(!atomic_load(&_init_done))) { \
 		pthread_mutex_lock(&_init_mutex); \
 		if (!atomic_load(&_init_done)) { \
-			if (_fr_atexit_global_once_funcs(_init, _free, _our_uctx) < 0) { \
-				*(_ret) = -1; \
-				pthread_mutex_unlock(&_init_mutex); \
-			} \
+			if (_fr_atexit_global_once_funcs(_init, _free, _our_uctx) < 0) *(_ret) = -1; \
```
alandekok, line 135: That sets a -1 return code, and the later code unconditionally sets a return code to 0.  I'll push a different fix.

PR #4865, opened 2023-01-26
pap: Only look for passwords amongst non-vendor attributes

terryburton:
Otherwise, the following will trigger a "Don't set control:User-Password" warning:

VENDOR ABC 42
BEGIN-VENDOR ABC
ATTRIBUTE Not-A-Password 2 string
END-VENDOR ABC

update { &control:Not-A-Password := "x" }
pap


alandekok:
Yeah, the v4 way of checking for attributes is to use the `da`.  But that makes it difficult to use `switch` statements.

PR #4868, opened 2023-01-26
v3.2.x status server attrs

terryburton:
Useful for periodic stats shipping or control flow based on load.


terryburton:
Rebase correctly.

PR #4875, opened 2023-01-31
add null pointer check

lifeiyu2022:


arr2036:
`talloc_free` checks if the pointer is NULL...

PR #4880, opened 2023-02-02
Index the headers starting with 1, just like all the data.

meadmaker:


arr2036:
Good catch.  Are you going down the CSV route for stats ingestion?

meadmaker:
> Good catch. Are you going down the CSV route for stats ingestion?

Yes, that seems much more straightforward than parsing the regular output :)

PR #4885, opened 2023-02-07
v4: Remove list_as_attr

ndptech:
These changes achieve the removal of `list_as_attr` and are to be followed up by further tmpl changes such as reworking `tmpl_attr_rules_t` to distinguish the different contexts for tmpl evaluation.


src/lib/server/tmpl_tokenize.c
```diff
@@ -396,6 +396,34 @@ void tmpl_debug(tmpl_t const *vpt)
...
 
+/** Parse one a single list reference
+ *
+ * @param[out] da_p	attribute representing a list.
+ * @param[in] in	Sbuff to read request references from.
+ * @return
+ *	- > 0 the number of bytes parsed.
+ *      - 0 no list qualifier found.
+ */
+fr_slen_t tmpl_attr_list_from_substr(fr_dict_attr_t const **da_p, fr_sbuff_t *in)
```
arr2036, line 393: Why is this needed?  There shouldn't be a special case for list attributes anywhere.

src/lib/server/request.c
```diff
@@ -38,18 +38,18 @@ fr_dict_autoload_t request_dict[] = {
...
 
 fr_dict_attr_t const *request_attr_root;
-fr_dict_attr_t const *request_attr_request;
-fr_dict_attr_t const *request_attr_reply;
-fr_dict_attr_t const *request_attr_control;
-fr_dict_attr_t const *request_attr_state;
+fr_dict_attr_t const *fr_request_attr_request;
+fr_dict_attr_t const *fr_request_attr_reply;
+fr_dict_attr_t const *fr_request_attr_control;
+fr_dict_attr_t const *fr_request_attr_state;
```
arr2036, line 44: Not sure we want the `fr_` prefix, it's only usually used in libraries we might want to use in other applications, so it's not traditionally been added to things in the server library.

src/lib/server/request.h
```diff
@@ -80,10 +80,10 @@ typedef	void (*fr_request_process_t)(request_t *, fr_state_signal_t);	//!< Funct
...
 
 extern HIDDEN fr_dict_attr_t const *request_attr_root;
-extern HIDDEN fr_dict_attr_t const *request_attr_request;
-extern HIDDEN fr_dict_attr_t const *request_attr_reply;
-extern HIDDEN fr_dict_attr_t const *request_attr_control;
-extern HIDDEN fr_dict_attr_t const *request_attr_state;
+extern fr_dict_attr_t const *fr_request_attr_request;
+extern fr_dict_attr_t const *fr_request_attr_reply;
+extern fr_dict_attr_t const *fr_request_attr_control;
+extern fr_dict_attr_t const *fr_request_attr_state;
```
arr2036, line 86: The reason why these were hidden was because it caused an ODR violation.  I guess this is is no longer happening.

src/lib/server/tmpl_tokenize.c
arr2036: Mostly for the unit test utilities

src/lib/unlang/xlat_builtin.c
```diff
@@ -2744,6 +2750,8 @@ static xlat_action_t xlat_func_pairs(TALLOC_CTX *ctx, fr_dcursor_t *out,
 				&(tmpl_rules_t){
 					.attr = {
 						.dict_def = request->dict,
+						.list_def = fr_request_attr_request,
+						.allow_wildcard = true,
```
arr2036, line 2754: Hmm, what does allow_wildcard so, and when would we want to disallow it?

src/lib/server/tmpl_tokenize.c
```diff
@@ -1790,7 +1800,7 @@ static inline int tmpl_attr_afrom_attr_substr(TALLOC_CTX *ctx, tmpl_attr_error_t
 	 *	Don't alter the fr_strerror buffer, may contain useful
 	 *	errors from the dictionary code.
 	 */
-	if (!at_rules->allow_unresolved) {
+	if (!at_rules->allow_unresolved && !(at_rules->allow_wildcard && fr_sbuff_is_char(name, '['))) {
```
arr2036, line 1803: It this more automatic wildcard so if an index isn't provided we return all the instances? Or what's meant to be going on here.

alandekok: IIRC this is a sanity check for `&request[*]`.  There's only one `request`, so adding filters on it should be forbidden

arr2036: If that's the case then check if it's a list attribute instead of passing rules down would be better?

arr2036: Unless you can think of another situation where we need to allow/deny filters?

src/lib/server/tmpl_tokenize.c
ndptech: There are two places it's used - when compiling `update` sections into `edit` sections and in `_map_afrom_cs()` when parsing `update` sections

src/lib/unlang/xlat_builtin.c
ndptech: It allows things such as `&request.[*]` to parse where `allow_unresolved` is false

src/lib/server/tmpl_tokenize.c
ndptech: There is another use in `rlm_exec` but that is easily removable

src/lib/server/tmpl_tokenize.c
ndptech: It allows `&request.[*]` when `allow_unresolved` is false.  The first parsing pass will have extracted `&request`, then we are left with a blank attr ref which won't be found in the dictionaries.  Testing for `allow_wildcard` and the next character being `[` allows this to fall through to call `tmpl_attr_ref_afrom_unresolved_substr` which will add the undefined attr ref and parse the filter.

src/lib/server/request.c
ndptech: I'll take that out - I misunderstood the reasoning behind the `HIDDEN` and thought it was to do with the naming.

arr2036:
Merged the commits manually

PR #4888, opened 2023-02-09
v4: Use slab allocator to manage connection handles for rlm_rest

ndptech:


src/modules/rlm_rest/rlm_rest.c
```diff
@@ -1192,6 +1201,12 @@ static int instantiate(module_inst_ctx_t const *mctx)
 		return -1;
 	}
 
+	cs = cf_section_find(conf, "connections", NULL);
+	if (!cs) cs = cf_section_alloc(conf, conf, "connections", NULL);
+
+	if (cf_section_rules_push(cs, conn_config) < 0) return -1;
+	if (cf_section_parse(inst, &inst->conn_config, cs) < 0) return -1;
```
arr2036, line 1237: This shouldn't be needed, if it is, that's a bug in the cf_parse code... Or it looks like you might just be missing the FR_TYPE_SUBSECTION entry in `module_config`?

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -135,6 +135,15 @@ static const CONF_PARSER xlat_config[] = {
...
 
+static const CONF_PARSER conn_config[] = {
+	{ FR_CONF_OFFSET("min", FR_TYPE_UINT32, rlm_rest_conn_config_t, min_elements), .dflt = "10" },
+	{ FR_CONF_OFFSET("max", FR_TYPE_UINT32, rlm_rest_conn_config_t, max_elements), .dflt = "100" },
+	{ FR_CONF_OFFSET("limit", FR_TYPE_BOOL, rlm_rest_conn_config_t, limit), .dflt = "no" },
+	{ FR_CONF_OFFSET("cleanup_interval", FR_TYPE_TIME_DELTA, rlm_rest_conn_config_t, interval), .dflt = "30s" },
+	{ FR_CONF_OFFSET("connect_timeout", FR_TYPE_TIME_DELTA, rlm_rest_conn_config_t, connect_timeout), .dflt = "3.0" },
+	CONF_PARSER_TERMINATOR
+};
+
```
arr2036, line 146: It may be better to move this into our libcurl library.  I think there's already some generic CONF_PARSER arrays in there.  I think the conn config is going to be pretty much identical for all the curl based modules?

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -1096,6 +1096,31 @@ static int parse_sub_section(rlm_rest_t *inst, CONF_SECTION *parent, CONF_PARSER
...
+static int rest_conn_alloc(fr_curl_io_request_t *randle, void *uctx)
+{
+	rlm_rest_t const	*inst = talloc_get_type_abort(uctx, rlm_rest_t);
+	rlm_rest_curl_context_t	*curl_ctx = NULL;
+
+	randle->candle = curl_easy_init();
+	if (!randle->candle) return -1;
+
+	curl_ctx = talloc_zero(randle, rlm_rest_curl_context_t);
```
arr2036, line 1132: MEM9)

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -1096,6 +1096,31 @@ static int parse_sub_section(rlm_rest_t *inst, CONF_SECTION *parent, CONF_PARSER
...
+	return 0;
+}
+
+static int rest_conn_alloc(fr_curl_io_request_t *randle, void *uctx)
+{
+	rlm_rest_t const	*inst = talloc_get_type_abort(uctx, rlm_rest_t);
+	rlm_rest_curl_context_t	*curl_ctx = NULL;
+
+	randle->candle = curl_easy_init();
+	if (!randle->candle) return -1;
```
arr2036, line 1130: Either MEM() or wrap the condition in unlikely() and say what the error was with `fr_strerror`

src/modules/rlm_rest/rlm_rest.c
ndptech: Makes sense - should be identical for IMAP and SMTP

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -1167,6 +1167,13 @@ static int mod_thread_instantiate(module_thread_inst_ctx_t const *mctx)
 		return -1;
 	}
 
+	if (fr_rest_slab_list_alloc(t, &t->slab, mctx->el, inst->conn_config.interval, inst->conn_config.min_elements,
```
arr2036, line 1167: This many arguments really indicates there should be a configuration structure containing the configurable elements of the slab allocator.   Then you could just nest it inside the generic curl config structure, and not have to pass in five arguments manually.

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -141,6 +141,8 @@ static const CONF_PARSER module_config[] = {
 	{ FR_CONF_OFFSET("http_negotiation", FR_TYPE_VOID, rlm_rest_t, http_negotiation),
 	  .func = cf_table_parse_int, .uctx = &(cf_table_parse_ctx_t){ .table = http_negotiation_table, .len = &http_negotiation_table_len }, .dflt = "default" },
 
+	{ FR_CONF_OFFSET("connections", FR_TYPE_SUBSECTION, rlm_rest_t, conn_config), .subcs = (void const *) fr_curl_conn_config },
```
arr2036, line 144: connection singular.  We don't usually pluralise config sections.

src/modules/rlm_rest/rlm_rest.c
```diff
@@ -1095,6 +1075,77 @@ static int parse_sub_section(rlm_rest_t *inst, CONF_SECTION *parent, CONF_PARSER
...
+
+	MEM(curl_ctx = talloc_zero(randle, rlm_rest_curl_context_t));
+	curl_ctx->headers = NULL;
+	curl_ctx->request.instance = inst;
+	curl_ctx->response.instance = inst;
+
+	randle->uctx = curl_ctx;
+	talloc_set_destructor(randle, _mod_conn_free);
+
+	fr_fr_curl_io_request_t_slab_set_destructor(randle, rest_request_cleanup, NULL);
```
arr2036, line 1144: fr_fr?

src/lib/curl/base.c
```diff
@@ -72,6 +72,15 @@ CONF_PARSER fr_curl_tls_config[] = {
 	CONF_PARSER_TERMINATOR
 };
 
+CONF_PARSER fr_curl_conn_config[] = {
+	{ FR_CONF_OFFSET("min", FR_TYPE_UINT32, fr_curl_conn_config_t, min_elements), .dflt = "10" },
+	{ FR_CONF_OFFSET("max", FR_TYPE_UINT32, fr_curl_conn_config_t, max_elements), .dflt = "100" },
```
arr2036, line 77: This is a little confusing.  We have `max` and then we have `limit`... So max isn't really the maximum.  It's more like `pool_min`, `pool_max`, or `cache_min`, `cache_max`.  I don't know @alandekok  any comments?  The basic issue here is when `limit = yes` we limit the number of connections to `max`, if `limit = no` the only limit comes from the limit sections in unlang.  I think we might not want to expose limit at all, and ONLY allow limiting from unlang, and keep the limit arg for internal uses.

src/lib/util/slab.h
```diff
@@ -160,34 +167,29 @@ extern "C" {
...
+		(*out)->elements_per_slab = (config->elements_per_slab > 0 ? config->elements_per_slab : \
+					     (config->min_elements ? config->min_elements : 1)); \
+		(*out)->min_elements = config->min_elements; \
+		(*out)->max_elements = config->max_elements; \
+		(*out)->at_max_fail = config->at_max_fail; \
 		(*out)->alloc = alloc; \
-		(*out)->num_children = num_children; \
-		(*out)->child_pool_size = child_pool_size; \
+		(*out)->num_children = config->num_children; \
+		(*out)->child_pool_size = config->child_pool_size; \
```
arr2036, line 186: Just use the config internally too in the `_slab_list_t` then we don't have to add more boilerplate every time we add a config item :P

PR #4890, opened 2023-02-13
fix talloc_bstrdup

lifeiyu2022:
an error will occur when inlen is 1


src/lib/util/talloc.c
```diff
@@ -402,18 +402,16 @@ char *talloc_bstrdup(TALLOC_CTX *ctx, char const *in)
 	char	*p;
 	size_t	inlen = talloc_array_length(in);
 
-	if (inlen == 0) inlen = 1;
-
-	p = talloc_array(ctx, char, inlen);
+	p = talloc_array(ctx, char, inlen + 1);
```
arr2036, line 405: That's incorrect.  That's allocate an extra \0 bytes

PR #4899, opened 2023-02-15
Destroy MD context, causing leaks with OpenSSL >= 3.0

jpereira:


src/lib/radius.c
```diff
@@ -4862,6 +4862,8 @@ ssize_t rad_tunnel_pwdecode(uint8_t *passwd, size_t *pwlen, char const *secret,
 			reallen = passwd[2] ^ digest[0];
 			if (reallen > encrypted_len) {
 				fr_strerror_printf("tunnel password is too long for the attribute");
+				fr_md5_destroy(&old);
+				fr_md5_destroy(&context);
```
mcnewton, line 4866: fr_md5_destroy() ultimately calls free(), and calling free() on the same pointer twice is undefined behaviour. As there's a fr_md5_copy above, then this might actually be happening? (Same with the existing destroys on lines 4889/4890 below, too)

jpereira: The problem is because that statement is `return -1` without destroying the contexts causing a leak. if you see the next line `4867`, the `return -1` is being called.

mcnewton: I understand why the destroys have been added there. The point I was making what that it looks like they could sometimes be the same pointer and, if that is the case, it will result in a double free.

alandekok: There's a mass of ifdef's, but they all do safe things, in one of two ways:

1) md5_copy() just does `dst = src`, which is a straight copy.  But md5_destroy() is a NOOP

2) md5_copy() calls `EVP_MD_CTX_copy_ex(_dst, _src)` which should create a full copy, and md5_destroy calls EVP_MD_CTX_destroy(*)

If we're worried, just run it under asan / lsan and send it packets.  It will complain pretty quickly if there's anything wrong.

PR #4902, opened 2023-02-16
(Almost) always check request->reply->code

jejones3141:
The exception is at send_reply; all paths leading there assign a constant good value to request->reply->code.

### DIFF
--- a/src/process/tacacs/base.c
+++ b/src/process/tacacs/base.c
@@ -603,6 +603,7 @@ RESUME(auth_start)
 			break;
 
 		default:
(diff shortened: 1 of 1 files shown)


alandekok:
Some of the TACACS tests fail after this change.  It's not clear why.

jejones3141:
Looking into it.

arr2036:
Still seems to fail :(

jejones3141:
Bummer. Will check it out; locally it passed.

jejones3141:
Closing; I think I see a better way to do this, because all these tacacs issues go back to `reply_code()`.

PR #4905, opened 2023-02-17
Annotate dead code false positive

jejones3141:
Coverity doesn't see the "goto error" hidden in the invocations of FR_CURL_REQUEST_SET_OPTION(), and hence thinks that at error, randle will always be NULL, making the fr_imap_slab_release() invocation dead code.


jejones3141:
Yes, it is... and a better way to do it, too. Thanks.

PR #4908, opened 2023-02-23
v3.2: Define max_retries for pool connections

ndptech:
Allows control over the number of times a connection operation can be retried before the module call fails.

Previously this was always set to the number of connections in the pool
- so on a system with a large number of open connections, and a remote server going slow, this would easily block threads.


jpereira:
@ndptech Maybe must be good to have that new _option_ described in the `ChangeLog`

PR #4915, opened 2023-03-03
Initial pass at per module call tmpl evaluation

ndptech:
This is not complete yet - so is purely for code review.

The changes to rlm_files are not complete either - just used as a test bench to check the working of the other changes.

### DIFF
--- a/src/lib/unlang/compile.c
+++ b/src/lib/unlang/compile.c
@@ -4657,6 +4657,116 @@ check_for_loop:
 	return subcs;
 }
 
+/** Parse per call module env
+ *
+ * Used for config options which must be parsed in the context in which
+ * the module is being called.
+ *
+ * @param[in] single		Module call being compiled.
+ * @param[in] unlang_ctx	Current compilation context.
+ * @param[in] cs		Module config.
+ * @param[in] module_env	to parse.
+ * @return
+ *	- 0 on success;
+ *	- <0 on failure;
+ */
+static int method_env_parse(unlang_module_t *single, unlang_compile_t *unlang_ctx, CONF_SECTION const *cs,
+			    module_env_t const *module_env) {
+	CONF_PAIR const		*cp, *next;
+	module_env_parsed_t	*module_env_parsed;
+	ssize_t			len, opt_count, multi_index;
+	char const		*value;
+	fr_token_t		quote;
+	fr_type_t		type;
+
+	while (module_env->name) {
+		if (FR_BASE_TYPE(module_env->type) == FR_TYPE_SUBSECTION) {
+			CONF_SECTION const *subcs;
+			subcs = cf_section_find(cs, module_env->name, module_env->section.ident2);
+			if (!subcs) goto next;
+
+			if (method_env_parse(single, unlang_ctx, subcs, module_env->section.subcs) < 0) return -1;
+			goto next;
+		}
+
+		cp = cf_pair_find(cs, module_env->name);
+
+		if (!cp && !module_env->dflt) {
+			if (!module_env->pair.required) goto next;
+
+			cf_log_err(cs, "Module %s missing required option %s", single->self.name, module_env->name);
+			return -1;
+		}
+
+		/*
+		 *	Check for additional conf pairs and error
+		 *	if there is one and multi is not allowed.
+		 */
+		if (!module_env->pair.multi && ((next = cf_pair_find_next(cs, cp, module_env->name)))) {
+			cf_log_err(cf_pair_to_item(next), "Invalid duplicate configuration item '%s'", module_env->name);
+			return -1;
+		}
+
+		opt_count = cf_pair_count(cs, module_env->name);
+		if (opt_count == 0) opt_count = 1;
+
+		for (multi_index = 0; multi_index < opt_count; multi_index ++) {
+			MEM(module_env_parsed = talloc_zero(single->mod_env_ctx, module_env_parsed_t));
+			module_env_parsed->rule = module_env;
+			module_env_parsed->opt_count = opt_count;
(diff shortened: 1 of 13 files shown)


src/modules/rlm_files/rlm_files.c
```diff
@@ -683,12 +695,17 @@ module_rlm_t rlm_files = {
...
+		  .method_env = &method_env	},
+
+		{ .name1 = "recv",		.name2 = "accounting-request",	.method = mod_preacct,
+		  .method_env = &method_env	},
+		{ .name1 = "recv",		.name2 = CF_IDENT_ANY,		.method = mod_authorize,
+		  .method_env = &method_env	},
+		{ .name1 = "authenticate",	.name2 = CF_IDENT_ANY,		.method = mod_authenticate,
+		  .method_env = &method_env	},
+		{ .name1 = "send",		.name2 = CF_IDENT_ANY,		.method = mod_post_auth,
+		  .method_env = &method_env	},
```
arr2036, line 708: Remove the `method_` prefix so the struct name isn't repeated in the fields.  Smurfs and all that...

src/lib/unlang/module.c
```diff
@@ -905,6 +986,44 @@ static unlang_action_t unlang_module(rlm_rcode_t *p_result, request_t *request,
 		goto done;
 	}
 
+	if (mc->method_env) {
+		/*
+		 *	We have a method env and this is the first call.
+		 *	Allocate the data structure.
+		 */
+		if (!state->last_expanded) {
```
arr2036, line 994: Could you do an enum/switch here just to make it really explicit what's going on, `MOD_ENV_EXP_INIT`, `MOD_ENV_EXP_DONE`, `MOD_CALL`...  Or you can split them into different functions.  I think that's what @alandekok ended up doing for the init code.

src/lib/unlang/module.c
```diff
@@ -864,6 +871,85 @@ static void unlang_module_event_retry_handler(UNUSED fr_event_list_t *el, fr_tim
 	frame->signal(request, frame, FR_SIGNAL_TIMEOUT);
 }
 
+/** Parse the result of module_env tmpl expansion
+ */
+static int module_env_value_parse (request_t *request, unlang_frame_state_module_t *state) {
```
arr2036, line 876: \s

arr2036: `static inline CC_HINT(always_inline)` if it only gets called from one place.

src/lib/unlang/module.c
```diff
@@ -864,6 +871,85 @@ static void unlang_module_event_retry_handler(UNUSED fr_event_list_t *el, fr_tim
...
+	/*
+	 *	Find the location of the output
+	 */
+	out = ((uint8_t *)state->env_data) + env->rule->offset;
+
+	/*
+	 *	If this is a multi pair option, the output is an array.
+	 *	On the first pass allocate the array.
+	 */
+	if (env->rule->pair.multi) {
```
arr2036, line 921: I think you'd want to do this first so the CTX for the boxes is the array?

src/lib/unlang/module.c
```diff
@@ -864,6 +871,85 @@ static void unlang_module_event_retry_handler(UNUSED fr_event_list_t *el, fr_tim
...
+		out = &array[env->multi_index];
+	}
+
+	while ((vb = fr_value_box_list_pop_head(&state->tmpl_expanded))) {
+		switch(type) {
+		case FR_TYPE_VOID:
+			break;
+
+		case FR_TYPE_LEAF:
+			if (fr_value_box_cast_in_place(state->env_data, vb, type, NULL) < 0) {
```
arr2036, line 937: Same here, should probably be the multi array if one's configured

arr2036: It seems a bit simpler to initialise the output array outside of this function before the tmpl is expanded, pass it as the ctx to the tmpl expansion code, and pass a pointer to the correct output element into this function.

src/lib/unlang/compile.c
```diff
@@ -4657,10 +4657,127 @@ static CONF_SECTION *virtual_module_find_cs(CONF_ITEM *ci, rlm_components_t *pco
...
+ *
+ * @param[in] single		Module call being compiled.
+ * @param[in] unlang_ctx	Current compilation context.
+ * @param[in] cs		Module config.
+ * @param[in] module_env	to parse.
+ * @return
+ *	- 0 on success;
+ *	- <0 on failure;
+ */
+static int parse_method_env(unlang_module_t *single, unlang_compile_t *unlang_ctx, CONF_SECTION const *cs,
```
arr2036, line 4673: It might be worth using a slab allocator to hold the `tmpl_t` produced so they occur sequentially in memory.  I think it's a good trade off between difficulty of implementation and at least some speedups.

src/lib/unlang/compile.c
```diff
@@ -4657,10 +4657,165 @@ static CONF_SECTION *virtual_module_find_cs(CONF_ITEM *ci, rlm_components_t *pco
...
+				cf_log_perr(cp, "Failed to parse '%s' for %s", cf_pair_value(cp), module_env->name);
+				return -1;
+			}
+
+			/*
+			 *	Ensure only valid TMPL types are produced.
+			 */
+			switch (module_env_parsed->tmpl->type) {
+			case TMPL_TYPE_UNRESOLVED:
+				if (tmpl_cast_in_place(module_env_parsed->tmpl, module_env->type, NULL) < 0) goto error;
```
arr2036, line 4746: It'd be better to pass the cast in the tmpl rules, you shouldn't get "UNRESOLVED" tmpls then.

PR #4920, opened 2023-03-09
v4: Add libcurl tuneables to control open / cached connection counts

ndptech:
These libcurl multi handle options allow admins to avoid having too many connections open to remote servers at the same time.  Especially beneficial for rlm_imap where it appears that even a curl_easy_cleanup() doesn't cause the session to log out of the IMAP server.

### DIFF
--- a/raddb/mods-available/imap
+++ b/raddb/mods-available/imap
@@ -230,5 +230,38 @@ imap {
 			cleanup_interval = 30s
 
 		}
+
+		#
+		# curl_opts { .. }:: Options for resource limiting connections
+		# handled by libcurl.  These represent per thread limits.
+		#
+		curl_opts {
+			#
+			#  max_connects:: The libcurl option CURLMOPT_MAXCONNECTS.
+			#
+			#  Sets the maximum number of connections to maintain in
+			#  the connection cache.  If this number is exceeded, then
+			#  the oldest connection is closed when a new one is
+			#  requested.
+			#
+			max_connects = 100
+
+			#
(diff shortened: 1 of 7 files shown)


arr2036:
don't think we should expose these toggles to the user.  Does IMAP just keep opening new connections indefinitely if it's fed requests in serial?

It's not like we can do anything useful by setting a per host limit... it just cause the requests to randomly fail.  Same with the maximum number of active connections... and unless the connection pool grows to greater than the maximum number of requests running in parallel, then it does really make sense to restrict it here?

If the user wants to limit the maximum number of requests running through a module they should use the limit section.  It doesn't really make sense to place limits in the module itself.

ndptech:
If CURLMOPT_MAX_TOTAL_CONNECTIONS and CURLMOPT_MAX_HOST_CONNECTIONS are not set then they mean unlimited connections.  If they are set then they cause connections which hit one or other of those limits to queue until a connection is available.

CURLMOPT_MAXCONNECTS is more complex.  If it is not set then the connection cache size automatically increases to be 4 times the number of handles added to a mulit handle.

Testing rlm_imap has shown that even if, following an authentication `curl_easy_cleanup()` is called which is supposed to reset the connection, sessions remain logged into the IMAP server, until the connection is cleared from the cache.

So, with a limit section in place, the maximum number of open connections in the cache would be 4 x limit x thread count.  Whether these remain open indefinitely is likely down to the configuration of the IMAP server, though RFC 2060 states that idle timeouts should be at least 30 minutes.

ndptech:
Makes sense - on the to-do list after async ldap

arr2036:
OK, I'll close this PR for now.

PR #4938, opened 2023-03-23
add option to load the private_key_password from a file

ottigeda:


src/main/tls.c
```diff
@@ -3851,6 +3852,23 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client, char const *chain_
...
 	}
+	if (conf->private_key_password_file) {
+		FILE* passwordfile = fopen(conf->private_key_password_file, "r");
+		if (passwordfile) {
+			char password[256];
+			if(fgets(password, sizeof(password), passwordfile))	{
+				SSL_CTX_set_default_passwd_cb_userdata(ctx, password);
+				SSL_CTX_set_default_passwd_cb(ctx, cbtls_password);
+			}
+			else	{
```
alandekok, line 3863: Random formatting is not recommended.  There's a consistent set of formatting in the server, for good reason.

src/main/tls.c
```diff
@@ -3851,6 +3852,23 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client, char const *chain_
...
+				ERROR(LOG_PREFIX ": Error reading private_key_password_file %s", conf->private_key_password_file);
+			}
+			fclose(passwordfile);
```
alandekok, line 3866: This will call `fclose()` when `passwordfile` is `NULL`.  That is not recommended.

alandekok:
It's not clear why this is useful?  There isn't a lot of difference between the password being in a FreeRADIUS configuration file, or being in a different file on disk.

What's the use-case here?  Is this security?  Or simplicity, or...??

i.e. I can see that the patch does something, but it would be good to have an explanation as to *why* the patch is needed.

src/main/tls.c
alandekok: I guess.  The random indentation and weird formatting is confusing.

src/main/tls.c
ottigeda: You are perfectly right, I tried to follow the existing style but it seems I was not doing a good job, sorry :-)
I just tried to improve it a bit.
I also tried to read https://wiki.freeradius.org/contributing/coding-standards#function-naming-conventions. 
If it is still not good, could you provide me where I could find the information about the formatting ?
Is there any formatting tool you might use ?

src/main/tls.c
```diff
@@ -3851,6 +3852,23 @@ SSL_CTX *tls_init_ctx(fr_tls_server_conf_t *conf, int client, char const *chain_
 			SSL_CTX_set_default_passwd_cb(ctx, cbtls_password);
 		}
 	}
+	if (conf->private_key_password_file) {
+		FILE* passwordfile = fopen(conf->private_key_password_file, "r");
+		if (passwordfile) {
+			char password[256];
+			if(fgets(password, sizeof(password), passwordfile))	{
```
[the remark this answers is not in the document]
ottigeda: I tried to improve that, to remove CR/LF at the end of the string.

ottigeda:
I use freeradius together with stepca certificate authority. Since stepca supports password stored in files, and we already use stepca that way, this patch would remove the need to have the password stored in multiple files. For me this is more the practical point of view, not so much security.

src/main/tls.c
ottigeda: Sorry, must have overlooked this one, I just improved on it.
Hope the formatting is now ok ?

PR #4945, opened 2023-03-28
v4: Use async LDAP queries for user, group and profile object lookups

ndptech:
Partial conversion of rlm_ldap mod_authorize and mod_authenticate to use async LDAP queries.

In addition, use module environment for tmpl expansion and restore user SASL binds

### DIFF
--- a/src/lib/ldap/base.h
+++ b/src/lib/ldap/base.h
@@ -486,6 +486,29 @@ typedef struct {
 	int			msgid;
 } fr_ldap_bind_ctx_t;
 
+/** Holds arguments for the async SASL bind operation
+ *
+ */
+typedef struct {
+	fr_ldap_connection_t	*c;			//!< to bind.
+	char const		*mechs;			//!< SASL mechanisms to run
+	char const		*dn;			//!< to bind as.
+	char const		*identity;		//!< of the user.
+	char const		*password;		//!< of the user, may be NULL if no password is specified.
+	char const		*proxy;			//!< Proxy identity, may be NULL in which case identity is used.
+	char const		*realm;			//!< SASL realm (may be NULL).
+	LDAPControl		**serverctrls;		//!< Controls to pass to the server.
+	LDAPControl		**clientctrls;		//!< Controls to pass to the client (library).
+
+	int			msgid;			//!< Last msgid.
+	LDAPMessage		*result;		//!< Previous result.
+	char const		*rmech;			//!< Mech we're continuing with.
+} fr_ldap_sasl_ctx_t;
+
+typedef enum {
+	LDAP_BIND_SIMPLE		= 0,
+	LDAP_BIND_SASL
+} fr_ldap_bind_type_t;
 
 /** Holds arguments for async bind auth requests
  *
@@ -497,7 +520,11 @@ typedef struct {
 	fr_ldap_thread_t	*thread;	//!< This bind is being run by.
 	int			msgid;		//!< libldap msgid for this bind.
 	request_t		*request;	//!< this bind relates to.
-	fr_ldap_bind_ctx_t	*bind_ctx;	//!< Data relating to the user being bound.
+	fr_ldap_bind_type_t	type;		//!< type of bind.
+	union {
+		fr_ldap_bind_ctx_t	*bind_ctx;	//!< User data for simple binds.
+		fr_ldap_sasl_ctx_t	*sasl_ctx;	//!< User data for SASL binds.
+	};
 	fr_ldap_result_code_t	ret;		//!< Return code of bind operation.
 } fr_ldap_bind_auth_ctx_t;
 
--- a/src/lib/ldap/base.h
+++ b/src/lib/ldap/base.h
@@ -711,8 +711,7 @@ unlang_action_t fr_ldap_trunk_modify(rlm_rcode_t *p_result,
 				     TALLOC_CTX *ctx,
 				     fr_ldap_query_t **out, request_t *request, fr_ldap_thread_trunk_t *ttrunk,
 				     char const *dn, LDAPMod *mods[],
-				     LDAPControl **serverctrls, LDAPControl **clientctrls,
-				     bool is_async);
+				     LDAPControl **serverctrls, LDAPControl **clientctrls);
 
 /*
  *	base.c - Wrappers arounds OpenLDAP functions.
--- a/src/lib/ldap/base.h
+++ b/src/lib/ldap/base.h
@@ -14,6 +14,7 @@
 #include <freeradius-devel/server/global_lib.h>
 #include <freeradius-devel/server/map.h>
 #include <freeradius-devel/server/trunk.h>
+#include <freeradius-devel/unlang/function.h>
 #include <freeradius-devel/util/dlist.h>
 
 #define LDAP_DEPRECATED 0	/* Quiet warnings about LDAP_DEPRECATED not being defined */
--- a/src/lib/ldap/base.h
+++ b/src/lib/ldap/base.h
@@ -856,6 +856,14 @@ int		fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
 			    		char const *proxy,
 			    		char const *realm,
 			    		LDAPControl **serverctrls, LDAPControl **clientctrls);
+
+int		fr_ldap_sasl_bind_auth_async(request_t *request,
+					     fr_ldap_thread_t *thread,
+					     char const *mechs,
+					     char const *dn,
+					     char const *identity,
+					     char const *password,
+					     char const *proxy, char const *realm);
 #endif
 
 /*
--- a/src/lib/ldap/sasl.c
+++ b/src/lib/ldap/sasl.c
@@ -354,3 +354,158 @@ int fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
 	return 0;
 }
 
+
+/** Submit an async SASL LDAP auth bind
+ *
+ * @param[in] p_result	Unused.
+ * @param[in] priority	Unused.
+ * @param[in] request	being processed.
+ * @param[in] uctx	bind auth ctx.
+ * @return	unlang action.
+ */
+static unlang_action_t ldap_async_sasl_auth_bind_start(UNUSED rlm_rcode_t *p_result, UNUSED int *priority,
+						       request_t *request, void *uctx)
+{
+	fr_ldap_bind_auth_ctx_t	*bind_auth_ctx = talloc_get_type_abort(uctx, fr_ldap_bind_auth_ctx_t);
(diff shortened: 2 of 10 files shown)


src/lib/ldap/sasl.c
```diff
@@ -372,3 +354,162 @@ int fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
...
+	case LDAP_PROC_NOT_PERMITTED:
+		RDEBUG2("Bind as user \"%s\" not permitted", sasl_ctx->identity);
+		break;
+
+	case LDAP_PROC_REJECT:
+		RDEBUG2("Bind as user \"%s\" rejected", sasl_ctx->identity);
+		break;
+
+	case LDAP_PROC_CONTINUE:
+		return unlang_function_push(request, ldap_async_sasl_auth_bind_start, ldap_async_sasl_auth_bind_results,
```
jpereira, line 438: Maybe we are missing to release the pointers here:

```
	ldap_msgfree(sasl_ctx->result);
	talloc_free(bind_auth_ctx);
```

src/lib/ldap/sasl.c
```diff
@@ -372,3 +354,162 @@ int fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
...
+				    ldap_async_sasl_auth_bind_cancel, UNLANG_SUB_FRAME, bind_auth_ctx);
+
+	default:
+		break;
+	}
+
+	ldap_msgfree(sasl_ctx->result);
+	talloc_free(bind_auth_ctx);
+
+	switch (bind_auth_ctx->ret) {
```
jpereira, line 448: We're releasing the `talloc_free(bind_auth_ctx);` above and accessing the pointer here. It might _crash_ here....

src/lib/ldap/sasl.c
```diff
@@ -372,3 +354,162 @@ int fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
...
+ *	- 0 on success.
+ *	- -1 on failure.
+*/
+int fr_ldap_sasl_bind_auth_async(request_t *request, fr_ldap_thread_t *thread, char const *mechs, char const *dn,
+				 char const *identity, char const *password, char const *proxy, char const *realm)
+{
+	fr_ldap_bind_auth_ctx_t *bind_auth_ctx;
+	fr_ldap_connection_t	*ldap_conn = talloc_get_type_abort(thread->conn->h, fr_ldap_connection_t);
+
+	if (ldap_conn->state != FR_LDAP_STATE_RUN) {
```
jpereira, line 489: Maybe a single statement like: `if (ldap_conn->state != FR_LDAP_STATE_RUN || ldap_conn->fd < 0) { ... }`

src/lib/unlang/interpret.c
```diff
@@ -578,8 +578,8 @@ unlang_frame_action_t frame_eval(request_t *request, unlang_stack_frame_t *frame
 		 *	called the interpreter.
 		 */
 		case UNLANG_ACTION_YIELD:
-			unlang_frame_perf_yield(frame);
-			yielded_set(frame);
+			unlang_frame_perf_yield(&stack->frame[stack->depth]);
+			yielded_set(&stack->frame[stack->depth]);
```
arr2036, line 582: Not sure that's right?  The yielding frame should be the frame at the point frame_eval was called?

ndptech: I found that if I used `unlang_function_push` to push a function onto the stack, return `UNLANG_ACTION_PUSHED_CHILD`, then that function returned `UNLANG_ACTION_YIELD`, `frame` was not pointing at the frame which had just yielded.  It may well be that something else is amis, or my methodology is not quite right.

arr2036: Methodology sounds fine, but I'm not sure why the frame would be incorrect.  I've now added asserts to make sure the stack hasn't grown when yield is returned.

src/lib/ldap/sasl.c
arr2036: No, the bind_auth_ctx gets reused in the call to unlang_function_push, and is parented from a stack frame so can never leak.  You may need to release `sasl_ctx->result` though @ndptech?

src/lib/ldap/sasl.c
arr2036: Yeah, seems like a legit issue.

ndptech: Thanks @jpereira - not sure how I didn't get a "use after free" when I ran it with sanitizers!

src/lib/ldap/sasl.c
ndptech: No - at that point `sasl_ctx->result` points at the last result, which is passed back to `ldap_sasl_interactive_bind`.
When you have a multi-round SASL auth, the result of the previous round is freed in `_ldap_bind_auth_io_read()` in `rlm_ldap.c` before the next result is retrieved.

This follows the same model as multi round SASL binds as the admin user.

arr2036: OK

arr2036: You'll still leak `sasl_ctx->result` if the request is cancelled between yielding and getting a response back from LDAP as there's no cleanup code in the signal callback.  It might be worth adding destructors to any ctxs that hold an `ldap_result` to make sure that if the pointer isn't NULL, that it's always passed to `ldap_msgfree` if the ctx gets freed.

arr2036: Would probably also leak on I/O timeout?

src/lib/ldap/sasl.c
```diff
@@ -372,3 +354,160 @@ int fr_ldap_sasl_bind_async(fr_ldap_connection_t *c,
...
+	fr_ldap_bind_auth_ctx_t *bind_auth_ctx;
+	fr_ldap_connection_t	*ldap_conn = talloc_get_type_abort(thread->conn->h, fr_ldap_connection_t);
+
+	if ((ldap_conn->state != FR_LDAP_STATE_RUN) || (ldap_conn->fd < 0)) {
+		fr_connection_signal_reconnect(ldap_conn->conn, FR_CONNECTION_FAILED);
+		return -1;
+	}
+
+	MEM(bind_auth_ctx = talloc_zero(request, fr_ldap_bind_auth_ctx_t));
+	MEM(bind_auth_ctx->sasl_ctx = talloc_zero(bind_auth_ctx, fr_ldap_sasl_ctx_t));
```
arr2036, line 496: You don't need to zero the memory if you're immediately overwriting it with a compound literal

src/modules/rlm_ldap/groups.c
```diff
@@ -266,37 +383,120 @@ static unlang_action_t rlm_ldap_group_dn2name(rlm_rcode_t *p_result, rlm_ldap_t
...
+		     vp = fr_pair_list_next(&group_ctx->groups, vp)) {
+			RDEBUG2("&control.%s += \"%pV\"", group_ctx->inst->cache_da->name, &vp->data);
+		}
+	}
+
+	fr_pair_list_append(list, &group_ctx->groups);
+	REXDENT();
+
+	talloc_free(group_ctx);
+	RETURN_MODULE_RCODE(RLM_MODULE_OK);
```
arr2036, line 416: `RETURN_MODULE_OK`

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -949,18 +1015,64 @@ static int rlm_ldap_groupcmp(void *instance, request_t *request, fr_pair_t const
...
+		ldap_auth_mod_env_t *mod_env = auth_ctx->mod_env;
+		if (fr_ldap_sasl_bind_auth_async(request, auth_ctx->thread, mod_env->user_sasl_mech.vb_strvalue,
+						 auth_ctx->dn, mod_env->user_sasl_authname.vb_strvalue,
+						 auth_ctx->password, mod_env->user_sasl_proxy.vb_strvalue,
+						 mod_env->user_sasl_realm.vb_strvalue) < 0) goto fail;
+	} else {
+		if (fr_ldap_bind_auth_async(request, auth_ctx->thread, auth_ctx->dn, auth_ctx->password) < 0) goto fail;
+	}
+
+	RETURN_MODULE_RCODE(unlang_interpret_synchronous(unlang_interpret_event_list(request), request));
```
arr2036, line 1065: Guessing this is temporary? :)

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -2035,6 +2177,73 @@ static int mod_instantiate(module_inst_ctx_t const *mctx)
...
+	.inst_size = sizeof(ldap_auth_mod_env_t),
+	.inst_type = "ldap_auth_mod_env_t",
+	.env = authenticate_module_env
+};
+
+static const module_method_env_t authorize_method_env = {
+	.inst_size = sizeof(ldap_autz_mod_env_t),
+	.inst_type = "ldap_autz_mod_env_t",
+	.env = authorize_module_env
+};
```
arr2036, line 2245: Can these live at the top of the file with the CONF_PARSER sections, preferably interleaved with them...

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -1131,221 +1203,291 @@ static unlang_action_t rlm_ldap_map_profile(rlm_rcode_t *p_result, rlm_ldap_t co
...
+	RETURN_MODULE_RCODE(rcode);
+}
+
+static void mod_authorize_cancel(UNUSED request_t *request, fr_state_signal_t action, void *uctx)
+{
+	ldap_autz_ctx_t	*autz_ctx = talloc_get_type_abort(uctx, ldap_autz_ctx_t);
+
+	if (action != FR_SIGNAL_CANCEL) return;
+
+	talloc_free(autz_ctx->expanded.ctx);
```
arr2036, line 1436: Does LDAP need to be signalled to abandon the search?

src/modules/rlm_ldap/rlm_ldap.c
ndptech: Yes, until the bind auth is switched to use a trunk so it can be properly async

src/lib/ldap/base.h
```diff
@@ -485,6 +486,29 @@ typedef struct {
...
+
+	int			msgid;			//!< Last msgid.
+	LDAPMessage		*result;		//!< Previous result.
+	char const		*rmech;			//!< Mech we're continuing with.
+} fr_ldap_sasl_ctx_t;
+
+typedef enum {
+	LDAP_BIND_SIMPLE		= 0,
+	LDAP_BIND_SASL
+} ldap_bind_type_t;
```
arr2036, line 511: Can this get an `fr_` prefix too.

PR #4950, opened 2023-03-31
Simplify behaviour in REST encoding

herwinw:
The logic of the code was pretty much like this:
```c
while (alloc <= limit) {
  // Allocate `alloc` bytes, move the old buffer over to the new buffer, free old buffer
  len = func(current + used, alloc - used, 1, userdata);
  if (!len) {
    return; // This means we're finished
  }
  alloc = alloc * 2; // Double the buffer size and try again
}
// If we get here, we ran out of buffer space and cause an error
```
This means once the last part has been written by `func`, the returned length is larger than zero, so we run the loop once again, allocate a bigger chunk of memory, move the contents and free the old buffer, then call `func` again to have it return a 0 immediately so we can return the buffer we already had in the previous iteration.

Another effect of this logic is that the limit is effectively cut in half: the limit argument is 8K, but if the output is more than 4K and less than 8K, the loop gets into another iteration, but now alloc is more than limit, so it gets into the error state because it thinks it does not have enough space. We ran into this issue with JSON encoding (which is the most verbose output in the module) and EAP-TLS (which add a lot of virtual attributes with information about the certificate).


arr2036:
Those functions get passed to `CURLOPT_READFUNCTION` so need that specific signature.

arr2036:
Final issue is we need commits to be signed for this branch.  Can you sign it, then I'd be happy to merge.  Thanks!

arr2036:
It's now come up with a signature error, saying the commit email doesn't match the GPG signature email and still won't let me merge the commit :(

herwinw:
> It's now come up with a signature error, saying the commit email doesn't match the GPG signature email and still won't let me merge the commit :(

I know, I've been trying to push a new signed commit a few times, but it keeps returning this error, even though they're the same addresses.
I'll just keep continue trying, it might accidentally match some time.

arr2036:
Did you add your GPG key to GitHub? That may the issue if not...

mcnewton:
Merged manually.

PR #4958, opened 2023-04-03
Fix missing phtread destroy

jpereira:
As we are calling pthread_mutex_init(), we should call pthread_mutex_destroy() to release all initialized mutexs.


alandekok:
I just merged the other commit to v3.0.x, and then cherry-picked that commit to v3.2.x.  Thanks

Issue #3501, opened 2020-06-18
outbound RADIUS/TLS connection stalls


pauldekkers:
# Issue type

- Defect - Unexpected behaviour (obvious or verified by project member).

# Defect
## How to reproduce the issue

Using the freeradius/freeradius-server:3.0.21 Docker image (on Ubuntu 18.04 LTS with it's Docker 18.09.7):

Because of remote network problems, one of our RadSec peers was unreachable in such a way that the TCP sessions set up slow, and the TLS challenge never happens (before closing the TCP session) or after a huge delay. Also `openssl's s_client` timed out after a while.

The result is that the entire server stalled and no requests were processed at all (with some 100s reqs/s coming in). This happened fairly quickly after a restart as well, probably because of the high traffic volume, so a restart didn't resume the handling of (other) requests.

## Output of ``[radiusd|freeradius] -X`` showing issue occurring

```
Trying SSL to port 2083
Requiring Server certificate
(0) (other): before SSL initialization
(0) TLS_connect: before SSL initialization
(0) >>> send TLS 1.2  [length 00b1]
(0) TLS_connect: SSLv3/TLS write client hello
```

... and that's where it stalled.

I have no backtrace, the daemons didn't crash (or at least not within the 15 minutes, and I had to act). Inspected with strace (that I unfortunately didn't keep), and all workers were just blocked I guess, not much happening.

I believe FreeRADIUS 4 would be better in the sense that it does some things asynchronous, but I also believe it doesn't have RADIUS/TLS support yet (otherwise I'd give it a go, also because we may have a higher chance of making things dynamic as well, if we can dynamically create home_servers - but that's an entirely different topic).

I *think* I would be able to reproduce when I connect to a simple destination that accepts the TCP session, but doesn't do anything after. Happy to give it a go if you have more ideas about debugging this.

alandekok:
> Because of remote network problems, one of our RadSec peers was unreachable in such a way that the TCP sessions set up slow, and the TLS challenge never happens (before closing the TCP session) or after a huge delay. Also openssl's s_client timed out after a while.

The short answer is "if the network is broken, things will stall".

For v4, the outbound connections are done asynchronously.  We don't yet have TLS support, but it's possible with a bit of work.

Unless there are mitigating reasons, I'm going to close this as "known, can't fix in v3".   If we make every write asynchronous, that requires major changes to the server core.  hence all of the work on v4.  :(

pauldekkers:
[...]

PR #4961, opened 2023-04-03
Backport major fixes from v3.2.x

jpereira:
Major changes related to LSAN/ASAN fixes.


alandekok:
I've pulled over most of the changes.

The `event.c` and `pthread.h` changes need to be done after the event code is fixed with the patch for #3501.

Some of other CI fixes don't quite merge correctly, so I've left them for now.

PR #4963, opened 2023-04-04
Don't pass NULL buffer to _fr_syserror()

jejones3141:
The issue turns up if _fr_syserror_buffer() returns NULL.


arr2036:
Maybe make equivalent changes for all callers of the buffer function... Coverity will probably complain about them next.

jejones3141:
Good idea. Will seek them out.

jejones3141:
Caught the other use, and parenthesized the first one a la FreeRADIUS.

PR #4968, opened 2023-04-07
Fix runtime LSAN/ASAN error in map.c

jpereira:
Such error:

src/main/map.c:1228:11: runtime error: signed integer overflow: -2147483648 - 1 cannot be represented in type 'int' SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/main/map.c:1228:11 in


alandekok:
Fixed in a slightly different way.

PR #4984, opened 2023-05-09
Replace LDAP-Group pair compare with an xlat

ndptech:
Includes adding support for module environments to xlats and removal of the remaining synchronous LDAP code


src/lib/unlang/xlat.c
```diff
@@ -304,8 +308,95 @@ static unlang_action_t unlang_xlat_repeat(rlm_rcode_t *p_result, request_t *requ
 	xlat_action_t			xa;
 	xlat_exp_head_t const		*child = NULL;
 
+	/*
+	 *	If the xlat is a function with a method_env, expand it before calling the function.
+	 */
+	if ((state->exp->type == XLAT_FUNC) && state->exp->call.func->method_env) {
```
arr2036, line 314: Move this to a separate static inline.  It's all nicely self contained and does one thing, so it's clearer to put it in a separate function.

arr2036:
Now the code is used in two places calling it `module_env` is confusing.  Could it be renamed `call_env` and placed in a separate source file with a separate header.  Could that work done by separately from this PR, so there's one set of changes for re-arranging code, and one for adding the LDAP xlat.

src/modules/rlm_ldap/rlm_ldap.c
```diff
@@ -1911,6 +2109,11 @@ static int mod_bootstrap(module_inst_ctx_t const *mctx)
 	xlat = xlat_func_register_module(NULL, mctx, mctx->inst->name, ldap_xlat, FR_TYPE_STRING);
 	xlat_func_mono_set(xlat, ldap_xlat_arg);
 
+	snprintf(memberof_xlat, sizeof(memberof_xlat), "%s_memberof", mctx->inst->name);
+	xlat = xlat_func_register_module(NULL, mctx, memberof_xlat, ldap_memberof_xlat, FR_TYPE_BOOL);
```
arr2036, line 2110: Can you change this so the function name is always created as `<inst>.<func name>`.  Maybe as a separate PR.  I'm guessing there's a bunch of xlat registrations which use the same pattern.

PR #4987, opened 2023-05-11
amending thread_pool_stop() to free a TALLOC_CTX blob/chunk

martinsta:
thread_pool_stop() will free a TALLOC_CTX blob/chunk via fr_atomic_queue_free()

### DIFF
--- a/src/lib/atomic_queue.c
+++ b/src/lib/atomic_queue.c
@@ -136,7 +136,7 @@ void fr_atomic_queue_free(fr_atomic_queue_t *aq)
 {
 	if (!aq) return;
 
-	talloc_free((aq->chunk);
+	talloc_free(aq->chunk);
 	aq = NULL;
 }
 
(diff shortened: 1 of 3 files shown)


src/main/threads.c
```diff
@@ -1238,7 +1238,7 @@ void thread_pool_stop(void)
 
 	for (i = 0; i < NUM_FIFOS; i++) {
 #ifdef HAVE_STDATOMIC_H
-		talloc_free(thread_pool.queue[i]);
+		fr_atomic_queue_free(thread_pool.queue[i]);
```
alandekok, line 1241: This should have an `&` added: `fr_atomic_queue_free(&thread_pool.queue[i]);`

Doing that makes the rest of the patches unnecessary.  I'll push a fix.

src/lib/atomic_queue.c
```diff
@@ -132,12 +132,12 @@ fr_atomic_queue_t *fr_atomic_queue_alloc(TALLOC_CTX *ctx, size_t size)
...
-void fr_atomic_queue_free(fr_atomic_queue_t **aq)
+void fr_atomic_queue_free(fr_atomic_queue_t *aq)
 {
-	if (!*aq) return;
+	if (!aq) return;
 
-	talloc_free((*aq)->chunk);
-	*aq = NULL;
+	talloc_free((aq->chunk);
+	aq = NULL;
```
alandekok, line 140: This will give an error with `unused assignment`

PR #4998, opened 2023-05-17
Change module environment to more generic call environment

ndptech:
And add framework to xlat code


src/lib/unlang/xlat.c
```diff
@@ -298,14 +302,122 @@ int unlang_xlat_push_node(TALLOC_CTX *ctx, bool *p_success, fr_value_box_list_t
 	return unlang_xlat_push_internal(ctx, p_success, out, request, NULL, node, UNLANG_TOP_FRAME);
 }
 
+static unlang_action_t unlang_xlat_repeat(rlm_rcode_t *p_result, request_t *request, unlang_stack_frame_t *frame);
+
+/** Expand an xlat's call environment
+ *
+ */
+static inline unlang_action_t xlat_call_env_expand(request_t *request, unlang_stack_frame_t *frame,
```
arr2036, line 310: Is there any way we could merge this with the module method expand so we don't have duplicateish code?

arr2036: We could theoretically push the xlat or module method call, and then the call_env expansion call onto the stack separately, and just have the call_env expansion write out the result and status to somewhere that the xlat or module method call can find them...

arr2036:
Looks good, much cleaner :)

PR #5007, opened 2023-05-25
make SoH use flat or nested attributes

jpereira:


mcnewton:
`rlm_soh` should really just be removed. It was removed from Windows 10 a few years back, I don't think anyone else implements it.

PR #5008, opened 2023-05-25
make yubikey use flat or nested attributes.

jpereira:


arr2036:
This shouldn't use a convert to fr_pair_find_by_da_nested, this should use the module env interface

PR #5013, opened 2023-05-25
Port TLS nonblocking fixes from 3.2.x to 3.0.x, including TLS tests

mrw42:
Changes to fix TLS non-blocking issue were hand-selected and moved back to the 3.0.x code, omitting changes for other features added in 3.2.x (COA_TUNNEL, haproxy, RADIUSV11, etc.).  The src/tests/tls files were moved in their entirety.


mrw42:
@alandekok, should I just close this pull request?

alandekok:
I've pulled this over manually and reworked some of the commits to have a more straightforward git history.

PR #5014, opened 2023-05-25
Add static IP assignements to redis_ippool

ndptech:


src/modules/rlm_redis_ippool/rlm_redis_ippool.c
```diff
@@ -211,14 +211,23 @@ static char lua_alloc_cmd[] =
 	 */
 	"exists = redis.call('GET', owner_key);" EOL							/* 8 */
 	"if exists then" EOL										/* 9 */
-	"  local expires_in = tonumber(redis.call('ZSCORE', pool_key, exists) - ARGV[1])" EOL		/* 10 */
-	"  if expires_in > 0 then" EOL									/* 11 */
-	"    ip = redis.call('HMGET', '{' .. KEYS[1] .. '}:"IPPOOL_ADDRESS_KEY":' .. exists, 'device', 'range', 'counter')" EOL	/* 12 */
+	"  local expires = tonumber(redis.call('ZSCORE', pool_key, exists))" EOL			/* 10 */
+	"  local static = expires >= 2^" STRINGIFY(IPPOOL_STATIC_BIT) EOL
```
arr2036: Lua  5.1 isn't smart enough to pre-calculate the exponent, so this should be a static value

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
```diff
@@ -211,14 +211,23 @@ static char lua_alloc_cmd[] =
...
 	"      if expires_in < tonumber(ARGV[2]) then" EOL						/* 14 */
-	"        redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2], exists)" EOL			/* 15 */
+	"        redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2] + (static and 2^" STRINGIFY(IPPOOL_STATIC_BIT) " or 0), exists)" EOL	/* 15 */
 	"        expires_in = tonumber(ARGV[2])" EOL							/* 16 */
 	"      end" EOL											/* 17 */
+
+	/*
+	 *	Ensure gateway is set correctly
+	 */
+	"      if ARGV[4] ~= ip[4] then" EOL
```
arr2036: Ah yes, good catch.  I guess in the majority of cases the gateway doesn't change, but it might if you were moving around between APs or something

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
```diff
@@ -295,16 +304,18 @@ static char lua_update_cmd[] =
 	 *	Update the expiry time
 	 */
 	"pool_key = '{' .. KEYS[1] .. '}:"IPPOOL_POOL_KEY"'" EOL			/* 14 */
-	"redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2], ARGV[3])" EOL		/* 15 */
+	"local expires = tonumber(redis.call('ZSCORE', pool_key, ARGV[3]))" EOL
```
arr2036: What's going on here?  If the IP wasn't re-used, then we don't care about its previous ZSCORE?

arr2036: Oh, the staticness...

arr2036: Just change the scope of the first expires var?

src/modules/rlm_redis_ippool/rlm_redis_ippool_tool.c
```diff
@@ -1588,6 +1687,26 @@ do { \
 	}
 		continue;
 
+	case IPPOOL_TOOL_ASSIGN:
```
arr2036, line 1690: I think we need an unassign function too to remove the staticness?  release kinda does something different.  You also need to update the stats, to differentiate between lease times with the static flag set.

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
ndptech: This is the only use of `expires` in this script

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
```diff
@@ -211,14 +211,23 @@ static char lua_alloc_cmd[] =
...
-	"    ip = redis.call('HMGET', '{' .. KEYS[1] .. '}:"IPPOOL_ADDRESS_KEY":' .. exists, 'device', 'range', 'counter')" EOL	/* 12 */
+	"  local expires = tonumber(redis.call('ZSCORE', pool_key, exists))" EOL			/* 10 */
+	"  local static = expires >= 2^" STRINGIFY(IPPOOL_STATIC_BIT) EOL
+	"  local expires_in = expires - (static and 2^" STRINGIFY(IPPOOL_STATIC_BIT) " or 0) - ARGV[1]" EOL
+	"  if expires_in > 0 or static then" EOL							/* 11 */
+	"    ip = redis.call('HMGET', '{' .. KEYS[1] .. '}:"IPPOOL_ADDRESS_KEY":' .. exists, 'device', 'range', 'counter', 'gateway')" EOL	/* 12 */
 	"    if ip and (ip[1] == ARGV[3]) then" EOL							/* 13 */
 	"      if expires_in < tonumber(ARGV[2]) then" EOL						/* 14 */
-	"        redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2], exists)" EOL			/* 15 */
+	"        redis.call('ZADD', pool_key, 'XX', ARGV[1] + ARGV[2] + (static and 2^" STRINGIFY(IPPOOL_STATIC_BIT) " or 0), exists)" EOL	/* 15 */
```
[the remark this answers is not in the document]
ndptech: Unfortunately needed to make the trick with `and` ... `or` come up with the correct result

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
ndptech: And for static assignments - they never hit the gateway assignment at the end of the script

PR #5017, opened 2023-05-26
Let's use flat or nested attributes

jpereira:


src/modules/rlm_ldap/user.c
```diff
@@ -230,11 +230,11 @@ void rlm_ldap_check_reply(module_ctx_t const *mctx, request_t *request, fr_ldap_
 	*/
 	if (!inst->expect_password || !RDEBUG_ENABLED2) return;
 
-	if (!fr_pair_find_by_da(&request->control_pairs, NULL, attr_cleartext_password) &&
-	    !fr_pair_find_by_da(&request->control_pairs, NULL, attr_nt_password) &&
+	if (!fr_pair_find_by_da_nested(&request->control_pairs, NULL, attr_cleartext_password) &&
```
alandekok, line 233: This could arguably be changed to look for `parent` as the `Password` attribute.  If that isn't found, set `parent = request->control_pairs`

and then just do `fr_pair_find_by_da(parent, NULL, attr_cleartext_password)` etc.  That avoids some extra work.

src/modules/rlm_ldap/user.c
```diff
@@ -221,20 +221,24 @@ rlm_rcode_t rlm_ldap_check_access(rlm_ldap_t const *inst, request_t *request, LD
...
 	*
 	*	Expect_password is set when we process the mapping, and is only true if there was a mapping between
 	*	an LDAP attribute and a password reference attribute in the control list.
 	*/
 	if (!inst->expect_password || !RDEBUG_ENABLED2) return;
 
-	if (!fr_pair_find_by_da(&request->control_pairs, NULL, attr_cleartext_password) &&
-	    !fr_pair_find_by_da(&request->control_pairs, NULL, attr_nt_password) &&
+	parent = fr_pair_find_by_da_nested(&request->control_pairs, NULL, attr_password);
+	if (!parent) parent = request->control_ctx;
```
arr2036, line 235: As I said before, this likely won't work until we implement the conversion fully

src/modules/rlm_totp/rlm_totp.c
```diff
@@ -239,7 +239,7 @@ static unlang_action_t CC_HINT(nonnull) mod_authenticate(rlm_rcode_t *p_result,
 	uint8_t buffer[80];	/* multiple of 5*8 characters */
 
 
-	password = fr_pair_find_by_da(&request->request_pairs, NULL, attr_totp_user_password);
+	password = fr_pair_find_by_da_nested(&request->request_pairs, NULL, attr_totp_user_password);
```
arr2036, line 242: These should all be call env

src/modules/rlm_sqlippool/rlm_sqlippool.c
```diff
@@ -516,7 +516,7 @@ static unlang_action_t CC_HINT(nonnull) mod_alloc(rlm_rcode_t *p_result, module_
 		return do_logging(p_result, inst, request, inst->log_exists, RLM_MODULE_NOOP);
 	}
 
-	if (fr_pair_find_by_da(&request->control_pairs, NULL, attr_pool_name) == NULL) {
+	if (fr_pair_find_by_da_nested(&request->control_pairs, NULL, attr_pool_name) == NULL) {
```
arr2036, line 519: should be call env

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
```diff
@@ -1269,7 +1269,7 @@ static unlang_action_t CC_HINT(nonnull) mod_accounting(rlm_rcode_t *p_result, mo
 	/*
 	 *	IP-Pool.Action override
 	 */
-	vp = fr_pair_find_by_da(&request->control_pairs, NULL, attr_pool_action);
+	vp = fr_pair_find_by_da_nested(&request->control_pairs, NULL, attr_pool_action);
```
arr2036, line 1272: these should all be call env

src/modules/rlm_sqlippool/rlm_sqlippool.c
arr2036: This ones weirder and requires more rework in sqlippool

src/modules/rlm_redis_ippool/rlm_redis_ippool.c
arr2036: Probably ok to leave as static

PR #5022, opened 2023-05-31
enable compression for mysql

rouzier:


alandekok:
The MySQL documentation says:  https://dev.mysql.com/doc/c-api/8.0/en/mysql-options.html

`As of MySQL 8.0.18, MYSQL_OPT_COMPRESS is deprecated.`

What benefits are there from compression?

PR #5032, opened 2023-06-04
add 'if not exists' to mysql create table queries

ludomikula:
Added **IF NOT EXISTS** to all mysql CREATE TABLE sql queries to make it consistent.


arr2036:
Commits need to be signed before they can be merged.

PR #5033, opened 2023-06-05
Add "use_tls = yes" support for redis (if hiredis supports SSL)

mkhon:
Requires hiredis 1.0.x

The code related to HAVE_REDIS_SSL_OPTIONS is a placeholder for the option to skip certificate verification (is only possible with hiredis 1.1.0 where the appropriate API method was added). This can be added later if the patch itself is considered to be good and merged.

### DIFF
--- a/src/lib/redis/cluster.c
+++ b/src/lib/redis/cluster.c
@@ -1481,13 +1481,17 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 		return NULL;
 	}
 
+	conn = talloc_zero(ctx, fr_redis_conn_t);
+	conn->handle = handle;
+	talloc_set_destructor(conn, _cluster_conn_free);
+
 #ifdef HAVE_REDIS_SSL
 	if (node->cluster->ssl_ctx != NULL) {
-		fr_tls_session_t *tls_session = fr_tls_session_alloc_client(ctx, node->cluster->ssl_ctx);
+		fr_tls_session_t *tls_session = fr_tls_session_alloc_client(conn, node->cluster->ssl_ctx);
 		if (!tls_session) {
 			fr_tls_strerror_printf("%s - [%i]", log_prefix, node->id);
 			ERROR("%s - [%i] Failed to allocate TLS session", log_prefix, node->id);
-			redisFree(handle);
+			talloc_free(conn);
 			return NULL;
 		}
 
@@ -1496,7 +1500,7 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 		if (redisInitiateSSL(handle, tls_session->ssl) != REDIS_OK) {
 			ERROR("%s - [%i] Failed to initiate SSL: %s", log_prefix, node->id, handle->errstr);
 			SSL_free(tls_session->ssl);
-			redisFree(handle);
+			talloc_free(conn);
 			return NULL;
 		}
 	}
@@ -1518,7 +1522,7 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 			ERROR("%s - [%i] Failed authenticating: %s", log_prefix, node->id, handle->errstr);
 		error:
 			if (reply) fr_redis_reply_free(&reply);
-			redisFree(handle);
+			talloc_free(conn);
 			return NULL;
 		}
 
@@ -1573,10 +1577,6 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 			goto error;
 		}
 	}
-
-	conn = talloc_zero(ctx, fr_redis_conn_t);
-	conn->handle = handle;
-	talloc_set_destructor(conn, _cluster_conn_free);
 
 	return conn;
 }
--- a/src/lib/redis/cluster.c
+++ b/src/lib/redis/cluster.c
@@ -154,9 +154,15 @@
 #include <freeradius-devel/util/rand.h>
 #include <freeradius-devel/util/time.h>
 
+#include "config.h"
 #include "base.h"
 #include "cluster.h"
 #include "crc16.h"
+
+#ifdef HAVE_REDIS_SSL
+#include <freeradius-devel/tls/strerror.h>
+#include <hiredis/hiredis_ssl.h>
+#endif
 
 #define KEY_SLOTS		16384			//!< Maximum number of keyslots (should not change).
 
@@ -252,6 +258,9 @@ struct fr_redis_cluster {
 
 	fr_redis_conf_t		*conf;			//!< Base configuration data such as the database number
 							//!< and passwords.
+#ifdef HAVE_REDIS_SSL
+	SSL_CTX			*ssl_ctx;		//!< SSL context.
+#endif
 
 	fr_redis_cluster_node_t	*node;			//!< Structure containing a node id, its address and
 							//!< a pool of its connections.
@@ -806,7 +815,7 @@ static int cluster_map_node_validate(redisReply *node, int map_idx, int node_idx
 		return FR_REDIS_CLUSTER_RCODE_BAD_INPUT;
 	}
 
-	if (fr_inet_pton(&ipaddr, node->element[0]->str, node->element[0]->len, AF_UNSPEC, false, true) < 0) {
+	if (fr_inet_pton(&ipaddr, node->element[0]->str, node->element[0]->len, AF_UNSPEC, true, true) < 0) {
 		return FR_REDIS_CLUSTER_RCODE_BAD_INPUT;
 	}
 
@@ -1471,6 +1480,27 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 		ERROR("%s - [%i] Connection failed", log_prefix, node->id);
 		return NULL;
 	}
+
+#ifdef HAVE_REDIS_SSL
+	if (node->cluster->ssl_ctx != NULL) {
+		fr_tls_session_t *tls_session = fr_tls_session_alloc_client(ctx, node->cluster->ssl_ctx);
+		if (!tls_session) {
+			fr_tls_strerror_printf("%s - [%i]", log_prefix, node->id);
+			ERROR("%s - [%i] Failed to allocate TLS session", log_prefix, node->id);
+			redisFree(handle);
+			return NULL;
+		}
+
+		// redisInitiateSSL() takes ownership of SSL object on success
+		SSL_up_ref(tls_session->ssl);
+		if (redisInitiateSSL(handle, tls_session->ssl) != REDIS_OK) {
+			ERROR("%s - [%i] Failed to initiate SSL: %s", log_prefix, node->id, handle->errstr);
+			SSL_free(tls_session->ssl);
+			redisFree(handle);
+			return NULL;
+		}
+	}
+#endif
 
 	if (node->cluster->conf->password) {
 		if (node->cluster->conf->username) {
@@ -2287,6 +2317,37 @@ fr_redis_cluster_t *fr_redis_cluster_alloc(TALLOC_CTX *ctx,
 	 */
 	if (!cf_section_find(module, "pool", NULL)) {
 		(void) cf_section_alloc(module, module, "pool", NULL);
+	}
+
+	/*
+	 *	Parse TLS configuration
+	 */
+	if (conf->use_tls) {
+#ifdef HAVE_REDIS_SSL
+		CONF_SECTION *tls_cs;
+		fr_tls_conf_t *tls_conf;
+
+		tls_cs = cf_section_find(module, "tls", NULL);
+		if (!tls_cs) {
+			tls_cs = cf_section_alloc(module, module, "tls", NULL);
+		}
+
+		tls_conf = fr_tls_conf_parse_client(tls_cs);
+		if (!tls_conf) {
+			ERROR("%s - Failed to parse TLS configuation", cluster->log_prefix);
+			talloc_free(cluster);
+			return NULL;
+		}
+
+		cluster->ssl_ctx = fr_tls_ctx_alloc(tls_conf, true);
+		if (!cluster->ssl_ctx) {
+			ERROR("%s - Failed to allocate SSL context", cluster->log_prefix);
+			talloc_free(cluster);
+			return NULL;
+		}
+#else
+		WARN("%s - No redis SSL support, ignoring \"use_tls = yes\"", cluster->log_prefix);
+#endif
 	}
 
 	if (conf->max_nodes == UINT8_MAX) {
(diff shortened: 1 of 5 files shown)


src/lib/redis/cluster.c
```diff
@@ -1472,6 +1480,31 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
 		return NULL;
 	}
 
+#ifdef HAVE_REDIS_SSL
+	if (node->cluster->conf->ssl) {
+		redisSSLContextError ssl_error = REDIS_SSL_CTX_NONE;
+#ifdef HAVE_REDIS_SSL_OPTIONS
+		redisSSLOptions opt = { .verify_mode = REDIS_SSL_VERIFY_PEER };
+
+		ssl_context = redisCreateSSLContextWithOptions(&opt, &ssl_error);
```
arr2036, line 1489: The redis SSL functions don't expose many useful options, but as the whole thing is based around OpenSSL it looks like it's mostly just taking an `SSL_CTX *`.

https://github.com/redis/hiredis/blob/master/ssl.c#L64

You can get an `SSL_CTX` using the generic TLS client config code https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/tls/conf.c#L495 and another function which transforms that into a configured `SSL_CTX`.

https://github.com/FreeRADIUS/freeradius-server/blob/master/src/lib/tls/ctx.c#L610

arr2036: We would somehow need to get at the `SSL *` to call `fr_tls_session_request_bind` and `fr_tls_session_request_unbind`, but after that all the async cert validation code should work properly too.

arr2036: Nice, looks like we can alloc the `SSL *` too, so we can configure it the same as all the other `SSL *` objects we create: https://github.com/redis/hiredis/blob/master/ssl.c#L395

mkhon: @arr2036 yes, you can merge with changes - I don't mind as long as it works.

On a side note I would not rely on contexts to be the same as they may want to change that "opaque" object later

mkhon:
Additionally, we get 2x DORA sustained rate with native redis ssl compared to stunnel proxy

alandekok:
I think it's worth merging the PR.  I would also poke the hiredis people to add a public API which exposes the `SSL_CTX *` and the `SSL*` from their internal structures.

This lets us get the functionality today.  And if the hiredis APIs get updated, we can update our code to add the async certificate validation.

But without the hiredis API changes, it's difficult to get at those SSL data structures.

arr2036:
@alandekok there's already a public API.  I posted a link to it in my previous comment.  You can initialise the `SSL *` fully independently of the Redis connection.

mkhon:
> @alandekok there's already a public API. I posted a link to it in my previous comment. You can initialise the `SSL *` fully independently of the Redis connection.

Let me check that

src/lib/redis/cluster.c
```diff
@@ -1472,6 +1480,31 @@ void *fr_redis_cluster_conn_create(TALLOC_CTX *ctx, void *instance, fr_time_delt
...
+#else
+		ssl_context = redisCreateSSLContext(NULL, NULL, NULL, NULL, NULL, &ssl_error);
+#endif
+		if (!ssl_context || ssl_error != REDIS_SSL_CTX_NONE) {
+			ERROR("%s - [%i] Failed to create SSL context: %s", log_prefix, node->id, redisSSLContextGetError(ssl_error));
+			redisFree(handle);
+			return NULL;
+		}
+
+		if (redisInitiateSSLWithContext(handle, ssl_context) != REDIS_OK) {
```
alandekok, line 1499: This could be `redisInitiateSSL(handle, SSL*)`, where the `SSL*` is taken from the FreeRADIUS APIs.

arr2036: Yeah, and then we get a bunch of configuration knobs and complex validation behaviour for free :)

arr2036: If you look through the Redis code, it's configuring the I/O layer for you, so all you have to do is pass in an SSL * with various configuration options set.

https://github.com/redis/hiredis/blob/master/ssl.c#L356

mkhon: So you suggest I just use SSL_new()?

arr2036: The thing we're looking for here is using the same configuration file layout as the other uses of TLS.  The reason why we don't use that function elsewhere is because other libraries (like libldap) don't expose the details of their TLS library, so we can't pass in an `SSL *`, and outbound RADSEC functionality hasn't been forward ported from v3 yet.

arr2036: As I said before, the redis code sets the FD, so that's not why the handshake is failing
```
static int redisSSLConnect(redisContext *c, SSL *ssl) {
    if (c->privctx) {
        __redisSetError(c, REDIS_ERR_OTHER, "redisContext was already associated");
        return REDIS_ERR;
    }

    redisSSL *rssl = hi_calloc(1, sizeof(redisSSL));
    if (rssl == NULL) {
        __redisSetError(c, REDIS_ERR_OOM, "Out of memory");
        return REDIS_ERR;
    }

    c->funcs = &redisContextSSLFuncs;
    rssl->ssl = ssl;

    SSL_set_mode(rssl->ssl, SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER);
    SSL_set_fd(rssl->ssl, c->fd);
    SSL_set_connect_state(rssl->ssl);
```

https://github.com/redis/hiredis/blob/master/ssl.c#L356

arr2036: If you look at the hiredis code, it's not doing anything special to prepare the `SSL *` https://github.com/redis/hiredis/blob/master/ssl.c#L415

mkhon: @arr2036 redis code not only prepares `SSL *` but also does `SSL_connect()` again (first time it is done in `fr_tls_session_alloc_client()` so I get `SSL_connect failed: Connection reset by peer` error when it is invoked for the second time

alandekok: Since nothing else uses `fr_tls_session_alloc_client()`, it's probably worth removing the call to `SSL_connect()` from that function.  The callers can then set up whatever IO handlers are necessary.

Please submit a separate PR for `fr_tls_session_alloc_client()`, which fixes the issue you found in it.  We can then have a separate PR for the redis SSL patches.

mkhon: @alandekok there is another issue: I can't use `SSL_CTX` created by `fr_tls_ctx_alloc()` because it sets msg, info and verify callbacks (which get called when redis tries to use `SSL` object created using that `SSL_CTX` and break because there are no any bound FR requests and such.

I do not see reasons to set these callbacks in `SSL_CTX` becasue the only users of `fr_tls_ctx_alloc()` are EAP modules which all use `eap_tls_session_init()` which in turn uses `fr_tls_session_alloc_server()` which sets own msg, info and verify callbacks.

So I would suggest to not set any callbacks for `SSL_CTX` in `fr_tls_ctx_alloc()`. What do you think?

alandekok: the `fr_tls_session_alloc_client()` function should probably not call `fr_tls_session_request_unbind(tls_session->ssl)` at the end of the function.  It's fine to have an internal (and otherwise unused) request associated with the client.

That way the callbacks will always have a `request` associated with them.  And it will be the same `request` for all of the callbacks.

mkhon: @alandekok do we really need these `SSL_CTX` callbacks for generic TLS clients? At least verify callback (as it is implemented now) breaks client TLS handshake for example.

I propose to keep the common code simpler:
- do not set up any callbacks when we create `SSL_CTX` - all the existing users of `fr_tls_ctx_alloc()` currently use `fr_tls_session_alloc_server()` which sets up necessary callbacks for them on its own (on `SSL` object)
- do not do `SSL_connect()` in `fr_tls_session_alloc_client()` like you suggested, or do it only when we pass in `fd` - this is the change that I have in my working tree

What do you think?

arr2036: Agree `SSL_connect()` should be removed from the alloc_client function.  If you want to ifdef the verify callbacks out and submit the PR that'd be fine with me.  Then we'll go back and figure out adding a fake request to the `SSL *` to run them later.

mkhon: @alandekok @arr2036 please review #5046 which contains src/lib/tls changes that are required

arr2036:
OK, you should be good to continue now.  The only change I didn't merge was removing the verify callback from the tls ctx alloc function, as that'd break validation in all the EAP methods.  If setting the callback to NULL on the `SSL *` doesn't override the function from the `SSL_CTX *` then you can remove it from the `SSL_CTX *` but it needs to be set in server `SSL *` alloc function.

mkhon:
> OK, you should be good to continue now. The only change I didn't merge was removing the verify callback from the tls ctx alloc function, as that'd break validation in all the EAP methods. If setting the callback to NULL for the `SSL *` doesn't override the function from the `SSL_CTX *` then you can remove it from the `SSL_CTX *` but it needs to be set in server `SSL *` alloc function.

@arr2036 seems that you also haven't merged the commit that removes SSL_connect() call

mkhon:
> Ahh, ok, fine. I've pushed a fix to remove the call from the `SSL_CTX *` alloc function. Sorry, I assumed the code was sane/non-duplicative.

There's two more issues with this approach:

1) You haven't fixed msg callback to cope with NULL request. This fix is like this:

```
diff --git a/src/lib/tls/session.c b/src/lib/tls/session.c
index 98fb43ff82..07cf5fdc45 100644
--- a/src/lib/tls/session.c
+++ b/src/lib/tls/session.c
@@ -615,7 +615,7 @@ static void session_msg_log(request_t *request, fr_tls_session_t *tls_session, u
         *      Don't print this out in the normal course of
         *      operations.
         */
-       if (!RDEBUG_ENABLED2) return;
+       if (!ROPTIONAL_ENABLED(RDEBUG_ENABLED2, DEBUG_ENABLED3)) return;
 
        if (((size_t)tls_session->info.version >= NUM_ELEMENTS(tls_version_str)) ||
            !tls_version_str[tls_session->info.version]) {
@@ -699,10 +699,10 @@ static void session_msg_log(request_t *request, fr_tls_session_t *tls_session, u
         *      Print out information about the record and print the
         *      data at higher debug levels.
         */
-       if (RDEBUG_ENABLED4) {
-               RHEXDUMP4(data, data_len, "%s", tls_session->info.info_description);
+       if (ROPTIONAL_ENABLED(RDEBUG_ENABLED4, DEBUG_ENABLED4)) {
+               ROPTIONAL(RHEXDUMP4, HEXDUMP4, data, data_len, "%s", tls_session->info.info_description);
        } else {
-               RDEBUG2("%s", tls_session->info.info_description);
+               ROPTIONAL(RDEBUG2, DEBUG2, "%s", tls_session->info.info_description);
        }
 }
 
```

2) redisInitiateSSL() takes ownership of the passed `SSL *` object. fr_tls_session_t can't actually use `.ssl` pointer because of this (because its lifetime is controlled by redis connection). So there is no reason in using `fr_tls_session_alloc()` and the only option is to use just `SSL_new()` on `SSL_CTX *` object obtained using `fr_tls_ctx_alloc()`

What do you think?

mkhon:
> Ahh, ok, fine. I've pushed a fix to remove the call from the `SSL_CTX *` alloc function. Sorry, I assumed the code was sane/non-duplicative.

> > Ahh, ok, fine. I've pushed a fix to remove the call from the `SSL_CTX *` alloc function. Sorry, I assumed the code was sane/non-duplicative.
> 
> @arr2036 There's two more issues with this approach:
> 
> 1. You haven't fixed msg callback to cope with NULL request. This fix is like this:

Alternatively, I'd omit setting msg callback for `SSL_CTX *` object (and in `fr_tls_session_alloc_client`)

mkhon:
@arr2036 Please check #5051 for additional fixes

mkhon:
@arr2036 there's one unmerged commit from #5051. Please check #5056

The issue is that redisInitiateSSL() takes ownership of the passed `SSL *` object. So `fr_tls_session_t` can't actually use .ssl pointer because of this (because its lifetime is controlled by redis connection). So `fr_tls_session_alloc()` can not be used and the only option is to use just `SSL_new()` on `SSL_CTX *` object obtained using `fr_tls_ctx_alloc()`

arr2036:
Looks like we can upref the `SSL *` https://www.openssl.org/docs/manmaster/man3/SSL_new.html `SSL_up_ref(SSL *s);` before passing it to Redis.  Redis' attempt to free will just become a decref.

mkhon:
@alandekok @arr2036 please review the updated PR

Please note that "use_cluster_map = no" is required to support stunnel-based TLS: stunnel-based TLS was functioning in our setup (before native TLS) only because
1) lib/redis has not been using DNS resolution (which is fixed in this PR - otherwise native TLS does not work with AWS ElastiCache)
2) AWS ElastiCache returns hostnames (not IP addresses) in "CLUSTER NODES" response

So, in this case cluster map initialization failed before (this was a non-fatal error) with the following error:

```
rlm_redis (redis) - Bootstrap server "localhost" returned invalid data: Not IPv4/6 address, and asked not to resolve
```

And redis modules were operating without a cluster map.

Now that DNS resolution is fixed, stunnel-based TLS has become broken because (when initializing cluster map) freeradius tries to reconnect to cluster nodes (which accept only TLS) using plaintext (because in stunnel-based TLS freeradius initially connects to redis as plaintext).

arr2036:
Looks better.  The tls session should be alloc'd from the conn not the context, so that'll need to be allocated higher in the function... Then when the conn is freed, the tls session should also be freed, and the destructor should decref the `SSL *`.

You can just reference `OPENSSL_LIBS`, `OPENSSL_LDFLAGS`, `OPENSSL_CPPFLAGS` in the libfreeradius-redis' all.mk file, no need to search for that again in the configure script.

Configuration for disabling cluster map seems fine.

```
SRC_CFLAGS := $(OPENSSL_CPPFLAGS)
TGT_LDLIBS  := $(OPENSSL_LIBS)
TGT_LDFLAGS := $(OPENSSL_FLAGS)
```

mkhon:
@arr2036 

> Looks better. The tls session should be alloc'd from the conn not the context, so that'll need to be allocated higher in the function... Then when the conn is freed, the tls session should also be freed, and the destructor should decref the `SSL *`.

This should be fixed now

> You can just reference `OPENSSL_LIBS`, `OPENSSL_LDFLAGS`, `OPENSSL_CPPFLAGS` in the libfreeradius-redis' all.mk file, no need to search for that again in the configure script.
> 
> ```
> SRC_CFLAGS := $(OPENSSL_CPPFLAGS)
> TGT_LDLIBS  := $(OPENSSL_LIBS)
> TGT_LDFLAGS := $(OPENSSL_FLAGS)
> ```

hiredis_ssl configure check requires OPENSSL_LDFLAGS at least, so this needs to be set at configure stage, doing that in all.mk is not enough

PR #5036, opened 2023-06-07
Fix missing close() in listen_bind()

jpereira:


alandekok:
Fixed with a few other `close()` added.

PR #5041, opened 2023-06-07
dhcpclient: Add missing '-r <retries>' help message.

jpereira:


alandekok:
Can you sign the commits?  GitHub says that they're not signed

arr2036:
I sign all my commits... but it seems like the ones made to PRs directly are no longer signed by GitHub

PR #5046, opened 2023-06-09
TLS changes required for native redis TLS support

mkhon:
There is more than one way to implement the changes so please merge with changes where required

### DIFF
--- a/src/lib/tls/session.c
+++ b/src/lib/tls/session.c
@@ -1624,7 +1624,6 @@ static void session_init(fr_tls_session_t *session)
  */
 fr_tls_session_t *fr_tls_session_alloc_client(TALLOC_CTX *ctx, SSL_CTX *ssl_ctx)
 {
-	int			ret;
 	int			verify_mode;
 	fr_tls_session_t	*tls_session = NULL;
 	request_t		*request;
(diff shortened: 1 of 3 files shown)


src/lib/tls/session.c
```diff
@@ -1657,7 +1657,7 @@ fr_tls_session_t *fr_tls_session_alloc_client(TALLOC_CTX *ctx, SSL_CTX *ssl_ctx)
 	DEBUG2("Requiring Server certificate");
 	verify_mode = SSL_VERIFY_PEER;
 	verify_mode |= SSL_VERIFY_FAIL_IF_NO_PEER_CERT;
-	SSL_set_verify(tls_session->ssl, verify_mode, fr_tls_verify_cert_cb);
+	SSL_set_verify(tls_session->ssl, verify_mode, NULL);
```
alandekok, line 1660: If there are no verify callbacks, do we still need to add a `request` and associate it with the `SSL*`?

The default for new `SSL*` is also to have no verify callbacks.  So there's not much point in calling `SSL_CTX_set_verify()` with a `NULL` argument.

Is there any _danger_ in leaving the verify callbacks in place, with an associated `request`?  What impact would that have?

arr2036: There needs to be something in there to free the internal request when the `SSL *` is freed, else we leak memory.

mkhon: @alandekok 

1) we still have msg and info callbacks

2) verify callback require eap_tls_handshake_push() context (there's `tls_session->can_pause` check in the beginning of this function) which can not be provided for client sessions. So client session handshake fails

arr2036: @alandekok It's fairly complex to make async validation work correctly, so I'm ok just dropping the verification callback and internal request, and doing the work ourselves.

PR #5051, opened 2023-06-12
Additional TLS fixes required for native redis SSL support

mkhon:


mkhon:
@arr2036 is this PR good to go?

arr2036:
I've made equivalent changes that check for request in all the logging functions

PR #5052, opened 2023-06-12
Clean up remains of SoH protocol

jpereira:
Related to the previous 309f347 commit.


jpereira:
That changes must fix the CI for now.

PR #5054, opened 2023-06-13
fix compiler warnings when WITH_TCP id not defined

chiouss:
Fix some compiler warnings here.

### DIFF
--- a/src/main/listen.c
+++ b/src/main/listen.c
@@ -3205,7 +3205,9 @@ static rad_listen_t *listen_parse(CONF_SECTION *cs, char const *server)
 	char const	*value;
 	fr_dlhandle	handle;
 	CONF_SECTION	*server_cs;
+#ifdef WITH_TCP
 	char const	*p;
+#endif
 	char		buffer[32];
 
 	cp = cf_pair_find(cs, "type");
--- a/src/main/process.c
+++ b/src/main/process.c
@@ -5394,7 +5394,9 @@ static void event_new_fd(rad_listen_t *this)
 		this->print(this, buffer, sizeof(buffer));
 		ERROR("Failed adding event handler for socket %s: %s", buffer, fr_strerror());
 		this->status = RAD_LISTEN_STATUS_EOL;
+	#ifdef WITH_TCP
 		goto listener_is_eol;
+	#endif
 	} /* end of INIT */
 
 	if (this->status == RAD_LISTEN_STATUS_PAUSE) {


alandekok:
I've manually merged the patch to `listen.c`

The patch to `process.c` doesn't make any sense.  I don't see why it's necessary, and it changes the logic of the code completely.

chiouss:
> I've manually merged the patch to `listen.c`

Thanks.

> The patch to `process.c` doesn't make any sense. I don't see why it's necessary, and it changes the logic of the code completely.

because the `listener_is_eol` tag is inside the `WITH_TCP` defined
https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/main/process.c#L5449

alandekok:
@chious Removing the `goto` means that the listener state will be `RAD_LISTEN_STATUS_EOL`, which then isn't handled when building without TCP.  As a result, the listener will stick around, and won't be cleaned up.

The solution is to set the listener state to `RAD_LISTEN_REMOVE_NOW`.  I'll push a fix.

chiouss:
@alandekok thanks.

chiouss:
Hi @alandekok ,
71c8188 makes compiler error.

```
src/main/process.c: In function 'event_new_fd':
src/main/process.c:5526:43: error: 'proxy_eol_cb' undeclared (first use in this function)
     fr_packet_list_walk(proxy_list, this, proxy_eol_cb);
                                           ^
src/main/process.c:5526:43: note: each undeclared identifier is reported only once for each function it appears in
src/main/process.c:5550:29: error: 'listen_socket_t {aka struct listen_socket_t}' has no member named 'ev'
           this, &when, &sock->ev)) {
                             ^
scripts/boiler.mk:644: recipe for target 'build/objs/src/main/process.lo' failed
make[3]: *** [build/objs/src/main/process.lo] Error 1
```
is `proxy_eol_cb` OK to move outsite `WITH_TCP` define?

chiouss:
> Hi @alandekok , [71c8188](https://github.com/FreeRADIUS/freeradius-server/commit/71c8188f3b1b8e12633800dde456b20124101e3c) makes compiler error.
> 
> ```
> src/main/process.c: In function 'event_new_fd':
> src/main/process.c:5526:43: error: 'proxy_eol_cb' undeclared (first use in this function)
>      fr_packet_list_walk(proxy_list, this, proxy_eol_cb);
>                                            ^
> src/main/process.c:5526:43: note: each undeclared identifier is reported only once for each function it appears in
> src/main/process.c:5550:29: error: 'listen_socket_t {aka struct listen_socket_t}' has no member named 'ev'
>            this, &when, &sock->ev)) {
>                              ^
> scripts/boiler.mk:644: recipe for target 'build/objs/src/main/process.lo' failed
> make[3]: *** [build/objs/src/main/process.lo] Error 1
> ```

and there is not `&sock->ev` if `WITH_TCP` is not defined below:
https://github.com/FreeRADIUS/freeradius-server/blob/v3.0.x/src/main/process.c#L5488

alandekok:
I'll take a look, but building without TCP isn't a high priority.  There are really few reasons to do this unless you have a_very_ constrained environment.

chiouss:
> I'll take a look, but building without TCP isn't a high priority. There are really few reasons to do this unless you have a_very_ constrained environment.

Thanks. I'm running it on a embeded device so I've disabled/removed all unnecessary modules and feature flags to make it smaller to fit the device's flash.

I'll change the status `RAD_LISTEN_STATUS_EOL` to `RAD_LISTEN_STATUS_REMOVE_NOW` for a quick fix without TCP build.

PR #5055, opened 2023-06-13
Auth Lost: Free the state's opaque and ctx storage

terryburton:


alandekok:
Manually merged because github was complaining about GPG signatures.

PR #5063, opened 2023-06-15
Fix wrong ifdef

jpereira:
... It should be only when build with EVFILT_LIBKQUEUE

Fixing the below _unused_ warning on Linux

```
src/lib/util/event.c:96:12: error: unused variable 'log_conf_kq' [-Werror,-Wunused-variable]
static int log_conf_kq;
           ^
1 error generated.
make: *** [scripts/boiler.mk:737: build/objs/src/lib/util/event.lo] Error 1
```

### DIFF
--- a/src/lib/util/event.c
+++ b/src/lib/util/event.c
@@ -92,7 +92,7 @@ static fr_table_num_sorted_t const kevent_filter_table[] = {
 };
 static size_t kevent_filter_table_len = NUM_ELEMENTS(kevent_filter_table);
 
-#ifdef __linux__
+#ifdef EVFILT_LIBKQUEUE
 static int log_conf_kq;
 #endif
 


arr2036:
Better to add a configure check to require 2.6.1 I think, given that older versions are just broken.

jpereira:
@arr2036 for sure, therefore if you take a look at https://github.com/FreeRADIUS/freeradius-server/blob/92cdbf374d0b7303a92f58247e57f0434d05ce07/src/lib/util/event.c#L2842 we are using `log_conf_kq` around `#ifdef EVFILT_LIBKQUEUE ... #endif`

arr2036:
Just writing a simple configure check which verifies if the custom EVFILT is present would be enough.  Versions of libkqueue which ship with Ubuntu do not work, do I don't see the point in adding support for these versions.  If you wanted to submit patches for libkqueue which added version.h and didn't cause the Windows build to fail (that was the last problem I had) then that'd also be fine.

PR #5071, opened 2023-06-20
Some dhcpv6 fixes

jpereira:


arr2036:
Initialising a fr_pair_list_t like that shouldn't leak memory because it's allocated on the stack... but there's no issue moving the initialisation call later.

PR #5073, opened 2023-06-20
Fix dhcpv6 example

jpereira:
* Command 'Offer' dont't exist in dhcpv6.
* Add better sample for 'Reply' based on received commands.

### DIFF
--- a/raddb/sites-available/dhcpv6
+++ b/raddb/sites-available/dhcpv6
@@ -296,14 +296,6 @@ recv Request {
 	ok
 }
 
-send Offer {
-	#
-	#  All replies MUST include a Server-ID option
-	#
-	&reply.Server-ID.DUID = UUID
-	&reply.Server-ID.DUID.UUID.Value := 0x00000000000000000000000000000000
-}
-
 #
 #  We also support recv / send for all normal DHCPv6 packet types.
 #
@@ -319,8 +311,34 @@ send Reply {
 	#
 	#  All replies MUST include a Server-ID option
 	#
+	&reply.Transaction-ID = &request.Transaction-ID
+
 	&reply.Server-ID.DUID = UUID
 	&reply.Server-ID.DUID.UUID.Value := 0x00000000000000000000000000000000
+
+	switch &request.Packet-Type {
+		case Request {
+			&reply += {
+				#&SOL-Max-RT = 60
+				#&DNS-Servers = fd85:d2bb:92c::1
+				#&DNS-Servers = fd85:d2bb:92c::2
+				#&Domain-List = "example.com"
+			}
+		}
+
+		case Release {
+			&reply += {
+				#&Client-ID.DUID = UUID
+				#&Client-ID.DUID.UUID.Value = 0x000102030405060708090a0b0c0d0e0f
+				#&Status-Code.Value = Success
+				#Status-Code.Message = "Release received."
+			}
+		}
+
+		case {
+			failed
+		}
+	}
 }
 
 }


raddb/sites-available/dhcpv6
```diff
@@ -319,8 +311,34 @@ send Reply {
...
+				#&DNS-Servers = fd85:d2bb:92c::1
+				#&DNS-Servers = fd85:d2bb:92c::2
+				#&Domain-List = "example.com"
+			}
+		}
+
+		case Release {
+			&reply += {
+				#&Client-ID.DUID = UUID
+				#&Client-ID.DUID.UUID.Value = 0x000102030405060708090a0b0c0d0e0f
```
arr2036, line 332: ... why would we add a static DUID to the example config?  All the client ID copying gets handled automatically by the DHCPv6 process module.

raddb/sites-available/dhcpv6
```diff
@@ -319,8 +311,34 @@ send Reply {
...
 	&reply.Server-ID.DUID.UUID.Value := 0x00000000000000000000000000000000
+
+	switch &request.Packet-Type {
+		case Request {
+			&reply += {
+				#&SOL-Max-RT = 60
+				#&DNS-Servers = fd85:d2bb:92c::1
+				#&DNS-Servers = fd85:d2bb:92c::2
+				#&Domain-List = "example.com"
+			}
```
arr2036, line 326: What's SOL-Max-RT?

raddb/sites-available/dhcpv6
```diff
@@ -319,8 +311,34 @@ send Reply {
...
+		}
+
+		case Release {
+			&reply += {
+				#&Client-ID.DUID = UUID
+				#&Client-ID.DUID.UUID.Value = 0x000102030405060708090a0b0c0d0e0f
+				#&Status-Code.Value = Success
+				#Status-Code.Message = "Release received."
```
arr2036, line 334: The status-code value is populated automatically from the section rcode.  Setting it in policy is bad practice unless there's a very good reason.

raddb/sites-available/dhcpv6
```diff
@@ -319,8 +311,34 @@ send Reply {
 	#
 	#  All replies MUST include a Server-ID option
 	#
+	&reply.Transaction-ID = &request.Transaction-ID
+
```
arr2036, line 315: No.  This is already handled by the DHCPv6 process module.

raddb/sites-available/dhcpv6
jpereira: well, the current code is not doing that automatically.

arr2036:
If you want to make some useful changes to the DHCPv6 code, then we need to following issues fixing:

- Server ID needs to be configurable on a per-virtual server basis.  Ideally this would be a configuration section with all the different DUID fields broken out, the DUID type, UUID type and UUID fields.  But for simplicity the following would be enough for now.

```
server_id {
    uuid {
        raw = <hex string>
    }
}
```
- This Server ID should be inserted into the response to Solicits, and other packet types which aren't to a specific DHCPv6 server.
- The listeners in the virtual server should check the server ID against incoming packets and drop packets which do not match.  This is because DHCPv6 relays are stupid, and will forward packets to all upstreams, and if multiple DHCPv6 servers reply, DHCPv6 clients are also stupid, and get confused, and fail to acquire leases.

raddb/sites-available/dhcpv6
arr2036: - https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L351
- https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L410
- https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L436
- https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L486
- https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L612

Feel free to submit a test case which shows that it's not doing this automatically, and we can figure out a fix.  It's probably an issue with the state table. https://github.com/FreeRADIUS/freeradius-server/blob/master/src/process/dhcpv6/base.c#L751

arr2036: ...but using a switch statement like this is just wrong.  There's send sections for the various reply types, if you did want to add canned responses you'd add them there.

alandekok: This patch is going into a `send Reply { ... }` section.  The issue is that a `Reply` can be sent in response to a `Confirm`, `Solicit`, `Renew`, `Rebind` ... section.  See https://datatracker.ietf.org/doc/html/rfc8415, and look for `/Reply`.

So we may need to change what's in the reply based on the type of the request.

Perhaps the simple way is to just document this, and in the `recv Confirm` etc. sections, put the "update reply" examples there.  And then also document `send Reply` that is is processed for multiple kinds of request packets.  And therefore the `send Reply` should only contain things which are relevant for all replies.

arr2036: Yeah that works...

arr2036: Or add `send reply.confirm` or `send reply_to_confirm` sections...

arr2036:
I think the majority of this PR needs changing, so closing it for now.  Feel free to reopen when changes have been made.

PR #5079, opened 2023-06-21
Don't use "git archive --add-file" which is implemented only in git 2.29

mkhon:
Ubuntu Focal 20.04 LTS has only git 2.25:

```
$ git --version
git version 2.25.1
```


arr2036:
We don't support < RHEL8 for v4.0.x.  I think this one is best kept in your local set of patches.

PR #5087, opened 2023-06-28
Make offer_time and lease_time optional as they are not required for RELEASE

mkhon:


arr2036:
@ndptech  commit fixes this.  As discussed in slack, using separate callenvs means the unused fields won't be populated.  The side effect of this change (or more the couple of commits before) is that if you were using `control.Pool-Action` before, you'll need to switch to using the module method call syntax (`pool.alloc`, `pool.release`, etc...) https://github.com/FreeRADIUS/freeradius-server/commit/56d7de8e1401120e64aa8345fd7e5b0bc935d102

PR #5088, opened 2023-06-28
Add process tests for dhcpv6

jpereira:


src/tests/process/dhcpv6/Request
```diff
@@ -0,0 +1,32 @@
...
+		&IA-PD-Prefix.Preferred-Lifetime = 7200
+		&IA-PD-Prefix.Valid-Lifetime = 7500
+		&IA-PD-Prefix.IPv6-Prefix = 2a00:1:1:100::/56
+	}
+
+	call dhcpv6 {
+		if (&reply.Packet-Type != Reply) {
+			test_fail
+		}
+	}
```
arr2036, line 31: You need to check that Client-ID and Server-ID are echo'd back from the server in the reply.

src/tests/process/dhcpv6/Request
```diff
@@ -0,0 +1,32 @@
...
+	&Server-ID.DUID.LLT.Hardware-Type.Ethernet.Address = 00:11:22:33:44:55
+
+	&Option-Request = DNS-Servers
+	&Option-Request = AFTR-Name
+
+	&Elapsed-Time = 0
+
+	&IA-PD.IAID = 33752069
+	&IA-PD.T1 = 3600
+	&IA-PD.T2 = 5400
```
arr2036, line 20: These would usually be set in the reply not the request

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+#		move_failure_message_to_parent = yes
+	}
+
+	recv Solicit {
+		ok
+	}
+
+	send Advertise {
+		&reply += {
+			&Transaction-ID = 0xd81eb8
```
arr2036, line 120: You shouldn't be setting this manually.  The whole point of the test is to verify that the process state machine does this for us.

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+			&IA-PD.IAID = 33752069
+			&IA-PD.T1 = 150
+			&IA-PD.T2 = 250
+
+			&IA-PD.Options = {
+				&IA-PD-Prefix.Preferred-Lifetime = 250
+				&IA-PD-Prefix.Valid-Lifetime = 300
+				&IA-PD-Prefix.IPv6-Prefix = 2a00:1:1:100::/56
+			}
+			&Client-ID.DUID = LL
```
arr2036, line 130: Again, should be set by the state machine, not manually

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+			&Client-ID.DUID.LL.Hardware-Type = Ethernet
+			&Client-ID.DUID.LL.Hardware-Type.Ethernet.Address = 00:01:02:03:04:05
+			
+			&Server-ID.DUID = LLT
```
arr2036, line 134: and this too...

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+	}
+
+	recv Solicit {
+		ok
+	}
+
+	send Advertise {
+		&reply += {
+			&Transaction-ID = 0xd81eb8
+			&IA-PD.IAID = 33752069
```
arr2036, line 121: I'm not sure if we currently auto-populate the IAID but we should

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+	recv Request {
+		ok
+	}
+
+	recv Information-Request {
+		ok
+	}
+
+	send Reply {
+		&Transaction-ID = 0x1e291d
```
arr2036, line 156: same as above, transaction ID IAID, client-id, server-id should all automatically be copied across by the state machine

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,179 @@
...
+#	radiusd -Xd src/tests/ -i 127.0.0.1 -p 12340 -n test
+#
+server dhcpv6 {
+	namespace = dhcpv6
+
+	all_dhcp_servers_and_relays = FF02::1:2
+	interface = eth0
+	port = 547
+
+	listen all_dhcp_servers_and_relays {
```
arr2036, line 81: Do we need these?  We're just running the process module

arr2036: I guess something needs to compile the sections, but maybe we can add a null transport @alandekok

PR #5090, opened 2023-06-30
Check for NULL pointers

jpereira:
It makes the CI happy again.

### DIFF
--- a/src/main/tmpl.c
+++ b/src/main/tmpl.c
@@ -778,7 +778,7 @@ finish:
 	 *	sure what we were passed is valid.
 	 */
 	memcpy(&vpt->data.attribute, &attr, sizeof(vpt->data.attribute));
-	if ((vpt->type == TMPL_TYPE_ATTR) && attr.da->flags.is_unknown) {
+	if ((vpt->type == TMPL_TYPE_ATTR) && attr.da && attr.da->flags.is_unknown) {
 		vpt->tmpl_da = (DICT_ATTR *)&vpt->data.attribute.unknown.da;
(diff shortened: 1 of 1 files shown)


src/main/tmpl.c
```diff
@@ -778,7 +778,7 @@ ssize_t tmpl_from_attr_substr(vp_tmpl_t *vpt, char const *name,
 	 *	sure what we were passed is valid.
 	 */
 	memcpy(&vpt->data.attribute, &attr, sizeof(vpt->data.attribute));
-	if ((vpt->type == TMPL_TYPE_ATTR) && attr.da->flags.is_unknown) {
+	if ((vpt->type == TMPL_TYPE_ATTR) && attr.da && attr.da->flags.is_unknown) {
```
alandekok, line 781: Hmm... the `da` can be in `vpt->data.attribute.unknown.da`, or in `attr.da`.  If it's in `vpt->data.attribute.unknown.da`, then `attr.da` is _always_ `NULL`, and this check is always skipped.

The field `attr.da` may be `NULL` if the execution runs through the block which ends in `goto do_num`.

In which case that's the only reason for `attr.da` to be NULL.  But `attr.unknown.da` is then not NULL, so it can be used instead.

jpereira: @alandekok the CI is complaining saying that it could be a NULL https://github.com/FreeRADIUS/freeradius-server/actions/runs/5425971300/jobs/9867417180#step:18:232

alandekok:
I've pushed a different fix.

PR #5093, opened 2023-07-03
Keep too-long lines from overwriting output_buffer

jejones3141:
This came up because of CID #1533664, but almost certainly won't placate coverity. It just makes sure that it won't ask for more than will fit in output_buff, at the extremely rare cost of not printing all of the remaining part of the line.


arr2036:
But the whole point of the sbuffs is that they don't allow overflow.?

jejones3141:
Not in this case; it's using fixed size buffers for the line  but more importantly for the output.

arr2036:
But the sbuffs have an end pointer which is one byte after the end of the fixed length buffer?

PR #5095, opened 2023-07-04
tests: Add process tests for dhcpv4 and dhcpv6

jpereira:


src/tests/process/dhcpv6/Request
```diff
@@ -0,0 +1,63 @@
...
+	&Elapsed-Time = 0
+
+	call dhcpv6 {
+		# 'Request' expects 'Reply'
+		if (&reply.Packet-Type != Reply) {
+			test_fail
+		}
+
+		# We should reply the below attributes...
+		if (!&reply.Transaction-ID) {
```
alandekok, line 27: Most of these checks could go into a policy?

```
policy dhcpv6_reply_matches_request {
   ...
```

and then the individual test just runs the policy.

That should simplify things a bit.

I think once that's done, we can merge this

src/tests/process/dhcpv6/server.conf
```diff
@@ -0,0 +1,112 @@
...
+		ok
+	}
+
+	recv Request {
+		ok
+	}
+
+	send Reply {
+		&reply += {
+			&IA-PD.IAID = 33752069
```
arr2036, line 98: Is the IAID not auto-populated?  We should definitely fix that.  Something like if there's a `IA-` container in the reply, and only one container of that type if the request, copy the IAID over automatically...

PR #5096, opened 2023-07-05
Added TLS Load Testing

delexagon:
Thoughts:
- Given the README in the tests directory, I assume src/tests/Makefile is the main testing harness (as compared to all.mk, which doesn't seem to be mentioned but does seem to run a variety of tests). I added the load test to run on `make all` (but not on `make tests`).
- The Docker build process can't be cached because the entire point of testing is to use the most recent version of Freeradius, so rebuilding takes a very long time to do. Unless the test is told to rebuild, which takes a very long time, the test will run off of the old image and not the most current version.
- The load testing I created seems very different than the rest of the existing tests, especially because it requires Docker which isn't a requirement for anything else. It doesn't really fit any of the existing categories. Because of that, I really wasn't sure how to integrate it. I have no idea what the right decision is, so adding it to the Makefile like this is a temporary solution.
- It seems likely that Docker redownloading/rebuilding Freeradius may be able to be avoided entirely by mounting build/ on /etc/freeradius and running it directly. I like the idea of this, but it deviates pretty heavily from what might be the expected behavior so I haven't implemented it.
- src/tests/Makefile did not run, so I changed it. First, I changed the generation of test.conf to prevent trying to link to a '//certs' directory (this seems to be an error, and still present in 3.2.x HEAD). Second, I changed it so that certs/server.pem and certs/ca.pem are generated before the server is started (this may not be desirable behavior).
- Docker generates an excessive amount of output, which may not be desirable.
Otherwise, a description of the script is in the src/tests/tls-load/README.md file.


arr2036:
All commits need to be signed.  Please merge the "Update README.md" commits into other related commits.

alandekok:
You don't need to close the PR.  You can rewrite your local commits in `git`, and then do a `git push --force` to update the PR.

delexagon:
Alright, I can do that. I would like to know whether you would prefer to run the test directly from the /build directory, rather than doing anything like rebuilding images for new tests? It seems like that's feasible and it seems like a much better solution, but it has the downside of me having to create a custom Dockerfile (to install necessary libraries, specifically libtalloc). It may take me a little while longer to get that working though.

PR #5097, opened 2023-07-06
Alternative TLS Load Test

delexagon:
- I changed src/tests/Makefile because it did not run. First, I fixed an apparent error in generating test.conf (two /s are generated in a directory path). Second, I changed it so that certs/server.pem and certs/ca.pem are generated before the server is started (this may not be desirable behavior). I would recommend looking over the Makefile to see if it's acceptable.
- I do not know the 'proper way' to add this test to the Makefile, because it seems very different from existing tests. I set it to run under 'all' but not under 'tests' in the testing Makefile.
- Docker generates an excessive amount of output, which may not be desirable when also running many other tests in a Makefile.
- I created a different pull request because this is a significantly different, and I think superior, strategy towards using Docker than what I had before.

### DIFF
--- a/src/tests/Makefile
+++ b/src/tests/Makefile
@@ -336,7 +336,11 @@ tests.runtests:  test.conf | radiusd.kill radiusd.pid
 	${Q}BIN_PATH="$(BIN_PATH)" PORT="$(PORT)" ./runtests.sh $(TESTS)
 
 # probably not the best way to do this
-tests.load: $(RADDB_PATH)certs/ca.pem $(RADDB_PATH)certs/server.pem $(RADDB_PATH)certs/client.pem
-	${Q}(cd tls-load && BUILD_DIR=${BUILD_PATH} LIB_DIR=${LIB_PATH} DICT_DIR=${top_builddir}/share RADDB_DIR=${top_builddir}/raddb ./run_test.sh -o output 10)
+.PHONY: $(BUILD_PATH)/tests/tls-load
+$(BUILD_PATH)/tests/tls-load:
+	${Q}mkdir -p $@
+
+tests.load: $(BUILD_PATH)/tests/tls-load $(RADDB_PATH)certs/ca.pem $(RADDB_PATH)certs/server.pem $(RADDB_PATH)certs/client.pem
+	${Q}(cd tls-load && BUILD_DIR=${BUILD_PATH} LIB_DIR=${LIB_PATH} DICT_DIR=${top_builddir}/share RADDB_DIR=${top_builddir}/raddb ./run_test.sh -o $(BUILD_PATH)/tests/tls-load 10)
 
 tests: tests.runtests tests.eap
--- a/src/tests/Makefile
+++ b/src/tests/Makefile
@@ -3,7 +3,7 @@
 ## Makefile -- Build and run tests for the server.
 ##
 ##	http://www.freeradius.org/
-##	$Id: f7ab1f9b400f3e3bee1d80ba97ae24b4c6bd7f2a $
+##	$Id$
 ##
 #
 include ../../Make.inc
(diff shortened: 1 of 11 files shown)


delexagon:
It seems like *.pem files are ignored by git, so certs does not have the necessary files to run the test off the bat. Also, the output directory is empty and isn't created which is awkward, so I'll fix that. Seeing as the Makefile also didn't work because of the lack of a .pem file, I'm not sure whether it's good form for me to generate them myself for the test, even if they will not be used by the application itself.

alandekok:
the PEM files are ignored by git.  They eventually expire, and then CI fails.

It's better to just regenerate them for the tests.

The test outputs should usually go into `${BUILD_PATH}/tests/tls-load`.  You can then add a rule:

```
.PHONY: $(BUILD_PATH)/tests/tls-load
$(BUILD_PATH)/tests/tls-load:
	${Q}mkdir -p $@
```

And the tests can then depend on `$(BUILD_PATH)/tests/tls-load`.  The `make` framework will take care of creating the output directory.  So you don't need to create an empty output directory and commit it to git.

src/tests/Makefile
```diff
@@ -188,6 +188,7 @@ config/eap-test: $(RADDB_PATH)mods-available/eap config/eap-test-inner-tunnel
 radiusd.pid: test.conf $(RADDB_PATH)certs/ca.pem $(RADDB_PATH)certs/server.pem
 	${Q}rm -rf $(TEST_PATH)/gdb.log $(TEST_PATH)/radius.log $(TEST_PATH)/tlscache
 	${Q}mkdir -p $(TEST_PATH)/tlscache
+	$(info ${RADIUSD_BIN} -Pxxxxml ${TEST_PATH}/radius.log -d ${top_builddir}/src/tests -n test -i 127.0.0.1 -p ${PORT} -D ${DICT_PATH})
```
alandekok, line 191: IIRC `$(info...)` prints out things when the the Makefile is loaded, not when the rule is executed.

If this is for Makefile debugging, it can be deleted.  If it's needed as log output, it should be an `echo`

delexagon: Yes, that was for me trying to understand how the build directory was structured and should be deleted.

delexagon: I will do so.

src/tests/tls-load/test/home-entrypoint.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+echo /etc/freeradius/lib/local/.libs >> /etc/ld.so.conf
```
alandekok, line 2: This isn't necessary.  If you're running the server out of the local source tree, you can wrap the execution of the scripts with `jlibtool`.  See `scripts/run/radiusd`.  You can just do

```
/path/to/source/scripts/run/radiusd ....
```

and it will pass in the correct command-line arguments too.

The `jlibtool` wrapper sets up a bunch of environment variables which allow things to be run correctly.

I'm also not sure why the server binary and libraries are in `/etc`?  That directory is for configuration.  It may be better to use a different directory, like `/build`, or even `/home/freeradius`.

This just makes it clearer where everything is.

delexagon: Alright, fixed.

delexagon:
Alright, I have resolved all of the discussed issues. I am still not confident that this is the right way to deal with Making the test.

delexagon:
Also, because I am running .so files directly generated by make within a Docker image, I do not think it will work if the base Docker image used for the test is different than the OS that the user is currently running, which seems to me to be a severe constraint of the script as is. I'm not sure whether it would be worthwhile writing something that attempts to find a docker image associated with the user's current OS and using it as the base?

PR #5103, opened 2023-07-11
Check fr_pair_afrom_da() for NULL return (CID #1503939)

jejones3141:


alandekok:
I've pushed a slightly different fix:

* use `vp` instead of longer name.  We don't access the `fr_pair_t` after it's been created, and we don't care about the name
* use `TEST_ASSERT()` instead of `if` and `TEST_CHECK`.  If the `vp` isn't created, that's a fatal error.

PR #5109, opened 2023-07-13
Make clear to fr_radius_ok() that totallen is validated (CID #1455247)

jejones3141:
Coverity thinks that calculating totallen via byte-swapping taints the buffer holding the packet to be decoded. Modeling fr_radius_ok() doesn't seem to have worked, so we'll try making it clear to coverity that we validate totallen.

### DIFF
--- a/src/protocols/radius/base.c
+++ b/src/protocols/radius/base.c
@@ -517,7 +517,7 @@ bool fr_radius_ok(uint8_t const *packet, size_t *packet_len_p,
 	 *
 	 *	i.e. No response to the NAS.
 	 */
-	if (packet_len < totallen) {
+	if (totallen > packet_len) {
 		FR_DEBUG_STRERROR_PRINTF("packet is truncated (received %zu <  packet header length of %zu)",
 					 packet_len, totallen);
 		failure = DECODE_FAIL_MIN_LENGTH_MISMATCH;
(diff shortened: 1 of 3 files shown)


src/protocols/radius/base.c
```diff
@@ -517,7 +517,7 @@ bool fr_radius_ok(uint8_t const *packet, size_t *packet_len_p,
 	 *
 	 *	i.e. No response to the NAS.
 	 */
-	if (packet_len < totallen) {
+	if (totallen > packet_len) {
```
alandekok, line 520: Why does changing the order of comparisons help?  Is Coverity that stupd?

jejones3141: Found the case; it's CID 1503921. The code in question (starting at src/protocols/dns/base.c:235)

`

				/*
				 *	Can't point to the header.
				 */
				if (offset < 12) {
					DECODE_FAIL(POINTER_TO_HEADER);
					return false;
				}

				/*
				 *	Can't point to the current label.
				 */
				if ((packet + offset) >= start) {
					DECODE_FAIL(POINTER_LOOPS);
					return false;
				}

				/* coverity[tainted_data] */
				if (!fr_dns_marker[offset]) {
					DECODE_FAIL(POINTER_TO_NON_LABEL);
					return false;
				}
`

That insists that offset be in [12, start - packet), but coverity doesn't realize it and gives the tainted {data, scalar} error.

PR #5110, opened 2023-07-14
dhcpv6: Fix {IANA,IATA,IAPD}.IA-ID fields

jpereira:
dhcpv6: Fix IA{N,P,T}A.IA-ID fields

    We should echo back IAID, and any other IDs like that, if they're in the request


alandekok:
The client should be creating and sending the IAID.  See RFC 8415 in many places.  The server must echo the IAID back, but the server should not create an IAID attribute.

PR #5112, opened 2023-07-18
dhcpv6: Fix {IANA,IATA,IAPD}.IA-ID fields

jpereira:
We should echo back IAID, and any other IDs like that, if they're in the request

### DIFF
--- a/src/process/dhcpv6/base.c
+++ b/src/process/dhcpv6/base.c
@@ -103,6 +103,9 @@ typedef struct {
 	fr_pair_t	*transaction_id;
 	fr_pair_list_t	client_id;
 	fr_pair_list_t	server_id;
+	fr_pair_list_t	iana_iaid;
+	fr_pair_list_t	iata_iaid;
+	fr_pair_list_t	iapd_iaid;
 } process_dhcpv6_client_fields_t;
 
 /** Records fields from the original relay-request so we have a known good copy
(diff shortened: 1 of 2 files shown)


alandekok:
The `fr_pair_copy_by_ancestor()` is used earlier in `base.c` to copy the `Client-ID` structure, _or_ the members of hte `Client-ID structure`.

In contrast, the `IAID` attribute is `uint32`, to the `fr_pair_copy_by_ancestor()` will just copy it alone.  Or maybe not even do that.

The client can also send multiple address requests.  See https://datatracker.ietf.org/doc/html/rfc8415#section-6.6.  So we'd need to copy the `IAID` attribute for each one.

Then when the attribute is restored, it has to be restored to the correct encapsulating structure.

I think the solution has to be more complex than what's shown here.  :(  It looks like the solution is more along the lines of "loop over the IA_FOO options which contain IAID, allocate addresses as per the contents of those options, and then echo the IAID back in the reply, along with those options"

I think the simplest thing to do here which isn't wrong is to just loop over all of the structures which might contain IAID, and ensure that (a) they actually contain IAID (and fail if not), and then perhaps also check that the IAIDs don't overlap.

RFC 8415 also says:  `A binding containing information about an IA is indexed by the tuple <DUID, IA-type, IAID>`.  i.e. the addresses are bound to `Client-ID + IA type + IAID` as the lease identifier.

arr2036:
@alandekok agreed.  As IAID is usually a field it shouldn't be missing *BUT* because we now support subrequests and calls, all bets are off.  I'd be in favour of failing outright if one of these group attributes come in without an IAID value, because they're so critical to how the protocol functions.  You can't do a per group attribute status-code, because you can't identify which group attribute in the response matched the request.

arr2036:
Not sure why 'upstream defect' was added...

PR #5118, opened 2023-07-24
Fix typo

jpereira:


alandekok:
`Base branch requires signed commits. Rebase merges cannot be automatically signed by GitHub`

PR #5121, opened 2023-07-26
add Calix-Role in Calix Dictionary

asakapab0i:
Some customers uses `Calix-Role` attribute and when the Calix dictionary was added to the v3.2.x branch it created a conflict error on our end and with lacking `Calix-Role` attribute.

### DIFF
--- a/share/dictionary.calix
+++ b/share/dictionary.calix
@@ -12,6 +12,7 @@
 
 VENDOR		Calix				6321
 
+ATTRIBUTE	Calix-Role			1 string
 ATTRIBUTE	Calix-CMS-User-Group			220	string
 ATTRIBUTE	Calix-CMS-Alarm-Filter			221	integer
 


aren:
The `Calix-Role` attribute is present on the `dictionary.calix` in `master`: https://github.com/FreeRADIUS/freeradius-server/blob/master/share/dictionary/radius/dictionary.calix

This same change should probably be made in v3.0.x branch as well, if you are still accepting changes there.

arr2036:
Commits to v3.2.x and v3.0.x need to be signed.  If you sign the commits i'll merge them.

arr2036:
Also be aware, if you're using Calix gear their RADIUS implementation is all pretty broken currently.  We're working with a customer to get the issues resolved... but don't plan on getting a sane User-Name in accounting packets, or being able to use User-Names longer than ~40chars in CoAs.

asakapab0i:
Hi @arr2036 , the commit has been signed. Thank you!

PR #5131, opened 2023-08-02
Initialize local value boxes for some coercion results (CIDs below)

jejones3141:
CIDs: #1503917, #1503948, #1503o5o, #1503989

These are in calc_{string, octet}(), where coercion results stored in locals must be cleared before returning. Currently coverity claims that the value box's type is not set. One would think that modeling fr_value_box_cast() to say it writes the destination on success would suffice, but that might not be the case, so we will initialize the value boxes just as we have done with some local buffers used in sbuffs.

### DIFF
--- a/src/lib/util/calc.c
+++ b/src/lib/util/calc.c
@@ -845,7 +845,8 @@ static int calc_octets(TALLOC_CTX *ctx, fr_value_box_t *dst, fr_value_box_t cons
 {
 	uint8_t *buf;
 	size_t len;
-	fr_value_box_t one, two;
+	fr_value_box_t one = (fr_value_box_t){};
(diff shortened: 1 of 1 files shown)


src/lib/util/calc.c
```diff
@@ -845,7 +845,8 @@ static int calc_octets(TALLOC_CTX *ctx, fr_value_box_t *dst, fr_value_box_t cons
 {
 	uint8_t *buf;
 	size_t len;
-	fr_value_box_t one, two;
+	fr_value_box_t one = (fr_value_box_t){};
+	fr_value_box_t two = (fr_value_box_t){};
```
arr2036, line 849: The compound literal is redundant here... could just be `{}` but whatever...

PR #5133, opened 2023-08-03
Deal with remaining uninitialized local issues (CIDs below)

jejones3141:
(CIDs: 1503918, 1503958, 1504020, 1504041, 1504436, 1506689, 1506890

### DIFF
--- a/src/bin/radsniff.c
+++ b/src/bin/radsniff.c
@@ -463,7 +463,7 @@ static void rs_packet_print_fancy(uint64_t count, rs_status_t status, fr_pcap_t 
 		}
 
 		if (conf->print_packet && (fr_debug_lvl >= L_DBG_LVL_2)) {
-			char vector[(RADIUS_AUTH_VECTOR_LENGTH * 2) + 1];
+			char vector[(RADIUS_AUTH_VECTOR_LENGTH * 2) + 1] = "";
 
 			fr_pair_list_sort(list, fr_pair_cmp_by_da);
 			fr_pair_list_log(&default_log, 4, list);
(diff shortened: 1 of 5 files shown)


alandekok:
These kinds of things make me wary.  Coverity is too stupid to notice that a `digest` buffer is initialized by a call top `fr_md5_calc(digest, ...)`.  So we have to mangle our code with unnecessary initializers, just so that Coverity won't complain.

Is there a way to annotate the `fr_md5_calc()` function for Coverity, to say "initializes output buffer" ?

arr2036:
Agreed as this is the OUTPUT buffer and we're not reading anything there shouldn't be an issue :/

jejones3141:
One very common schema: declare a local array of either char or uint8_t,
with a size determined by knowledge of the relevant RFC or API or by
deciding N is probably enough for some value of N. The code creates an
[sd]buff using that array as a buffer and pass that [sd]buff to various
other functions, ending by
using the local array directly At that point, coverity complains that the
array is uninitialized, because it doesn't recognize that the [sd]buff
calls do fill in the leading portion of the local array.

Perhaps coverity can infer when the [sd]buff functions fill in the array,
but clearly they do so iff the function returns success. There are a fair
number of places where the return isn't checked. (One function uses
FR_DBUFF_REMAINING_RETURN(dbuff, total # of bytes written by the next N
calls) and then doesn't check the returns of those calls. A human can infer
that those checks are unnecessary, but coverity can't. Yet.)

Coverity suggests the following mechanism for writing a model of the form
we need. You declare a local of the type the function returns, but don't
initialize it. You then test its value to determine whether to call one of
the __coverity_foo__() functions and which one and what parameters to pass
it.

So all these functions have the body

int result;
if (result > 0) __coverity_sbuff_write buffer_bytes(out->p, result);
return result;

Depending on the particular function, that may be >= 0, and for sbuffs,
write result + 1 for the terminating NUL.

The problem here: no example in coverity docs or in open source coverity
models I've seen make reference to a member of a structure. There are a few
special __coverity_foo__() functions that let one do something to an entire
struct without having to use sizeof, I think to let one use

typedef struct {
} foo_t;

for all structs in modeling functions. So, since the data being written is
pointed at by a member of an sbuff or dbuff,
it looks like [sd]buff functions can't be modeled in coverity.l

So... what I will do is close the pull request and try  the following:

   1. Pull the [sd]buff function models.
   2. Take one of the functions the pull request wanted to change and
   instead change it to check ALL [sd]buff calls that write to the local array.
   3. See whether coverity will give its approval to the changed function.

and of course continue to work on the other annotations.

PR #5135, opened 2023-08-06
add check for illumos

hadfl:
illumos, like FreeBSD does not define `SOL_TCP`


alandekok:
I've added this manually as this commit wasn't signed.

PR #5138, opened 2023-08-11
Attempt at a simpler uninit local buffer workaround (CIDs below)

jejones3141:
CIDs: 1506690, 1506689, 1504436, 1504041, 1504020, 1503918

This doesn't name the uninitialized local array, but instead goes via the sbuff/dbuff API to get to the data.


arr2036:
That's a pretty big change in coding style just to satisfy coverity.  It's not dealing with compound literals correctly?

arr2036:
OK, I see, thanks for the explanation.  It is slightly more within the spirit of dbuffs and sbuffs.

PR #5139, opened 2023-08-14
We don't need scripts/build/lsan_leaks_osx as we have all suppression in __lsan_default_suppressions()

jpereira:
* we don't need scripts/build/lsan_leaks_osx as we have all suppression in __lsan_default_suppressions()
* sorted in alphabetic order.

### DIFF
--- a/src/lib/util/debug.c
+++ b/src/lib/util/debug.c
@@ -155,22 +155,22 @@ char const CC_HINT(used) *__lsan_default_suppressions(void)
 	return
 		"leak:CRYPTO_THREAD_lock_new\n"		/* OpenSSL init leak - reported by heaptrack */
 #if defined(__APPLE__)
+		"leak:*gmtsub*\n"
+		"leak:ImageLoaderMachO::doImageInit\n"
+		"leak:_st_tzset_basic\n"
+		"leak:attachCategories\n"
+		"leak:fork\n"
 		"leak:getaddrinfo\n"
 		"leak:getpwuid_r\n"
-		"leak:*gmtsub*\n"
-		"leak:tzsetwall_basic\n"
-		"leak:ImageLoaderMachO::doImageInit\n"
 		"leak:libSystem_atfork_child\n"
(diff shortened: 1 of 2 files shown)


src/lib/util/debug.c
```diff
@@ -155,22 +155,26 @@ char const CC_HINT(used) *__lsan_default_suppressions(void)
...
 		"leak:load_images\n"
-		"leak:realizeClassWithoutSwift\n"
 		/* Perl >= 5.32.0 - Upstream bug, tracked by https://github.com/Perl/perl5/issues/18108 */
 		"leak:perl_construct"
+		"leak:newlocale\n"
+		"leak:realizeClassWithoutSwift\n"
+		"leak:tzset\n"
+		"leak:tzsetwall_basic\n"
+		"leak:wrap_calloc\n"
+		"leak:wrap_malloc\n"
```
arr2036, line 177: I mean yeah, if you just ignore leaks from the memory allocation functions themselves LSAN'll definitely won't report anything.

@jpereira on a personal level I have no issue with you, but your complete lack of attention to detail is not only frustrating, it's downright dangerous.  C requires you to understand what you're doing,  just having code that compiles and _appears_ to work isn't enough.  The reason why it takes so long for your Pull Requests to be merged is we don't trust that the code you've submitted is correct. One of the senior engineers has to take the time to go through line by line carefully checking that you've not made any catastrophic errors. 

It's not like (we've|I've) been silent on this either, my feedback is always around explaining your changes, demonstrate understanding... but it's been years at this point and you still rarely explain the reason behind your changes, or demonstrate an understanding of _why_ a particular fix works.

At this point my knee jerk reaction when you raise a PR or start working on an area of code is of concern and extreme caution.  When you start asking questions this turns into frustration... and finally there's a kind of bafflement, that you're still submitting pull requests after having so many closed, with so much negative feedback, and that, despite all that, you're still somehow making the same basic mistakes over and over again.

PR #5141, opened 2023-08-15
Set option_len before calling fr_nbo_to_uint32(). (CID #1503954)

jejones3141:


alandekok:
I'm not clear why this is necessary.  What is Coverity complaining about?

jejones3141:
> I'm not clear why this is necessary. What is Coverity complaining about?

For the fr_nbo_to_foo() functions, Coverity flags the passed pointer as tainted (or if you pass ptr + <const>, it considers ptr as tainted. Thereafter, any copy of ptr or data referenced via ptr or a copy thereof is also considered tainted. This is an attempt to keep option_len from being tainted.

jejones3141:
Better to deal with fr_nbo_to_foo() directly and deal with multiple coverity issues. Closing.

PR #5142, opened 2023-08-15
mrc/totp

mrw42:
Adds loopback support to TOTP, also makes the TOTP module configurable (including Time-Step, OTP-Length and loopback variables).  Adds a src/tests/modules/totp directory with TOTP testing instructions in the README.


src/modules/rlm_totp/rlm_totp.c
```diff
@@ -327,8 +341,13 @@ int main(int argc, char **argv)
 			(void) sscanf(argv[2], "%llu", &now);
 		}
 
-		if (totp_cmp(NULL, (time_t) now, (uint8_t const *) argv[3], strlen(argv[3]), argv[4]) == 0) {
-			return 0;
+		printf ("=== Time = %llu, TIME_STEP = %d, BACK_STEPS = %d, BACK_STEP_SECS = %d ===\n",
```
arr2036, line 428: Was this for debug output whilst you were developing?  If not and you want to emit log messages, then it's better to use the `RDEBUG` macros.  Most of these types of messages use `RDEBUG2(<fmt>,<args...>`.

src/modules/rlm_totp/rlm_totp.c
```diff
@@ -226,7 +240,7 @@ static rlm_rcode_t CC_HINT(nonnull) mod_authenticate(UNUSED void *instance, REQU
 	uint8_t const *key;
 	size_t keylen;
 	uint8_t buffer[80];	/* multiple of 5*8 characters */
-
+	uint64_t now = time(NULL);
```
arr2036, line 312: @alandekok are we still on non-monotonic time for v3?

src/modules/rlm_totp/rlm_totp.c
mrw42: No, I intended for it to stay here.  This is in the main() function, that is only complied when the stand-alone test application is built.  I thought I shouldn't use RDEBUG3 here, because I didn't have a request.  However, it appears that "RDEBUG3" is redefined as "printf" when TESTING is defined, so I could have used RDEBUG3, after all.  The point of this line is to make it clear when a group of backstops is performed as part of a single TOTP authentication.

alandekok:
I've pulled this over manually with some tweaks.

PR #5143, opened 2023-08-15
Add missing 'Interim-Update' accounting statement

jpereira:


arr2036:
It's already there as "Alive".  I'm not sure why it's called "Alive" and not Interim-Update.  @alandekok ?

arr2036:
Ignoring the underlying problem, submitting the quick fix

alandekok:
`Alive` is the old name from RFC 2038 or so.  `Interim-Update` is the new name.  So arguably the `accounting Alive` section should be replaced with `accounting Interim-Update`

PR #5167, opened 2023-09-01
Fix typo

jpereira:


jpereira:
weird, the editor is showing a syntax error. therefore, it is related to the _macro_ `fr_value_box_list_foreach_safe`

alandekok:
The macro is a little weird. :(

arr2036:
Yeah you need the extra closing curly brace.  Not really any other way to fix it unfortunately.

arr2036:
Unless maybe have a loop close macro or something?

jpereira:
I'd suggest something like top macro. it looks reasonable.

PR #5172, opened 2023-09-08
Get rid of needless check (CID #1544258)

jejones3141:
The while loop condition is that vp is non-NULL, and no code in the loop body changes vp, it will always pass vp->da to fr_proto_da_stack_build(), and the conditional expression is unnecessary.


alandekok:
Fixed as part of other rewrites.

PR #5175, opened 2023-09-12
Set pair list verified on initialization (true) and append (false)

jejones3141:
CIDs: #1544620 through #1544624


src/lib/util/pair.c
```diff
@@ -53,6 +53,7 @@ void fr_pair_list_init(fr_pair_list_t *list)
 	 */
 	fr_pair_order_list_talloc_init(&list->order);
 
+	list->verified = true;
```
alandekok, line 56: these should be protected with `#ifdef WITH_VERIFY_PTR`, as with the other uses of the `list->verified` field.

jejones3141: Will do.

jejones3141: Done.

PR #5181, opened 2023-09-15
rlm_totp: Backport changes from v3.2.x

jpereira:
* Backport recent changes in rlm_totp from v3.2.x
* Move the TOTP functionality to libfreeradius-totp.a
* Add scripts/totp/totp-gen.py script to generate real token
* Add a new test against rlm_totp.

e.g:

```
$ make test.totp
TOTP-TEST INPUT=now_token_6digits.txt TOKEN=085509 TOTP_GEN_ARGV="-d 6 -D sha1"
TOTP-TEST INPUT=now_token_8digits.txt TOKEN=43085509 TOTP_GEN_ARGV="-d 8 -D sha1"
$
```


raddb/mods-available/totp
```diff
@@ -43,4 +43,27 @@
 #  The module takes no configuration items.
```
alandekok: that comment should be fixed.  :)  I'll push a change for 3.2

src/tests/totp/bin/Makefile
```diff
@@ -3,26 +3,29 @@
...
 #
 .PHONY: src
 src:
-	@ln -sf ../../../src
+	@ln -sf ../../../../src
 
 .PHONY: freeradius-devel
 freeradius-devel:
-	@ln -sf ../../../src/include freeradius-devel
+	@ln -sf ../../../../src/include freeradius-devel
```
alandekok, line 28: I'm not sure why this is necessary?  the `src` directory is only 3 directories up.  If I do:

```
$ cd src/modules/rlm_totp
$ ls -al ../../../src/
```

I get the contents of the `src` directory, as expected.

doc/antora/modules/raddb/nav.adoc
```diff
@@ -67,6 +67,7 @@
 *** xref:mods-available/sqlippool.adoc[SQL-IP-Pool Module]
 *** xref:mods-available/sradutmp.adoc[sRadutmp Module]
 *** xref:mods-available/stats.adoc[Stats Module]
+*** xref:mods-available/totp.adoc[TOTP Module]
```
alandekok, line 70: you likely also need to do `git add doc/antora/..../totp.adoc`, otherwise the file won't exist.

src/tests/totp/bin/Makefile
jpereira: @alandekok That `Makefile` is in `src/tests/totp/bin/`.

so, we need to go up 4 levels. e.g: 

```
$ (cd src/tests/totp/bin/ && ls -adl ../../../../src/)
drwxr-xr-x  16 jpereira  staff  512 Sep 14 22:28 ../../../../src/
$
```

raddb/mods-available/totp
jpereira: Ops! done

src/tests/totp/bin/Makefile
jpereira: I replaced by `$(top_srcdir)`

alandekok:
I've pulled most of this over, except for a few pieces:

* I didn't revert the `call_env` stuff.  That is likely not too hard
* the TOTP dictionary needs to be moved, but I ran into other issues with the dictionaries
* the last commit for tests needs the rlm_totp without the `call_env` stuff, so that's been left.

But the module builds and seems to work, so that's a step forward.

PR #5187, opened 2023-09-23
Fix Antora examples to not use update

jpereira:


alandekok:
Merged manually with a few corrections.

PR #5194, opened 2023-09-29
add check against NULL before calling EVP_DigestInit_ex to avoid NULL…

icy17:
… deref.


alandekok:
I'm not sure why these patches are necessary.  If the EVP call fails, then something is majorly wrong with the server and with OpenSSL.  In that case, perhaps the best thing to do is to just exit.

It would also be good to explain what real-world problem is fixed by adding this code.  If the issue is largely theoretical, it's much less interesting.

We also have very strong opinions on code formatting.  It's not good to new code with completely different formatting than the rest of the server.  The end result of that would be that the server is a random collection of randomly formatted code.

As a result, we cannot merge this.

PR #5200, opened 2023-10-04
move to %func(args) everywhere

jpereira:


raddb/mods-available/mschap
```diff
@@ -336,7 +336,7 @@ mschap {
 		#  TIP: We give both examples here, but *only one should be used*.
 		#
 #		local_cpw = "%exec(/path/to/script %mschap(User-Name) %{MS-CHAP-New-Password.Cleartext})"
-#		local_cpw = "%sql(UPDATE radcheck set value='%{MS-CHAP-New-NT-Password}' where username='%{User-Name}' and attribute='Password.NT'}"
+#		local_cpw = "%sql(UPDATE radcheck set value='%{MS-CHAP-New-NT-Password}' where username='%{User-Name}' and attribute='Password.NT')"
```
alandekok, line 339: The argument parser will get upset over spaces in the arguments.  This should be:
```
local_cpw = %sql("UPDATE radcheck set value='%{MS-CHAP-New-NT-Password}' where username='%{User-Name}' and attribute='Password.NT'")
```

i.e. the outer quotes aren't necessary.  But the inner quotes are.  The `"UPDATE... where ..."` statement is one string, and has to be passed to to the `%sql()` function as one string.

raddb/mods-available/mschap
```diff
@@ -336,7 +336,7 @@ mschap {
 		#  TIP: We give both examples here, but *only one should be used*.
 		#
 #		local_cpw = "%exec(/path/to/script %mschap(User-Name) %{MS-CHAP-New-Password.Cleartext})"
```
alandekok, line 338: That should be 

```
%exec(`/path/to/script`, %mschap(User-Name). %{MS-CHAP-New-Password.Cleartext})
```

commas, not spaces between the arguments.

raddb/mods-config/sql/main/cassandra/queries.conf
```diff
@@ -49,7 +49,7 @@ sql_user_name = "%{User-Name}"
 # distrust the provided Event-Timestamp.
 #event_timestamp_epoch = "%l"
 
-event_timestamp_epoch = "%{%(integer:%{Event-Timestamp}):-%l}"
+event_timestamp_epoch = "%integer(%{&Event-Timestamp || %l})"
```
alandekok, line 52: That can now be

```
event_timestamp_epoch = %{((integer) &Event-Timestamp) || %l}
```

_cast_ to integer, as we don't need an `%integer()` function any more.  And the outer quotes aren't necessary.

raddb/mods-config/sql/main/mssql/queries.conf
```diff
@@ -323,10 +323,10 @@ accounting {
 				SET \
 					AcctStopTime = ${....event_timestamp}, \
 					AcctSessionTime = '%{Acct-Session-Time}', \
-					AcctInputOctets = convert(bigint, '%{%{Acct-Input-Gigawords}:-0}' * POWER(2.0, 32)) | '%{%{Acct-Input-Octets}:-0}', \
-					AcctOutputOctets = convert(bigint, '%{%{Acct-Output-Gigawords}:-0}' * POWER(2.0, 32)) | '%{%{Acct-Output-Octets}:-0}', \
+					AcctInputOctets = convert(bigint, '%{&Acct-Input-Gigawords || 0}' * POWER(2.0, 32)) | '%{&Acct-Input-Octets || 0}', \
```
alandekok, line 326: Maybe just do all this in FR?

```
AcctInputOctets = %{(((uint64) &Acct-Input-Gigawords) << 32) | (uint64) &Acct-Input-Octets}
```

That's a _lot_ simpler.  It means that we don't need the `|| 0` everywhere.  The code now goes "No Acct-Input-Gigawords?  OK, the default value is zero"

PR #5205, opened 2023-10-16
Avoid reference to uninitialized local used for sbuff (CID #1547825)

jejones3141:
Referring to fr_sbuff_start() rather than the local buffer used in a temporary sbuff avoids the issue of coverity not associating the local buffer with the FR_SBUFF_OUT() using it.


alandekok:
I've pushed a slightly different fix.  We can use `fr_sbuff_used()`, and avoid a `strlen()`

PR #5208, opened 2023-10-24
Fix build errors with Python 3.13

antoniotorresm:
According to [1], the `PyEval_InitThreads()` (deprecated 3.9), `PyEval_AcquireLock()` and `PyEval_ReleaseLock()` (deprecated 3.2). functions, are removed in Python 3.13. `Py_SetProgramName()` is deprecated since 3.11 as well.

`PyEval_InitThreads` can be deleted if running Python >= 3.10 as it simply does nothing from Python 3.9 [2].

Lock functions are replaced with `PyEval_RestoreThread` and `PyEval_SaveThread` as advised by Python docs [3], [4].

As for `Py_SetProgramName`, it should be replaced with PyConfig. The config is isolated to replicate behavior of `InitializeEx` [5].

[1]: https://docs.python.org/3.13/whatsnew/3.13.html
[2]: https://docs.python.org/3/c-api/init.html#c.PyEval_InitThreads
[3]: https://docs.python.org/3/c-api/init.html#c.PyEval_AcquireLock
[4]: https://docs.python.org/3/c-api/init.html#c.PyEval_ReleaseLock
[5]: https://docs.python.org/3/c-api/init_config.html#isolated-configuration

### DIFF
--- a/src/modules/rlm_python3/rlm_python3.c
+++ b/src/modules/rlm_python3/rlm_python3.c
@@ -679,12 +679,10 @@ finish:
 
 static void python_interpreter_free(PyThreadState *interp)
 {
-DIAG_OFF(deprecated-declarations)
-	PyEval_AcquireLock();
+	PyEval_RestoreThread(interp);
 	PyThreadState_Swap(interp);
 	Py_EndInterpreter(interp);
-	PyEval_ReleaseLock();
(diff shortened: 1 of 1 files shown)


alandekok:
I've done some formatting and cleanups.  Please double-check.

PR #5218, opened 2023-11-14
Avoid fr_nbo_to_uint16() to get EAP packet length in eap_identity() (CID #1243443)

jejones3141:
Moved to talloc_array_length() instead, which worked much better after I looked at the source and saw it's a macro that uses the type of the argument, so it requires casting to pointer to the type it was allocated with.


alandekok:
OK, this works becaue `eap_packet_to_vp()` checks that the length in the header == talloc_array_length(packet)

PR #5226, opened 2023-11-22
Deal with remaining missing locks (CIDs listed below)

jejones3141:
CIDs: 1551700, 1551701, 1551704, 1551705


arr2036:
Github is being weird, can you rebase

jejones3141:
Will do.

jejones3141:
Rebasing shows conflict; will manually merge.

jejones3141:
It was a couple of places where master had &foo->mutex and I had &(foo->mutex). I think I'd seen that elsewhere in FreeRADIUS source, which has some parentheses present only to clearly show the semantics, C being notorious for many precedence levels. I made it consistent with master and pushed.

PR #5234, opened 2023-11-29
Removed extra closing brace

47Cid:


alandekok:
Read the definition of `fr_heap_foreach` in `heap.h`.  Also note that if you do delete the extra brace, the build fails.

PR #5244, opened 2023-12-12
Start converting remaining unlang function calls to new syntax

jejones3141:


src/tests/modules/client/xlat.unlang
```diff
@@ -25,8 +25,8 @@ if (&Tmp-String-0 != 'b_type') {
 #
 #  Test non-existent client properties
 #
-&Tmp-String-3 := "%(client:non-existent-attr)"
-&Tmp-String-4 := "%(client:non-existing-attr2)"
+&Tmp-String-3 := "%client(non-existent-attr)"
```
alandekok, line 28: these should be quoted like the first parameter to the other `%client()` calls

src/tests/modules/exec/async.unlang
```diff
@@ -2,7 +2,7 @@
 #  Async calls should always return a zero length string
 #  because we don't wait for the response.
 #
-&Tmp-String-0 := "%(exec_async:/bin/sh -c 'echo -n hello')"
+&Tmp-String-0 := "%exec_async('/bin/sh -c \'echo -n hello\'')"
```
alandekok, line 5: The parameters should be different arguments, I think, like this other test:

```
src/tests/modules/linelog/linelog.unlang:&Tmp-Integer-0 := %exec('/bin/sh', '-c', "wc -l < $ENV{MODULE_TEST_DIR}/test_e.log")
```

src/tests/modules/client/xlat.unlang
jejones3141: Done.

src/tests/modules/exec/async.unlang
jejones3141: Done.

alandekok: ? patch still shows one argument.  it should be

```
&Tmp-String-0 := %exec_async('/bin/sh', '-c', "'echo -n hello'")
```

i.e. commas between arguments, and there's no need to wrap everything in double quotes.

jejones3141: Thanks. With that change, Tmp-String-0 gets assigned null, and the following check gives this debugging output (time/date stamp stripped):
`Debug : (0)      if !(&Tmp-String-0 == "")  {
Debug : (0)        | !
Debug : (0)            | ==
Debug : (0)                | &Tmp-String-0
Debug : (0)                  | &Tmp-String-0
Debug : (0)                  (null)
Debug : (0)            | %cmp_eq({}{})
Debug : (0)            | --> false
Debug : (0)        | %unary_not({false})
Debug : (0)        | --> true
`
and the test fails. `"&Tmp-String-0"` in the comparison made no difference.

alandekok: You can change the comparison to:

```
if &Tmp-String-0  {
```

which checks if it exists.  Or maybe

```
if &Tmp-String-0 != "" {
```

There are still unresolved issues with comparing empty boxes `%cmp_eq({}{})` as you see above.  I'll see if I can track those down.

PR #5249, opened 2023-12-20
Fix typos in documentation

DimitriPapadopoulos:
Misspellings found by [codespell](https://github.com/codespell-project/codespell).

### DIFF
--- a/doc/antora/modules/raddb/pages/mods-available/files.adoc
+++ b/doc/antora/modules/raddb/pages/mods-available/files.adoc
@@ -6,7 +6,7 @@
 
 The `users` file as located in `raddb/mods-config/files/authorize`. (Livingston-style format).
 
-See the xref:raddb:mods-config/files/users.adoc[users] file documention for information
+See the xref:raddb:mods-config/files/users.adoc[users] file documentation for information
 on the format of the input file, and how it operates.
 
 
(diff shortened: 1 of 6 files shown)


arr2036:
Please submit against the config files themselves not the auto-generated documentation see 'raddb'

DimitriPapadopoulos:
See #5252.

DimitriPapadopoulos:
However most typos originate in the `*.adoc` themselves files as far as I can see. Why close the whole pull request?

DimitriPapadopoulos:
I have read [`raddb/README.adoc`](https://github.com/FreeRADIUS/freeradius-server/blob/master/raddb/README.adoc) and it states:
> All of the documentation, READMEs, etc. for this directory are now stored in the doc/raddb directory, which is a soft link to the Asciidoc source

@arr2036 I have not modified `doc/raddb`: I changed files under the symlink target `doc/antora/modules/raddb/pages`  and more generally under `doc/antora/modules/reference`. I fail to find the same typos elsewhere, so I cannot find source files that would be used to generate the ones I have fixed. Could you please provide an additional hint?

mcnewton:
Merged manually, thanks.

@DimitriPapadopoulos I think it was actually only one file that was generated - if you edit files in `raddb` and then run `make asciidoc` it will recreate the Antora raddb documentation automatically.

alandekok:
The files in `doc/antora/modules/raddb/pages` are auto-generated from the files in `raddb`.  Please submit changes against the originals, not against the auto-generated versions.

I've pushed a minor fix to the build in commit b935f9a8ae28

You can edit the files such as `raddb/mods-enabled/ldap`, and then do:

```
$ make doc.raddb
```

and it will be automatically converted.

PR #5250, opened 2023-12-21
fix the  Null pointer dereference: prefix

zengwei2000:
fix the  Null pointer dereference: prefix

Signed-off-by: zengwei zengwei1@uniontech.com

### DIFF
--- a/src/lib/util/strerror.c
+++ b/src/lib/util/strerror.c
@@ -754,7 +754,12 @@ void fr_perror(char const *fmt, ...)
 		talloc_free(prefix);
 	} else {
 		if (!error) return;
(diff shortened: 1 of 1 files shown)


arr2036 (commented):
Prefix will always be NULL in this branch so the patch is pointless.

zengwei2000:
> Prefix will always be NULL in this branch so the patch is pointless.

So the fprintf content here should be error instead of prefix？

zengwei2000:
> Prefix will always be NULL in this branch so the patch is pointless.

https://github.com/FreeRADIUS/freeradius-server/pull/5251  I have resubmitted the PR, please review whether it should be modified like this

arr2036:
Yes

PR #5286, opened 2024-01-29
Fix issues with RADIUS over TCP

Fredi-raspall:
We observed that under load (when multiple RADIUS messages were received on the same TCP segment  and locally delivered on same read()) the server becomes stuck, unable to process more messages from then on.

Debugging showed that:
*  the file descriptor associated to the client TCP connection was blocking; instead of the presumed non-blocking mode. 
* Setting the socket to be non-blocking also showed that, when there's data "leftover" (i.e. read from a prior read, but smaller than a RADIUS message), it does not end up being processed / stored.

So:
* made the TCP connection non-blocking
* ensured that, even if no new data is read from the socket, any left-over is processed.

### DIFF
--- a/src/lib/io/master.c
+++ b/src/lib/io/master.c
@@ -1364,6 +1364,19 @@ redo:
 			return 0;
 		}
 
+		/*
+		 * Set the descriptor for the new child connection as
+		 * non-blocking.
+		 */
+		if (fr_nonblock(accept_fd) < -1) {
+			DEBUG("proto_%s - failed to set new TCP descriptor %d as non-blocking. Closing...",
+					inst->app->common.name, accept_fd);
+			close(accept_fd);
+			return 0;
+		}
+
+
+
 #ifdef STATIC_ANALYZER
 		saremote.ss_family = AF_INET; /* static analyzer doesn't know that accept() initializes this */
 #endif
--- a/src/listen/radius/proto_radius_tcp.c
+++ b/src/listen/radius/proto_radius_tcp.c
@@ -113,8 +113,16 @@ static ssize_t mod_read(fr_listen_t *li, UNUSED void **packet_ctx, fr_time_t *re
 	 */
 	data_size = read(thread->sockfd, buffer + *leftover, buffer_len - *leftover);
 	if (data_size < 0) {
-		PDEBUG2("proto_radius_tcp got read error %zd", data_size);
-		return data_size;
+		if (errno != EAGAIN && errno != EWOULDBLOCK) {
+		    fr_strerror_printf("proto_radius_tcp got read error: %s", fr_syserror(errno));
+		    return -1;
+		}
+		/*
+		 * we're done if had no leftovers. Otherwise, continue to ensure that any data left
+		 * over gets processed.
+		 */
+		if (*leftover == 0)
+		    return 0;
 	}
 
 	/*
@@ -140,7 +148,10 @@ static ssize_t mod_read(fr_listen_t *li, UNUSED void **packet_ctx, fr_time_t *re
 		return -1;
 	}
 
-	in_buffer = data_size + *leftover;
+	/* data_size could be -1 if we did not manage to read */
+	in_buffer = *leftover;
+	if (data_size > 0)
+	    in_buffer += data_size;
 
 	/*
 	 *	Not enough for one packet.  Tell the caller that we need to read more.


alandekok:
Thanks.  The `read()` definitely has to handle `EWOULDBLOCK`.

I've pulled it in with some changes:

* we don't check the return of `fr_nonblock()` in other places, so it's fine to do that here, too.  Pretty much all moderns OS's support the NONBLOCK call.  Realistically, if `fr_nonblock()` fails, then the file descriptor is bad.
* I've changed the code after `read()` to check for `EWOULDBLOCK` a bit differently.  On some systems `EAGAIN` and `EWOULDBLOCK` have the same value.  The `ifdef` and `switch` method is what we use elswhere in the code.  So for consistency, I've made it the same here
* if we do get `EWOULDBLOCK`, then there's no point in doing anything but returning `0` bytes read.  Whatever data is already in the buffer is left over from a previous call to `read()`.  So it was already checked for being a partial packet, and we don't need to do that again.
* I've added the same code to the TACACS+ socket handler.  It needs the same fix.

We're in the process of moving to a new API for sockets, in `src/lib/bio/fd.c`.  That code already handles `EWOULDBLOCK` correctly.  It's also abstracting a ton of extra work around TLS, etc.  The hope is that we can replace the _very_ complex code in `src/lib/io/master.c` with much cleaner / simpler code in `src/lib/bio`.

Please check the patch and verify that it fixes the issue.

Fredi-raspall:
> Thanks. The `read()` definitely has to handle `EWOULDBLOCK`.
> 
> I've pulled it in with some changes:
> 
> * we don't check the return of `fr_nonblock()` in other places, so it's fine to do that here, too.  Pretty much all moderns OS's support the NONBLOCK call.  Realistically, if `fr_nonblock()` fails, then the file descriptor is bad.
> * I've changed the code after `read()` to check for `EWOULDBLOCK` a bit differently.  On some systems `EAGAIN` and `EWOULDBLOCK` have the same value.  The `ifdef` and `switch` method is what we use elswhere in the code.  So for consistency, I've made it the same here

Ok. I think that makes the code less readable, but that's not my call.

> * if we do get `EWOULDBLOCK`, then there's no point in doing anything but returning `0` bytes read.  Whatever data is already in the buffer is left over from a previous call to `read()`.  So it was already checked for being a partial packet, and we don't need to do that again.

I believe you need to keep that processing since that is what fixes the problem. 
If I am not mistaken, the code can request to read default_message_size=4096 worth of data. It may therefore receive multiple messages in the same read. It may read a complete message and a fraction of another message; and that's fine because when the missing chunk will be received, the code will resume with that leftover. However, it may also happen that you read N complete messages and nothing else later on. Since the code processes one message at a time, there will be N-1 messages pending to be processed. If no other read is made later on (or if other ones are attempted but there's no more data available) those N-1 messages will be in the buffer but never processed.

> * I've added the same code to the TACACS+ socket handler.  It needs the same fix.

Great. There's something I don't understand, though. In the tcp one, I added logging of fr_syserror(errno) to report what may cause read errors. That's been removed, but in the TACACS code that's logged. Any reason not to have that logged in case of TCP?

> We're in the process of moving to a new API for sockets, in `src/lib/bio/fd.c`. That code already handles `EWOULDBLOCK` correctly. It's also abstracting a ton of extra work around TLS, etc. The hope is that we can replace the _very_ complex code in `src/lib/io/master.c` with much cleaner / simpler code in `src/lib/bio`.
> 
> Please check the patch and verify that it fixes the issue.

alandekok:
On Jan 30, 2024, at 5:42 AM, Fredi Raspall ***@***.***> wrote:
> Ok. I think that makes the code less readable, but that's not my call.

  It's important for the code to look then same everywhere.  This includes minor things like comment formatting and spaces between code elements.

  If we have consistent formatting, then it's easier to see the code as "patterns".  i.e. "this block of text does the same thing as the other 5 times we've seen the text".

  The alternative is random formatting, random indentation, etc.  Adding randomly formatted code very quickly turns a clean code base into an un-maintainable mess.

> I believe you need to keep that processing since that is what fixes the problem.
> If I am not mistaken, the code can request to read default_message_size=4096 worth of data. It may therefore receive multiple messages in the same read.

  Ah, OK.  I've pushed a fix.

  The real solution there is to skip the read() if we have a full packet in the buffer.  We only re-do the read() if there's insufficient data for a full packet.

> Great. There's something I don't understand, though. In the tcp one, I added logging of fr_syserror(errno) to report what may cause read errors. That's been removed, but in the TACACS code that's logged. Any reason not to have that logged in case of TCP?

  It was just missed.

PR #5289, opened 2024-02-02
Deal with another dereference after null check (CID #1530398)

jejones3141:
Once again an earlier test for NULL makes Coverity suspicious of later unchecked uses.


jejones3141:
After correcting the conflict, there was no change to  make. I suspect that other changes have corrected the issue.

PR #5297, opened 2024-02-21
Add some debugging output for test_connection_levels_max().

jejones3141:
This is solely for testing for now at least. Running locally doesn't trip up the issue, so this will let it run where it breaks.


arr2036:
Now conflicts?

PR #5299, opened 2024-03-01
Fix eap tls preload certificate chains for realms

mihmantis:
Fix expected incorrect key name when pem file name contains FQDN
Fix trying to load key when it is absent and not loading key when it's present


alandekok:
Fixed, thanks.

PR #5312, opened 2024-04-06
RADIUS: Updates to dictionary.huawei

linalan:
Several attributes had incorrect field types. Also added a couple missing attributes. 

Reference: https://support.huawei.com/enterprise/en/doc/EDOC1100279002/ecd24420/appendix-radius-attributes#EN-US_CONCEPT_0000202324910809

### DIFF
--- a/share/dictionary/radius/dictionary.huawei
+++ b/share/dictionary/radius/dictionary.huawei
@@ -152,9 +152,9 @@ ATTRIBUTE	Framed-Pool-Group			157	string
 ATTRIBUTE	Framed-IPv6-Address			158	ipv6addr
 ATTRIBUTE	Acct-Update-Address			159	integer
 ATTRIBUTE	NAT-Policy-Name				160	string
-ATTRIBUTE	NAT-Public-Address			161	string
-ATTRIBUTE	NAT-Start-Port				162	string
-ATTRIBUTE	NAT-End-Port				163	string
+ATTRIBUTE	NAT-Public-Address			161	ipaddr
+ATTRIBUTE	NAT-Start-Port				162	integer
+ATTRIBUTE	NAT-End-Port				163	integer
 ATTRIBUTE	NAT-Port-Forwarding			164	string
 ATTRIBUTE	NAT-Port-Range-Update			165	integer
 ATTRIBUTE	DS-Lite-Tunnel-Name			166	string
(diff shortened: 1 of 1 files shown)


share/dictionary/radius/dictionary.huawei
```diff
@@ -189,12 +189,12 @@ VALUE	Port-Mirror			Enable			3
 
 ATTRIBUTE	Account-Info				184	string
 ATTRIBUTE	Service-Info				185	string
-ATTRIBUTE	DHCP-Option				187	octets
+ATTRIBUTE	DHCP-Option				187	string
```
alandekok, line 192: This should be left at `octets`, or even `group ref=..DHCPv4` as per the RFC 9445 dictionary, and the `DHCP-Options` attribute.

share/dictionary/radius/dictionary.huawei
```diff
@@ -189,12 +189,12 @@ VALUE	Port-Mirror			Enable			3
...
 ATTRIBUTE	AVpair					188	string
 ATTRIBUTE	Delegated-IPv6-Prefix-Pool		191	string
-ATTRIBUTE	IPv6-Prefix-Lease			192	octets
-ATTRIBUTE	IPv6-Address-Lease			193	octets
-ATTRIBUTE	IPv6-Policy-Route			194	ipv6prefix # manual says string?
+ATTRIBUTE	IPv6-Prefix-Lease			192	string
```
alandekok, line 195: `string` is _printable text_.  From looking at the link, this attribute is actually a structure with multiple fields.  So this should either be left as `octets`, or it should be defined as a `struct`, and given child  `MEMBER` fields.

alandekok:
I will review the dictionary against the link to Huawei support, and will likely pull in the changes manually.  As noted in the comments above, a number of the attributes should have different data types.

share/dictionary/radius/dictionary.huawei
```diff
@@ -228,6 +228,8 @@ ATTRIBUTE	MS-Maximum-MAC-Study-Number		222	octets # ether??
 ATTRIBUTE	GGSN-Vendor				232	string
 ATTRIBUTE	GGSN-Version				233	string
 ATTRIBUTE	Ext-Specific				238	string
+ATTRIBUTE	Huawei-USR-GRP-NAME			251	string
```
alandekok, line 231: In the `master` branch there's no need to add the vendor name

linalan:
Thanks for taking the time to review and correct my changes.I was initially fixing attributes 161-163 when I noticed a few others than needed fixing. For the others, I should have checked a packet capture and decode like I did w/161-163. Apologies for not being more thorough.

PR #5313, opened 2024-04-15
add define for Solaris.

ipv6labs:
Solaris, like illumos and FreeBSD does not define SOL_TCP


ipv6labs:
Thank you for fixed the issue.

PR #5318, opened 2024-05-06
Reorder fr_command_debug() parameters

jejones3141:
This makes it consistent with fr_{atomic_queue, message_set, queue, ring_buffer, time_tracking, worker}_debug(), to avoid needless special cases when invoking it from lldb.


jejones3141:
This is a trivial difference to the lldb script; no need to bother.

PR #5320, opened 2024-05-16
hook up Challenge for Post-Auth-Type for inner tunnel processing

jimdigriz:
For inner-tunnel processing it is not possible to use `Post-Auth-Type Challenge`, for example with `rlm_detail`, as it is completely skipped.

This PR hooks up the doll to allow this.

Works For Me(tm)


alandekok:
I think this looks good.  Since there's no existing `Post-Auth Challenge` section, this new functionality shouldn't affect existing systems.

mcnewton:
LGTM; pulled in manually, thanks.

PR #5325, opened 2024-05-28
Add missing quotes

tvdijen:


alandekok:
This works without the quotes in CI and in my testing.  What problem is solved by adding quotes?

tvdijen:
I got the following error on CentOS 9:

> /etc/raddb/mods-config/sql/main/mysql/queries.conf[52]: Syntax error: Expected comma after '%{((integer)': &Event-Timestamp) || %l}
> Errors reading or parsing /etc/raddb/radiusd.conf
> 

I should have mentioned that in the first post, sorry :)

PR #5335, opened 2024-06-06
Support async SQL queries and update mysql driver with async calls

ndptech:


src/modules/rlm_sql/drivers/rlm_sql_db2/rlm_sql_db2.c
```diff
@@ -106,50 +106,38 @@ static sql_rcode_t sql_socket_init(rlm_sql_handle_t *handle, rlm_sql_config_t co
...
-
-	conn = handle->conn;
-	SQLNumResultCols(conn->stmt, &c);
-	return c;
+	query_ctx->rcode = RLM_SQL_OK;
+	RETURN_MODULE_OK;
 }
 
-static sql_rcode_t sql_fields(char const **out[], rlm_sql_handle_t *handle, UNUSED rlm_sql_config_t const *config)
+static sql_rcode_t sql_fields(char const **out[], fr_sql_query_t *query_ctx, UNUSED rlm_sql_config_t const *config)
```
arr2036, line 138: Should really be `rlm_sql_rcode_t` and `rlm_sql_query_t`, or `sql_rcode_t` and `sql_query_t` the `fr_` prefix is meant to be reserved for the utility and protocol libraries to allow reuse outside of freeradius.

PR #5337, opened 2024-06-09
amend policy insert_acct_class/acct_unique to work with multiple Class attributes

jimdigriz:
Some [deployments](https://lists.freeradius.org/pipermail/freeradius-users/2017-March/087021.html) may wish to use multiple `Class` attributes and continue to use FR's existing provided accounting policies.

We also take the opportunity to remove needless hex encoding on the Class attribute due to the string to octet conversion originally in place.


Gman42412:
Never subscribed to your company

alandekok:
My one question is does this change any existing behavior?  If so, it would have to be added as a separate policy.

jimdigriz:
Only problem I can foresee is when someone updates to this alternative policy and they have existing old hex encoded `Class` attributes out there in the wild.

So maybe the 'encoder' is good, but for the 'decoder' we need to check if the preamble is either `hex("ai:")` or if it is `hex(hex("ai:"))` to determine what to do with it.

alandekok:
Thanks, I'd be happier with being compatible with existing systems.  I have a large fear of breaking peoples networks.

jimdigriz:
Sorry it took a while, but I think the functionality to maintain backwards compatibility now works.

jimdigriz:
> won't let me merge without signed commits. Could you sign these with your GPG key @jimdigriz. Thanks.

Sorry for the tardiness, cooking a signing key means I have to do some physical gymnastics with USB keys and it has been a busy week.

Should be okay now, thanks.

PR #5344, opened 2024-06-20
radclient: fix debug to not show source port in hex but as integer

jimdigriz:
Noticed when running `radtest` I was seeing hex used to represent the source port.

No idea if this was intention (appeared during some RADIUSv1.1 work), as the destination port is shown as an integer, but this commit makes the source port an integer.


jimdigriz:
should this be closed since you added 59f5cf14e7ef97c8fc52d51c881bb13f5b2600b8?

PR #5359, opened 2024-07-03
Annotate false positive overflow_sink (CID #1604608)

jejones3141:
Coverity docs say it has a model of `write()`. Ideally it should be able to infer that given the initial value of `done` and the loop invariant, `len - done` will never be negative and thus will be representable as a `size_t`.


jejones3141:
Closing to replace with Coverity-only check.

PR #5360, opened 2024-07-08
Annotate false positive overflow_const (CID #1604609)

jejones3141:
Despite the `while` loop condition that passes iff `pos` is nonzero, Coverity thinks that inside the loop `pos` might equal zero so that the decrement would make it 255, making `(1 << pos)` have undefined behavior.


jejones3141:
Closing to switch to a solution without annotation.

PR #5361, opened 2024-07-09
Actually fill fr_rand_pool.randrsl

jejones3141:
In `fr_rand_init()`, where `/dev/urandom` is available, the code as written would always read into the beginning of `fr_rand_pool.randrsl`, so that only the first max(read return value) bytes would be filled.

(Not associated with CID #1604611, but found while looking into it.)

### DIFF
--- a/src/lib/util/rand.c
+++ b/src/lib/util/rand.c
@@ -44,10 +44,11 @@ void fr_rand_init(void)
 	if (fd >= 0) {
 		size_t total;
 		ssize_t this;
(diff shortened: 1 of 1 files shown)


jejones3141:
Don't know; where can I find the v3 code?

jejones3141:
No need to be inconsistent with "fill the random pool" (https://github.com/FreeRADIUS/freeradius-server/commit/d5276a878367d2f1e13183a4834a984ea7220f24); will resubmit.

PR #5362, opened 2024-07-09
Don't crash when processing stats for requests lacking packet data

terryburton:


alandekok:
Looks good, thanks.  Probably should also be cherry-picked to v3.0.x

PR #5370, opened 2024-07-12
Correct filling of random pool

jejones3141:
Catches an edge case of not filling `fr_rand_pool.randrsl` in a single `read()`. This version is simpler than our first attempt and consistent with #d5276a8, and I suspect it will also take care of CID #1604611.

### DIFF
--- a/src/lib/util/rand.c
+++ b/src/lib/util/rand.c
@@ -34,6 +34,8 @@ static _Thread_local bool fr_rand_initialized = false;
 void fr_rand_init(void)
 {
 	int fd;
+	uint8_t *p = (uint8_t *) &fr_rand_pool.randrsl[0];
+	uint8_t *end = p + sizeof(fr_rand_pool.randrsl);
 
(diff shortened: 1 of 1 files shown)


src/lib/util/rand.c
```diff
@@ -34,6 +34,8 @@ static _Thread_local bool fr_rand_initialized = false;
 void fr_rand_init(void)
 {
 	int fd;
+	uint8_t *p = (uint8_t *) &fr_rand_pool.randrsl[0];
+	uint8_t *end = p + sizeof(&fr_rand_pool.randrsl);
```
arr2036, line 38: Shouldn't this be `sizeof(fr_rand_pool.randrsl)`?

jejones3141: Will double check; thanks.

jejones3141: Indeed it should; correcting. Thanks.

PR #5373, opened 2024-07-16
Annotate overflow_const in  fr_isaac_init() (CID #1604614)

jejones3141:
overflow_const strikes again in the ISAAC random number generator. The shift is intentional, and it's an unsigned type so the behavior is defined.


jejones3141:
No choice but annotation here. Coverity is right; the first time through `mix()` it knows the values, and `h <= 9` will have bits falling off the most significant end--but that's defined behavior, and that's standard behavior for hash functions.

PR #5375, opened 2024-07-17
Replace radsecret with a Bash variant

panlinux:
This avoids adding two extra perl dependencies just because of this script, and trigger a component mismatch in Ubuntu (where a package in main [freeradius] cannot depend on packages from universe [these two new perl modules]).

Use 13 bytes instead of 12, due to
https://bugs.launchpad.net/ubuntu/+source/freeradius/+bug/2073269/comments/6

Fixes Ubuntu bug [LP: #2073269](https://bugs.launchpad.net/ubuntu/+source/freeradius/+bug/2073269/comments/6)
Fixes Debian bug [#1076458](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1076458)

### DIFF
--- a/src/main/radsecret
+++ b/src/main/radsecret
@@ -1,7 +1,3 @@
-#!/usr/bin/env perl
-#
-#  A tool which generates strong shared secrets.
-#
-use Convert::Base32;
-use Crypt::URandom();
-print join('-', unpack("(A4)*", lc encode_base32(Crypt::URandom::urandom(12)))), "\n";
+#!/bin/bash
+data=$(dd if=/dev/urandom bs=1 count=13 2>/dev/null| base32 | tr 'A-Z' 'a-z')
+echo ${data:0:4}-${data:4:4}-${data:8:4}-${data:12:4}-${data:16:4}


panlinux:
I'm trying to fix the "unverified" commit status.

panlinux:
I'm using gpg subkeys, and am having trouble getting GH to recognize the signature on the commit, which was made using a subkey. I'm a bit stuck.

alandekok:
I'll take a look, and perhaps merge it manually.

My man concern is the `/bin/bash` dependency.  it works on Debian, but may be less useful on other platforms.

panlinux:
fixed the commit signature. I had to remove and re-upload the public gpg key, otherwise GH wouldn't pick up the subkey that I added since (see https://www.hjdskes.nl/blog/github-pgp-response/)

panlinux:
`/bin/sh` doesn't have that substring notation indeed:
```
$ sh
$ data=376yfuycd5a76sjzuhwa
$ echo $data
376yfuycd5a76sjzuhwa
$ echo ${data:0:4}-${data:4:4}-${data:8:4}-${data:12:4}-${data:16:4}
dash: 3: Bad substitution
```
I also tried some `printf` "%.4s" shenanigans with `/bin/sh`, but didn't get far

mcnewton:
Agreed, relying on `bash` isn't a great idea. Or `base32` for that matter either.

    #! /bin/sh
    
    DATA=$(dd if=/dev/urandom bs=1 count=13 2>/dev/null| base32 | tr 'A-Z' 'a-z')
    echo "$DATA" | awk '{print substr($0,0,4) "-" substr($0,5,4) "-" substr($0,9,4) "-" substr($0,13,4) "-" substr($0,17,4)}'

Then add something to behave correctly if `base32` is not installed (fallback to `base64`, then possibly `od` for hex as better than nothing at all). If `awk` isn't there (what system is this?) just print `$DATA`, the hyphens are only to make it easier to read anyway.

panlinux:
In linux systems (citation needed), `tr`, `base32`, `od`, and `dd` come from `coreutils`, which at least on debian and ubuntu is guaranteed to be installed (priority: required). `awk` is not, and there are multiple implementations too: `mawk`, `gawk`.

How about using `openssl` from the command line? Is that more common than the alternatives? Something like:
```
$ openssl rand -base64 12
l4Ni7usKpXzrIrU5
```
We lose the dashes. which ironically seems to be the most difficult part for this script :)

Or just drop the dashes indeed, this output will probably be copied and pasted anyway.

mcnewton:
@panlinux We're not just talking Linux systems. `base32` isn't on a FreeBSD system I'm looking at here (and neither is `base64`). I would be very surprised if `awk` wasn't available on pretty much any system the admin runs the script on (or we could just add a package dependency for that), and `substr` is POSIX so will be in all variants. But TBH `cut` would work just as well; it would just need multiple invocations.

panlinux:
> @panlinux We're not just talking Linux systems. `base32` isn't on a FreeBSD system

Ok, that's fair. How about openssl, or how do you declare dependencies on freebsd?

alandekok:
I've pushed changes which try a bunch of different things, and then complains if none of them are found.

This tool is just a helper script, and isn't run that often.  So performance doesn't matter.

panlinux:
I see you also switched to `/dev/random`, which can block (but is a better source of randomness). Was that on purpose?

alandekok:
typo. :(  urandom is generally better.

panlinux:
`echo "Failed finding 'openssl', 'base64', or 'base32' tools".`

You could include `hexdump` there now:
`echo "Failed finding 'openssl', 'base64', 'base32', or 'hexdump' tools".`

panlinux:
I believe this is sorted now, closing the PR.

PR #5376, opened 2024-07-17
Annotate false positive overflow_const (CID #1604617)

jejones3141:
Given what `ret` is assigned, its value is necessarily between 1 and 8 inclusive, rather than 2305843009213693952 which Coverity thinks it is, so that `sizeof(uint64_t) - ret` will not underflow as Coverity thinks it will.


jejones3141:
Rewriting as conditional Coverity-only check rather than annotation,.

PR #5342, opened 2024-06-18
cbtls_info() to report "Need to write more data" when called on exit …

martinsta:
…of a handshake function

The current code does seem to miss to cater for the cases when a TLS handshake exits and there is data left that needs either to be extracted from the BIO and/or sent to the other side. Differentiating and logging such cases within cbtls_info() might aid in further debugging any stalled handshakes followed up by timing out requests we experience with the up to date 3.2.x head.

PR #5377, opened 2024-07-19
fr_socket_client_tcp() needs literal nonblock value to use it as an a…

martinsta:
…sync parameter

fd is currently missing O_NONBLOCK property that should be set via nested fr_socket_client_tcp() ->  fr_nonblock() -> fcntl() calls.

this will fix the issue #5342 

https://github.com/FreeRADIUS/freeradius-server/pull/5342


alandekok:
Thanks,  We've pushed that along with some more fixes in f44086315776 and fb8310dc9ea6

PR #5396, opened 2024-08-08
Revise `write_all()` parameters to avoid overflow (CID #1604608)

jejones3141:
`write_all()` parameters changed to resemble `write()` in return type and type for the number of bytes to be written, to try to avoid an over or underflow Coverity claims occurs.


alandekok:
while you're in the code, there's a lot of

```
if (write_all(a,b,size) != size) {
   fail
}
```

it would be simpler and shorter if `write_all` returned `0` for success, and `-1` for error.  The code can then do:

```
if (write_all(a,b,size) < 0) {
   ...
}
```

which seems clearer.

jejones3141:
Good point; will do.

PR #5408, opened 2024-08-23
Attempt Coverity-only overflow check in xlat_print() (CID #1604615)

jejones3141:
The error return subtracts a `size_t` from an `ssize_t`, which C does in `size_t`, and then casts the result to `ssize_t` for return, which in theory raises the possibility of overflow. We add a Coverity-only check that tries to catch it. If Coverity doesn't recognize it, we'll have to fall back to annotation.


alandekok:
I'm wary of adding this kind of stuff.  It pollutes the source code quite a bit, and is a corner case which will _never_ happen.

I think an annotation would be better.

jejones3141:
Closing to resubmit with just the annotation after backing out the Coverity-only check.

PR #5422, opened 2024-09-10
Update sql

RigzTech:


alandekok:
Why would we merge this into the main distribution of FreeRADIUS?  This is a change for your local system, and belongs in revision control for your local system.

PR #5428, opened 2024-09-10
Huawei dictionary update

gk-fschubert:
The following pull request defines attributes for Huawei devices which are visible under https://support.huawei.com/enterprise/en/doc/EDOC1000178178/35071f9a/radius-aaa

The definition from https://github.com/FreeRADIUS/freeradius-server/blob/master/share/dictionary/radius/dictionary.huawei
(https://support.huawei.com/enterprise/en/doc/EDOC1100279002/ecd24420/appendix-radius-attributes#EN-US_CONCEPT_0172358597) 
is not reachable anymore. 

Not every attribute which was present in https://github.com/FreeRADIUS/freeradius-server/blob/master/share/dictionary/radius/dictionary.huawei was also available in the docs of https://support.huawei.com/enterprise/en/doc/EDOC1000178178/35071f9a/radius-aaa. We changed only attributes which are documented. All other have been untouched.


alandekok:
We need this PR for 3.2, also.

the new Huawei dictionary uses different names / data types for previously defined attributes.  So we must add this one as a 'conflicting' dictionary in a separate file.

once we support multiple dictionaries for a NAS, we can include it in the main list.  until then, people will have to manually edit the `$INCLUDE` lines to get the "correct" dictionary for their environment.

arr2036:
Different names are fine.  From the sampling I looked through I saw mostly consistent types.  So for v4 we can take this as an update.  For v3.2 it should be a separate dictionary.

arr2036:
People can always define aliases in their site-specific dictionary if they want the old names back, right?

alandekok:
People can define aliases for names.  From discussion on the mailing list, the issue was that the types were inconsistent.

gk-fschubert:
Is something required on my part or was there simply no time to accept the PR?

alandekok:
As we noted in earlier comments, the dictionary names changed in v4.  There is no need for a `Huawei-` prefix.

I've pulled the dictionary into the v3.2.x branch.  We can look at moving it to v4 later.

PR #5435, opened 2024-09-16
Avoid Coverity overflow in write_all() (CID #1604608)

jejones3141:
Revises `write_all()` loop analogously to `fr_rand_init()` loop (CID #1604611)


alandekok:
fixed in commit 1aec8ee

PR #5448, opened 2024-10-14
Update listen.c for illumos and Solaris in v3.0.x

ipv6labs:
add define for illumos and Solaris.

The same as Fixes #5135 #5313 in v3.2.x


alandekok:
Added a patch.

PR #5449, opened 2024-10-18
V3.2.x Add handling for Subject Alternative Name (SAN) URIs

ablanken:
Add handling for Subject Alternative Name (SAN) URIs.

These URIs can be leveraged by network access control systems to make authentication and authorization decisions based on external device management services, such as Intune, where device certs may include the "IntuneDeviceId://guid" SAN URI.

Once this is included, we can look at calling the Intune Graph API to query device compliance status & group membership etc.


ablanken:
I have assumed that the cert attributes in the dictionary can't be reordered. Please let me know if this is not the case or if the attribute numbers need to be changed.

ATTRIBUTE	TLS-Cert-Subject-Alt-Name-Uri	1934	string
ATTRIBUTE	TLS-Client-Cert-Subject-Alt-Name-Uri	1935	string


One thing to be mindful of is 'GEN_URI' already appears in the following method in tls.c: static int ocsp_parse_cert_url
This may or may not be an issue.

Neustradamus:
Replaced by:
- https://github.com/FreeRADIUS/freeradius-server/pull/5450

PR #5455, opened 2024-10-30
Update the base Alpine Linux container image

fpoirotte:
End of support was reached for Alpine Linux 3.13 on November 2022.
Use the latest stable version instead (supported until April 2026).


mcnewton:
merged manually - build needed fixing due to alpine > 3.13 having newer incompatible couchbase libraries

PR #5457, opened 2024-11-04
Re-enable EAP-TTLS

ndptech:


src/lib/eap/base.c
```diff
@@ -386,80 +388,81 @@ void eap_add_reply(request_t *request, fr_dict_attr_t const *da, uint8_t const *
...
-	vp = fr_pair_find_by_da(&request->control_pairs, NULL, attr_virtual_server);
-	server_cs = vp ? virtual_server_find(vp->vp_strvalue) : virtual_server_find(virtual_server);
+	fr_assert(request->parent);
 
-	if (server_cs) {
-		RDEBUG2("Running request through virtual server \"%s\"", cf_section_name2(unlang_call_current(request)));
-	} else {
-		RDEBUG2("Running request in virtual server");
-	}
+	vs = virtual_server_find(virtual_server);
```
arr2036, line 435: Might be something to fix.  I think in a lot of places we resolve virtual servers on startup now.  There's actually a struct that represents a virtual server, whereas there didn't use to be.

src/lib/eap/base.c
```diff
@@ -386,80 +388,81 @@ void eap_add_reply(request_t *request, fr_dict_attr_t const *da, uint8_t const *
...
+	if (eap_session->child_state) {
+		fr_pair_append_by_da(request->request_ctx, &vp, &request->request_pairs, attr_state);
+		fr_value_box_memdup(vp, &vp->data, NULL, eap_session->child_state,
+				    talloc_array_length(eap_session->child_state), false);
+		RDEBUG4("Restoring child state %pV", &vp->data);
 	}
 
-	return rcode;
-#endif
+	vp = fr_pair_find_by_da(&request->parent->request_pairs, NULL, attr_packet_type);
```
alandekok, line 456: Nit: this should always be Access-Request, right?  TTLS is pretty hard-coded to RADIUS.  :(

src/modules/rlm_eap/types/rlm_eap_ttls/ttls.c
```diff
@@ -629,32 +657,26 @@ fr_radius_packet_code_t eap_ttls_process(request_t *request, eap_session_t *eap_
...
-		code = FR_RADIUS_CODE_ACCESS_REJECT;
-		goto finish;
+		return UNLANG_ACTION_FAIL;
 	}
 
-	if (!diameter_verify(request, data, data_len)) {
-		code = FR_RADIUS_CODE_ACCESS_REJECT;
-		goto finish;
-	}
+	if (!diameter_verify(request, data, data_len)) return UNLANG_ACTION_FAIL;
```
alandekok, line 671: Nit: do we eventually want to add a `src/protocols/diameter`, and move the verify / encode / decode there?

Not for this  commit, but something to think about.

arr2036: Yeah I was thinking exactly the same.  Probably not too much work to shift the code over and fixup up the function signature.

arr2036: ...then we could stick someone on starting to write tests, and making the code more generic.

PR #5459, opened 2024-11-07
support tcp packets in radsniff

derde:
Radius over TCP generally appears in complete packets with the PSH flag set, and generally one radius packet per TCP PSH packet.  This code is to enable the reading of TCP packets for this one common case.  Tested for IP4.

The handling of `#include <netinet/tcp.h>` might not be correct - sorry.

Testing with pcap file:

    DIR=. ; PROGRAM=radsniff ; $DIR/build/make/jlibtool --mode=execute $FR_DEBUGGER $DIR/build/bin/local/$PROGRAM -d $DIR/raddb -D $DIR/share/dictionary -x -f  'tcp and port 1812 or 1813' -I tcp.pcap

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1020,6 +1020,7 @@ AC_CHECK_HEADERS( \
   net/if_dl.h \
   netdb.h \
   netinet/in.h \
+  netinet/tcp.h \
   netpacket/packet.h \
   prot.h \
   pwd.h \
--- a/src/bin/radsniff.c
+++ b/src/bin/radsniff.c
@@ -1275,6 +1275,7 @@ static void rs_packet_process(uint64_t count, rs_event_t *event, struct pcap_pkt
 	ip_header_t const	*ip = NULL;		/* The IP header */
 	ip_header6_t const	*ip6 = NULL;		/* The IPv6 header */
 	udp_header_t const	*udp;			/* The UDP header */
+	struct tcphdr const     *tcp=NULL;		/* The TCP header */
 	uint8_t			version;		/* IP header version */
 	bool			response;		/* Was it a response code */
 
@@ -1316,12 +1317,17 @@ static void rs_packet_process(uint64_t count, rs_event_t *event, struct pcap_pkt
 		ip = (ip_header_t const *)p;
 		len = (0x0f & ip->ip_vhl) * 4;	/* ip_hl specifies length in 32bit words */
 		p += len;
+		if (ip->ip_p == IPPROTO_TCP) {
+			tcp=(const struct tcphdr*)p;
+		}
 		break;
 
 	case 6:
 		ip6 = (ip_header6_t const *)p;
 		p += sizeof(ip_header6_t);
-
+		if (ip6->ip_next == IPPROTO_TCP) {
+			tcp=(const struct tcphdr*)p;
+		}
 		break;
 
 	default:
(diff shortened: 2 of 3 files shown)


alandekok:
I'm not sure this will work.  It assumes that all of the RADIUS packet is in one TCP packet.  That might happen, but it is not guaranteed to happen.

The correct way to do this is to reassemble the TCP data, and then decode that.  But that involves re-implementing much of TCP, with windows, offsets, etc.

derde:
Yes, the correct way would be to reassemble the TCP stream, and have `radsniff` be able to act on a stream - but in practice that's an overkill, and the complete TCP stream is sometimes not available anyhow.  Strictly `radsniff` should do IP fragment reassembly too, but it's not necessary in the normal case.  For the common case of TCP, which is no Nangle algorithm, no oversized packets, the assumption of 1 radius packet at the beginning of the TCP packet holds well enough to be useful.  

I wrote the patch to diagnose a problem yesterday, and it showed that FreeRADIUS was doing the right thing, and the problem was downstream.  Having radius in TCP is great, but not being able to look into it is problematic.  There really isn't any other tool that allows inspection of the data, even poorly.  `tcpdump` certainly has no idea.  

The code could be trivially improved by some reporting of the edge cases, but I only have a pcap file with data that is easy to work with.  It really does work ... here's a (censored) sample

    2024-11-07 08:56:21.825164 (181) Access-Accept Id 9 tcp.pcap:9.22.15.40:36579 <- 33.30.77.186:1812 +10.043 +0.050
            Framed-Protocol = PPP
            Framed-Compression = Van-Jacobson-TCP-IP
            Reply-Message = "Access-Accept"
            Session-Timeout = 250000
            Proxy-State = 0x323431
            Proxy-State = 0x313437
            Acct-Interim-Interval = 3600
            Authenticator-Field = 0xd390b79f5c8c00f0e68dbe48a9674289
    2024-11-07 08:56:21.857550 (183) Access-Reject Id 24 tcp.pcap:9.22.15.40:36579 <- 33.30.77.186:1812 +10.075 +1.022
            Reply-Message = "Access-Reject"
            Proxy-State = 0x3433
            Authenticator-Field = 0x2e4955cd9199556335f896c9963f2542
    2024-11-07 08:56:21.882926 (185) Access-Request Id 110 tcp.pcap:9.22.15.40:36579 -> 33.30.77.186:1812 +10.100
            User-Name = "username3@realm"
            User-Password = "\r\3476\220\3215d13d"
            NAS-IP-Address = 33.43.47.20
            NAS-Port = 6758160
            Service-Type = Framed-User
            Framed-Protocol = PPP
            Calling-Station-Id = "2c:0f:ff:aa:e0:a0"
            Proxy-State = 0x313736
            Proxy-State = 0x313733
            NAS-Port-Type = PPPoEoQinQ
            Acct-Session-Id = "A829762E652F"
            Event-Timestamp = "2024-11-07T06:56:21Z"
            Message-Authenticator = 0xf95549568ddd9d4ec4f478b94270268d
            NAS-Port-Id = "4/2/3:132.655"
            Authenticator-Field = 0x14477dbef0859972ae862bb00c222bc4
    2024-11-07 08:56:21.890839 (186) Access-Reject Id 200 tcp.pcap:9.22.15.40:36579 <- 33.30.77.186:1812 +10.108 +1.023
            Reply-Message = "User does not exist"
            Reply-Message = "Access-Reject"
            Proxy-State = 0x323335
            Proxy-State = 0x333612
            Authenticator-Field = 0x96dc0d26ac6f02ff7a7ee547caa1bede
    Done reading packets (tcp.pcap)

alandekok:
> There really isn't any other tool that allows inspection of the data, even poorly. 

wireshark should work with RADIUS/TCP.

We can't accept this patch, because it works only in some circumstances, and will be wrong or misleading in other circumstances.

derde:
Wireshark cannot decode RADIUS/TCP.  This code always works.  I'm sorry you didn't understand it.

alandekok:
I'm sorry you're abusive.

arr2036:
Reassembly isn't usually a problem with UDP, because the only payloads large enough to require it get fragmented on the supplicant side, and are generally lower than the link MTU of the interface radsniff is running over.

With TCP (a stream protocol), we can very easily have RADIUS packets split over a TCP packet boundary.  It'll occur in any reasonably high load situation.  Even with with naggle's algorithm disabled, efficient code will be using `writev`, and asynchronous code will be using a write event on the TCP Socket, where multiple outbound RADIUS packets will likely be queued until the next time the event loop is serviced...  You _WILL_ have split RADIUS packets, and this code _WILL NOT_ work.  So broadly from a technical perspective I agree with Alan.

> This code always works. I'm sorry you didn't understand it.

Is both false, and antagonising.

PR #5465, opened 2024-11-21
fix confusing message in rlm_sql when user mangles query

jimdigriz:
830c3dece7c missed fixing this too, but also changed from RDEBUG to WARN to match the `client_query` warning


alandekok:
commits need to be signed?

jimdigriz:
It is?
image

alandekok:
merged manually

PR #5476, opened 2024-12-10
Close path opened by fr_mkdir() to avoid fd leak (CID #1636671)

jejones3141:
In `fr_bio_fd_reopen()`, if `cfg->mkdir` is true, it tries to create a directory and returns -1 if it fails. If it succeeds, the path number of the created directory is stored in `fd`...and the next use of `fd` is assigning the result of `open()` to it, leaving the directory open and accumulating inaccessible open files.


alandekok:
I already have a different fix, and will push it in a bit.  We should be using `openat()` if possible.

PR #5479, opened 2024-12-17
feat: Added `auto_begin` FLAG to dictionary parse.

ethan-thompson:
This flag will allow the toggle between automatically beginning tlv's and struct's (enabled by default).


alandekok:
Based on offline discussion, we can likely hold this as "maybe not needed"

Issue #5486, opened 2025-01-06
[defect]: Proxying EAP sessions fails when using eap in pre-proxy


meadmaker:
### What type of defect/bug is this?

Unexpected behaviour (obvious or verified by project member)

### How can the issue be reproduced?

This issue applies to the `3.2.x` branch.

1. Add `eap` to the `pre-proxy` section of the `default` site.
2. Create a `home_server`, `home_server_pool`, and `realm` entry in `proxy.conf` to define proxying requests to a `home_server`.
3. Create a `client` in `clients.conf` to accept connections.
4. Generate a RADIUS/EAP authentication session for the realm that will be proxied.  I used `rad_eap_test`.

Any EAP authentication to be proxied will fail before the proxying happens, in the first packet of the EAP/RADIUS authentication.

### Log output from the FreeRADIUS daemon

```shell
The most relevant line of the output is:

eap: EAP length does not match attribute length
```

The full output is:
```
root@bf3e4ff0358d:/etc/freeradius# raddebug -t 0
(721) Mon Jan  6 04:04:48 2025: Debug: Received Access-Request Id 0 from 172.16.128.101:42788 to 172.16.128.10:1812 length 145
(721) Mon Jan  6 04:04:48 2025: Debug:   User-Name = "bob@testing.id"
(721) Mon Jan  6 04:04:48 2025: Debug:   NAS-IP-Address = 127.0.0.1
(721) Mon Jan  6 04:04:48 2025: Debug:   Calling-Station-Id = "70-6F-6C-69-73-68"
(721) Mon Jan  6 04:04:48 2025: Debug:   Framed-MTU = 1400
(721) Mon Jan  6 04:04:48 2025: Debug:   NAS-Port-Type = Wireless-802.11
(721) Mon Jan  6 04:04:48 2025: Debug:   Service-Type = Framed-User
(721) Mon Jan  6 04:04:48 2025: Debug:   Connect-Info = "rad_eap_test + eapol_test"
(721) Mon Jan  6 04:04:48 2025: Debug:   EAP-Message = 0x02b7001301626f624074657374696e672e6964
(721) Mon Jan  6 04:04:48 2025: Debug:   Message-Authenticator = 0x27def39f84d9349d5e9a4d61034900d6
(721) Mon Jan  6 04:04:48 2025: Debug: # Executing section authorize from file /etc/freeradius/sites-enabled/default
(721) Mon Jan  6 04:04:48 2025: Debug:   authorize {
[...]

PR #5487, opened 2025-01-06
Fix EAP preproxy

meadmaker:
The calculation of EAP message length needs to total the length of all the EAP message attributes, rather than the length of the first EAP message attribute and all subsequent attributes of all types.  Also, enhance the error message to include the calculated lengths.

Fixes https://github.com/FreeRADIUS/freeradius-server/issues/5486


alandekok:
The patch is unsigned, so I added it manually as commit 31e9fc79205

PR #5492, opened 2025-01-13
fix: da->dict == NULL

ethan-thompson:
Updated reference to attr.type_parse to use the dict of the decode context, since the da->dict is not set until shortly after this call is made in dict_read_process_common


src/lib/util/dict_tokenize.c
```diff
@@ -437,12 +437,12 @@ static int dict_process_type_field(dict_tokenize_ctx_t *dctx, char const *name,
 	if (fr_type_is_null(type)) {
 		fr_dict_attr_t *da = *da_p;
```
alandekok, line 438: You can nuke this line, too.  Otherwise the compiler will complain about unused variables

PR #5494, opened 2025-01-15
fix: Protocol type_size and type_length checks use protocol set values

ethan-thompson:
Updated the type_size check to correctly use what is set in the protocol instead of a hard-coded value.


ethan-thompson:
This fix has already been added to master by @alandekok

PR #5497, opened 2025-01-21
Feature/der encoder decoder

ethan-thompson:
Wrote an encoder and decoder for DER. This adds the ability to decoder DER encoded X509 certificates and CSR's into dictionary attributes, as well as encode into DER.


src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+
+/**
+ * $Id$
+ *
+ * @file protocols/der/decode.c
+ * @brief Functions to decode DER encoded data.
+ *
+ * @author Ethan Thompson (ethan.thompson@inkbridge.io)
+ *
+ * @copyright 2024 Inkbridge Networks SAS.
```
alandekok, line 25: Oh boy!

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+ */
+
+#include "include/build.h"
```
alandekok, line 28: nits: we usually put `#include "foo"` after all of the common includes.  that way any location definitions don't affect the other header files.

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+#include "der.h"
```
alandekok, line 29: `#include <freeradius-devel/util/table.h>` etc.

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+extern fr_dict_autoload_t libfreeradius_der_dict[];
+fr_dict_autoload_t	  libfreeradius_der_dict[] = { { .out = &dict_der, .proto = "der" }, { NULL } };
+
+extern fr_dict_attr_autoload_t libfreeradius_der_dict_attr[];
+fr_dict_attr_autoload_t	       libfreeradius_der_dict_attr[] = {
+	       { NULL }
+};
+
+fr_der_tag_constructed_t tag_labels[] = {
+	[FR_DER_TAG_BOOLEAN]	      = FR_DER_TAG_PRIMATIVE,
```
alandekok, line 50: `PRIMATIVE`?  Is that a DER thing, or is it `PRIMITIVE`?

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+	if (--instance_count != 0) return;
+
+	fr_dict_autofree(libfreeradius_der_dict);
+}
+
+static int dict_flag_tagnum(fr_dict_attr_t **da_p, char const *value, UNUSED fr_dict_flag_parser_rule_t const *rules)
+{
+	fr_der_attr_flags_t *flags = fr_dict_attr_ext(*da_p, FR_DICT_ATTR_EXT_PROTOCOL_SPECIFIC);
+
+	flags->tagnum = (uint8_t)atoi(value);
```
alandekok, line 106: do we want sanity checks here for overflow, or for bad formats?

Admins make mistakes. An error of "you screwed up" is usually better than silently doing the wrong thing.

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+	flags->sequence_of = type;
+	flags->is_sequence_of = true;
+
+	return 0;
+}
+
+static int dict_flag_set_of(fr_dict_attr_t **da_p, char const *value, UNUSED fr_dict_flag_parser_rule_t const *rules)
+{
+	static fr_table_num_sorted_t const table[] = {
+		{ L("bitstring"), FR_DER_TAG_BITSTRING },
```
alandekok, line 234: with repeated tables like this, it may be useful to put them into a common macro?  depending on whether or not the C preprocessor likes the `L()` macro inside another macro :(

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+
+	flags->class = FR_DER_CLASS_CONTEXT;
+
+	flags->tagnum = (uint8_t)atoi(value);
+
+	return 0;
+}
+
+static fr_dict_flag_parser_t const der_flags[] = {
+						   { L("class"), { .func = dict_flag_class } },
```
alandekok, line 340: whitespace / indentation

src/protocols/der/base.c
```diff
@@ -0,0 +1,518 @@
...
+
+static bool attr_valid(fr_dict_attr_t *da)
+{
+	if (da->flags.subtype && !fr_type_to_der_tag_valid(da->type, da->flags.subtype)) {
+		return false;
+	}
+
+	if (fr_der_flag_is_sequence_of(da->parent) || fr_der_flag_is_set_of(da->parent)) {
+		static fr_table_num_sorted_t const table[] = {
+			{ L("bitstring"), FR_DER_TAG_BITSTRING },
```
alandekok, line 453: repeated repetition... :(

src/protocols/der/der.h
```diff
@@ -0,0 +1,171 @@
+#include "include/build.h"
```
alandekok, line 1: use `<freeradius-devel/...`

src/protocols/der/der.h
```diff
@@ -0,0 +1,171 @@
...
+} fr_der_tag_class_t;
+
+extern fr_der_tag_constructed_t tag_labels[];
+
+/*
+ *	Create a mapping between FR_TYPE_* and valid FR_DER_TAG_*'s
+ */
+static bool *fr_type_to_der_tags[] = {
+	[FR_TYPE_MAX] = NULL,
+	[FR_TYPE_BOOL] = (bool []){[FR_DER_TAG_BOOLEAN] = true, [FR_DER_TAG_INTEGER] = true, [FR_DER_TAG_NULL] = true, [FR_DER_TAG_MAX] = false},
```
alandekok, line 57: maybe put line breaks after the `,` which can make the lines shorter, and a bit easier to read.

src/protocols/der/der.h
```diff
@@ -0,0 +1,171 @@
...
+	[FR_TYPE_STRING] = FR_DER_TAG_UTF8_STRING,
+	[FR_TYPE_DATE] = FR_DER_TAG_GENERALIZED_TIME,
+	[FR_TYPE_TLV] = FR_DER_TAG_SEQUENCE,
+	[FR_TYPE_STRUCT] = FR_DER_TAG_SEQUENCE,
+	[FR_TYPE_GROUP] = FR_DER_TAG_SEQUENCE
+};
+
+static inline CC_HINT(always_inline) fr_der_tag_num_t fr_type_to_der_tag_default(fr_type_t type)
+{
+	return fr_type_to_der_tag_defaults[type];
```
alandekok, line 106: what about data types like `FR_TYPE_IPADDR` which aren't listed in the above table?  The C compiler will initialize the array entry to `0`.  Is there a `FR_DER_TAG_INVALID` which is defined as `0` ?

src/protocols/der/der.h
```diff
@@ -0,0 +1,171 @@
...
+
+#define DER_TAG_CONTINUATION 0x1f	 //!< Mask to check if the tag is a continuation.
+
+#define DER_LEN_MULTI_BYTE 0x80	 //!< Mask to check if the length is multi-byte.
+
+#define DER_BOOLEAN_FALSE 0x00	 //!< DER encoded boolean false value.
+#define DER_BOOLEAN_TRUE 0xff	 //!< DER encoded boolean true value.
+
+typedef struct {
+	uint8_t tagnum;
```
alandekok, line 129: maybe format these so that the names are further to the right, and then all lined up vertically.  that makes it easier to read.

src/protocols/der/der.h
```diff
@@ -0,0 +1,171 @@
...
+	bool is_choice;
+} fr_der_attr_flags_t;
+
+static inline fr_der_attr_flags_t const *fr_der_attr_flags(fr_dict_attr_t const *da)
+{
+	return fr_dict_attr_ext(da, FR_DICT_ATTR_EXT_PROTOCOL_SPECIFIC);
+}
+
+#define fr_der_flag_tagnum(_da) 	(fr_der_attr_flags(_da)->tagnum)
+#define fr_der_flag_class(_da) 	(fr_der_attr_flags(_da)->class)
```
alandekok, line 153: whitespace nits :)

src/protocols/der/all.mk
```diff
@@ -0,0 +1,15 @@
+#
+# Makefile
+#
+# Version:      $Id$
+#
+TARGET		:= libfreeradius-der$(L)
+
+SOURCES		:= base.c \
+		   decode.c \
```
alandekok, line 9: maybe include push this commit until after the `encode.c` and `decode.c` files have been added.  otherwise the build is broken.

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+#include <freeradius-devel/util/strerror.h>
+#include <freeradius-devel/util/types.h>
+
+#include <freeradius-devel/util/proto.h>
+#include <freeradius-devel/util/sbuff.h>
+#include <freeradius-devel/util/struct.h>
+#include <freeradius-devel/util/time.h>
+#include <freeradius-devel/util/value.h>
+#include <stdlib.h>
+#include <time.h>
```
alandekok, line 53: `util/time.h` already includes `time.h`

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+	bool	 oid_value_pairs;
+} fr_der_decode_ctx_t;
+
+#define IS_DER_TAG_CONTINUATION(_tag) (((_tag) & DER_TAG_CONTINUATION) == DER_TAG_CONTINUATION)
+#define IS_DER_TAG_CONSTRUCTED(_tag) (((_tag) & 0x20) == 0x20)
+#define IS_DER_LEN_MULTI_BYTE(_len) (((_len) & DER_LEN_MULTI_BYTE) == DER_LEN_MULTI_BYTE)
+
+typedef ssize_t (*fr_der_decode_oid_t)(uint64_t subidentifier, void *uctx, bool is_last);
+
+static ssize_t fr_der_decode_oid(fr_pair_list_t *out, fr_dbuff_t *in, fr_der_decode_oid_t func, void *uctx);
```
alandekok, line 66: maybe add `CC_HINT(nonnull)` for the function prototypes

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+					  .decode      = fr_der_decode_generalized_time },
+	[FR_DER_TAG_VISIBLE_STRING]   = { .constructed = FR_DER_TAG_PRIMATIVE, .decode = fr_der_decode_visible_string },
+	[FR_DER_TAG_GENERAL_STRING]   = { .constructed = FR_DER_TAG_PRIMATIVE, .decode = fr_der_decode_general_string },
+	[FR_DER_TAG_UNIVERSAL_STRING] = { .constructed = FR_DER_TAG_PRIMATIVE,
+					  .decode      = fr_der_decode_universal_string },
+
+	[UINT8_MAX] = { .constructed = FR_DER_TAG_PRIMATIVE, .decode = NULL },
+};
+
+static int decode_test_ctx(void **out, TALLOC_CTX *ctx, UNUSED fr_dict_t const *dict)
```
alandekok, line 173: move the various `test_ctx` stuff to the bottom of the file, where it's located for the other protocols

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+	size_t len = fr_dbuff_remaining(&our_in);
+
+	if (!fr_type_is_bool(parent->type)) {
+		fr_strerror_printf("Boolean found in non-boolean attribute %s of type %s", parent->name,
+				   fr_type_to_str(parent->type));
+		return -1;
+	}
+
+	/*
+	 * 	ISO/IEC 8825-1:2021
```
alandekok, line 204: comments and references to standards make me happy.

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+		fr_strerror_printf("Boolean has incorrect length (%zu). Must be 1.", len);
+		return -1;
+	}
+
+	if (unlikely(fr_dbuff_out(&val, &our_in) < 0)) {
+		fr_strerror_const("Insufficient data for boolean");
+		return -1;
+	}
+
+	if (unlikely(val != DER_BOOLEAN_FALSE && val != DER_BOOLEAN_TRUE)) {
```
alandekok, line 226: brackets around things.  It's not required by C, but it cam avoid typos

`if ((a == b) || (c == d))`

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+		fr_strerror_printf("Boolean is not correctly DER encoded (0x%02x or 0x%02x).", DER_BOOLEAN_FALSE,
+				   DER_BOOLEAN_TRUE);
+		return -1;
+	}
+
+	vp = fr_pair_afrom_da(ctx, parent);
+	if (unlikely(vp == NULL)) {
+		fr_strerror_const("Out of memory for boolean pair");
```
alandekok, line 234: just `Out of memory` is fine, like all of the other uses.  arguably we could have a macro for this somewhere else in the code.

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+		return -1;
+	}
+
+	fr_pair_append(out, vp);
+
+	return fr_dbuff_set(in, &our_in);
+}
+
+typedef struct {
+	TALLOC_CTX	     *ctx; /**< Allocation context */
```
alandekok, line 581: comments can be aligned vertically, and use `//!<` which is more consistent with the rest of the code.

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+
+static ssize_t fr_der_decode_printable_string(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_dict_attr_t const *parent,
+					      fr_dbuff_t *in, UNUSED fr_der_decode_ctx_t *decode_ctx)
+{
+	fr_pair_t *vp;
+	fr_dbuff_t our_in = FR_DBUFF(in);
+	char	  *str	  = NULL;
+
+	static bool const allowed_chars[] = {
+		[' '] = true, ['\''] = true, ['('] = true, [')'] = true, ['+'] = true,	     [','] = true, ['-'] = true,
```
alandekok, line 1421: format as 4 per line, which makes it a bit easier to read

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+
+static ssize_t fr_der_decode_t61_string(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_dict_attr_t const *parent,
+					fr_dbuff_t *in, UNUSED fr_der_decode_ctx_t *decode_ctx)
+{
+	fr_pair_t *vp;
+	fr_dbuff_t our_in = FR_DBUFF(in);
+	char	  *str	  = NULL;
+
+	static bool const allowed_chars[] = {
+		[0x08] = true, [0x0A] = true, [0x0C] = true,	  [0x0D] = true, [0x0E] = true, [0x0F] = true,
```
alandekok, line 1493: see `src/lib/util/token.c`.  GCC and clang support ranges, which makes these tables a _lot_ easier

```
/*
 *	This is fine.  Don't complain.
 */
#ifdef __clang__
#pragma clang diagnostic ignored "-Wgnu-designator"
#endif
...
	[ 0 ... T_HASH ] = '?',	/* GCC extension for range initialization, also allowed by clang */
```

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2852 @@
...
+	constructed = IS_DER_TAG_CONSTRUCTED(tag_byte);
+
+	/*
+	 *	Decode the tag
+	 */
+	if (IS_DER_TAG_CONTINUATION(tag_byte)) {
+		/*
+		 *	We have a multi-byte tag
+		 *
+		 *	Note: Mutli-byte tags would mean having a tag number that is greater than 30 (0x1E) (since tag
```
alandekok, line 2073: multi

share/dictionary/der/dictionary
```diff
@@ -0,0 +1,21 @@
...
+#
+#	The FreeRADIUS Vendor-Specific dictionary for TLS operations.
+#
+# Version:	$Id$
+#
+
+PROTOCOL	DER		11354911
+BEGIN-PROTOCOL DER
+
+$INCLUDE dictionary.test
```
alandekok, line 14: probably don't want test dictionaries in the default build.

the `src/tests/unit` stuff allows for local dictionaries to be loaded

share/dictionary/der/dictionary.common
```diff
@@ -0,0 +1,55 @@
+DEFINE GeneralName      choice
+BEGIN GeneralName
+
+ATTRIBUTE otherName     0       sequence option=0
```
alandekok, line 4: run `./scripts/dict/format.dl share/dictionary/der/dictionary*`

for consistent formatting

src/protocols/der/base.c
ethan-thompson: Addressed in new commits

src/tests/unit/protocols/der/dictionary.test
```diff
@@ -0,0 +1,183 @@
+# -*- text -*-
+# Copyright (C) 2025 The FreeRADIUS Server project and contributors
+# This work is licensed under CC-BY version 4.0 https://creativecommons.org/licenses/by/4.0
+# Version $Id$
+DEFINE	Certificate-Extensions				x509_extensions ref=OID-Tree
```
alandekok, line 5: run the dicts through `scripts/dict/format.pl`.  I like pretty things

share/dictionary/der/dictionary
```diff
@@ -0,0 +1,19 @@
...
+# This work is licensed under CC-BY version 4.0 https://creativecommons.org/licenses/by/4.0
+# Version $Id$
+#
+#	The FreeRADIUS Vendor-Specific dictionary for TLS operations.
+#
+# Version:	$Id$
+#
+
+PROTOCOL	DER		11354911
+BEGIN-PROTOCOL DER
```
alandekok, line 12: Delete those two lines, and replace them with

`BEGIN PROTOCOL DER 11354911`

I _think_ the server supports 64-bit protocol numbers?  I haven't checked recently.

what's the significance of `11354911` ?

src/tests/unit/protocols/der/dictionary.test
ethan-thompson: That was the format after having run the `format.pl` script on the dictionaries

share/dictionary/der/dictionary
ethan-thompson: I will update that line. There was no significance in 11354911. I believe @arr2036 may have used this number to avoid collision

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2845 @@
...
+};
+
+static ssize_t fr_der_decode_boolean(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_dict_attr_t const *parent, fr_dbuff_t *in,
+				     UNUSED fr_der_decode_ctx_t *decode_ctx)
+{
+	fr_pair_t *vp;
+	fr_dbuff_t our_in = FR_DBUFF(in);
+	uint8_t	   val;
+
+	size_t len = fr_dbuff_remaining(&our_in);
```
arr2036, line 172: You should check the return code of `fr_dbuff_out` instead of manually checking the length

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2845 @@
...
+static ssize_t fr_der_decode_integer(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_dict_attr_t const *parent, fr_dbuff_t *in,
+				     UNUSED fr_der_decode_ctx_t *decode_ctx)
+{
+	fr_pair_t *vp;
+	fr_dbuff_t our_in = FR_DBUFF(in);
+	int64_t	   val	  = 0;
+	uint8_t	   sign	  = 0;
+	size_t	   i;
+
+	size_t len = fr_dbuff_remaining(&our_in);
```
arr2036, line 231: Same here

src/protocols/der/decode.c
```diff
@@ -0,0 +1,2845 @@
...
+static ssize_t fr_der_decode_bitstring(TALLOC_CTX *ctx, fr_pair_list_t *out, fr_dict_attr_t const *parent,
+				       fr_dbuff_t *in, fr_der_decode_ctx_t *decode_ctx)
+{
+	fr_pair_t *vp;
+	fr_dbuff_t our_in      = FR_DBUFF(in);
+	uint8_t	   unused_bits = 0;
+	uint8_t	  *data;
+
+	ssize_t data_len = 0, index = 0;
+	size_t	len = fr_dbuff_remaining(&our_in);
```
arr2036, line 331: Same here

arr2036:
All instances of `fr_dbuff_remaining` should be removed, and the output of the value copying function should be used instead.  `fr_dbuff_remaining` just shows what's available in your window, not what's available in the stream.

arr2036:
Default value flag should produce a value box.  Using a `DEFAULT` enum value is counterintuitive.

arr2036:
Where possible you should still remove fr_dbuff_remaining where fr_dbuff_out would perform the same check

arr2036:
Have some bad news `...` is a GCC extension, so probably wants to be converted to the explicit range

arr2036:
Even with the artificial windows, it's probably better to use fr_dbuff_extend_lowat and check the return values.  That lets the artificial "end" point to unrealised data.

src/protocols/der/encode.c
```diff
@@ -0,0 +1,2014 @@
...
+
+	fr_dbuff_set(dbuff, length_start);
+
+	fr_dbuff_in(dbuff, (uint8_t)(0x80 | len_len));
+
+	for (; i < len_len; i++) {
+		fr_dbuff_in(dbuff, (uint8_t)((slen) >> ((len_len - i - 1) * 8)));
+	}
+
+	fr_dbuff_set(dbuff, fr_dbuff_current(length_start) + len_len + 1 + slen);
```
alandekok, line 1756: This doesn't seem right?  The function encodes the length, and then skips over the length + the data which will be encoded?

and almost all calls to this function do

```
 ...fr_der_encode_len(&our_dbuff, &length_start, fr_dbuff_behind(&length_start) - 1)
```

except for one...

src/protocols/der/encode.c
```diff
@@ -0,0 +1,2014 @@
...
+	 *
+	 */
+
+	if (fr_der_flag_has_default(vp->da)) {
+		/*
+		 *	Skip encoding the default value, as per ISO/IEC 8825-1:2021 11.5
+		 */
+		fr_dict_enum_value_t const *evp;
+
+		evp = fr_dict_enum_by_name(vp->da, "DEFAULT", strlen("DEFAULT"));
```
alandekok, line 1868: I'll leave this in for now, but it could be in the `dict_ext` stuff?

arr2036: I made similar comments on the review call.  It can just be a flag.

PR #5507, opened 2025-01-31
Moved Installation module to under the Howto module

nolade:
2nd set of toplology undates. The Installation module now located under the Howto module and refs and nav panel updated.


nolade:
2nd topo update

alandekok:
merged manually with a few minor edits

PR #5518, opened 2025-02-14
adds a hint message for users testing TEAP using eapol_test

stgmsa:
Describe the limitation of using some version of eapol_test in TEAP testing.
Give possible suggestions.


alandekok:
We require commits to be signed.  I'll just pull this in manually.

PR #5523, opened 2025-02-21
docs: Terminology updates

nolade:
First pass through all docs to modify/change/update words and terms as defined in HIVE 3149. Note - I didnt change terms/words in the coding sections of the *.c type files as defs/labels already set. Some comments were updated.


doc/antora/modules/concepts/pages/wpa.adoc
```diff
@@ -7,21 +7,21 @@
...
+WPA2 implements the required elements of 802.11i. In particular, in addition to TKIP and the Michael algorithm, it introduces a new Advanced Encryption Standard based algorithm, CCMP, that is considered fully secure. Official support for WPA2 in Microsoft Windows XP was rolled out on 1 May 2005. Driver upgrades for network cards may be required. Apple Computer supports WPA2 on all AirPort Extreme-enabled Apple Macintosh, the AirPort Extreme Base Station, and the AirPort Express. Firmware upgrades needed are included in AirPort 4.2, released July 14, 2005. Note that from March 13, 2006, WPA2 certification is required for all new devices wishing to be Wi-Fi certified.
```
arr2036, line 24: To me these have different meanings.  "required" would be the necessary bits, and "mandatory" would be the parts explicitly marked as being needed, i.e. the "MUST"s in an RFC.

doc/antora/modules/howto/pages/installation/dependencies.adoc
```diff
@@ -1,7 +1,7 @@
 = FreeRADIUS Dependencies
 
 Some external dependencies must be installed before building or
-running FreeRADIUS. The core depends on two mandatory libraries:
+running FreeRADIUS. The core depends on two required libraries:
```
arr2036, line 4: Just remove required

alandekok:
I'll pull this in after some tweaks.

* bad words can stay in the source code.  We're happy with that.
* The directory doc/rfc is taken from external, public, standards.  We just keep those as-is
* Most of the directory doc/antora/modules/reference/raddb/ is taken from the originals in raddb/.  The originals should generally be modified, not the auto-converted outputs.

alandekok:
I've merged everything else in, with tweaks as noted by Arran.

Sometimes words like "mandatory" have technical meaning, and are a required part of the specifications.

PR #5527, opened 2025-02-28
Virtual Server section 

nolade:
Reference Documentation > Configuration Files > Virtual Servers section reworked
- organized into vm types (authentication/datastore/policy/io/protocol)
- added top level page for each sub-section,
- nav updated


alandekok:
I've pushed a different fix, which organizes the virtual servers by protocol.  It also adds DNS and DHCPv6, which were missing from the nav bar.

I've also removed a few old virtual servers which were wrong, or out of date.

I didn't add an `index.adoc` file for each protocol.  That takes more time.  For now, the navbar is at least organized.

PR #5530, opened 2025-03-07
docs: Module section update and formatting.

nolade:
Partials added to replace lists - formatted as tables
Docs sub-folder contains new files (landing pages)


alandekok:
Merged with two minor changes:

* no need to repeat "SQL" or "REDIS" for IP pools.  The "IP Pool" item is already within the "SQL" parent, and the other nav entries under "SQL" don't repeat "SQL"
* use `*` instead of a bullet character for bullet lists

PR #5533, opened 2025-03-16
Draft: eap:dedup_key should have PW_TYPE_XLAT

jimdigriz:
The example shows an xlat ('%{Calling-Station-Id}') but on startup logs the following which seems somewhat a tease...
```
Found dynamic expansion in string which will not be dynamically expanded
```

This make sense or look correct?

COMPILE TESTED ONLY


alandekok:
I think this is correct.  Let me double check, and I'll merge it in

PR #5538, opened 2025-03-28
docs: Add howto/Datastore section & AD/LDAP/SQL/REDIS subsections. 

nolade:
The <type>_datastore.adoc files are stubs with headings only. If topology OK - AD info will be copied over.


alandekok:
Pulled with a few tweaks.  Changes should show up after CI runs.

PR #5541, opened 2025-04-02
Docker misc improvements

jwarnier:


alandekok:
We've re-done the docker rules and cleaned them up with these changes and more.

PR #5542, opened 2025-04-03
Fix typos in CI

jwarnier:
In GitHub CI code (actions and workflows):
* Fixes mistypes in des*c*ription tags
* Fixes depend*a*ncies in comments and steps names


alandekok:
Thanks.  I've pulled this is as one commit, to avoid "back and forth" changes in the git history.

PR #5545, opened 2025-04-05
Update arista dictionary

hzhongmj:
This brings Arista dictionary up to date


alandekok:
Thanks.  I've pulled it in with a few tweaks.

PR #5557, opened 2025-04-18
Update dictionary.ciena

jbruynoo:
Added missing attributes and values


alandekok:
I've pulled this in manually.

If you'll note, we do _not_ use the `Ciena-` prefix in v4.  There is no need to add it back.

PR #5564, opened 2025-04-28
Adding support for 389-ds/FreeIPA PBKDF2 passwords

gvde:
This adds support for the PBKDF2 password hashes as used by 389-ds, i.e. with headers `{PBKDF2-SHA1}`, `{PBKDF2-SHA256` and `{PBKDF2-SHA512}`.

In addition, 389-ds also supports a legacy format with header `{PBKDF2_SHA256}` (note the underscore instead of the dash).


gvde:
I have added support for the legacy PBKDF2_SHA256 hash format in 389ds. As far as I understand this was the first implementation of PBKDF2 using a binary format for storage.

https://fy.blackhats.net.au/blog/2022-11-25-why-are-pbkdf2-sha256-and-pbkdf2-sha256-different-in-389-ds/

It uses fixed lengths and iterations as defined in the [389 source code](https://github.com/389ds/389-ds-base/blob/4e62f543ea656aa622687594912e6fdc31327521/ldap/servers/plugins/pwdstorage/pbkdf2_pwd.c).

I have condensed the code from pap_auth_pbkdf2_parse_digest into pap_auth_pbkdf2_sha256_legacy. I also used a local struct to write the base64 data directly into instead of splitting the parts into three separate variables. A struct seem liked the logical choice as that's what it's effectively is.

I think that completes the implementation of the 389ds hashes support in the master branch. All module tests for the new schemes are successful.

I hope coding style is O.K. If I need to make any changes, please let me know.

If my changes are accepted I would take a look at the v3.2 branch for backporting.

arr2036:
This looks good to me.  Only minor style nit is you should put your sub-expressions in brackets, i.e. `if ((end - p) > 2)` not `if (end - p > 2)`.  If you fix that I think this is good to merge.

arr2036:
...also,  could you rebase off of master to avoid the merge commits.

gvde:
The failing master-self-ubuntu24-linux-clang* checks fail because of this:

```
src/modules/rlm_pap/rlm_pap.c:904:26: warning: 1st function call argument is an uninitialized value [core.CallAndMessage]
  904 |         pbkdf2_buf.iterations = ntohl(pbkdf2_buf.iterations);
      |                                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

The pbkdf2_buf struct is not used uninitialized. The contents comes from the fr_base64_decode and is checked for required, fixed length, i.e. the whole size of the struct is written by the decode.

Should I (redundantly) initialize the struct anyway? I would probably have to initialize all three members?

The alternative, I guess, would be to create a temporary buffer, decode into that and copy each member from there, which seems like overkill to achieve the exact same thing...

gvde:
> This looks good to me. Only minor style nit is you should put your sub-expressions in brackets, i.e. `if ((end - p) > 2)` not `if (end - p > 2)`. If you fix that I think this is good to merge.

O.K. Done. I hope I didn't miss some. Although the minimum length checks could probably be improved to a higher value, if it was clear to me how long each component must be. I also wasn't sure if "PASSWORD_HASH_VARIABLE" is the best type for each of them. In particular the legacy hash has a fixed length...

gvde:
> ...also, could you rebase off of master to avoid the merge commits.

O.K. After fighting with git for a while I have managed to rebase to current master. I have also squashed the commits into two remaining commits: one for the implementation of the legacy hash and one for the other hashes.

arr2036:
Yes, just redundantly initialise the structure

gvde:
The first check `master-self-ubuntu24-linux-clang-lean` failed with

```
Unable to connect to cache.servers.ca.internal.networkradius.com:3128
```

The `master-macos-clang` fails at the `EAPOL-TEST tls`. I don't think I have touched anything related to that test. It succeeds on my build machine.

PR #5566, opened 2025-04-28
docs: Reorganized sqlippool section and 1st pass edit on content.

nolade:
Reorganize SQL-IP-POOL section and xref/link updates


alandekok:
Added with tweaks, thanks.

PR #5568, opened 2025-05-01
docs: Add raduat tool info to Antora repo

nolade:
docs: import raduat information from wiki, salt repo, and raduat script


alandekok:
Pulled in with minor tweaks

PR #5576, opened 2025-05-08
Implement support for 389ds pbkdf2 password hashes in v3.2.x

gvde:
Implement support for 389ds password hashes PBKDF2-SHA{1,256,512} and legacy PBKDF2_SHA256 in v.3.2.x

Backport of PR #5564


gvde:
A remark: this is a notable change in cases, when there are multiple hashes for divergent passwords available. For instance, if there is a NT hash and PBKDF2 hash available in the ldap directory, before freeradius could only evaluate the NT hash. Now, it can evaluate the PBKDF2 hash and in case that hash is for a different password, it would fail now, if the PBKDF2 hash is checked first. I don't expect that to be a common scenario but technically it's two different attributes in ldap which may or may not hash the same plaintext password.

PR #5578, opened 2025-05-09
Docs xlat reorg

nolade:
1st pass at reorganzing xlats directory into small discrete chunks of info. Added summary file
See individual commits for more details of updated files.


alandekok:
Merged with tweaks, thanks.

PR #5583, opened 2025-05-16
TLS 1.3 support for LDAP module

torinsapp:
Originally, I was receiving the error message:
Invalid 'tls.tls_min_version' value "1.3"

Test setup:
Freeradius server with mods-enabled/ldap configured. 
tls_start must be no and tls_min_version must be 1.3 to replicate the error message.

Decided to check it out and found it basically hardcoded to not accept 1.3. Added the option and freeradius will start now.

This code resolves that issue.


alandekok:
Thanks, I've pulled in the patch as one commit.

PR #5588, opened 2025-05-20
[v3.2.x] Debian systemd: MemoryLimit -> MemoryMax

matsimon:
MemoryMax has been introduced with systemd 231
MemoryLimit has been marked deprecated with systemd 252


matsimon:
This avoids a warning like this one (on Debian 12, using systemd 231):

```bash
# systemd-analyze verify /lib/systemd/system/freeradius.service
/lib/systemd/system/freeradius.service:23: Unit uses MemoryLimit=; please use MemoryMax= instead. Support for MemoryLimit= will be removed soon.```

matsimon:
Forgot a bit about this PR, but thank you for merging it. A local modification has been working for some time here on Debian, so looking forward for the upstream inclusion. 👍

PR #5589, opened 2025-05-20
[master] Debian/RedHat systemd: MemoryLimit -> MemoryMax

matsimon:
Why the change?

* MemoryMax has been introduced with [systemd 231](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#MemoryMax=bytes)
* MemoryLimit has been marked deprecated with [systemd 252](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#History)

Thoughts based on the systemd version shipped with Debian and RHEL distributions:
* For Debian-based distributions: Debian 10 already shipped with systemd 241 and is [past its long term support](https://www.debian.org/News/2024/20240615), Ubuntu 18.04 LTS shipped with 239 and is also pasts is regular support (only paid support options) so for future releases of FreeRADIUS we should be pretty safe.
* For EL-based distributions: Considering that RHEL 7 has been out of regular support and that RHEL 8 ships with systemd 239 this change should be safe to.


matsimon:
This avoids a warning like this one (on Debian 12, using systemd 231)

```bash
# systemd-analyze verify /lib/systemd/system/freeradius.service
/lib/systemd/system/freeradius.service:23: Unit uses MemoryLimit=; please use MemoryMax= instead. Support for MemoryLimit= will be removed soon.```

arr2036:
Is there any change in behaviour?  Or the systemd guys just like renaming things to keep people on their toes?

mcnewton:
As well as potential change of behaviour, this also needs checking that it doesn't break any of the older Debian-based distros as they are all built using the same source.

matsimon:
@arr2036: It seems to be related to this [info in the v252 release of systemd](https://github.com/systemd/systemd/releases/tag/v252) and b) them dropping cgroupsv1 at some point as described in the "history" section of [``systemd.resouce-control(5)``](https://www.freedesktop.org/software/systemd/man/latest/systemd.resource-control.html#History). Also the [`NEWS`](https://github.com/systemd/systemd/blob/caecff5fa2ebc4ad7470aebd564cf7c1a705f184/NEWS#L7) file mentions that they plan on dropping completely with ### v258.

Considering that (likely) until 4.x gets released, very old distributions based of Debian or RHEL are likely long gone, but for 3.2 it might be worth to do some more checking.

For what I can see on Debian 12, it's pretty much a drop-in replacement, but I agree with @mcnewton that it needs proper checking. There is one other example I found from a package from Red Hat where they updated it for RHEL 9 in ``redhat-cloud-client-configuration`` ([RHEL-17009](https://issues.redhat.com/browse/RHEL-17009)).

## How distributions using the InkBridge repositories might be affected

First: InkBridge Networks builds FreeRADIUS packages for RHEL currently builds packages off the [3.0 & 3.2 releases](https://packages.inkbridgenetworks.com/#) master / 4.x are shouldn't be affected IMO. (Thank you for this service!)

**Debian-based distributions**
* The **oldest Debian** builds are for version _9 / stretch_, It shipped with systemd [**232**](https://archive.debian.org/debian/pool/main/s/systemd/) (``systemd_232*deb9*.deb``). 
No new packages have been built since 2022 [for 3.0]( http://packages.networkradius.com/freeradius-3.0/debian/stretch/pool/main/f/freeradius/) and 2023 [for 3.2](http://packages.networkradius.com/freeradius-3.2/debian/stretch/pool/main/f/freeradius/)
* The **oldest Ubuntu** builds are for version __18.04 / bionic__, it shipped with systemd [**237**](https://launchpad.net/ubuntu/bionic/+source/systemd/+changelog).
No new packages have been built since 2022 [for 3.0]( http://packages.networkradius.com/freeradius-3.0/ubuntu/bionic/pool/main/f/freeradius/) and 2023 [for 3.2](http://packages.networkradius.com/freeradius-3.2/ubuntu/bionic/pool/main/f/freeradius/)

**(RH)EL-based distributions**
* Rocky 8 is likely not  affected, it shipped with systemd 239
* The oldest RHEL-based/-derived distribution is CentOS 7, it shipped with systemd [**219**](https://archive.kernel.org/centos-vault/centos/7/updates/x86_64/Packages/). 
New packages have been built in [2025 ](http://packages.networkradius.com/freeradius-3.2/centos/7Server/) **so this one would be affected by a backport to the 3.2 branch**.

matsimon:
Should I update the to only change the behaviour for Debian?

The reasoning is, that since the initial PR things have changed as follows:
- master branch includes the modification for Redhat (with: https://github.com/FreeRADIUS/freeradius-server/commit/87a68423297b4d55ee3d47641cda8dc52fd6643e)
- v3.2.x branch includes the modification for both Redhat (with: https://github.com/FreeRADIUS/freeradius-server/commit/87a68423297b4d55ee3d47641cda8dc52fd6643e) and Debian (with: https://github.com/FreeRADIUS/freeradius-server/commit/b9d57d39584fc1616c09fca598f9baeac01ea2c3) systems

PR #5590, opened 2025-05-20
Update dictionary.adva to align with the latest product release

mwolodkiewicz:
Updated UUM-User-Level list to align with the latest release of Fiber Service Platform product.

Please notice that ADVA (Optical Networking) and Adtran have been merged into Adtran Optical Networks and ADVA does not exist anymore, but this dictionary remains valid for ADVA's Fiber Service Platform product.

Commented out lines are for deprecated UUM-User-Level values. Maybe those should be removed?

### DIFF
--- a/share/dictionary/radius/dictionary.adva
+++ b/share/dictionary/radius/dictionary.adva
@@ -28,7 +28,10 @@ VALUE	User-Level			Retrieve		0
 
 ATTRIBUTE	UUM-User-Level				102	integer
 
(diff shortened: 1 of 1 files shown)


alandekok:
I've merged it manually.

We don't delete entries from the dictionaries.  They are still needed for interaction with historic systems.

mwolodkiewicz:
> I've merged it manually.

Thank you very much.

> We don't delete entries from the dictionaries. They are still needed for interaction with historic systems.

Understood.

PR #5593, opened 2025-05-23
change spelling of pbkdf2 files from pbkfd2

gvde:
Test  files for pbkdf2 were spelled pbkfd2. Renaming them to the correct spelling and using the correct spelling in some of the tests, too.

PR #5594, opened 2025-05-23
change spelling of pbkdf2 files from pbkfd2

gvde:
Test files for pbkdf2 were spelled pbkfd2. Renaming them to the correct spelling and using the correct spelling in some of the tests, too. Backport of #5593


arr2036:
Looks like we're missing a signed commit for the first one

gvde:
> Looks like we're missing a signed commit for the first one

git can really be confusing sometimes... I squashed those two commits into one and sign it. That should fit now.

PR #5597, opened 2025-05-26
docs: buildfix-howto-reference - updated xref links and nav files to …

nolade:
…point to the correct locations.


alandekok:
Merged with some tweaks, thanks.

PR #5599, opened 2025-06-02
v3.2.x rlm_json: Option to render dates as seconds since Unix epoch

terryburton:


alandekok:
I have no comments here.  Unless there are objections, we can merge it in.

mcnewton:
LTGM; merged manually, thanks.

PR #5622, opened 2025-07-11
docs-v4: Update Concepts landing page and nav panel HIVE 4414/6

nolade:
I've gone back to spelling authorization as it's spelled in the RFCs. Also, our code uses `authorize` and it's better to keep the same spelling so to confuse our users.


doc/antora/modules/concepts/pages/aaa/authz.adoc
```diff
@@ -1,20 +1,20 @@
-= Authorisation
+= Authorization
```
alandekok, line 1: the rest of the code uses `authorize`, but the guidelines says `authorize`?  I think it's best to stay with `authorize`, as that's what's in the configuration files.

arr2036: There shouldn't be anything in v4 that uses authorize anymore?  At least not user facing?

alandekok: It's been `authorize` in the rest of the server for 25 years.

arr2036: Sure, but we don't have authorize sections, we don't have authorize methods, it's all gone...  So it doesn't really matter.  We decided to us the British/Canadian spelling elsewhere in the server, there's no reason for make an exception here.

nolade:
Updated this PR and resubmitted as https://github.com/FreeRADIUS/freeradius-server/pull/5674

PR #5650, opened 2025-09-24
xylan(dictionnary): Add new attribute (Priv-F-R3/R4/W3/W4)

alagoutte:
Hi,

From an [issue ](https://gitlab.com/wireshark/wireshark/-/merge_requests/21457)get on wireshark projet, also update the FreeRADIUS xylan Dictionnary


alandekok:
These are already in the dictionary file.

PR #5653, opened 2025-10-07
[master] rlm_pap: allow 389ds legacy PBKDF2_SHA256 to use arbitrary iteration count

gvde:
After some research into lots of existing password hashes with PBKDF2_SHA256 it seems there has been some other iteration counts than 8192 implemented in earlier 389ds versions, at least for RHEL based systems.

This pull request now accepts the iteration count given in the hash as it is, instead of assuming it must be 8192 which is the fixed setting for 389ds for years now, allowing hashes with deviant iteration count.


src/modules/rlm_pap/rlm_pap.c
```diff
@@ -912,10 +912,10 @@ static inline unlang_action_t CC_HINT(nonnull) pap_auth_pbkdf2_sha256_legacy(unl
...
 	pbkdf2_buf.iterations = ntohl(pbkdf2_buf.iterations);
 
-	if (pbkdf2_buf.iterations != PBKDF2_SHA256_LEGACY_ITERATIONS) {
-		REDEBUG("Password.With-Header {PBKDF2_SHA256} has unexpected number of iterations %d instead of %d.", pbkdf2_buf.iterations, PBKDF2_SHA256_LEGACY_ITERATIONS);
-		RETURN_UNLANG_INVALID;
-	}
+	/*
+	 *      0 iterations is invalid (we need at least one)
+	 */
+	if (pbkdf2_buf.iterations == 0) pbkdf2_buf.iterations = 1;
```
arr2036: Just return in invalid here.  A single iteration isn't really valid either.

PR #5675, opened 2025-11-06
pointer to Calling-Station-ID length field in the key is off by one

martinsta:


alandekok:
I've pushed a fix, thanks.

PR #5691, opened 2025-12-05
Fix: Resolved Incorrect Parse Return

ethan-thompson:
Corrected the return values used when applying default configuration values. Previously, when using macros such as `FR_CONF_OFFSET_IS_SET`, the parser would mark _is_set even when the value came from a default rather than the actual configuration. The behaviour is now fixed so `_is_set` reflects only whether the value was explicitly provided in the config.


arr2036:
I think these are spurious failures

PR #5696, opened 2025-12-12
Feature/open flags

ethan-thompson:
Added parameter to pass flags to `open` calls and refactored the existing calls to use the flags that were previously hard coded in the `open` calls.

### DIFF
--- a/src/lib/server/exfile.c
+++ b/src/lib/server/exfile.c
@@ -232,11 +232,11 @@ void exfile_enable_triggers(exfile_t *ef, CONF_SECTION *conf, char const *trigge
  *	Try to open the file. It it doesn't exist, try to
  *	create it's parent directories.
  */
-static int exfile_open_mkdir(exfile_t *ef, char const *filename, mode_t permissions)
+static int exfile_open_mkdir(exfile_t *ef, char const *filename, mode_t permissions, int flags)
 {
 	int fd;
 
-	fd = open(filename, O_RDWR | O_CREAT, permissions);
+	fd = open(filename, O_RDWR | O_CREAT | flags, permissions);
 	if (fd < 0) {
 		mode_t dirperm;
 		char *p, *dir;
@@ -271,7 +271,7 @@ static int exfile_open_mkdir(exfile_t *ef, char const *filename, mode_t permissi
 		}
 		talloc_free(dir);
 
-		fd = open(filename, O_RDWR | O_CREAT, permissions);
+		fd = open(filename, O_RDWR | O_CREAT | flags, permissions);
 		if (fd < 0) {
 			fr_strerror_printf("Failed to open file %s: %s", filename, fr_syserror(errno));
 			return -1;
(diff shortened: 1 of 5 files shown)


ethan-thompson:
Re-based to current master HEAD

src/lib/server/exfile.c
```diff
@@ -232,11 +232,11 @@ void exfile_enable_triggers(exfile_t *ef, CONF_SECTION *conf, char const *trigge
  *	Try to open the file. It it doesn't exist, try to
  *	create it's parent directories.
  */
-static int exfile_open_mkdir(exfile_t *ef, char const *filename, mode_t permissions)
+static int exfile_open_mkdir(exfile_t *ef, char const *filename, mode_t permissions, int inflags)
```
alandekok: I'd just call this `flags`

src/lib/server/exfile.c
```diff
@@ -232,11 +232,11 @@ void exfile_enable_triggers(exfile_t *ef, CONF_SECTION *conf, char const *trigge
...
 {
 	int fd;
 
-	fd = open(filename, O_RDWR | O_CREAT, permissions);
+	fd = open(filename, inflags, permissions);
```
alandekok: It's better to do `O_RDWR | O_CREAT | flags`

The idea is that these flags are _extra_ flags.  The exfile code still needs it's assumptions to hold true.  i.e. that the file is opened read/write, and that the file is created if it doesn't already exist.

src/modules/rlm_detail/rlm_detail.c
```diff
@@ -319,7 +319,7 @@ static unlang_action_t CC_HINT(nonnull) detail_do(unlang_result_t *p_result, mod
 
 	RDEBUG2("%s expands to %pV", env->filename_tmpl->name, &env->filename);
 
-	outfd = exfile_open(inst->ef, env->filename.vb_strvalue, inst->perm, NULL);
+	outfd = exfile_open(inst->ef, env->filename.vb_strvalue, inst->perm, O_RDWR | O_CREAT, NULL);
```
alandekok: just make this `0`, as these are extra flags.

src/modules/rlm_linelog/rlm_linelog.c
```diff
@@ -405,7 +405,7 @@ static int linelog_write(rlm_linelog_t const *inst, linelog_call_env_t const *ca
 			*p = '/';
 		}
 
-		fd = exfile_open(inst->file.ef, path, inst->file.permissions, &offset);
+		fd = exfile_open(inst->file.ef, path, inst->file.permissions, O_RDWR | O_CREAT, &offset);
```
alandekok: same here

src/modules/rlm_sql/sql.c
```diff
@@ -389,7 +389,7 @@ void rlm_sql_query_log(rlm_sql_t const *inst, char const *filename, char const *
 	size_t len;
 	bool failed = false;	/* Write the log message outside of the critical region */
 
-	fd = exfile_open(inst->ef, filename, 0640, NULL);
+	fd = exfile_open(inst->ef, filename, 0640, O_RDWR | O_CREAT, NULL);
```
alandekok: and here

src/lib/server/exfile.c
```diff
@@ -432,7 +432,7 @@ static int exfile_open_lock(exfile_t *ef, char const *filename, mode_t permissio
 		}
 
 		close(ef->entries[i].fd);
-		ef->entries[i].fd = open(filename, O_RDWR | O_CREAT, permissions);
+		ef->entries[i].fd = open(filename, inflags, permissions);
```
alandekok: `OD_RDWR | O_CREAT | flags`

alandekok:
it's also worth mashing the commits into one, otherwise one commit will fail to build, and then the other one will fix the build error.

src/lib/server/exfile.c
ethan-thompson: Revised in recent commit.

PR #5703, opened 2026-01-07
Add support for directoryName subjectAltNames

Firstyear:
This adds support to extract and provide directoryName's from the subjectAltName extension to modules.

The primary question I have during review is the changes to `share/dictionary.freeradius.internal`. There were no more ID's left in the 193x range, so I needed to use some from 1960. Is that valid and okay?


Firstyear:
Hi, it would be great to have this reviewed so that I can also contribute it to 4.x as well.

Firstyear:
Just a reminder that it'd be great to have this reviewed :)

arr2036:
The commit would need to be signed to merge, otherwise we need to pull in manually.

Firstyear:
Done, I think that should satisfy the git signing requirements.

Firstyear:
@arr2036 As a reminder, I have signed the commit per your request.

PR #5707, opened 2026-01-13
 Freeradius Documentation - Tutorials/New User

tharkadharshana:


alandekok:
Some of the commits are wrong and not needed.

* don't commit extra `..` with no purpose
* don't delete the .gitignore file
* don't edit / commit files in `raddb`.  the files are there for a reason, and the organization is needed.

doc/antora/modules/tutorials/pages/new_user.adoc
```diff
@@ -96,6 +96,13 @@ The error will usually be one of the following problems:
 (0)  ERROR: Failed decoding packet: invalid Message-Authenticator (shared secret is incorrect)
```
alandekok: This part of the output should be updated, too.

alandekok:
Also, the date on the commits is from 11 months ago?  The commits should have the correct time.

arr2036:
There should be no files added in `mods-enabled` the fact it was in .gitignore should've been a clue that was wrong.

arr2036:
In the always module, there are deprecated config items that have been removed (mpp, simulcount), and the function syntax is incorrect.  Should `%func('<arg>')`.

arr2036:
and agreeing with Alan, the tutorial contents shouldn't be pulled back into the config files.  These configs are meant to form the basis of an actual configuration, we really don't want to have users need to remove a wall of text in order to see the config items.

tharkadharshana:
I have mistakenly pushed some of the things I have done in 2025. Today I only edited "doc/antora/modules/tutorials/pages/new_user.adoc" file only.

alandekok:
@tharkadharshana Please use a new, clean branch for PRs

PR #5715, opened 2026-01-16
Consider 'c' an expected character in an IP(v6) address; output "IPv6 address" rather than "IPv 6 address"

gclinch:
Without these, debug output looks surprisingly different for addresses containing 'c':
ipv6addr = 1111:111:3c:c000::139 IPv 6 address [1111:111:3c:c000::139]
vs
ipv6addr = 1111:111:1:132::34


alandekok:
Commits should be signed, so I've added it manually and pushed a fix.

PR #5726, opened 2026-01-23
oss-fuzz: Add new fuzzer targetting base16/32/64 encode/decode

arthurscchan:
This PR introduces a new fuzzer for freeradius, targeting base16/base32/base64 encoding and decoding utilities.

Remark: This is the first in a series of PRs intended to expand fuzzing coverage for the FreeRadius project. Subsequent PRs will continue to address further coverage gaps and enhance fuzzing effectiveness, thereby improving the detection of issues across the codebase.


alandekok:
I've merged it manually based on moving the fuzzer files to `src/fuzzer`

PR #5728, opened 2026-01-25
Add TLS-Certificate.Serial-Number-And-Issuer

minfrin:
The Serial Number and Issuer is a string format defined in RFC4523 called a CertificateExactAssertion that allows a unique instance of an X509 certificate to be identified independently of the contents of the certificate.

This interoperates with the SSL_CLIENT_CERT_RFC4523_CEA variable in Apache httpd's mod_ssl.

### DIFF
--- a/src/lib/tls/attrs.h
+++ b/src/lib/tls/attrs.h
@@ -39,6 +39,7 @@ extern HIDDEN fr_dict_attr_t const *attr_tls_certificate_serial;
 extern HIDDEN fr_dict_attr_t const *attr_tls_certificate_signature;
 extern HIDDEN fr_dict_attr_t const *attr_tls_certificate_signature_algorithm;
 extern HIDDEN fr_dict_attr_t const *attr_tls_certificate_issuer;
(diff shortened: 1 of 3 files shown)


arr2036:
@ndptech did you put the DER decoder behind a toggle yet?

alandekok:
@arr2036 The DER decoder is behind `der_decode = yes` in the TLS configuration.

alandekok:
This patch isn't needed any more.  The DER code will now decode any arbitrary fields in a certificate.

We don't want to add more special-purpose code to v4.

minfrin:
> This patch isn't needed any more. The DER code will now decode any arbitrary fields in a certificate.
> 
> We don't want to add more special-purpose code to v4.

I don't follow, the Serial-Number-And-Issuer is not a field in a certificate, but rather two fields in a certificate that are escape and encoded together in a special way described in RFC4523.

Right now today, neither the serial number nor the issuer is encoded in a way compatible with RFC4523, thus the need for a dedicated encoding.

alandekok:
I'll re-open this, but the unlang in v4 allows for fairly complex manipulation of attributes.  It's possible that this kind of manipulation can be done with just normal unlang.

PR #5730, opened 2026-01-28
Tutorials/dynamic translation

98manudi:


alandekok:
I've merged most of this in with some changes.

PLEASE fix the formatting in ALL future commits:

* word wrap the text at 80 columns
* don't use `---------------------------`, use `[source,text]` or `[source,bash]` as needed
* Edit the debug output to remove ` ./scripts/bin/../../`  No one needs to see that

PR #5732, opened 2026-01-29
Feature/batched logging

ethan-thompson:
Reworked rlm_linelog to allow for batched logging. Other minor change to allow for exfile_t entries cleanup to happen faster, which is useful for testing (and required for the tests I added).

### DIFF
--- a/src/modules/rlm_linelog/file.c
+++ b/src/modules/rlm_linelog/file.c
@@ -1,0 +1,355 @@
+/*
+ * file.c
+ *
+ * Version:	$Id$
+ *
+ *   This program is free software; you can redistribute it and/or modify
+ *   it under the terms of the GNU General Public License as published by
+ *   the Free Software Foundation; either version 2 of the License, or
+ *   (at your option) any later version.
+ *
+ *   This program is distributed in the hope that it will be useful,
+ *   but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *   GNU General Public License for more details.
+ *
+ *   You should have received a copy of the GNU General Public License
+ *   along with this program; if not, write to the Free Software
+ *   Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ *
+ * @copyright 2025 Network RADIUS SAS (legal@networkradius.com)
+ */
+
+RCSID("$Id$")
+
+#include <freeradius-devel/util/iovec.h>
+
+#include <sys/stat.h>
+
+#include "file.h"
+
+static uint32_t filename_hash(void const *data)
+{
+	rlm_linelog_file_t const *file = data;
+
+	return fr_hash_case_string(file->filename);
+}
+
+static int8_t filename_cmp(void const *one, void const *two)
+{
+	rlm_linelog_file_t const *a = one;
+	rlm_linelog_file_t const *b = two;
+
+	return CMP(strcmp(a->filename, b->filename), 0);
+}
+
+unlang_action_t file_batching_mod_resume(UNUSED unlang_result_t *p_result, module_ctx_t const *mctx, request_t *request)
+{
+	rlm_linelog_file_entry_t	*entry = (rlm_linelog_file_entry_t *)mctx->rctx;
+
+	if (entry->failed) {
+		REDEBUG("Write failed with error %s", fr_syserror(entry->error));
+		return UNLANG_ACTION_FAIL;
+	}
+
+	return UNLANG_ACTION_EXECUTE_NEXT;
+}
+
+xlat_action_t file_batching_xlat_resume(TALLOC_CTX *ctx, fr_dcursor_t *out, xlat_ctx_t const *xctx,
+					   request_t *request, UNUSED fr_value_box_list_t *in)
+{
+	fr_value_box_t			*wrote;
+	rlm_linelog_file_entry_t 	*entry = (rlm_linelog_file_entry_t *)xctx->rctx;
+
+	if (entry->failed) {
+		REDEBUG("Write failed with error %s", fr_syserror(entry->error));
+		return XLAT_ACTION_FAIL;
+	}
+
+	MEM(wrote = fr_value_box_alloc(ctx, FR_TYPE_SIZE, NULL));
+	wrote->vb_size = entry->data_len;
+
+	fr_dcursor_insert(out, wrote);
+
+	return XLAT_ACTION_DONE;
+}
+
+static inline CC_HINT(always_inline) void _batching_handle_signal(NDEBUG_UNUSED request_t *request, NDEBUG_UNUSED fr_signal_t action, rlm_linelog_file_entry_t *entry)
+{
+	fr_assert(action == FR_SIGNAL_CANCEL);
+	fr_assert(entry->request == request);
+
+	entry->request = NULL;
+}
+
+void file_batching_mod_handle_signal(module_ctx_t const *mctx, request_t *request, fr_signal_t action)
+{
+	_batching_handle_signal(request, action, (rlm_linelog_file_entry_t *)mctx->rctx);
+}
+
+void file_batching_xlat_handle_signal(xlat_ctx_t const *xctx, request_t *request, fr_signal_t action)
+{
+	_batching_handle_signal(request, action, (rlm_linelog_file_entry_t *)xctx->rctx);
+}
+
+static int _file_free(rlm_linelog_file_t *file)
+{
+	fr_hash_table_delete(file->thread_inst->file_table, file);
+	talloc_free(file);
+
+	return 0;
+}
+
+static void _batching_cleanup_timer(UNUSED fr_timer_list_t *tl, UNUSED fr_time_t now, void *uctx)
+{
+	talloc_free(uctx);
+}
+
+static inline CC_HINT(always_inline) void batching_mark_entries_failed(rlm_linelog_file_t *file, rlm_linelog_file_entry_t *start_entry, int error)
+{
+	rlm_linelog_file_entry_t *entry;
+
+	for (entry = start_entry; entry < file->entry_p; entry++) {
+		entry->failed = true;
+		entry->error = error;
+
+		if (unlang_interpret_is_resumable(entry->request)) {
+			unlang_interpret_mark_runnable(entry->request);
+		}
+	}
+}
+
+static void _batch_write(rlm_linelog_file_t *file)
+{
+	int			ret, written, fd = -1;
+	bool			with_delim;
+	size_t			len = 0, header_len = 0;
+	off_t			offset, file_size;
+	int			write_error = 0;
+	struct stat		stat_buf;
+	struct iovec		to_write[3];
+	struct iovec		*to_write_p = to_write;
+	rlm_linelog_t const	*inst = talloc_get_type_abort_const(file->mod_inst, rlm_linelog_t);
+
+	fr_assert(inst->file.buffer_write);
+	fr_assert(file);
+	fr_assert(file->filename);
+
+	if (file->write) {
+		FR_TIMER_DISARM(file->write);
+	}
+
+	with_delim = (inst->delimiter_len > 0);
+
+	fd = exfile_open(inst->file.ef, file->filename, inst->file.permissions, 0, &offset);
+	if (fd < 0) {
+		PERROR("Failed to open %s", file->filename);
+	error:
+		batching_mark_entries_failed(file, file->entry, errno);
+		goto done;
+	}
+
+	if (inst->file.group_str && (chown(file->filename, -1, inst->file.group) == -1)) {
+		PERROR("Unable to change system group of \"%s\": %s", file->filename, fr_syserror(errno));
+		goto error;
+	}
+
+	/*
+	 *	If a header format is defined and we are at the beginning
+	 *	of the file then expand the format and write it out before
+	 *	writing the actual log entries.
+	 */
+	if (!fr_type_is_null(file->log_header->type) && (offset == 0)) {
+		to_write_p->iov_base = UNCONST(char *, file->log_header->vb_strvalue);
+		header_len = to_write_p->iov_len = file->log_header->vb_length;
+		to_write_p++;
+
+		if (with_delim) {
+			to_write_p->iov_base = UNCONST(char *, inst->delimiter);
+			header_len += to_write_p->iov_len = inst->delimiter_len;
+			to_write_p++;
+		}
+	}
+
+	to_write_p->iov_base = fr_dbuff_start(&file->dbuff);
+	to_write_p->iov_len = fr_dbuff_used(&file->dbuff);
+	to_write_p++;
+
+	if (fstat(fd, &stat_buf) < 0) {
+		ERROR("Failed to stat file %s: %s", file->filename, fr_syserror(errno));
+		goto error;
+	}
+	file_size = stat_buf.st_size;
+
+	errno = 0;
+	ret = fr_writev(fd, to_write, to_write_p - to_write, fr_time_delta_from_sec(0));
+	written = ret - header_len;
+	write_error = errno;
+	if (ret >= 0) {
+		if (ret < (ssize_t)(header_len)) {
+			ret = ftruncate(fd, 0);
+			if (ret < 0) {
+				PERROR("Failed truncating file \"%s\" after partial header write: %s", file->filename, fr_syserror(errno));
+			}
+
+			PERROR("Failed writing header to \"%s\"", file->filename);
+			goto error;
+		}
+	} else {
+		/*
+		 *	 Check for partial write
+		 */
+		if (fstat(fd, &stat_buf) < 0) {
+			ERROR("Failed to stat file %s: %s", file->filename, fr_syserror(errno));
+			goto error;
+		}
+
+		written = stat_buf.st_size - file_size - header_len;
+
+		if ((errno == ENOSPC) && (written > 0)) {
+			ERROR("No space left on device when writing to \"%s\". Not all data was written", file->filename);
+		}
+		else {
+			ERROR("Failed writing to \"%s\": %s", file->filename, fr_syserror(errno));
+			goto error;
+		}
+	}
+
+	if (fsync(fd) < 0) {
+		ERROR("Failed syncing \"%s\" to persistent storage: %s", file->filename, fr_syserror(errno));
+		goto error;
+	}
+
+	for (rlm_linelog_file_entry_t *entry = file->entry; entry < file->entry_p; entry++) {
+		len += entry->data_len;
+
+		if (len > (size_t)written) {
+			request_t *request = entry->request;
+			ROPTIONAL(RWARN, WARN, "Buffered log write failed. Expected %zu bytes, but only %zu bytes were written", len, (size_t)written);
+			ret = ftruncate(fd, offset + (len - entry->data_len + header_len));
+			if (ret < 0) {
+				PERROR("Failed truncating file \"%s\" after partial write: %s", file->filename, fr_syserror(errno));
+			}
+
+			batching_mark_entries_failed(file, entry, write_error);
+			break;
+		}
+
+		/*
+		* Needed because the write function could be called either from a timer or directly when the buffer is full
+		* in which case the last entry would not have been yielded
+		*/
+		if (unlang_interpret_is_resumable(entry->request)) {
+			unlang_interpret_mark_runnable(entry->request);
+		}
+	}
+
+done:
+	if (exfile_close(inst->file.ef, fd) < 0) {
+		PERROR("Failed closing file %s", file->filename);
+	}
+
+	fr_dbuff_reset_talloc(&file->dbuff);
+	file->entry_p = file->entry;
+
+	if (fr_time_delta_gt(inst->file.buffer_expiry, fr_time_delta_wrap(0))) {
+		fr_timer_in(file,file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry, false, _batching_cleanup_timer, file);
+	}
+}
+
+static void _batching_handle_timeout(UNUSED fr_timer_list_t *tl, UNUSED fr_time_t now, void *uctx)
+{
+	_batch_write((rlm_linelog_file_t *)uctx);
+}
+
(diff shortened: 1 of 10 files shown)


arr2036:
FAIL

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+
+	return fr_hash_case_string(file->filename);
+}
+
+int8_t filename_cmp(void const *one, void const *two)
+{
+	rlm_linelog_file_t const *a = one;
+	rlm_linelog_file_t const *b = two;
+
+	return strcmp(a->filename, b->filename);
```
alandekok: this should be `return CMP(strcmp(a->filename, b->filename), 0)`

the `foo_cmp()` routines return -1, 0, +1.  strcmp() returns an `int` value, which can be just about anything.

The `CMP()` macro fixes that.

See the other `foo_cmp()` functions for examples:

```
$ git grep CMP src | grep strcmp
src/lib/ldap/base.h:	int8_t uricmp = CMP(strcmp(a->uri, b->uri), 0);
src/lib/ldap/base.h:	return CMP(strcmp(a->bind_dn, b->bind_dn), 0);
...
```

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+
+	return XLAT_ACTION_DONE;
+}
+
+static inline CC_HINT(always_inline) void _batching_handle_signal(request_t *request, fr_signal_t action, rlm_linelog_file_entry_t *entry)
+{
+	fr_assert(action == FR_SIGNAL_CANCEL);
+	fr_assert(entry->request == request);
+
+	RDEBUG("Request associated with buffered log has been cancelled");
```
alandekok: there's likely no need for this.  the debug messages usually inform something that an action has been taken.  "taken no action" can just be omitted.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+
+void batching_xlat_handle_signal(xlat_ctx_t const *xctx, request_t *request, fr_signal_t action)
+{
+	_batching_handle_signal(request, action, (rlm_linelog_file_entry_t *)xctx->rctx);
+}
+
+inline CC_HINT(always_inline) void batching_cleanup(rlm_linelog_file_t *file)
+{
+	if (file->write) {
+		fr_timer_delete(&file->write);
```
alandekok: if the timer has `file` passed as the talloc ctx, then the talloc library will do all of this cleanup.  so less code.

the same goes for `file->expire`, and `file->log_header`

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+{
+	_batching_handle_signal(request, action, (rlm_linelog_file_entry_t *)mctx->rctx);
+}
+
+void batching_xlat_handle_signal(xlat_ctx_t const *xctx, request_t *request, fr_signal_t action)
+{
+	_batching_handle_signal(request, action, (rlm_linelog_file_entry_t *)xctx->rctx);
+}
+
+inline CC_HINT(always_inline) void batching_cleanup(rlm_linelog_file_t *file)
```
alandekok: this should just be `_file_free()`, and not an `inline` function.

you can then just do `talloc_free(file)` instead of running the explicit `batching_cleanup(file)` everywhere.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+		file->write = NULL;
+		file->expiry = NULL;
+		file->total_data_len = 0;
+		fr_dbuff_init_talloc(file, &file->dbuff, &file->tctx, 1024, SIZE_MAX);
+
+		if (!fr_hash_table_insert(thread->file_table, file)) {
+			RERROR("Failed tracking buffered log file %pV", call_env->filename);
+			goto error;
+		}
+	}
```
alandekok, line 304: end then do

```
talloc_destructor(file, _file_free)
```

src/modules/rlm_linelog/file.c
arr2036: should be `_file_free()` as you commented below :)

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+	file->total_data_len = 0;
+	talloc_free(file->log_header);
+
+	fr_hash_table_delete(file->thread_inst->file_table, file);
+	talloc_free(file);
+}
+
+static void _batching_cleanup_timer(UNUSED fr_timer_list_t *tl, UNUSED fr_time_t now, void *uctx)
+{
+	batching_cleanup(talloc_get_type_abort(uctx, rlm_linelog_file_t));
```
alandekok: just `talloc_free(uctx)`

etc. for the other, too

src/modules/rlm_linelog/file.c
arr2036: agreed, or make it RDEBUG4, definitely not RDEBUG

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+}
+
+xlat_action_t batching_xlat_resume(TALLOC_CTX *ctx, fr_dcursor_t *out, xlat_ctx_t const *xctx,
+					   request_t *request, UNUSED fr_value_box_list_t *in)
+{
+	fr_value_box_t			*wrote;
+	rlm_linelog_file_entry_t 	*entry = (rlm_linelog_file_entry_t *)xctx->rctx;
+
+	if (entry->failed) {
+		RDEBUG2("Write failed");
```
arr2036: REDEBUG and... reason?

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+		}
+	}
+
+	if (fsync(fd) < 0) {
+		ERROR("Failed syncing \"%s\" to persistent storage: %s", file->filename, fr_syserror(errno));
+		failed = true;
+		goto error;
+	}
+
+	// If we didn't write all data, keep the file open for truncation later
```
alandekok: nit: we use `//` at the end of lines, and `/* ... */` at the top of code blocks.  see the rest of the code for examples

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+
+	if (file->expiry) {
+		fr_timer_delete(&file->expiry);
+	}
+
+	file->total_data_len = 0;
+	talloc_free(file->log_header);
+
+	fr_hash_table_delete(file->thread_inst->file_table, file);
+	talloc_free(file);
```
arr2036, line 98: and you'd remove this if converting to a destructor

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+}
+
+static void _batching_cleanup_timer(UNUSED fr_timer_list_t *tl, UNUSED fr_time_t now, void *uctx)
+{
+	batching_cleanup(talloc_get_type_abort(uctx, rlm_linelog_file_t));
+}
+
+static inline CC_HINT(always_inline) void _batching_mark_entries_failed(rlm_linelog_file_t *file, rlm_linelog_file_entry_t *start_entry)
```
arr2036: Convention is callbacks or super private things get the underscore, not really clear why this is a separate function if it's only called once.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+		FR_TIMER_DISARM(file->write);
+	}
+
+	with_delim = (inst->delimiter_len > 0);
+
+	/* check path and eventually create subdirs */
+	p = strrchr(file->filename, '/');
+	if (p) {
+		*p = '\0';
+		if (fr_mkdir(NULL, file->filename, -1, 0700, NULL, NULL) < 0) {
```
arr2036: It'd be nice to fix this TOCTOU, if we could make fr_mkdir return the FD of the thing it opened or created.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+
+		error:
+			_batching_mark_entries_failed(file, file->entry);
+			failed = true;
+			goto done;
+		}
+		*p = '/';
+	}
+
+	fd = exfile_open(inst->file.ef, file->filename, inst->file.permissions, 0, &offset);
```
arr2036, line 144: ...and pass that FD in here and use openat

src/modules/rlm_linelog/file.c
arr2036: ...and wait, we're passing in `file->filename` to mkdir, and to `exfile_open`.... that can't be right?

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+	if (fd < 0) {
+		PERROR("Failed to open %s", file->filename);
+		goto error;
+	}
+
+	if (inst->file.group_str && (chown(file->filename, -1, inst->file.group) == -1)) {
+		PWARN("Unable to change system group of \"%s\"", file->filename);
```
arr2036: PERROR, and fail.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+
+		if (!with_delim) {
+			head_vector_len = 1;
+		} else {
+			memcpy(&head_vector_s[1].iov_base, &(inst->delimiter),sizeof(head_vector_s[1].iov_base));
+			head_vector_s[1].iov_len = inst->delimiter_len;
+			head_vector_len = 2;
+		}
+
+		if (writev(fd, &head_vector_s[0], head_vector_len) < 0) {
```
arr2036: Use the freeradius library wrapper that deals with signals and partial writes

arr2036: and why deref then pointer?  just `head_vector_s`?

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+		if (inst->file.fsync && (fsync(fd) < 0)) {
+			PERROR("Failed syncing \"%s\" to persistent storage", file->filename);
+			goto close_and_error;
+		}
+	}
+
+	fr_dbuff_set_end(&file->dbuff, fr_dbuff_current(&file->dbuff));
+	fr_dbuff_set_to_start(&file->dbuff);
+	do {
+		ret = write(fd, fr_dbuff_current(&file->dbuff), fr_dbuff_remaining(&file->dbuff));
```
arr2036: If there an issue with combining the dbuff buffer with the vector write above?

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+		if (ret > 0) {
+			fr_dbuff_advance(&file->dbuff, ret);
+		}
+	} while ((fr_dbuff_remaining(&file->dbuff) && (ret > 0)) || ((ret < 0) && ((errno == EINTR) || (errno == EAGAIN))));
+
+	if (fr_dbuff_remaining(&file->dbuff) > 0) {
+		// Check if no space left on device but some data was written (move to truncate later)
+		if ((errno == ENOSPC) && ((size_t)fr_dbuff_used(&file->dbuff) > 0)) {
+			PWARN("No space left on device when writing to \"%s\"", file->filename);
```
arr2036: Well isn't this also an error??

src/modules/rlm_linelog/file.c
arr2036: I remember we talked about this... but why are we fsyncing the header write separate from the rest?

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+
+	// If we didn't write all data, keep the file open for truncation later
+	if ((size_t)fr_dbuff_used(&file->dbuff) == file->total_data_len) {
+		if (exfile_close(inst->file.ef, fd) < 0) {
+			PERROR("Failed closing file %s", file->filename);
+		}
+	}
+
+done:
+	for (rlm_linelog_file_entry_t *entry = file->entry; entry < file->entry_p; entry++) {
```
arr2036, line 223: Against style guide to declare variables in a for loop.  @alandekok do we care about this?  I do prefer the tight scoping.

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+		/*
+		 * Needed because the write function could be called either from a timer or directly when the buffer is full
+		 * in which case the last entry would not have been yielded
+		 */
+		if (unlang_interpret_is_resumable(entry->request)) {
+			unlang_interpret_mark_runnable(entry->request);
+		}
+	}
+
+	// Reset the buffer for future use
```
arr2036: We don't use the `//` style of comments, either `/* */` or the multiline variant

src/modules/rlm_linelog/file.c
alandekok: `fr_mkdir()` does return the FD if needed, that's the first argument.

src/modules/rlm_linelog/file.h
```diff
@@ -0,0 +1,78 @@
...
+	fr_value_box_t			*log_header;		//!< Header to prepend to each log line.
+	rlm_linelog_t const		*mod_inst;		//!< Module instance this file belongs to.
+	rlm_linelog_thread_t		*thread_inst;		//!< Thread instance this file belongs to.
+	fr_dbuff_uctx_talloc_t		tctx;			//!< Talloc context for the dbuff.
+	fr_dbuff_t			dbuff;			//!< Talloced, dynamically resized dbuff containing all linelog content to write out.
+	fr_timer_t			*write;			//!< When we need to write out any pending data in the buffer.
+	fr_timer_t			*expiry;		//!< When we should cleanup file metadata due to inactivity.
+	rlm_linelog_file_entry_t	*entry_p;		//!< Last entry we wrote to.
+	rlm_linelog_file_entry_t	*entry_last;		//!< Last entry in the array.
+	size_t				total_data_len;		//!< Total data length in all entries.
```
arr2036: Isn't this known from dbuff?

src/modules/rlm_linelog/file.c
alandekok: I'm OK with this.

src/modules/rlm_linelog/file.h
```diff
@@ -0,0 +1,78 @@
...
+	request_t		*request;		//!< The request that created the data.
+	rlm_linelog_file_t	*file;			//!< The file being written to.
+	bool			failed;			//!< Write failed.
+	size_t			data_len;		//!< How much data this entry holds in the sbuff.
+} rlm_linelog_file_entry_t;
+
+typedef struct linelog_write_uctx_s linelog_write_uctx_t;
+
+struct rlm_linelog_file_s {
+	char const			*filename;		//!< Talloced filename string.
```
arr2036, line 45: Explicitly group the fields using whitespace.

src/modules/rlm_linelog/rlm_linelog.c
```diff
@@ -610,15 +548,35 @@ static xlat_action_t linelog_xlat(TALLOC_CTX *ctx, fr_dcursor_t *out,
 		vector[i].iov_len = inst->delimiter_len;
 		i++;
 	}
-	slen = linelog_write(inst, call_env, request, vector, i, with_delim);
+	if (inst->file.buffer_write) {
+		rlm_linelog_file_entry_t	*entry = NULL;
+
+		switch(batch_update(&entry, xctx->mctx, call_env, request, vector, i)) {
```
arr2036: file_enqueue_write

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,355 @@
...
+		fr_timer_in(file,file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry, false, _batching_cleanup_timer, file);
+	}
+}
+
+static void _batching_handle_timeout(UNUSED fr_timer_list_t *tl, UNUSED fr_time_t now, void *uctx)
+{
+	_batch_write((rlm_linelog_file_t *)uctx);
+}
+
+linelog_buffer_action_t batch_update(rlm_linelog_file_entry_t **entry_p, module_ctx_t const *mctx, linelog_call_env_t const *call_env, request_t *request, struct iovec *vector_p, size_t vector_len)
```
arr2036: rename to file_queue_write.  Do hard column break at 120, or just run clang-format with the rules in the repo.

src/modules/rlm_linelog/rlm_linelog.c
```diff
@@ -1070,6 +1074,42 @@ static int mod_bootstrap(module_inst_ctx_t const *mctx)
...
+{
+	rlm_linelog_t	*inst = talloc_get_type_abort(mctx->mi->data, rlm_linelog_t);
+
+	if (inst->log_dst == LINELOG_DST_FILE && inst->file.buffer_write) {
+		fr_hash_iter_t		iter;
+		rlm_linelog_file_t	*file = NULL;
+		rlm_linelog_thread_t 	*thread = talloc_get_type_abort(mctx->thread, rlm_linelog_thread_t);
+
+		for (file = fr_hash_table_iter_init(thread->file_table, &iter);
+		file;
```
arr2036: Weird indentation

src/modules/rlm_linelog/rlm_linelog.c
```diff
@@ -1070,6 +1074,42 @@ static int mod_bootstrap(module_inst_ctx_t const *mctx)
...
+		file = fr_hash_table_iter_next(thread->file_table, &iter)) {
+			batching_cleanup(file);
```
arr2036: Where's batching cleanup?  Style guide says all externally callable functions must start with the name of their source file.

arr2036: Also separation of concerns.  This all belongs in file.c with the rest of the file code, and then called from mod_thread_detach.

src/modules/rlm_linelog/rlm_linelog.c
```diff
@@ -1070,6 +1074,42 @@ static int mod_bootstrap(module_inst_ctx_t const *mctx)
...
+static int mod_thread_instantiate(module_thread_inst_ctx_t const *mctx)
+{
+	rlm_linelog_t	*inst = talloc_get_type_abort(mctx->mi->data, rlm_linelog_t);
+
+	if (inst->log_dst == LINELOG_DST_FILE && inst->file.buffer_write) {
+		rlm_linelog_thread_t *thread = talloc_get_type_abort(mctx->thread, rlm_linelog_thread_t);
+
+		thread->file_table = fr_hash_table_alloc(thread, filename_hash, filename_cmp, NULL);
+		thread->tl = mctx->el->tl;
+	}
```
arr2036, line 1085: Should be moved to file.c, and these just become dispatch functions.

src/modules/rlm_linelog/file.h
```diff
@@ -0,0 +1,78 @@
...
+
+unlang_action_t batching_mod_resume(UNUSED unlang_result_t *p_result, UNUSED module_ctx_t const *mctx, UNUSED request_t *request);
+
+xlat_action_t batching_xlat_resume(UNUSED TALLOC_CTX *ctx, UNUSED fr_dcursor_t *out, xlat_ctx_t const *xctx, request_t *request, UNUSED fr_value_box_list_t *in);
+
+void batching_mod_handle_signal(module_ctx_t const *mctx, request_t *request, fr_signal_t action);
+
+void batching_xlat_handle_signal(xlat_ctx_t const *xctx, request_t *request, fr_signal_t action);
+
+linelog_buffer_action_t batch_update(rlm_linelog_file_entry_t **entry_p, module_ctx_t const *mctx, linelog_call_env_t const *call_env, request_t *request, struct iovec *vector_p, size_t vector_len);
```
arr2036: These should all have a `file_` prefix.

src/modules/rlm_linelog/file.c
ethan-thompson: I recall us discussing this previously and determining that at this point the reason can't be determined, and previous logs would need to be used to see the reason, BUT we still wanted this message to be logged.

I'll look into this further

src/modules/rlm_linelog/file.c
ethan-thompson: This block (and the block you commented on earlier about passing in file->filename to mkdir, and to exfile_open) were taken from the existing rlm_linelog.c code (when my feature lived in the same source file). So changes to this block should also be made back there. I can look into this

src/modules/rlm_linelog/file.c
```diff
@@ -0,0 +1,356 @@
...
+		if (ret < (ssize_t)(header_len)) {
+			ret = ftruncate(fd, 0);
+			if (ret < 0) {
+				PERROR("Failed truncating file \"%s\" after partial header write: %s", file->filename, fr_syserror(errno));
+			}
+
+			PERROR("Failed writing header to \"%s\"", file->filename);
+			goto error;
+		}
+	}
```
alandekok: formatting... `} else {`

src/tests/modules/linelog/module.conf
```diff
@@ -165,6 +165,33 @@ linelog linelog_missingref {
...
+	destination = file
+	format = User-Name
+
+	file {
+		filename = "$ENV{MODULE_TEST_DIR}/test_batch.log"
+		fsync = true
+		buffer_count = 5
+		buffer_delay = 2
+	}
+
```
arr2036: extra space?

src/modules/rlm_linelog/file.c
ethan-thompson: Changes made to `exfile_open` address this

src/modules/rlm_linelog/file.h
ethan-thompson: I think this is a difference in Github's file viewer and our standards. So with Tab Size: 8, this would be aligned

alandekok:
pulled with some changes:

* code randomly did `ERROR()` vs `PERROR()`.  and used `PERROR()` while printing the actual error
* the thread detach routines aren't necessary.  The thread data will get freed on exit, and anything talloc'd from them will also get freed.  i.e. the hash table was parented from the thread data, so the hash table will get freed.  the `file` entries are parented from the hash table, so they'll get freed too
* various formatting / whitespace tweaks.

Let's see what CI says,.

alandekok:
Merged manually with tweaks.

PR #5738, opened 2026-02-06
feature: mkdir and rmdir xlats

ethan-thompson:
Added mkdir and rmdir xlats and updated docs and tests


ethan-thompson:
I'm not sure why those tests are failing, given that they wouldn't be using this feature yet

PR #5744, opened 2026-02-14
Batched Linelog Coverity Fixes

ethan-thompson:
Resolved a few issues where returns were not being checked or handled properly

### DIFF
--- a/src/modules/rlm_linelog/file.c
+++ b/src/modules/rlm_linelog/file.c
@@ -249,7 +249,7 @@ static void _batch_write(rlm_linelog_file_t *file)
 	}
 
 done:
-	if (exfile_close(inst->file.ef, fd) < 0) {
+	if ((fd >= 0) && (exfile_close(inst->file.ef, fd) < 0)) {
 		PERROR("Failed closing file %s", file->filename);
 	}
 
@@ -257,8 +257,10 @@ done:
 	file->entry_p = file->entry;
 
 	if (fr_time_delta_gt(inst->file.buffer_expiry, fr_time_delta_wrap(0))) {
-		fr_timer_in(file, file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry,
-			    false, _batching_cleanup_timer, file);
+		if (fr_timer_in(file, file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry,
+			    false, _batching_cleanup_timer, file) < 0) {
+			FATAL("Failed to insert timer for buffered log file cleanup for %pV", file->filename);
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


src/modules/rlm_linelog/file.c
```diff
@@ -290,14 +290,20 @@ linelog_buffer_action_t file_enqueue_write(rlm_linelog_file_entry_t **entry_p, m
 
 		file->filename = talloc_strdup(file, path);
 		file->log_header = fr_value_box_alloc_null(file);
```
arr2036: `MEM(file->log_header = fr_value_box_alloc_null(file))`

src/modules/rlm_linelog/file.c
```diff
@@ -249,16 +249,16 @@ static void _batch_write(rlm_linelog_file_t *file)
 	}
 
 done:
-	if (exfile_close(inst->file.ef, fd) < 0) {
+	if ((fd >= 0) && (exfile_close(inst->file.ef, fd) < 0)) {
```
arr2036, line 252: What was the exact issue here?

src/modules/rlm_linelog/file.c
```diff
@@ -249,16 +249,16 @@ static void _batch_write(rlm_linelog_file_t *file)
...
 	}
 
 	fr_dbuff_reset_talloc(&file->dbuff);
 	file->entry_p = file->entry;
 
 	if (fr_time_delta_gt(inst->file.buffer_expiry, fr_time_delta_wrap(0))) {
-		fr_timer_in(file, file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry,
-			    false, _batching_cleanup_timer, file);
+		MEM(fr_timer_in(file, file->thread_inst->tl, &file->expiry, inst->file.buffer_expiry,
+			    false, _batching_cleanup_timer, file));
```
arr2036: Switch to `FATAL()` when merged, and it should be `== 0`

src/modules/rlm_linelog/file.c
```diff
@@ -337,8 +343,12 @@ linelog_buffer_action_t file_enqueue_write(rlm_linelog_file_entry_t **entry_p, m
 	*entry_p = file->entry_p;
 
 	if (fr_time_delta_gt(inst->file.buffer_delay, fr_time_delta_wrap(0)) && !fr_timer_armed(file->write) ) {
-		fr_timer_in(file, file->thread_inst->tl, &file->write, inst->file.buffer_delay,
-			    false, _batching_handle_timeout, file);
+		if (fr_timer_in(file, file->thread_inst->tl, &file->write, inst->file.buffer_delay,
```
arr2036, line 342: You can probably just use FATAL here now as well

src/modules/rlm_linelog/file.c
ethan-thompson: We can jump to here after we fail to open the file with exfile_open, which means the fd < 0, which exfile_close does not like

alandekok:
Addressed by Nicks recent pushes.

PR #5758, opened 2026-03-02
Initial baseline multi-server tests for load-gen -> 5 homeserver, and load-gen -> 1 proxy -> 2 home server

marc-casavant:
This is the initial baseline multi-server tests for load-gen -> 5 homeserver, and load-gen -> 1 proxy -> 2 home server.


alandekok:
Pulled in with some tweaks.

PR #5759, opened 2026-03-04
Fix/dhcpv6 fixups

ethan-thompson:
Various DHCPv6 related fixes

### DIFF
--- a/src/protocols/dhcpv6/decode.c
+++ b/src/protocols/dhcpv6/decode.c
@@ -366,24 +366,39 @@ static ssize_t decode_option(TALLOC_CTX *ctx, fr_pair_list_t *out,
 		slen = fr_pair_array_from_network(ctx, out, da, data + 4, len, decode_ctx, decode_value);
 
 	} else if (da->type == FR_TYPE_VSA) {
-		bool append = false;
 		fr_pair_t *vp;
 
 		vp = fr_pair_find_by_da(out, NULL, da);
 		if (!vp) {
+			fr_pair_list_t staging_list;
+
 			vp = fr_pair_afrom_da(ctx, da);
 			if (!vp) return PAIR_DECODE_FATAL_ERROR;
 			PAIR_ALLOCED(vp);
 
-			append = true;
-		}
-
-		slen = decode_vsa(vp, &vp->vp_group, da, data + 4, len, decode_ctx);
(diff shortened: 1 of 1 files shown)


src/protocols/dhcpv6/base.c
```diff
@@ -37,6 +37,7 @@
 
 static uint32_t instance_count = 0;
 static bool	instantiated = false;
+static pthread_mutex_t init_mutex = PTHREAD_MUTEX_INITIALIZER;
```
arr2036, line 40: What was the justification for from claude for needing to make this thread safe?  I don't think this is needed.

arr2036: Alright, it didn't reference any call sites which actually needed this.  Remove this commit and add a `@note` saying that init and deinit are only called by the main thred and do not need to be thread-safe.  Ideally you'd add that to all init functions of the same type.

src/protocols/dhcpv6/decode.c
```diff
@@ -366,24 +366,39 @@ static ssize_t decode_option(TALLOC_CTX *ctx, fr_pair_list_t *out,
...
+			fr_pair_list_t staging_list;
+
 			vp = fr_pair_afrom_da(ctx, da);
 			if (!vp) return PAIR_DECODE_FATAL_ERROR;
 			PAIR_ALLOCED(vp);
 
-			append = true;
-		}
-
-		slen = decode_vsa(vp, &vp->vp_group, da, data + 4, len, decode_ctx);
```
arr2036, line 381: I think the correct thing here, is to do the same as we do in the raw label, but output the raw attribute within the VSA vp we already allocated, NOT free the vp.  @alandekok  @ndptech  do you agree?

alandekok: there's a lot of similar issues elsewhere.  I'll take a look.

alandekok:
I've grabbed some of the commit manually, and dropped others.

I've rewritten the "raw VSA" one, and hoisted the code out of the `decode_vsa()` function.  It turns out to do it slightly better, we need to track _both_ the VSA wrapper, and the vendor wrapper.  I've also added a test.

It still has to do a better job of decoding unknown / invalid VSAs, but that can wait for a future day.

PR #5764, opened 2026-03-07
expand scope of `suppress_secrets`

jimdigriz:
This PR includes:
 * defaulting `suppress_secrets` to on
 * add a number of other attributes as secret
     * mostly anything internal or IETF with 'Password' in the name; no vendor specific attributes touched though
     * decided that even the hashed passwords should be tagged as 'secret' too
 * introduce a macro to place in one location the logic for "should this be displayed"
 * add support to `xlat debug_attr`, `unlang`, `rlm_python3` and `rlm_python`; upstream you currently only support `rlm_perl`

I have tested this on `xlat debug_attr`, `unlang` `rlm_perl`, and `rlm_python3` (not `rlm_python`).

**N.B.** someone should check if the change to `pair.c` makes sense or if it should be dropped, not 100% confident about that.

# Smoke tests
## `xlat debug_attr`
```unlang
%{debug_attr:control:}"
```
outputs:
```
(0)     Attributes matching "control:"
(0)       &control:Cleartext-Password := <<< secret >>>
```

## Unlang
```unlang
update control { 
        Cleartext-Password := "%{control:Cleartext-Password}moo"
}
```
outputs:
```
(0)     update control {
(0)       EXPAND %{control:Cleartext-Password}moo
(0)          --> steakmoo
(0)       Cleartext-Password := <<< secret >>>
(0)     } # update control = noop
```

## python3
outputs:
```
authorize - 'config:Cleartext-Password' := '<<< secret >>>'
```

## perl
outputs (unchanged):
```
(0) perl: &control:Cleartext-Password = $RAD_CHECK{'Cleartext-Password'} -> '<<< secret >>>'
```

### DIFF
--- a/src/include/libradius.h
+++ b/src/include/libradius.h
@@ -207,6 +207,8 @@ typedef struct attr_flags {
 #define FLAG_ENCRYPT_TUNNEL_PASSWORD (2)
 #define FLAG_ENCRYPT_ASCEND_SECRET   (3)
 
+#define ATTRIBUTE_SECRET(_vp, _x) ((_vp->da->flags.secret && request->root->suppress_secrets && (rad_debug_lvl < 3)) ? "<<< secret >>>" : _x)
+
 extern const FR_NAME_NUMBER dict_attr_types[];
 extern const size_t dict_attr_sizes[PW_TYPE_MAX][2];
 extern const int fr_attr_max_tlv;
--- a/src/main/map.c
+++ b/src/main/map.c
@@ -1732,16 +1732,16 @@ void map_debug_log(REQUEST *request, vp_map_t const *map, VALUE_PAIR const *vp)
 	switch (map->lhs->type) {
 	case TMPL_TYPE_LIST:
 		RDEBUG("%.*s:%s %s %s", (int)map->lhs->len, map->lhs->name, vp ? vp->da->name : "",
-		       fr_int2str(fr_tokens, vp ? vp->op : map->op, "<INVALID>"), value);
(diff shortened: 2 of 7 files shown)


alandekok:
It's 2026 and we should use good practices.  But...  I'm less sure that we should enable `suppress_secrets` by default.  

I'm not looking forward to people asking questions about "why can't I see the passwords".

On a technical note, it's likely not necessary to suppress printing of salted / crypt'd / hashed passwords.  But it's not terrible to do so.

jimdigriz:
> But... I'm less sure that we should enable suppress_secrets by default.

When asking people to provide the output of `-X`, many will do so blindly and inadvertently harm themselves.

I have had this happen more than a few times on my OAuth2 rlm_perl project where they production passwords are dropped into a GitHub issue.

You probably can provide plenty of examples of this including those seen on the freeradius-users mailing list.

> I'm not looking forward to people asking questions about "why can't I see the passwords".

Instead of `<<< secret >>>` maybe we could use `<<< hidden by suppress_secrets config >>`?

I can also patch `rlm_pap` (and others) so that if `suppress_secrets` is set, the debug level is less than three and the password *mis-matches*, to print a warning "to see the actual suppressed password above, set `suppress_secrets` to `no`".

Sound like a workable option?

> On a technical note, it's likely not necessary to suppress printing of salted / crypt'd / hashed passwords. But it's not terrible to do so.

My guiding principal was "would you be happy with this up on Pastebin", if not, probably should be hidden.

Not sure anyone has the cojones to upload their `/etc/shadow` to Pastebin.

alandekok:
I've merged the commits individually, and added some additional debug messages to tell the user what's going on.

PR #5765, opened 2026-03-11
rlm_rest: support UNIX (including abstract) sockets

jimdigriz:
Allow for `rlm_rest` to connect to HTTP servers over a UNIX socket meaning you can sidestep explaining to auditors about non-TLS HTTP servers pointing to `localhost`[1], plus it can be faster to use a UNIX socket than TCP socket.

Tested with `socat` (both `UNIX-LISTEN` and `ABSTRACT-LISTEN`) proxying to `python3 -m http.server`:
```shell
$ socat UNIX-LISTEN:/tmp/test.sock,fork TCP:192.0.2.1:8000
```

```shell
$ python3 -m http.server
Serving HTTP on 0.0.0.0 port 8000 (http://0.0.0.0:8000/) ...
192.0.2.232 - - [11/Mar/2026 07:18:29] code 404, message File not found
192.0.2.232 - - [11/Mar/2026 07:18:29] "GET /user/bob/mac/?action=authorize HTTP/1.1" 404 -
```

Notes:
 * no testing for `NUL` in `connect_uri_socket` as `PW_TYPE_STRING` covers us for this
 * xlat not supported, as a bad actor against a bad config might be able to talk to other daemons
 * [curl ignores the configuration option](https://curl.se/libcurl/c/CURLOPT_UNIX_SOCKET_PATH.html) when `connect_uri_socket` is `NULL`
 * check for the length limit (effectively `UNIX_PATH_MAX`)
    * if we do not check it, we get back from curl an unclear error and the server refuses to start, so we check to bubble up a friendly actionable message
    * [curl provides an example as a way around this limit (for Linux)](https://curl.se/libcurl/c/CURLOPT_UNIX_SOCKET_PATH.html) but `rlm_smsotp` does not do this and sysadmins for now can just create a softlink if they need a workaround

[1] of course to run `tcpdump` you need privileges in a similar way to use `ptrace()` but with this we can be technically correct in stating "packet sniffing is not possible".

### DIFF
--- a/raddb/mods-available/rest
+++ b/raddb/mods-available/rest
@@ -50,6 +50,21 @@ rest {
 	connect_uri = "http://127.0.0.1"
 
 	#
+	# When set, connects to the server over a UNIX socket which may be
+	# helpful for those with (pseudo)security constraints that make it
+	# easier to use an UNIX socket than explain to an auditor the use of
+	# non-TLS HTTP localhost connections.
+	#
+	# NOTE: This is not allowed to be an xlat as it can lead to security issues.
+	#
+#	connect_uri_socket = "/run/myprogram.sock"
+
+	#
+	# When set, connect_uri_socket is an abstract UNIX socket
+	#
+#	connect_uri_socket_abstract = "no"
+
+	#
 	#  How long before new connection attempts timeout, defaults to 4.0 seconds.
 	#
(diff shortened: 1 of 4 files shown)


alandekok:
merged, thanks.

PR #5766, opened 2026-03-11
listen: resolve status sockets leak connection limits

jimdigriz:
Whilst using the virtual server 'aws-nlb', you very quickly hit:
```
Feb 16 22:19:41 service0 freeradius[31737]: Ignoring new connection from client localhost due to client max_connections (16)
Feb 16 22:19:41 service0 freeradius[31737]: Waking up in 28.8 seconds.
Feb 16 22:19:46 service0 freeradius[31737]: Ignoring new connection from client localhost due to client max_connections (16)
Feb 16 22:19:46 service0 freeradius[31737]: Waking up in 23.8 seconds.
```

This is just by using the simple client:
```
echo | nc -w0 192.0.2.232 8000
```

The issue is we are not calling `common_socket_free` for status sockets so when the client disconnects, we hit the `max_connections` limit very quickly.


alandekok:
merged, thanks

PR #5767, opened 2026-03-11
rlm_cache: fix documentation error

### LINKED ISSUE #5627: [defect]: rlm_cache return code mismatch with Cache-Read-Only and Cache-Status-Only in 3.2.X
### What type of defect/bug is this?

Unexpected behaviour (obvious or verified by project member)

### How can the issue be reproduced?

	1.	Define a cache module using rlm_cache_rbtree with a standard update block to populate session-state attributes (e.g., Custom OpenRoaming flags).
	2.	Use Cache-Read-Only := "yes" in the control list before calling the cache module.
	3.	Observe the module behavior using debug logs (radiusd -X).
	4.	Despite the cache returning data and merging into the request (as confirmed in the logs), the module returns ok instead of updated.

Sample Unlang:
```
update control {
    &Cache-Read-Only := "yes"
}
cache_realm_flags_local

if (!updated) {
    # This executes even though merge happened
    -realm_gate_flags
}
```

⸻

Log output from the FreeRADIUS daemon
```
cache_realm_flags_local: Merging cache entry into request
[cache_realm_flags_local] = ok
if (!updated) -> TRUE
```
Expected: updated
Actual: ok

⸻

Additional context

The [module documentation](https://github.com/FreeRADIUS/freeradius-server/blob/v3.2.x/raddb/mods-available/cache) in mods-available/cache explicitly says:

If Cache-Read-Only is set:
	•	Return updated if a cache entry was found
	•	Return notfound if no entry was found

This behavior is not matching what’s implemented (at least in 3.2.7), where merging occurs but return code is still ok.

This might be:
	•	A bug (i.e., the module should return updated), or
	•	A docs issue (if ok is now the intended behavior)

We are reporting it as a defect unless confirmed otherwise.

✅ Tested in 3.2.7

### Log output from the FreeRADIUS daemon

```shell
(437) cache_realm_flags_local: EXPAND %{NAS-Identifier}
(437) cache_realm_flags_local:    --> [REDACTED_NASID]
(437) cache_realm_flags_local: Mutex acquired
(437) cache_realm_flags_local: Found entry for "[REDACTED_NASID]"
(437) cache_realm_flags_local: Merging cache entry into request
(437) cache_realm_flags_local: Mutex released
(437)         modsingle[authorize]: returned from cache_realm_flags_local (rlm_cache)
(437)         [cache_realm_flags_local] = ok
(437)         if (!updated) {
(437)         if (!updated)  -> TRUE
```

### Relevant log output from client utilities

N/A

### Backtrace from LLDB or GDB

```shell
N/A
```

jimdigriz:
Reviewing the code, the return code is a statement about the status as well as any changes being made to the cache but not to the request.

For this 'ok' makes more sense than 'updated' as nothing was updated. We also avoid the perils of backwards compatibility if we were to change it. So this is just a documentation fix.

Resolves #5627

### DIFF
--- a/raddb/mods-available/cache
+++ b/raddb/mods-available/cache
@@ -20,6 +20,10 @@
 #	The module returns "noop" if it did nothing.
 #	The module returns "fail" on error.
 #
+#	NOTE: return codes are statements about the status as well as
+#	      any changes being made to the cache. They are not about
+#	      if any changes were made to the attributes of the request.
+#
 cache {
 	#  The backend datastore used to store the cache entries.
 	#  Current datastores are
@@ -137,8 +141,8 @@ cache {
 	#  &control:Cache-Read-Only - If present and set to 'yes' will
 	#  prevent a new entry from being created, but will allow existing
 	#  entries to be merged.  It will also alter the module's return codes.
-	#    - The module will return "updated" if a cache entry was found.
-	#    - The module will return "notfound" if no cache was found.
+	#    - The module will return "ok" if a cache entry was found.
+	#    - The module will return "notfound" if no cache entry was found.
 	#
 	#  &control:Cache-Merge - If present and set to 'yes' will merge new
 	#  cache entries into the current request. Useful if results


alandekok:
merged thanks (gpg signing issues)

PR #5768, opened 2026-03-13
OSS-Fuzz: Add fuzzer for json/jpath

arthurscchan:
This PR creates a new fuzzer, targeting the json and jpath parsing logic.


alandekok:
I'll merge, thanks.

PR #5773, opened 2026-03-20
Fix/der bugs

ethan-thompson:
A few bugs found in the DER protocol code

### DIFF
--- a/src/protocols/der/encode.c
+++ b/src/protocols/der/encode.c
@@ -590,6 +590,7 @@ static ssize_t fr_der_encode_oid_from_value(fr_dbuff_t *dbuff, uint64_t value, u
 {
 	fr_dbuff_t	our_dbuff;
 	int		i;
+	bool	wrote = false;
 	uint64_t	oid;
 
 	/*
@@ -633,11 +634,12 @@ static ssize_t fr_der_encode_oid_from_value(fr_dbuff_t *dbuff, uint64_t value, u
 		uint8_t more, part;
 
 		part = (oid >> i) & 0x7f;
-		if (!part) continue;
+		if (!part && !wrote && i > 0) continue;
 
 		more = ((uint8_t) (i > 0)) << 7;
 
 		FR_DBUFF_IN_RETURN(&our_dbuff, (uint8_t) (more | part));
+		wrote = true;
 	}
 
 	(*count)++;


src/protocols/der/encode.c
```diff
@@ -751,11 +751,8 @@ static int CC_HINT(nonnull) fr_der_encode_set_of_cmp(void const *one, void const
 	fr_der_encode_set_of_ptr_pairs_t const *a = one;
 	fr_der_encode_set_of_ptr_pairs_t const *b = two;
 
-	if (a->len >= b->len) {
-		return memcmp(a->data, b->data, a->len);
-	}
-
-	return memcmp(a->data, b->data, b->len);
+	size_t min_len = (a->len < b->len) ? a->len : b->len;
```
arr2036, line 754: Uh, yeah.   Just pass it directly to memcmp

src/protocols/der/encode.c
```diff
@@ -633,11 +634,12 @@ static ssize_t fr_der_encode_oid_from_value(fr_dbuff_t *dbuff, uint64_t value, u
 		uint8_t more, part;
 
 		part = (oid >> i) & 0x7f;
-		if (!part) continue;
+		if (!part && !wrote && i > 0) continue;
 
 		more = ((uint8_t) (i > 0)) << 7;
 
 		FR_DBUFF_IN_RETURN(&our_dbuff, (uint8_t) (more | part));
+		wrote = true;
```
arr2036, line 642: More explanation about why this is an issue?

alandekok:
There's a MEMCMP_FIELDS macro which takes care of memcmp and different length fieldsSent from my iPhoneOn Mar 20, 2026, at 5:46 PM, Arran Cudbard-Bell ***@***.***> wrote:﻿
@arr2036 commented on this pull request.



In src/protocols/der/encode.c:
> @@ -751,11 +751,8 @@ static int CC_HINT(nonnull) fr_der_encode_set_of_cmp(void const *one, void const
 	fr_der_encode_set_of_ptr_pairs_t const *a = one;
 	fr_der_encode_set_of_ptr_pairs_t const *b = two;
 
-	if (a->len >= b->len) {
-		return memcmp(a->data, b->data, a->len);
-	}
-
-	return memcmp(a->data, b->data, b->len);
+	size_t min_len = (a->len < b->len) ? a->len : b->len;

Uh, yeah.   Just pass it directly to memcmp

arr2036:
much better :)

src/protocols/der/encode.c
alandekok: The previous code skipped `0x00` in the middle of a number :(

alandekok:
merged, thanks

PR #5774, opened 2026-03-22
debian: parallel build

jimdigriz:
Lets make those Debian/Ubuntu builds a little bit faster by actually actioning on 'parallel'.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -50,6 +50,10 @@ else
 	CFLAGS += -O2
 endif
 
+ifneq (,$(filter parallel=%,$(DEB_BUILD_OPTIONS)))
+	MAKEFLAGS += -j$(patsubst parallel=%,%,$(filter parallel=%,$(DEB_BUILD_OPTIONS)))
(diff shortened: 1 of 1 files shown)


debian/rules
```diff
@@ -50,6 +50,11 @@ else
 	CFLAGS += -O2
 endif
 
+ifneq (,$(filter parallel=%,$(DEB_BUILD_OPTIONS)))
+	NUMJOBS = $(patsubst parallel=%,%,$(filter parallel=%,$(DEB_BUILD_OPTIONS)))
```
arr2036: Does `NUMJOBS` do anything special?  Made a very similar fix in master.  https://github.com/FreeRADIUS/freeradius-server/blob/master/debian/rules

jimdigriz: Nope, it can be folded down, I just kept it for readabilty?

Want me to flatten it out?

arr2036: Yes, thanks.

PR #5778, opened 2026-03-24
Include automask as a requirement of build.h

Firstyear:
build.h requires automask.h but it is not installed which prevents building 3rd modules against a standalone freeradius instance.


alandekok:
I've pulled it in manually, thanks.  We require commits to be signed.

PR #5795, opened 2026-03-30
docs-v4: transaction integrity check - added more info on keywords th…

nolade:
…at can be used, fix xref links, added rollback conditions (`notfound` / `timeout`)


alandekok:
the text is wrong:

> xref:unlang/group.adoc[`group`]`switch`, xref:unlang/group.adoc[`group`]`case`,

I'll tweak and push it.

PR #5802, opened 2026-03-31
docs-v4: continue integrity check - added more details about breaking…

nolade:
… out of a `foreach` loop


alandekok:
merged with fixes.  I've removed a duplicate paragraph.

PR #5810, opened 2026-04-01
docs-v4: added exec and xlat expansions can be used as a <key>

nolade:


alandekok:
We're deprecating the use of back-ticks / exec in v4.  So let's not add more references to it.

People can use `%exec(...)` instead.  It's just another xlat, and it has a number of better properties.

PR #5822, opened 2026-04-07
docs-v4: Intergrity check on condition.adoc - fix syntax, added truth…

nolade:
… table, wordsmithing


alandekok:
Merged, thanks

PR #5828, opened 2026-04-19
rlm_dpsk: generalize DPSK adapters and sources

hshimomura:
This PR reworks `rlm_dpsk` around configured adapters and sources so the module
can support multiple vendor-specific RADIUS request and reply layouts through
configuration instead of embedding those layouts directly in code.

The motivation is that "Dynamic PSK" in this space is not an IEEE or IETF
protocol term in the same sense as EAP, RADIUS, or WPA. The underlying
cryptographic inputs are standard WPA-style materials, but the way those inputs
are carried in RADIUS is vendor-specific.

In practice:

- RUCKUS DPSK is vendor technology rather than a generic standards term.
- Cisco IOS XE EasyPSK and Meraki Easy PSK solve a similar operational problem
  with different vendor attributes.
- Upstream code is easier to maintain if those attribute mappings live in
  configuration, while the module core handles generic parsing, lookup,
  verification, and reply generation.

### 1. Config-driven adapter model

- Adds configured `adapter` blocks for request and reply layouts:
  - `standard_attrs`
  - `key_value_vsa`
  - `named_vsa_attrs`
- Extends the existing `mods-available/dpsk` configuration so additional
  adapter definitions for IOS XE, RUCKUS, and Meraki can live inside the
  existing `rlm_dpsk` module configuration.
- Request detection, request extraction, and reply generation are now driven by
  adapter policy instead of hardcoded vendor branches.
- Vendor names were reduced in the code paths so runtime logic is organized
  around generic adapter types.
- The `standard` adapter remains as a generic reference example for custom
  integrations, while the shipped active adapters target current vendor
  implementations.
- The shipped RUCKUS adapter uses dictionary-defined child TLVs directly via
  `named_vsa_attrs`.

### 2. Config-driven source model

- Adds configured `source` blocks for PSK / PMK lookup backends.
- Supports CSV-backed sources, with direct attribute-based PSK / PMK mapping
  handled by adapter request policy.
- CSV filename resolution now uses configured source data instead of a
  hardcoded path.
- Source names are parsed from subsection names and are available to policy
  logic.

### 3. Configurable reply policies

- Reply generation is now explicit and adapter-driven.
- Supported reply modes include:
  - standard attributes
  - Tunnel-Password
  - Cisco AVPair hex PMK
  - MS-MPPE-Recv-Key
- Explicit `reply.mode` is now honored instead of being overwritten by adapter
  defaults.

### 4. VLAN reply support

- Adds configurable VLAN reply handling:
  - `Tunnel-Type`
  - `Tunnel-Medium-Type`
  - `Tunnel-Private-Group-Id`
- VLAN attributes can be emitted together with the selected PSK / PMK reply
  mode.
- This is an optional extension to the existing CSV-backed flow, not a
  requirement for existing `psk.csv` deployments.

### 5. PMK cache and source handling fixes

- File-backed entries are cached by `(filename, SSID)` instead of filename
  alone.
- This avoids cross-SSID PMK reuse when the same CSV file is used for multiple
  WLANs.
- Fixes source-name handling so configured source selection works as intended.
- Fixes CSV source resolution so configured filenames are used correctly.

### 6. Validation, safety, and hot-path cleanup

- Adds validation for request and reply policy completeness.
- Rejects unresolved configured attributes at instantiation time.
- Fixes cache lookup and file-entry lifetime bugs in the file-backed lookup
  path.
- Precompiles configured attribute templates so the hot path avoids repeated
  per-request parsing.
- Rejects octal escape sequences which overflow `UINT8_MAX`.
- Fixes file-backed lookup lifetime issues by copying transient strings before
  backing buffers can be freed.
- Tightens cache entry locking and uses defensive copies on cache hits to avoid
  lifetime issues during concurrent lookups.
- Removes packed blob adapter support now that the shipped and verified vendor
  paths are covered by `standard_attrs`, `key_value_vsa`, and
  `named_vsa_attrs`.

## Why this shape is useful upstream

- It reduces vendor-specific branching in `rlm_dpsk.c`.
- It makes future EasyPSK / DPSK-style integrations primarily an adapter /
  source configuration problem instead of a core-code problem.
- It lets deployments adapt vendor-specific attribute naming and reply layout
  changes in configuration, without changing module C code.
- In practice that shifts many EasyPSK / DPSK integration changes from
  rebuild-and-redeploy work to configuration update and restart.
- It keeps the module focused on generic behavior:
  - normalize request inputs
  - resolve PSK / PMK data
  - verify handshake MICs
  - emit configured reply attributes

## Compatibility notes

- The shipped RUCKUS adapter uses named child TLVs from
  `Vendor-Specific.Ruckus.DPSK-Params`.
- Cisco IOS XE and Meraki support are expressed through module configuration
  and README examples, not hardcoded vendor logic.
- Existing CSV-backed deployments remain supported, with VLAN handling added as
  an optional extension.

## Testing performed

- `make -j4` completed successfully.
- `freeradius -XC` passes with the new config model.
- Live authentication was verified for:
  - Cisco IOS XE EasyPSK
  - Meraki Easy PSK
  - RUCKUS DPSK

## Relevant references

- RUCKUS Dynamic PSK documentation:
  https://docs.ruckuswireless.com/smartzone/5.0/sz300-vszh-scg200-administrator-guide/GUID-DCEDE953-2206-4CF3-87E5-ADEA80F526D2.html
- RUCKUS Dynamic PSK overview:
  https://www.ruckusnetworks.com/technologies/RUCKUS-Security-Innovations/Dynamic-Pre-Shared-Key-PSK/
- Cisco IOS XE Easy PSK deployment guide:
  https://www.cisco.com/c/en/us/td/docs/wireless/controller/9800/technical-reference/easy-psk-deployment-guide.html
- Cisco Meraki documentation for RADIUS-backed PSK deployments:
  https://documentation.meraki.com/Wireless/Translated_Documents/IPSK_with_RADIUS_Authentication_jp

### DIFF
--- a/raddb/mods-available/dpsk
+++ b/raddb/mods-available/dpsk
@@ -7,55 +7,45 @@
 #
 #  = DPSK authentication Module
 #
-#  Calculate dynamic PSKs
+#  Calculate dynamic PSKs.
 #
-#  This module needs the following attributes as input:
+#  This module verifies DPSK / EasyPSK style requests by:
 #
-#	* control.Pre-Shared-Key -  the PSK for the user
-#	* User-Name - the supplicant MAC in hex format, e.g. "abcdef012345"
-#	* Called-Station-MAC - the AP MAC in binary
-#	  this attribute is set by the "rewrite_called_station_id" policy.
-#	* FreeRADIUS-802.1X-Anonce - from the AP
-#	* FreeRADIUS-802.1X-EAPoL-Key-Msg - from the AP
+#	* normalizing request attributes into generic inputs
+#	* resolving PSK / PMK material from configured sources
+#	* verifying the EAPOL-Key MIC
+#	* emitting configured reply attributes
 #
-#  The attributes where data is found can be amened using configuration options
-#  below, e.g. if you need to use different VSAs for the anonce and EAPoL key msg.
+#  The underlying PSK / PMK calculation is generic, but different NAS
+#  implementations send the request and expect the reply in different
+#  vendor-specific attributes.  This module therefore separates:
 #
-#  Note that you MUST run the "rewrite_called_station_id" policy before calling this module.
+#	* adapter policies - where request and reply attributes live
+#	* source policies  - where file-backed PSKs are found
 #
-#  That policy MUST also create the Called-Station-MAC attribute.
+#  The "standard" adapter uses generic FreeRADIUS attributes and serves as a
+#  reference example.  The additional adapters below show vendor-specific
+#  request / reply layouts used in practice.
 #
-#  Then place the following configuration into the "recv Access-Request" section:
+#  Note that you MUST run the "rewrite_called_station_id" policy before calling
+#  this module.
+#
+#  That policy SHOULD create Called-Station-SSID.  If you use the optional
+#  "standard" adapter example below, it may also create Called-Station-MAC.
+#
+#  Then place the following configuration into the "recv Access-Request"
+#  section:
 #
 #	recv Access-Request {
 #		...
 #		rewrite_called_station_id
-#
 #		dpsk
-#		if (updated) {
-#			control.PSK-Identity := "bob"
-#			control.Pre-Shared-Key := "this-is-super-secret"
-#		}
-#
 #	}
 #
-#  Alternatively, you can read control.PSK-Identity and
-#  control.Pre-Shared-Key from a database.  Just do that before
-#  running the "authenticate" section.  You can do the database lookup
-#  after running the "dpsk" module in the "recv Access-Request" section.
-#
-#  The database should look up the User-Name (i.e. MAC), and then set
-#  control,PSK-Identity and control.Pre-Shared-Key.  In general, it
-#  is easiest to just set PSK-Identity to be same as the User-Name.
-#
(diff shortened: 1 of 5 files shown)


hshimomura:
Thank you for the comment. I understand the architecture concern you raised.

Rather than keeping vendor-specific request/reply mapping inside the module via
an `adapter` model, I am going to rethink this in a direction where that logic
lives in `unlang`, the `users` file, or `FR_TYPE_TMPL`.

For the next step, I plan to reorganize the work in the following way:

- drop the adapter model
- leave SHA256 / PMF out of the first step and focus on the SHA1 path only because only Meraki support it as optional
- use RUCKUS and Meraki as policy-language-based configuration examples
- keep any `rlm_dpsk` changes vendor-neutral and minimal

The first thing I want to verify is whether the existing upstream v3.2
`rlm_dpsk` inputs and outputs are already sufficient for policy-side reply
generation.

Based on that, the next proposal will either focus on configuration examples
and documentation, or on a much smaller change with only the minimum generic
module-side changes required.

I think the best next step is to withdraw this PR for now and rework it into a
smaller proposal that fits the existing FreeRADIUS coding and configuration
patterns more closely.

PR #5829, opened 2026-04-19
rlm_dpsk: add generic adapters, VLAN replies, and SHA256 verification

hshimomura:
This PR extends `rlm_dpsk` in the FreeRADIUS 3.2 branch so the module can
support multiple vendor-specific DPSK / EasyPSK request and reply formats
through configuration while preserving existing v3.2 behavior.

The motivation is practical interoperability. The WPA inputs required to
verify a DPSK / EasyPSK exchange are standard, but vendors carry those inputs
in different RADIUS attributes and expect different reply formats.

In practice:

- RUCKUS DPSK uses vendor-specific attributes and expects PMK material in
  `MS-MPPE-Recv-Key`.
- Cisco IOS XE EasyPSK carries DPSK inputs in `Cisco-AVPair` key/value VSAs
  and expects the PMK back as Cisco AVPairs.
- Meraki Easy PSK uses vendor attributes and expects the PSK back via
  `Tunnel-Password`.

This PR keeps the module core focused on:

- normalizing request inputs into SSID / authenticator MAC / ANonce /
  EAPOL-Key data
- resolving PSK / PMK material from the existing control / CSV / cache flows
- verifying the MIC
- emitting vendor-appropriate reply attributes, optionally with VLAN
  attributes

### 1. Config-driven adapter model for v3.2

- Adds configured `adapter` blocks for request and reply layouts.
- Supports:
  - `standard_attrs`
  - `named_vsa_attrs`
  - `key_value_vsa`
- Ships working adapter definitions for:
  - `standard`
  - `ruckus`
  - `iosxe`
  - `meraki`
- Keeps request detection, extraction, and reply generation generic in the
  module core instead of adding one hardcoded vendor branch per product.

### 2. Vendor reply handling

- Adds explicit reply modes for:
  - standard PSK reply
  - `Tunnel-Password`
  - Cisco AVPair hex PMK reply
  - `MS-MPPE-Recv-Key`
- Lets one verifier path serve multiple vendor reply layouts.

### 3. VLAN reply support

- Adds optional VLAN reply emission using:
  - `Tunnel-Type = VLAN`
  - `Tunnel-Medium-Type = IEEE-802`
  - `Tunnel-Private-Group-Id = <vlan>`
- VLAN remains optional and is driven by the CSV data.
- CSV compatibility is preserved:
  - `identity,psk`
  - `identity,psk,client_mac`
  - `identity,psk,client_mac,vlan`

### 4. CSV / cache fixes

- Stores cached VLAN data as typed values instead of raw strings.
- Fixes cache string lifetime issues by preserving NUL termination.
- Fixes MAC parsing in CSV-backed lookup.
- Validates VLAN IDs before replying.

### 5. SHA256 / PMF verifier support

- Adds support for:
  - descriptor version 2 MIC verification (HMAC-SHA1)
  - descriptor version 3 MIC verification (AES-128-CMAC)
  - SHA1 PTK derivation
  - SHA256 PTK derivation
- Parses the AKM suite from the RSN IE inside the EAPOL-Key frame and selects
  the verifier path accordingly.
- Supports PMF-style hostap PTK handling for AKM `00-0f-ac-06`.

## Why this shape is useful upstream

- It adds practical multi-vendor DPSK support to the maintained v3.2 codebase.
- It preserves the existing v3.2 mental model:
  - control attributes
  - CSV fallback
  - PMK cache
- It reduces vendor-specific logic in the verification path and keeps vendor
  differences mostly in adapter configuration.
- It supports both SHA1 and SHA256 / PMF request types.

## Compatibility notes

- Existing v3.2 deployments without adapter customization continue to use the
  default adapter behavior.
- Legacy `ruckus = yes` compatibility is preserved.
- Existing CSV formats remain valid; the VLAN field is optional.
- The shipped adapter examples target current RUCKUS, Cisco IOS XE, and
  Meraki request / reply layouts.

## Testing performed

- `make -j1` completed successfully in the external v3.2 test build tree.
- `freeradius -XC` passes after rebuilding and installing the module.
- Live authentication was verified for:
  - RUCKUS DPSK, SHA1 path
  - Cisco IOS XE EasyPSK, SHA1 path
  - Meraki Easy PSK, SHA1 path
  - Meraki Easy PSK, SHA256 / PMF path
- VLAN replies were verified for all vendor paths above.

### DIFF
--- a/raddb/mods-available/dpsk
+++ b/raddb/mods-available/dpsk
@@ -191,4 +191,74 @@ dpsk {
 	#  search space, and thus the CPU requirements.
 	#
 #	filename = "${modconfdir}/${..:name}/psk.csv"
+	#
+	#  Multiple adapters can be defined.  The module picks the
+	#  first adapter whose request layout matches the packet.
+	#
+	adapter standard {
+		request {
+			type = standard_attrs
+			username = User-Name
+			ssid = Called-Station-SSID
+			called_station = Called-Station-MAC
+			anonce = FreeRADIUS-802.1X-Anonce
+			key_msg = FreeRADIUS-802.1X-EAPoL-Key-Msg
+		}
+		reply {
+			mode = standard
+			psk_attr = Pre-Shared-Key
+			psk_identity_attr = PSK-Identity
+		}
+	}
+
+	adapter ruckus {
+		request {
+			type = named_vsa_attrs
+			username = User-Name
+			ssid = Ruckus-SSID
+			called_station = Ruckus-BSSID
+			anonce = Ruckus-DPSK-Anonce
+			key_msg = Ruckus-DPSK-EAPoL-Key-Frame
+		}
+		reply {
+			mode = ms_mppe_recv_key
+			psk_attr = MS-MPPE-Recv-Key
+		}
+	}
+
+	adapter iosxe {
+		request {
+			type = key_value_vsa
+			username = User-Name
+			container_attr = Cisco-AVPair
+			ssid_key = cisco-wlan-ssid
+			called_station_key = cisco-bssid
+			anonce_key = cisco-anonce
+			key_msg_key = cisco-8021x-data
+			value_encoding = radius_escaped
+		}
+		reply {
+			mode = avpair_hex_pmk
+			avpair_attr = Cisco-AVPair
+			pmk_key = psk
+			extra_pairs = psk-mode=hex
+		}
+	}
+
+
+	adapter meraki {
+		request {
+			type = named_vsa_attrs
+			username = User-Name
+			ssid = Meraki-IPSK-SSID
+			called_station = Meraki-IPSK-BSSID
+			anonce = Meraki-IPSK-Anonce
+			key_msg = Meraki-IPSK-EAPOL
+		}
+		reply {
+			mode = tunnel_password
+			psk_attr = Tunnel-Password
+		}
+	}
+
 }
--- a/share/dictionary.meraki
+++ b/share/dictionary.meraki
@@ -3,6 +3,8 @@
 #
 #       For Meraki.
 #
+#       Version 2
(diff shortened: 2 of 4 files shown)


arr2036:
The majority of these enhancements belong in the policy language, not compiled into the DPSK module itself.  For other modules in v3, we define internal attributes that the module reads from and writes to.  The values from these can then be copied to VSAs using unlang or the users file.  For v4 there should be a callenv defined, with FR_TYPE_TMPL, and the user should be able to specify which attributes to read from and write to using attribute names.  We try and keep configuration patterns standard across the modules, and this "adapter" pattern matches no other modules in the server.

The various modes are likely useful to pull in, but they'd need tests for the crypto functions, preferably with test vectors pulled from applicable standards or vendor documentation.

hshimomura:
Thank you for the comment. I understand the architecture concern you raised.

Rather than keeping vendor-specific request/reply mapping inside the module via
an `adapter` model, I am going to rethink this in a direction where that logic
lives in `unlang`, the `users` file, or `FR_TYPE_TMPL`.

For the next step, I plan to reorganize the work in the following way:

- drop the adapter model
- leave SHA256 / PMF out of the first step and focus on the SHA1 path only because only Meraki support it as optional
- use RUCKUS and Meraki as policy-language-based configuration examples
- keep any `rlm_dpsk` changes vendor-neutral and minimal

The first thing I want to verify is whether the existing upstream v3.2
`rlm_dpsk` inputs and outputs are already sufficient for policy-side reply
generation.

Based on that, the next proposal will either focus on configuration examples
and documentation, or on a much smaller change with only the minimum generic
module-side changes required.

I think the best next step is to withdraw this PR for now and rework it into a
smaller proposal that fits the existing FreeRADIUS coding and configuration
patterns more closely.

PR #5830, opened 2026-04-20
rlm_dpsk: add generic reply attributes and optional VLAN replies

hshimomura:
## Summary
This change improves `rlm_dpsk` so that local policy can build vendor-specific replies from standard reply attributes, while keeping the module itself vendor-neutral.

In particular, the module now exposes `Pairwise-Master-Key` in addition to the existing `PSK-Identity` and `Pre-Shared-Key`, so local policy has access to all three reply attributes when it needs to construct vendor-specific responses.

This change also adds an optional fourth CSV column for VLAN assignment, adds Meraki IPSK test attributes to `dictionary.meraki`, and updates the comments in `raddb/mods-available/dpsk` and `raddb/sites-available/default` accordingly.

## Details
`rlm_dpsk` now accepts CSV entries in the following format:

`identity,psk[,mac[,vlanid]]`

This is backwards compatible with the existing file format:
- `identity,psk`
- `identity,psk,mac`

The VLAN field may also be used when the MAC field is empty, for example:

`identity,psk,,2065`

If `vlanid` is present, the module returns the standard tunnel reply attributes:
- `Tunnel-Type = VLAN`
- `Tunnel-Medium-Type = IEEE-802`
- `Tunnel-Private-Group-Id = "<vlanid>"`

The VLAN metadata is cached together with the other DPSK data, so cache hits return the same reply behavior as file lookups.

The sample `filename` expansion in `raddb/mods-available/dpsk` is also updated from `${..:name}` to `${.:name}` intentionally, so the example references the current module instance directly.

The Meraki dictionary update provides the IPSK test attributes needed for Meraki request handling and validation in the test environment.

## Fix included
While validating MAC-constrained CSV entries, I found that the MAC parsing path compared the decoded output length against `12` instead of `6`. This caused valid 12-hex-character MAC entries to be rejected. This patch corrects that check so MAC-constrained CSV rows work as intended.

## Documentation
The patch updates:
- `raddb/mods-available/dpsk`
- `raddb/sites-available/default`

The documentation now describes:
- the extended CSV format
- optional VLAN replies
- the updated sample `filename` expansion
- an example `Auth-Type dpsk` stanza showing the usual pattern for handling the module returning `updated`

## Validation
Tested against upstream v3.2.

Verified:
- legacy CSV rows without MAC or VLAN still work
- CSV rows with `identity,psk,,2065` return standard VLAN tunnel reply attributes
- CSV rows with `identity,psk,mac` now parse and match correctly
- Ruckus DPSK flow works with MAC-constrained rows
- Meraki IPSK flow works with optional VLAN replies alongside the existing `Tunnel-Password` reply
- the Meraki IPSK validation uses the attributes added in `dictionary.meraki`
- `freeradius -XC` passes with the updated module and configuration

### DIFF
--- a/raddb/mods-available/dpsk
+++ b/raddb/mods-available/dpsk
@@ -53,7 +53,7 @@
 #			if (updated) {
 #				... cache &reply:Pre-Shared-Key
 #				... cache &reply:PSK-Identity
-#
+#				ok
 #			}
 #		}
 #
@@ -75,13 +75,20 @@
 #  * updated	if there are DPSK attributes which match a PSK, and the
 #		PSK was read from 'filename'.
 #
-#		It also updates the attributes &reply:Pre-Shared-Key
-#		with the found PSK, along with &reply:PSK-Identity
-#		with the found identity.
+#		It also creates &reply:Pairwise-Master-Key.  When the
+#		matching PSK is known, it also creates the attributes
+#		&reply:Pre-Shared-Key with the found PSK, along with
+#		&reply:PSK-Identity with the found identity.
 #
-#		You can then check the return code for "updated", and
-#		write those attributes into a database.  This step
-#		ensures that 'filename' is read only as a last resort.
+#		When the PSK is read from 'filename', an optional VLAN ID
+#		can also be returned via the standard reply attributes
+#		Tunnel-Type = VLAN, Tunnel-Medium-Type = IEEE-802, and
+#		Tunnel-Private-Group-Id.
+#
+#		You can then check the return code for "updated",
+#		write those attributes into a database, and return
+#		ok to continue processing.  This step ensures that
+#		'filename' is read only as a last resort.
 #		Since the file is read from top to bottom for every
 #		packet, this process is much slower than using a
 #		database which is keyed to the PSK Identity.
@@ -111,7 +118,8 @@ dpsk {
 	#
 	#  The cache entry is the PSK-Identity and Pre-Shared-Key,
 	#  and/or the PMK which are used to verify the information in
-	#  the Access-Request.
+	#  the Access-Request.  When VLAN information is read from the
+	#  file, it is also cached with the entry.
 	#
 	#  Caching entries can help, even when using a database.  It
 	#  is very slow to calculate the PMK from the PSK.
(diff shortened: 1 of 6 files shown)


CWempe:
Any chance this will be implemented any time?
We tested this ourself and decided to use this in our campus wifi.
Would be great if this would be part of the official package ranther than depending on self compilation.

alandekok:
We'll take a look at this for the 3.2.11 release.

alandekok:
I've pulled the changes in with some tweaks.  Please validate.

hshimomura:
I will check it this and will update it to you.

hshimomura:
Thanks @alandekok . 

I validated the integrated changes from the current v3.2.x branch at e0151787c856c540a35dcc0bd40c22302eabe425.
I built and installed on Debian 13. FreeRADIUS starts successfully, and the dpsk-pmk and dpsk-psk tests pass.
I also validated the implementation with a Cisco Catalyst 9800 using EasyPSK. PSK authentication works correctly, including PSKs with leading zeroes.
During testing, I found that the VLAN reply attributes are added twice when the CSV entry contains a VLAN:
Tunnel-Type = VLAN
Tunnel-Medium-Type = IEEE-802
Tunnel-Private-Group-Id = "2065"
Tunnel-Type = VLAN
Tunnel-Medium-Type = IEEE-802
Tunnel-Private-Group-Id = "2065"
The current code calls dpsk_add_reply_vlan() twice in update_attributes:
lines 919–922
lines 939–942
https://github.com/FreeRADIUS/freeradius-server/blob/e0151787c856c540a35dcc0bd40c22302eabe425/src/modules/rlm_dpsk/rlm_dpsk.c#L918-L942
This also appears twice under Server Policies on the Catalyst 9800. Authentication still succeeds, but only one set of VLAN tunnel attributes should be returned.
Removing either duplicate block fixes the response. I suggest removing the first block at lines 919–922 and retaining the later block with the other reply attribute creation.
I could not find an existing DPSK test that verifies VLAN reply attributes. It would be useful to add a regression test which confirms that exactly one Tunnel-Type, Tunnel-Medium-Type, and Tunnel-Private-Group-Id is returned.
Would you prefer that I submit a small follow-up PR against the latest v3.2.x, or will you fix this directly in the integrated commits?

alandekok:
> I validated the integrated changes from the current v3.2.x branch at [e015178](https://github.com/FreeRADIUS/freeradius-server/commit/e0151787c856c540a35dcc0bd40c22302eabe425). I built and installed on Debian 13. FreeRADIUS starts successfully, and the dpsk-pmk and dpsk-psk tests pass. I also validated the implementation with a Cisco Catalyst 9800 using EasyPSK. PSK authentication works correctly, including PSKs with leading zeroes. 

Thanks.  That helps.

> During testing, I found that the VLAN reply attributes are added twice when the CSV entry contains a VLAN: Tunnel-Type = VLAN Tunnel-Medium-Type = IEEE-802 Tunnel-Private-Group-Id = "2065" Tunnel-Type = VLAN Tunnel-Medium-Type = IEEE-802 Tunnel-Private-Group-Id = "2065" The current code calls dpsk_add_reply_vlan() twice in update_attributes: lines 919–922 lines 939–942

I'll remove the first copy of the `dpsk_add_reply_vlan()` code.

> I could not find an existing DPSK test that verifies VLAN reply attributes. It would be useful to add a regression test which confirms that exactly one Tunnel-Type, Tunnel-Medium-Type, and Tunnel-Private-Group-Id is returned.

I'm OK with adding a unit test, but it's not critical.  The functionality is small and simple enough that it shouldn't be a problem to leave it.

> Would you prefer that I submit a small follow-up PR against the latest v3.2.x, or will you fix this directly in the integrated commits?

I'll push a fix, thanks.

PR #5832, opened 2026-04-21
docs-v4: update howto/modules/CHAP for v4, added diagram, and separat…

nolade:
…ed into pages for increased readability.


alandekok:
The text reads like AI summaries, and is a bit wrong:

* CHAP isn't really used this way in supplicants.
* "The client performs some computations and returns the result." is very vague, and doesn't mean anything to the reader
* the "3 phases" of CHAP usually refer to the data going over PPP, not the RADIUS portion.
* The explanations are also duplicated - the first paragraphs give an overview, and then the later text provides a little more detail.

The file `testchap.adoc` can just be called `testing.adoc`

PR #5835, opened 2026-04-22
Update noop for accounting handling in buffered-sql

Sam-Hutchings:
Added noop for buffered-sql that is present in 'default' site. Otherwise, this causes detail to hang on the below when encountering zero session length. 

This brings the same functionality from 'default' to 'buffered-sql' for ease.

For example, the below will be logged repeatably:detail (/var/log/freeradius/radacct/detail*): Read packet from /var/log/freeradius/radacct/detail.work
detail (/var/log/freeradius/radacct/detail*): Read packet from /var/log/freeradius/radacct/detail.work
detail (/var/log/freeradius/radacct/detail*): Read packet from /var/log/freeradius/radacct/detail.work
detail (/var/log/freeradius/radacct/detail*): Renaming /var/log/freeradius/radacct/detail -> /var/log/freeradius/radacct/detail.work
detail (/var/log/freeradius/radacct/detail*): Read packet from /var/log/freeradius/radacct/detail.work
detail (/var/log/freeradius/radacct/detail*): Polling for detail filedetail (/var/log/freeradius/radacct/detail*): Read packet from /var/log/freeradius/radacct/detail.work

### DIFF
--- a/raddb/sites-available/buffered-sql
+++ b/raddb/sites-available/buffered-sql
@@ -145,6 +145,19 @@ server buffered-sql {
 	#
 	accounting {
 		#
+		#  If you receive stop packets with zero session length,
+		#  they will NOT be logged in the database.  The SQL module
+		#  will print a message (only in debugging mode), and will
+		#  return "noop".
+		#
+		#  You can ignore these packets by uncommenting the following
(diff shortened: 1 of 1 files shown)


alandekok:
thanks.  I've merged it in.

PR #5836, opened 2026-04-30
CI test result matrix; fix latent dpsk module build issues

terryburton:
Runs all test, rather than stopping at first failure.

Reports failures in CI job summary:
- Pass/fail matrix
- Failure log extracts
- Full failure logs packaged as a downloadable assets

Example report here: https://github.com/terryburton/freeradius-server/actions/runs/25172640608


alandekok:
merged via cherry pick due to PGP issues.

PR #5839, opened 2026-05-01
updated valgrind.h path

marc-casavant:
Updates to the valgrind.h header file based on testing with new profiling build from [https://github.com/marc-casavant/freeradius-server/tree/new-profiling-image-build](https://github.com/marc-casavant/freeradius-server/tree/new-profiling-image-build)


alandekok:
I've updated `configure` and `src/include/autoconf.h.in` to match, and pushed it.

PR #5845, opened 2026-05-07
docs-v4: update howto/passwd section with more details, formatting, a…

nolade:
…nd headings


alandekok:
I've pulled this in with fixes:

* use back-ticks around filenames and attribute names

PR #5849, opened 2026-05-13
docs-v4: updated pam howto into sections, added details to steps

nolade:


alandekok:
I've pulled this in with fixes:

* the configuration files examples were wrong. There is no `process Access-Request` section.
* the formatting of the configuration files was all jumbled.  Most of this looks to be from the previous commits

PR #5852, opened 2026-05-15
Added new rlm_dotnet module

enteriaza:
New rlm_dotnet (.NET 8+) module for FreeRadius 3.2.x. Will also introduce a module for the new FreeRadius 4.x at a later stage.

This module was compiled and tested against 3.2.8 with no problems and/or concerns observed. A comprehensive C# Sdk is provided for 3rd parties to introduce .NET based applications into the FR pipeline.


arr2036:
What's the rationale for this?  FreeRADIUS doesn't run on Windows?.. and there's already a HTTP capable module in v3.

arr2036:
So a couple of notes, this is raised against master which is v4.0.x.  We have considered porting FreeRADIUS to windows, but it's not something we've invested a huge amount of effort it.   I was looking at fixing up the libkqueue dependency which would be the first step.

I'd be interested to understand how you're currently running master against Windows?  Is it using some kind of Linux emulation layer?

Generally for these kinds of integrations we'd recommend exposing your .net code as a web service and then using the rlm_rest module to communicate with it.  In v4 the rlm_rest module is extremely high performance, capable of 10s of thousands of request/s.

arr2036:
@claude review

PR #5853, opened 2026-05-15
protocols/der: unblock fuzzing harness by removing tautology

DavidKorczynski:
The fuzzing harness is currently blocked due to a conditional tautology. Coverage report shows no actual fuzzing of the der decoding logic: https://storage.googleapis.com/oss-fuzz-coverage/freeradius/reports/20260513/linux/src/freeradius-server/src/protocols/der/decode.c.html#L2694

This fixes it.

### DIFF
--- a/src/protocols/der/decode.c
+++ b/src/protocols/der/decode.c
@@ -2689,10 +2689,10 @@ static ssize_t fr_der_decode_proto(TALLOC_CTX *ctx, fr_pair_list_t *out, uint8_t
 {
 	fr_dbuff_t our_in = FR_DBUFF_TMP(data, data_len);
 
-	fr_dict_attr_t const *parent = fr_dict_root(dict_der);
+	fr_dict_attr_t const *parent = fr_dict_attr_by_name(NULL, fr_dict_root(dict_der), "Certificate");
 
-	if (unlikely(parent == fr_dict_root(dict_der))) {
-		fr_strerror_printf_push("Invalid dictionary. DER decoding requires a specific dictionary.");
+	if (unlikely(parent == NULL)) {
+		fr_strerror_printf_push("DER fuzzing: Certificate attribute not found in der dictionary");
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


arr2036:
@claude review Will changing the root to a specific attribute cause problems?

arr2036:
@claude seems like something that belongs in the fuzzer harness and passed in as a dictionary ctx to use?  Does the protocol ctx make it all the way through from the testpoint?

arr2036:
@claude Is there even a corpus for DER currently, or is it just going to flail?  Can you verify the output of the unit tests was folded into the corpus?  I'm assuming there they all expect to decode from the root?  Would changing the root invalidate the corpus?

alandekok:
the DER code requires that the encoding / decoding start from a particular part of the OID tree.  So the changes here are fine.

alandekok:
I've pushed some changes which will help.  The fuzzers now accept an environment variable FR_FUZZER_ROOT_ATTR which specifies the name of the root attribute to use for fuzzing.

the DER code has been updated to use it.

the fuzzers have also been moved to their own directory.  There are enough fuzzers now that the "bin" directory is getting complex.

alandekok:
I've pushed a number of changes which should help.  The only remaining need is to create a fuzzer_der.c, which has an explicit list of root attributes which are used for fuzzing.

DavidKorczynski:
Thanks @alandekok -- I can confirm the `der` fuzzer from here https://github.com/FreeRADIUS/freeradius-server/blob/c26dcfcd26678bf0e06486546f4f153513002184/src/fuzzer/all.mk#L8 is not running well. Closing this.

DavidKorczynski:
I might have spoken too early above, as I just reran the test on a fresh checkout and `fuzzer_der` wasn't performing well. Will follow up a bit later.

alandekok:
the `fuzzer_der` code won't do fuzzing in the current build.  We need to create a `fuzzer_der.c` like `fuzzer_xlat.c`. It can pass in specific `root_da`, based on the input.

Or, the fuzzer_der should just pick one root_da (e.g. `Certificate`), and use that

PR #5854, opened 2026-05-15
Add 'arp' to FUZZER_PROTOCOLS in all.mk

DavidKorczynski:
Enable fuzzing of the `arp` protocol logic


arr2036:
fuzzers fail because they expect corpus to exist.  Fixed now in master.

PR #5855, opened 2026-05-16
add new harness targeting value parsing

DavidKorczynski:
The value.c has around 25% code coverage at the moment: https://storage.googleapis.com/oss-fuzz-coverage/freeradius/reports/20260514/linux/src/freeradius-server/src/lib/util/value.c.html and this extends on that to achieve 70%+ from a local run.

This harness overlaps with fuzzer.c in the sense they both hit `fr_value_box_from_str`. However, this `fuzzer_value` significantly increase code coverage and I presume it's because it's more narrow than `fuzzer.c`. Let me know what you think.

### DIFF
--- a/src/bin/all.mk
+++ b/src/bin/all.mk
@@ -62,6 +62,11 @@ endef
 #
 SUBMAKEFILES += fuzzer_json.mk
 
+#
+#  Enable fuzzer_value
+#
+SUBMAKEFILES += fuzzer_value.mk
+
 $(foreach X,${FUZZER_PROTOCOLS},$(eval $(call FUZZ_PROTOCOL,${X})))
 
 .PHONY: fuzzer.help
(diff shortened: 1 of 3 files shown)


arr2036:
Nice, thanks!

arr2036:
A couple of minor nits:
- We shouldn't \0 terminate, we want things to explode :)
- If this is a duplicate of a subset of the main fuzzer case, could we pull the existing case out of fuzzer.c
- Could we poison memory around the value box and around the input string.  It's not going to be perfect, because __asan_poison_memory_region is 8 byte aligned, but it'd likely flag some reads into garbage memory given enough time.  We already do something similar in unit_test_attribute IIRC.

arr2036:
@claude review

alandekok:
I've merged it manually, thanks.

DavidKorczynski:
> I've merged it manually, thanks.

Just so I can follow, where was this added? I don't see it in `fuzzer_value.c` in src/bin?

alandekok:
I've just pushed the changes, after some other cleanups.  They'll take a pass through CI for verification, and then get merged.

PR #5858, opened 2026-05-17
add fuzzing harness to be consumed by OSS-Fuzz

DavidKorczynski:
Adds a fuzzing harness for the xlat* files, which are currently unfuzzed (see xplat* files here:
https://storage.googleapis.com/oss-fuzz-coverage/freeradius/reports/20260517/linux/src/freeradius-server/src/lib/unlang/report.html)

This harness will be consumed by OSS-fuzz to be run continuously.

### DIFF
--- a/src/fuzzer/fuzzer_xlat.mk
+++ b/src/fuzzer/fuzzer_xlat.mk
@@ -1,0 +1,8 @@
+TARGET			:= fuzzer_xlat$(E)
+SOURCES			:= fuzzer_xlat.c
+
+TGT_PREREQS		:= libfreeradius-unlang$(L) libfreeradius-util$(L) $(LIBFREERADIUS_SERVER)
+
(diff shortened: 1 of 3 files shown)


src/fuzzer/fuzzer_xlat.mk
```diff
@@ -0,0 +1,8 @@
+TARGET			:= fuzzer_xlat$(E)
+SOURCES			:= fuzzer_xlat.c
+
+TGT_PREREQS		:= libfreeradius-unlang$(L) libfreeradius-util$(L) $(LIBFREERADIUS_SERVER)
```
alandekok, line 4: Nit: `$(LIBFREERADIUS_SERVER)` is enough here.  It includes the unlang && util libraries.

alandekok:
The `unit_test_attribute` function could also be updated to call fuzzer_dump, too.

PR #5861, opened 2026-05-25
TEAP: optional method support for Windows 11

jimdigriz:
These commits allow some degree of flexibility in optional authentication with Windows 11. Limitations are described, including for a developer spelunking the codebase.

As a bonus, included are hooks that set `session-state:TEAP-Type-{Machine,User}` as the inner methods complete (we change the code to read administrative overrides from `config` instead) so unlang can be used to determine an authorization policy based on what actually happened during authentication.

There are hints we might be able to get Windows to allow those other currently unsupported optional method combinations, as well as support the Windows specific crypto-binding calculation, this needs more experimentation though and we are fighting with an opaque state machine.

A crucial note, this breaks support with `wpa_supplicant` (and thus also `eapol_test`), but as TEAP support there is behind a compile time flag marked "here be dragons", we need to implement the defacto standard. One day this may all be fixed by TEAPv2.

### DIFF
--- a/raddb/mods-available/eap
+++ b/raddb/mods-available/eap
@@ -1228,6 +1228,11 @@ eap {
 		#  chosen identity.  You can set &session-state:FreeRADIUS-EAP-TEAP-TLV-Identity-Type
 		#  manually, but doing that is substantially more complicated.
 		#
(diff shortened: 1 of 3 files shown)


alandekok:
Merged, thanks.

PR #5862, opened 2026-05-25
fuzzer: add radius_encode harness to be consumed by OSS-Fuzz

DavidKorczynski:
The goal is to increase code coverage of the folder here: https://storage.googleapis.com/oss-fuzz-coverage/freeradius/reports/20260515/linux/src/freeradius-server/src/protocols/radius/report.html

### DIFF
--- a/src/fuzzer/all.mk
+++ b/src/fuzzer/all.mk
@@ -49,13 +49,31 @@ FUZZER_ARTIFACTS ?= ${BUILD_DIR}/fuzzer
 FUZZER_TIMEOUT   ?= 10
 
 #
-#  Define a function to do all of the same thing.
+#  A handful of protocols ship a hand-written fuzzer_<proto>.c (committed to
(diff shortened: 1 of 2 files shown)


alandekok:
There is already an unused `encode` stub in `fuzzer.c`.  That code also uses the `common.c` file, which makes the protocol-specific fuzzers much simpler.

Since the encode path requires protocol-specific information, and the `fuzzer.c` file is now tiny, it might be worth just making protocol-specific encoders for many things:

* copy `fuzzer.c` to `fuzzer_radius.c`  (which removes a lot of the boilerplate of `fuzzer_radius_encode.c` in this PR)
* poke the build system to treat `fuzzer_radius` as a separate target, but which is still FUZZER_PROTOCOL
* update `fuzzer_radius` to call the `encode` path as is done in this PR.

Once that's done, it should be fairly straightforward to do the same for the other protocols.

PR #5867, opened 2026-05-29
docs-v4: update sql tutorial for v4, updated `add user into sql db` t…

nolade:
…utorial, added v4 samples


nolade:
Submitted an outdated version, will resubmit with all file changes.

PR #5870, opened 2026-06-02
fix detail listener failing to process any packets

bmork:
The refactoring of request_pre_handler() inadvertently changed its return value from 1 to 0 in the case where request->packet->vps already existed. This change broke the detail listener, failing to process any requests:

  detail (/foo/radius/spool/detail*): No response to request.  Will retry in 30 seconds

Fixes: 5691cbc615e5 ("rearrange request_pre_handler() to do less work")


alandekok:
merged, thanks.

PR #5871, opened 2026-06-02
docs-v4: Update sql tutorial for v4. Includes steps for connecting & …

nolade:
…testing db connections. Renamed file from sql.adoc to sql_db.adoc.


alandekok:
Merged, thanks.

I've left the `sql.adoc` name as-is.  I prefer to avoid redundancy.  SQL is a db, and the basic "connection" tutorial can just be "sql"

PR #5873, opened 2026-06-03
docs-v4: Updated for v4, Added local dictionary via `DEFINE`, custom …

nolade:
…dictionary implementation and test of vendor attributes, creating a local dictionary def witin a virtual server.


alandekok:
I'll pull this in, but there's a lot of explanatory text here on what the dictionaries are, and what they do.  That text should go into the main dictionary documentation.  And then the "howto" here can just say "go here to learn more".

PR #5874, opened 2026-06-04
skip psk checks when not supported

wolfrath-aws:
Releases 3.2.9 and 3.2.10 fail to build with OpenSSL that was compiled without PSK support. `psk_identity` is only defined if
`PSK_MAX_IDENTITY_LEN` is set.

```
CC src/main/regex.c
src/main/tls.c: In function ‘tls_new_client_session’:
src/main/tls.c:527:11: error: ‘fr_tls_server_conf_t {aka struct fr_tls_server_conf_t}’ has no member named ‘psk_identity’
  if (!conf->psk_identity) {
           ^~
```


alandekok:
Merged manually

PR #5882, opened 2026-06-11
docs: update wrap.py to handle any depth of nav list entries (4+). Ad…

nolade:
…ded xref: line exemption to prevent line wrapping of nav list entries.


alandekok:
Pulled in, thanks

PR #5883, opened 2026-06-12
docs-v4: add interval file to antora idex. Update nav.adoc 

nolade:
& policy_table.adoc to include new entry.


nolade:
merged & done

PR #5896, opened 2026-06-30
docs-v4: update format_raddb script so the .dotfiles are not included…

nolade:
… in the search for target files.

### DIFF
--- a/scripts/asciidoc/format_raddb.md
+++ b/scripts/asciidoc/format_raddb.md
@@ -19,6 +19,10 @@ should not be formatted.  The list of files to ignore is below:
 
 * ignore files in the `raddb/certs` directory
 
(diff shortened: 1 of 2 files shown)


scripts/asciidoc/format_raddb.md
```diff
@@ -82,7 +86,7 @@ converted to using `=`.  For example:
 Should get converted to:
 
 ```
-# = This is a title
+== This is a title
```
alandekok, line 89: This change is incorrect.  The intent is to say that there is a title within a comment, e.g.:

```
#
#  # Title
#
```

The comment block should be preserved.  The text within the comment block has a section 1 title: `= Title`, not `== Title`.

I'll pull in and tweak manually.

PR #5898, opened 2026-07-01
Update AI policy link in SECURITY.md

DavidKorczynski:
The referenced AI.md does not exist, so pointing this to the README.md section that discusses AI usage.


alandekok:
That's the wrong target for the link.  It should be AGENTS.md

PR #5899, opened 2026-07-07
docs-v4: Fix-level-0 headings in selected raddb files to correct docs…

nolade:
…ite outputs/rendering


alandekok:
I'll merge it manually, and keep the reformatting separate from the "get rid of leading dots" bit.

PR #5901, opened 2026-07-13
radtest: allow overriding NAS name via HOSTNAME env var

commodo:
Use HOSTNAME when set, retaining the portable hostname/uname fallback.

### DIFF
--- a/src/bin/radtest.in
+++ b/src/bin/radtest.in
@@ -105,7 +105,12 @@ if [ "$7" ]
 then
 	nas=$7
 else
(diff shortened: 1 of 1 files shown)


arr2036:
Baking Linux specific things in isn't great... feel free to add an override taking the hostname value from the `HOSTNAME` environmental variable.

commodo:
> Baking Linux specific things in isn't great... feel free to add an override taking the hostname value from the `HOSTNAME` environmental variable.

that also works

commodo:
updated

PR #5902, opened 2026-07-13
ci: publish profiling results to cinfra-ca.inkbridge.io

marc-casavant:
## Summary

The multi-server CI's profiling leg (`mode: profiling`) already produces a
Callgrind tree under `prof-results/` and uploads that tree as a short-lived
workflow artifact. The workflow artifact expires, so the history is lost and
nothing can browse profiling trends across runs.

The commit adds a publish step: after the artifact upload, the profiling leg
packs the same `prof-results/` tree into one gzipped tarball and POSTs it
over HTTPS to a durable profiling store, which ingests it server-side. The
results are then browsable per branch / commit / run in the profiling
dashboard:

- Store (raw data): https://cinfra-ca.inkbridge.io/profiling/data
- Dashboard (analyzer UI): https://cinfra-ca.inkbridge.io/profiling/dashboard
- Dashboard V2: https://cinfra-ca.inkbridge.io/profiling/dashboard/v2

Both endpoints are live. The ingest flow has been verified end to end against
the store container, including concurrent publishes for the same commit,
re-runs, and malformed archives.

## What changed

| File | Change |
|---|---|
| `.github/workflows/ci-multi-server-tests.yml` | Adds `permissions: id-token: write` to the job and a "Publish profiling results" step on the profiling leg (+23 lines) |
| `scripts/ci/publish-profiling-results.sh` | New POSIX shell script that tars the tree and POSTs it to the store's ingest endpoint; the header comment documents the behaviour |
| `src/tests/multi-server/scripts/profiling/start_valgrind_profiling.sh` | Callgrind flags: `--separate-threads=no` and `--separate-callers=6` |

## How the publish works

1. The runner mints a GitHub OIDC id-token whose audience is the store URL.
   The store validates the token in-process (signature against GitHub's JWKS,
   plus issuer, audience, and a repository allow-list). **No repository secret
   is added**; `id-token: write` is the only new permission.
2. The script tars an explicit file list (so empty run directories and editor
   droppings never travel), checks the tarball against the front proxy's
   request-body cap so an oversized publish fails with a clear message
   instead of an opaque 413, and POSTs it to the store's `_ingest` endpoint.
   That single request is the whole publish.
3. Everything stateful happens server-side in the store, not in CI: it
   validates and unpacks the tarball, renumbers the run indexes so each
   `<branch>/<sha>` reads as a gapless 1..K (under a lock, so two publishes
   for the same sha cannot collide; a re-run appends after the existing runs
   instead of overwriting them), records the publish in a ledger
   (`run-index-map.json`) using the run id/attempt claims from the verified
   token, and rebuilds the manifest the dashboard reads.
4. The store answers with a JSON summary of what it published (run-index
   mapping included), which lands in the CI log.

A missing or empty `prof-results/` tree skips the step cleanly (the step is
gated on the artifact upload having found files, and the script also exits 0
on a missing tree). The whole job is already gated on
`github.repository_owner == 'FreeRADIUS'`, so forks never attempt to publish.

## Callgrind flag change

`--separate-threads=yes` becomes `--separate-threads=no`, and
`--separate-callers=6` is added: per-thread output split the counts for the
same functions across thread-specific files, while the analyzer wants
combined per-function totals; the 6 levels of caller context let the analyzer
distinguish call paths when attributing cost.

## Failure behaviour

- No `prof-results/` tree: publish step is skipped, leg stays green.
- Tarball over the proxy's request-body cap: fails before uploading, with a
  message saying so.
- Token mint failure or any non-2xx from the store (bad archive, unreadable
  store state, ...): the script exits non-zero and the profiling leg fails,
  with the store's JSON error in the log.
- The store only refreshes its manifest when an ingest fully succeeds, so a
  retry of a half-landed publish re-derives the same run indexes and
  overwrites the half-landed files instead of duplicating them. The script
  deliberately never retries on its own: after a lost success response, a
  blind retry would append a duplicate set of runs.

## Store-side counterpart

The ingest endpoint (archive validation, locked run-index allocation, ledger,
manifest rebuild) lives in the store's own repo, `cinfra-profiling-server`,
and is deployed separately before this change is merged. The previous
revision of this PR did the numbering and index bookkeeping client-side in
the publish script; that (and the concurrent same-sha collision it could not
prevent) moved into the store.

### DIFF
--- a/.github/workflows/ci-multi-server-tests.yml
+++ b/.github/workflows/ci-multi-server-tests.yml
@@ -36,8 +36,9 @@ jobs:
     runs-on: self-hosted
     if: github.repository_owner == 'FreeRADIUS'
 
-    #  id-token lets the profiling leg mint a GitHub OIDC token to authenticate
-    #  uploads to the prof-results store which is validated by its access.lua.
+    #  id-token: write lets the profiling leg mint a GitHub OIDC token, which
+    #  authenticates the prof-results publish. Listing any permission drops
+    #  the unlisted ones, so contents: read is stated to keep checkout working.
     permissions:
       contents: read
       id-token: write
@@ -218,22 +219,14 @@ jobs:
           if-no-files-found: ignore
 
       #
-      #  Push the same prof-results/ tree into the durable prof-results store
-      #  so the browser UI can read it. The script tars the tree and POSTs the
-      #  tarball to the store's ingest endpoint, which numbers the runs,
-      #  updates its indexes, and refreshes its manifest server-side; see the
-      #  script header for the auth and proxy details. A failed publish fails
-      #  the leg.
-      #
-      #  Gated on the artifact step having found files: with if-no-files-found:
-      #  ignore, upload-artifact leaves artifact-id empty when the tree is
-      #  absent/empty, so this step is skipped rather than publishing nothing.
+      #  Publish the same prof-results/ tree to the durable store: the script
+      #  tars the tree and POSTs it, with a GitHub OIDC token, to the given
+      #  URL (see the script header). Skipped when the artifact step found no
+      #  files (empty artifact-id).
       #
       - name: Publish profiling results to the prof-results store
         if: ${{ matrix.mode == 'profiling' && always() && steps.upload-prof-results.outputs.artifact-id != '' }}
-        env:
-          PROF_RESULTS_URL: "https://cinfra-ca.inkbridge.io"
-        run: scripts/ci/publish-profiling-results.sh
+        run: scripts/ci/publish-profiling-results.sh "https://cinfra-ca.inkbridge.io/profiling/data"
 
       #
       #  If the CI has failed and the branch is ci-debug
--- a/.github/workflows/ci-multi-server-tests.yml
+++ b/.github/workflows/ci-multi-server-tests.yml
@@ -229,6 +229,25 @@ jobs:
         run: scripts/ci/publish-profiling-results.sh "https://cinfra-ca.inkbridge.io/profiling/data"
 
       #
+      #  Check latest profiling results vs each suite's previous run on the store.
+      #
+      #  cest-analyzer exit codes;
+      #   0 - clean
+      #   1 - latest result performance flagged
+      #   2 - tool error
+      #
+      - name: Check latest profiling results vs previous runs
+        if: ${{ matrix.mode == 'profiling' && success() }}
+        continue-on-error: true
+        run: |
+          curl -fsSL -o cest-analyzer \
+            "https://cinfra-ca.inkbridge.io/profiling/dashboard/bin/cest-analyzer"
+          chmod +x ./cest-analyzer
+          rc=0
+          ./cest-analyzer --latest -d prof-results --filter fr_ --gate 15 || rc=$?
+          echo "cest-analyzer exit code: $rc"
+
+      #
       #  If the CI has failed and the branch is ci-debug
       #  then start a tmate session for interactive debugging.
       #
--- a/.github/workflows/ci-multi-server-tests.yml
+++ b/.github/workflows/ci-multi-server-tests.yml
@@ -36,6 +36,12 @@ jobs:
     runs-on: self-hosted
     if: github.repository_owner == 'FreeRADIUS'
 
+    #  id-token lets the profiling leg mint a GitHub OIDC token to authenticate
+    #  uploads to the prof-results store which is validated by its access.lua.
+    permissions:
+      contents: read
+      id-token: write
+
     strategy:
       fail-fast: false
       matrix:
@@ -203,12 +209,31 @@ jobs:
       #  mode. Skipped in service mode (nothing to upload).
       #
       - name: Upload profiling results
+        id: upload-prof-results
         if: ${{ matrix.mode == 'profiling' && always() }}
         uses: actions/upload-artifact@v6
         with:
           name: prof-results-${{ github.sha }}
           path: prof-results/
           if-no-files-found: ignore
+
+      #
+      #  Push the same prof-results/ tree into the durable prof-results store
+      #  so the browser UI can read it. The script tars the tree and POSTs the
+      #  tarball to the store's ingest endpoint, which numbers the runs,
+      #  updates its indexes, and refreshes its manifest server-side; see the
+      #  script header for the auth and proxy details. A failed publish fails
+      #  the leg.
+      #
+      #  Gated on the artifact step having found files: with if-no-files-found:
+      #  ignore, upload-artifact leaves artifact-id empty when the tree is
+      #  absent/empty, so this step is skipped rather than publishing nothing.
+      #
+      - name: Publish profiling results to the prof-results store
+        if: ${{ matrix.mode == 'profiling' && always() && steps.upload-prof-results.outputs.artifact-id != '' }}
+        env:
+          PROF_RESULTS_URL: "https://cinfra-ca.inkbridge.io"
+        run: scripts/ci/publish-profiling-results.sh
 
       #
       #  If the CI has failed and the branch is ci-debug
--- a/scripts/ci/publish-profiling-results.sh
+++ b/scripts/ci/publish-profiling-results.sh
@@ -1,151 +1,64 @@
 #!/bin/sh
 #
-#  What this does
-#  --------------
-#  After a profiling CI run finishes, look for a prof-results/ directory and,
-#  if it is there, publish it to the long-lived prof-results store so the
-#  profiling web UI can show the new run. The store is run by the
-#  cinfra-profiling-server container, which the cinfra platform sits in front
-#  of.
+#  Tar up the prof-results/ directory (if any) and POST the tarball, with a
+#  GitHub OIDC bearer token, to the URL given as the only argument. A missing
+#  or empty prof-results/ is a quiet exit 0; any other failure exits non-zero
+#  so the CI leg fails with it.
 #
-#  The publish is a single request: we tar.gz the whole prof-results/ tree and
-#  POST it to the store's ingest endpoint. Everything stateful happens
-#  server-side, inside the container: it validates and unpacks the tarball,
-#  allocates the store run indexes (renumbering this run's local indexes to
-#  follow whatever the store already holds for the same <branch>/<sha>, under
-#  a lock so concurrent publishes cannot collide), records the publish in the
-#  run-index-map.json ledger, and rebuilds manifest.json for the UI. This
-#  script only packs and ships; the cinfra-profiling-server README documents
-#  the store side.
+#  The runner sets ACTIONS_ID_TOKEN_REQUEST_TOKEN / ACTIONS_ID_TOKEN_REQUEST_URL
+#  automatically when the job has "id-token: write"; the script uses them to
+#  mint an OIDC token whose audience is the target URL's origin.
 #
-#  A small routing detail worth knowing: we POST to /profiling/data/_ingest,
-#  but Traefik removes the /profiling prefix on the way in, so the container
-#  sees /data/_ingest.
-#
-#  How we authenticate
-#  -------------------
-#  No shared password. The runner mints a GitHub OIDC token whose audience is
-#  the store URL, and the store's access.lua checks the token's signature and
-#  the repo it came from. That is the only credential involved. The server
-#  also reads the GitHub run id / attempt / number for its ledger from that
-#  verified token, so nothing here needs to send them.
-#
-#  When we stop and fail
-#  ---------------------
-#  No prof-results/ directory (or no files in it)? Nothing to do, so we exit 0
-#  quietly. Otherwise the publish is all-or-nothing: any non-2xx response (or
-#  no response) fails this script and the CI leg with it. A failed publish is
-#  safe to re-run; the server derives run numbering so that a retry of a
-#  half-landed publish overwrites the half-landed files rather than
-#  duplicating them.
-#
-#  Why everything goes through a proxy
-#  -----------------------------------
-#  The runner cannot reach the store directly (egress is firewalled), so both
-#  the token mint and the upload go through the squid proxy. The connect
-#  timeout keeps us from hanging when the store is down; it does not limit the
-#  transfer itself.
-#
-#  The upload size cap
-#  -------------------
-#  The proxy in front of the store caps request bodies, so we check the
-#  tarball size before uploading and fail with a clear message instead of an
-#  opaque 413. The store's README documents the cap and the options if
-#  publishes outgrow it.
-#
-#  Usage
-#  -----
-#    publish-profiling-results.sh
-#
-#  Environment variables
-#  ---------------------
-#    PROF_RESULTS_URL                 Base URL of the store (required).
-#    ACTIONS_ID_TOKEN_REQUEST_TOKEN   } GitHub OIDC request credentials, set by
-#    ACTIONS_ID_TOKEN_REQUEST_URL     } the runner when id-token: write is on.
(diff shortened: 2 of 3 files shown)


scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,344 @@
+#!/bin/sh
```
arr2036, line 1: I'd rather not have this big blob of unintelligible shell in the freeradius repo.  Couldn't you just tar up the results and upload that to the container, and then have the container update all the indexes?  Then if you need to update auxilliary indexes, or change any other elements of processing, you don't need to raise a PR for FreeRADIUS.

marc-casavant: That's a good idea! On it. I don't see any reason why the solution you propose wouldn't work.

marc-casavant: The script now pushes a tarball, simplifying the script in the FreeRADIUS repo.  Updating indexes, etc. has also been moved to the profiling server container.  Verified working with the latest changes.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#
+echo "publishing $(du -h "$tarball" | cut -f1 | tr -d ' ') prof-results tarball"
+resp=$(mktemp)
+code=$(curl -sS --connect-timeout 10 -o "$resp" -w '%{http_code}' \
+	-X POST -H "Authorization: Bearer $token" \
+	--data-binary @"$tarball" \
+	"${PROF_RESULTS_URL}/profiling/data/_ingest") || code=000
+rm -f "$tarball"
+
+cat "$resp"; echo
```
arr2036: Always remove $resp, there's no point in putting in both branches

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  retry: if the server succeeded but the response got lost, a blind retry
+#  would append a duplicate set of runs, so leave retrying to a human re-run
+#  of the CI leg.
+#
+echo "publishing $(du -h "$tarball" | cut -f1 | tr -d ' ') prof-results tarball"
+resp=$(mktemp)
+code=$(curl -sS --connect-timeout 10 -o "$resp" -w '%{http_code}' \
+	-X POST -H "Authorization: Bearer $token" \
+	--data-binary @"$tarball" \
+	"${PROF_RESULTS_URL}/profiling/data/_ingest") || code=000
```
arr2036: You're posting to a resource so the path should just be `"${PROF_RESULTS_URL}/profiling/data`

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  the token mint and the upload go through the squid proxy. The connect
+#  timeout keeps us from hanging when the store is down; it does not limit the
+#  transfer itself.
+#
+#  The upload size cap
+#  -------------------
+#  The proxy in front of the store caps request bodies, so we check the
+#  tarball size before uploading and fail with a clear message instead of an
+#  opaque 413. The store's README documents the cap and the options if
+#  publishes outgrow it.
```
arr2036: If the message is opaque, fix the intermediary proxy to return a better body.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  When we stop and fail
+#  ---------------------
+#  No prof-results/ directory (or no files in it)? Nothing to do, so we exit 0
+#  quietly. Otherwise the publish is all-or-nothing: any non-2xx response (or
+#  no response) fails this script and the CI leg with it. A failed publish is
+#  safe to re-run; the server derives run numbering so that a retry of a
+#  half-landed publish overwrites the half-landed files rather than
+#  duplicating them.
+#
+#  Why everything goes through a proxy
```
arr2036: Not really necessary to document any of this, as the variables don't see to be passed explicitly to set a proxy server?

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  opaque 413. The store's README documents the cap and the options if
+#  publishes outgrow it.
+#
+#  Usage
+#  -----
+#    publish-profiling-results.sh
+#
+#  Environment variables
+#  ---------------------
+#    PROF_RESULTS_URL                 Base URL of the store (required).
```
arr2036: Pass this in as an argument

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#    ACTIONS_ID_TOKEN_REQUEST_TOKEN   } GitHub OIDC request credentials, set by
+#    ACTIONS_ID_TOKEN_REQUEST_URL     } the runner when id-token: write is on.
+#    PROF_MAX_UPLOAD_MB               Pre-upload tarball size limit in MB
```
arr2036: This should be removed

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#
+#  Usage
+#  -----
+#    publish-profiling-results.sh
+#
+#  Environment variables
+#  ---------------------
+#    PROF_RESULTS_URL                 Base URL of the store (required).
+#    ACTIONS_ID_TOKEN_REQUEST_TOKEN   } GitHub OIDC request credentials, set by
+#    ACTIONS_ID_TOKEN_REQUEST_URL     } the runner when id-token: write is on.
```
arr2036: This should jut be baked into the script, with possibility of ENV override, but it's very very unlikely to change.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  publishes outgrow it.
+#
+#  Usage
+#  -----
+#    publish-profiling-results.sh
+#
+#  Environment variables
+#  ---------------------
+#    PROF_RESULTS_URL                 Base URL of the store (required).
+#    ACTIONS_ID_TOKEN_REQUEST_TOKEN   } GitHub OIDC request credentials, set by
```
arr2036: This formatting is weird.  What is ACTIONS_ID_TOKEN_REQUEST_TOKEN?

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  but Traefik removes the /profiling prefix on the way in, so the container
+#  sees /data/_ingest.
+#
+#  How we authenticate
+#  -------------------
+#  No shared password. The runner mints a GitHub OIDC token whose audience is
+#  the store URL, and the store's access.lua checks the token's signature and
+#  the repo it came from. That is the only credential involved. The server
+#  also reads the GitHub run id / attempt / number for its ledger from that
+#  verified token, so nothing here needs to send them.
```
arr2036: Should be written in generic terms without knowledge of the server side.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#  allocates the store run indexes (renumbering this run's local indexes to
+#  follow whatever the store already holds for the same <branch>/<sha>, under
+#  a lock so concurrent publishes cannot collide), records the publish in the
+#  run-index-map.json ledger, and rebuilds manifest.json for the UI. This
+#  script only packs and ships; the cinfra-profiling-server README documents
+#  the store side.
+#
+#  A small routing detail worth knowing: we POST to /profiling/data/_ingest,
+#  but Traefik removes the /profiling prefix on the way in, so the container
+#  sees /data/_ingest.
```
arr2036: Implementation details don't belong in the public FreeRADIUS repo.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+#
+#  The publish is a single request: we tar.gz the whole prof-results/ tree and
+#  POST it to the store's ingest endpoint. Everything stateful happens
+#  server-side, inside the container: it validates and unpacks the tarball,
+#  allocates the store run indexes (renumbering this run's local indexes to
+#  follow whatever the store already holds for the same <branch>/<sha>, under
+#  a lock so concurrent publishes cannot collide), records the publish in the
+#  run-index-map.json ledger, and rebuilds manifest.json for the UI. This
+#  script only packs and ships; the cinfra-profiling-server README documents
+#  the store side.
```
arr2036: Not necessary.  I really don't like it when the AI slop comments are longer than the script itself.

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+: "${PROF_RESULTS_URL:?store URL required}"
+
+[ -d prof-results ] || { echo "no prof-results/ tree; skipping"; exit 0; }
+
+#
+#  Pack the tree. We tar an explicit file list (not the directory) so empty
+#  run directories - a test can claim a run index and then write nothing - and
+#  editor droppings never travel; the server only ever sees real result files.
+#
+list=$(mktemp)
```
arr2036: What's list?

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+tarball=$(mktemp).tar.gz
+find prof-results -type f ! -name '.DS_Store' | sed 's#^prof-results/##' >"$list"
+if ! [ -s "$list" ]; then
+	echo "prof-results/ holds no files; skipping"
+	rm -f "$list" "$tarball"
```
arr2036: Eh, doesn't this create the mktemp file (which is unused and never cleaned up) AND then uses the name of that file to create a tarball?

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+	exit 0
+fi
+#  COPYFILE_DISABLE stops tar implementations that honour it from adding
+#  synthetic ._* metadata members for extended attributes; GNU tar on the CI
+#  runner ignores the variable entirely.
+COPYFILE_DISABLE=1 tar -czf "$tarball" -C prof-results -T "$list"
```
arr2036: ok... so COPYFILE_DISABLE isnt actually needed then?

scripts/ci/publish-profiling-results.sh
```diff
@@ -0,0 +1,151 @@
...
+max_mb=${PROF_MAX_UPLOAD_MB:-45}
+size=$(wc -c <"$tarball")
+size_mb=$(( (size + 1048575) / 1048576 ))
+if [ "$size_mb" -gt "$max_mb" ]; then
+	echo "ERROR: publish tarball is ${size_mb} MB, over the ${max_mb} MB limit" >&2
+	echo "ERROR: the proxy in front of the store caps request bodies; raise that cap" >&2
+	echo "ERROR: (and PROF_MAX_UPLOAD_MB), or split the publish per run" >&2
+	rm -f "$tarball"
+	exit 1
+fi
```
arr2036: Remove this check, it's redundant.   Fix the infrastructure to return a reasonable error message.

src/tests/multi-server/scripts/profiling/start_valgrind_profiling.sh
```diff
@@ -40,7 +40,8 @@ valgrind \
   --log-file=/etc/prof-results/valgrind.log \
   --callgrind-out-file=/etc/prof-results/callgrind.out.%p \
   --trace-children=yes \
-  --separate-threads=yes \
+  --separate-threads=no \
```
arr2036, line 54: What does this do?

src/tests/multi-server/scripts/profiling/start_valgrind_profiling.sh
```diff
@@ -40,7 +40,8 @@ valgrind \
...
+  --separate-callers=6 \
```
arr2036, line 55: What does this do?

.github/workflows/ci-multi-server-tests.yml
```diff
@@ -36,6 +36,12 @@ jobs:
     runs-on: self-hosted
     if: github.repository_owner == 'FreeRADIUS'
 
+    #  id-token lets the profiling leg mint a GitHub OIDC token to authenticate
+    #  uploads to the prof-results store which is validated by its access.lua.
+    permissions:
+      contents: read
+      id-token: write
```
arr2036, line 44: What do these do?

.github/workflows/ci-multi-server-tests.yml
```diff
@@ -203,13 +209,32 @@ jobs:
...
+      #  the leg.
+      #
+      #  Gated on the artifact step having found files: with if-no-files-found:
+      #  ignore, upload-artifact leaves artifact-id empty when the tree is
+      #  absent/empty, so this step is skipped rather than publishing nothing.
+      #
+      - name: Publish profiling results to the prof-results store
+        if: ${{ matrix.mode == 'profiling' && always() && steps.upload-prof-results.outputs.artifact-id != '' }}
+        env:
+          PROF_RESULTS_URL: "https://cinfra-ca.inkbridge.io"
```
arr2036: I'd prefer the whole thing just be passed in as an arg, and there be where the script POSTs to.  That's a cleaner separation of concerns.  There's a script, it tars up the results, and posts it _somewhere_ with an OIDC token for autz.  That's nice and generic.  Then the implementation details are in how the script is called.

.github/workflows/ci-multi-server-tests.yml
marc-casavant: Claude's explanation:

(1) "id-token: write" lets the job mint a short-lived GitHub OIDC token: a signed statement of "I am a workflow run in FreeRADIUS/freeradius-server". The profiling leg's publish step sends that token to the prof-results store, whose access.lua verifies the signature and repo before accepting the upload. It's the only credential involved (no shared secret), and despite the name it grants no write access to the repo or anything else on GitHub.

(2) "contents: read" is what actions/checkout needs to clone the repo. It normally comes from the defaults, but declaring an explicit permissions: block sets every unlisted scope to none, so it has to be restated here.

.github/workflows/ci-multi-server-tests.yml
marc-casavant: Good idea.  Implementing the change.

src/tests/multi-server/scripts/profiling/start_valgrind_profiling.sh
marc-casavant: --separate-threads=no 
Merges all threads into one profile per process instead of one output file per thread; the analysis aggregates by function, so per-thread splits would only fragment the counts.

--separate-callers=6 
records a function's costs separately per calling chain of up to 6 callers, so a hot shared helper can be attributed to the call paths that actually drive it.  This gives us to "snip" or "fold" certain call trees based on results.  Increasing this value also substantially increases the size of the callgrind.out files generated during profiling.

scripts/ci/publish-profiling-results.sh
marc-casavant: Comments in script largely removed to only include a minimal amount of comments.

scripts/ci/publish-profiling-results.sh
marc-casavant: Removed details

scripts/ci/publish-profiling-results.sh
marc-casavant: Removed verbose comment.

scripts/ci/publish-profiling-results.sh
marc-casavant: Logic changed when implementing packaging results in a tarball.

PR #5903, opened 2026-07-13
rlm_krb5: assume thread-safety when cross-compiling

commodo:
When cross-compiling, the autoconf test program for krb5_is_thread_safe() cannot be run on the host. Previously this left krb5threadsafe empty, which is incorrect for all modern krb5 builds.

Fix by defaulting to thread-safety when the test cannot be performed.

### DIFF
--- a/src/modules/rlm_krb5/configure.ac
+++ b/src/modules/rlm_krb5/configure.ac
@@ -143,7 +143,8 @@ if test "$krb5threadsafe" != "no"; then
 	if test "x$ac_cv_lib_krb5_krb5_is_thread_safe" = xyes; then
 		AC_RUN_IFELSE([AC_LANG_PROGRAM([[#include <krb5.h>]], [[return krb5_is_thread_safe() ? 0 : 1]])],
 			[krb5threadsafe="-DKRB5_IS_THREAD_SAFE"], [AC_MSG_WARN([[libkrb5 is not threadsafe]])],
-			[AC_MSG_WARN(cross compiling: not checking)])
+			[AC_MSG_WARN(cross compiling: not checking)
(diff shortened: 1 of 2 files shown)


arr2036:
I have no issue with the change, but you need to apply it to the autoconf source first, and then regenerate the configure script.

commodo:
> I have no issue with the change, but you need to apply it to the autoconf source first, and then regenerate the configure script.

good point ; we've been patching this file in OpenWrt; i kind of went inertially to this

commodo:
> > I have no issue with the change, but you need to apply it to the autoconf source first, and then regenerate the configure script.
> 
> good point ; we've been patching this file in OpenWrt; i kind of went inertially to this

done

src/modules/rlm_krb5/configure.ac
```diff
@@ -143,7 +143,7 @@ if test "$krb5threadsafe" != "no"; then
 	if test "x$ac_cv_lib_krb5_krb5_is_thread_safe" = xyes; then
 		AC_RUN_IFELSE([AC_LANG_PROGRAM([[#include <krb5.h>]], [[return krb5_is_thread_safe() ? 0 : 1]])],
 			[krb5threadsafe="-DKRB5_IS_THREAD_SAFE"], [AC_MSG_WARN([[libkrb5 is not threadsafe]])],
-			[AC_MSG_WARN(cross compiling: not checking)])
+			[krb5threadsafe="-DKRB5_IS_THREAD_SAFE"])
```
arr2036: Well it's still not checking, it's just defaulting to thread_safe...

commodo: yep
that's the point when cross-compiling ; 

we can't check, since it's building with another target toolchain (we can't run it);

we could just leave the warning and default to thread_safe

PR #5905, opened 2026-07-16
eap_aka_sim: fix SIM-SQN check that rejected every value

paunadeu:
vector_umts_from_quintuplets() checks control.SIM-SQN by comparing vp_length
against MILENAGE_SQN_SIZE. SIM-SQN is FR_TYPE_UINT64, and vp_length is 0 for
numeric boxes, so the check can never pass:

    control.SIM-SQN incorrect length.  Expected 6 bytes, got 0 bytes
    Failed retrieving UMTS vectors

vector_umts_from_ki() reads the same attribute as vp_uint64 with no length
check, and the SQN/AK derivations just below this check read vp_uint64 too,
so it looks left over from when SQN was octets.

I replaced it with a range check rather than dropping it: SQN is 48 bits and
uint48_to_buff() masks off anything wider, which would silently give a bad AK.

This is reachable when the server has no local Ki and the quintuplets come
from somewhere else - in our case an HSS over SWx, which owns the SQN. It
also breaks SQN resync: with no Ki the server never sets SIM-SQN itself, but
RESUME(recv_aka_synchronization_failure) requires it, so recv
Synchronization-Failure has to set it, and then the retried challenge dies on
the length check.

Compiles clean. Tested against a live HSS with real test SIMs: before, the
resync MAR succeeded and the retried AKA-CHALLENGE was still rejected; after,
EAP-AKA and EAP-AKA' both recover from a desynchronised USIM.

### DIFF
--- a/src/lib/eap_aka_sim/vector.c
+++ b/src/lib/eap_aka_sim/vector.c
@@ -690,12 +690,14 @@ static int vector_umts_from_quintuplets(request_t *request, fr_pair_list_t *vps,
 
 	/*
 	 *	Fetch (optional) SQN
+	 *
+	 *	SQN is 48 bits.  uint48_to_buff() silently truncates anything
+	 *	wider, so reject it rather than derive AK from a mangled SQN.
 	 */
 	sqn_vp = fr_pair_find_by_da(vps, NULL, attr_sim_sqn);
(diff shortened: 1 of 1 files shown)


arr2036:
Yeah we used to populate it even for integers, thanks for the patch.

paunadeu:
Always welcome, thanks you for all the work.

PR #5906, opened 2026-07-20
docs-v4: remove duplicate dictionary info from the dictionary tutoria…

nolade:
…l. `xrefs` used in tutorial to guide user to reference files including the relevant configuration files.


alandekok:
Merged, thanks

PR #5907, opened 2026-07-21
Profiling CI: deterministic load pacing + automatic regression flagging (notify-only)

marc-casavant:
Two pieces that together make per-commit CPU-cycle (CEst) regression flagging trustworthy.

**Deterministic profiling pacing** (multi-server tests):

- Each suite's profiling run is its own params file (`profiling.short.ci.test.yml`,
  `profiling.extended.test.yml`)
- proto_load configuration adjusted to send packets at a slower rate which allows us to more consistent profiling results when using Valgrind. 
- all.mk partitions the targets: profiling targets run only the profiling.* tests.
- Results measured across 10 CI runs of one commit has now has a cycle estimate spread of 0.11-0.32% per suite. It was previously up to 7.4%, and the mysql profiling runs generated unusable results for accurate profiling analysis.

**Regression flagging** (workflow + `scripts/ci/flag-profiling-results.sh`):

- The profiling leg compares each suite's fresh results against the store's previous run,
  with thresholds calibrated from the measured noise floor; 

> --gate 15 sets the failure threshold at 15% (for example)
> --min-cest 10000000 (minimum amount of cycles a function needs in a run before its results get checked)
> --gate-scope \<value\>
>  - function (function cycles checked)
>  - total (total cycles of run checked)
>  - both (total cycles of the run and per function cycles checked)

- CI status includes a link to the dashboard UI
- Commit commenting is implemented (sticky comment, updated on re-runs, resolved by a clean
  run, degrades gracefully without a token) but switched OFF by default for now

**Verified**:

- Verified new make targets
- Analyzed the consistency of the profiling results on 10 CI runs
- Verified new script that checks latest CI run results with previous

### DIFF
--- a/.github/workflows/ci-multi-server-tests.yml
+++ b/.github/workflows/ci-multi-server-tests.yml
@@ -38,9 +38,11 @@ jobs:
 
     #  id-token: write lets the profiling leg mint a GitHub OIDC token, which
     #  authenticates the prof-results publish. Listing any permission drops
-    #  the unlisted ones, so contents: read is stated to keep checkout working.
+    #  the unlisted ones, so the others are stated explicitly.
+    #  contents: write (up from read) is needed so the profiling leg can post a
+    #  commit comment flagging a CEst regression; it also keeps checkout working.
     permissions:
-      contents: read
+      contents: write
       id-token: write
 
     strategy:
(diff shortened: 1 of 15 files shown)


arr2036:
- These do two very different things, they belong in different PRs.
- I'm not sure we want regression summaries in CI, especially as the output is so noisy. 

¿Que is esso? No entiendo:
Load is paced so proto_load never gates under valgrind: 25 pps (mysql 15 pps), and extended = double the send window.

PR #5910, opened 2026-07-25
Fix libbacktrace build when BUILD_DIR resolves to an absolute path

### LINKED ISSUE #5900: [defect]: version 4.0 unable to build libbacktrace
### What type of defect/bug is this?

incorrect 3rd party API usage

### How can the issue be reproduced?

Compile freeradius master branch (36b247dec2647641b3690d6fef645937e310e22a) from github with gcc 16.1.0 for OpenWrt x86_64 glibc, and getting the following : 
```
CC src/lib/util/backtrace.c
src/lib/util/backtrace.c:32:12: fatal error: freeradius-devel/backtrace/backtrace.h: No such file or directory
   32 | #  include <freeradius-devel/backtrace/backtrace.h>
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[3]: *** [scripts/boiler.mk:763: /home/nameless/works/openwrt/build_dir/target-x86_64_glibc_nameless/freeradius4-4.0.0/build/objs/src/lib/util/backtrace.lo] Error 1
```

Because of failed to compile libbacktrace, but I am unable to find the reason on the failure.

config.log from the ./src/lib/backtrace : 

[config.log](https://github.com/user-attachments/files/29928299/config.log)

In the ./src/lib/backtrace directory, Makefile and configure file are being created but somehow it doesn't reach/execute [this line](https://github.com/FreeRADIUS/freeradius-server/blob/36b247dec2647641b3690d6fef645937e310e22a/src/lib/util/libfreeradius-util.mk#L166).

### Log output from the FreeRADIUS daemon

```shell
Version 4.0 or commit 36b247dec2647641b3690d6fef645937e310e22a
```

### Relevant log output from client utilities

_No response_

### Backtrace from LLDB or GDB

```shell

```

vortexilation:
## Summary

Fixes #5900. `src/lib/util/libfreeradius-util.mk` hardcodes the literal
string `build/` for the libbacktrace object/library targets instead of
using `$(BUILD_DIR)`. `scripts/boiler.mk` only sets `BUILD_DIR` to the
relative `build` when `top_builddir` string-matches `$(PWD)`; otherwise
(e.g. cross-builds like OpenWrt/Buildroot, where `top_builddir` doesn't
match `$(PWD)` as a literal string) it canonicalizes `BUILD_DIR` to an
absolute path.

That mismatch means the order-only prerequisite meant to build the
`src/include/backtrace -> ../lib/backtrace` symlink before compiling
`backtrace.c` names a different make target than the one `boiler.mk`
actually builds under an absolute `BUILD_DIR`, so the symlink is
silently never created, and the build fails with:

    src/lib/util/backtrace.c:32:12: fatal error:
    freeradius-devel/backtrace/backtrace.h: No such file or directory

## Fix

Replace the hardcoded `build/` paths with `$(BUILD_DIR)` in the five
affected rules (the backtrace.c object prerequisite, and the four
libbacktrace.a/.la copy rules), matching the convention already used
elsewhere in the build (`scripts/install.mk`, `scripts/libtool.mk`).

## Test plan

- [x] Confirmed `WITH_BACKTRACE = yes` and `HAVE_BACKTRACE` defined
- [x] Confirmed `src/include/backtrace` symlink is created before
      `backtrace.c` is compiled when `BUILD_DIR` resolves absolute
- [x] Full build succeeds with libbacktrace enabled under an
      OpenWrt/Buildroot-style cross-compile where `top_builddir` is
      passed as an absolute path

### DIFF
--- a/src/lib/util/libfreeradius-util.mk
+++ b/src/lib/util/libfreeradius-util.mk
@@ -150,7 +150,7 @@ TGT_LDFLAGS	+= -L$(top_builddir)/build/lib/local/.libs
 src/include/backtrace:
 	cd src/include && ln -s ../lib/backtrace
 
-build/objs/src/lib/util/backtrace.$(OBJ_EXT): | src/include/backtrace
+$(BUILD_DIR)/objs/src/lib/util/backtrace.$(OBJ_EXT): | src/include/backtrace
 
 # Actually call the 'sub'-make to build libbacktrace.
 #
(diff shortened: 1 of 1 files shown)


vortexilation:
How to fix "Cannot update this protected ref." ?

alandekok:
I've merged this manually, thanks.

vortexilation:
@alandekok
Thanks a lot for considering my PR.

PR #5919, opened 2026-08-11
multi-server: one profiling run index per make invocation

marc-casavant:
This change ensures all profiling test results are generated and saved in a directory structure that uses a single run index per make invocation.  

Without this change, when tests run in parallel (-j), results from one invocation could scatter across several indexes.  This is because the previous implementation computed the index inside each test's recipe. 

The index is now calculated once at makefile parse time and before any tests are run, calculated from the workflow run number in CI, highest existing plus one locally.

There's also a minor edit in the README to fix the directory structure of the profiling results.

### DIFF
--- a/src/tests/multi-server/all.mk
+++ b/src/tests/multi-server/all.mk
@@ -12,7 +12,7 @@
 #                                   PROFILING_RESULT_PATH.
 # - PROFILING_RESULT_MODE=<ci|dev>      Profiling output layout (only meaningful when
 #                                   MODE=profiling). Default `ci`.
-#                                     ci:  PROFILING_RESULT_ROOT/<suite>/<test>/<branch>/<commit>/<run-index>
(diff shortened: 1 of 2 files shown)


src/tests/multi-server/all.mk
```diff
@@ -12,7 +12,7 @@
 #                                   PROFILING_RESULT_PATH.
 # - PROFILING_RESULT_MODE=<ci|dev>      Profiling output layout (only meaningful when
 #                                   MODE=profiling). Default `ci`.
-#                                     ci:  PROFILING_RESULT_ROOT/<suite>/<test>/<branch>/<commit>/<run-index>
+#                                     ci:  PROFILING_RESULT_ROOT/<branch>/<commit>/<run-index>/<suite>/<test>
```
arr2036, line 15: Should just be `<suite>/<test>`

Issue #5912, opened 2026-07-30
[feature]: Support 802.11r BSS Fast Transition (FT-PSK) for DPSK module


CWempe:
### What type of feature is needed?

Proposal for new functionality

### What is the feature?

We are currently looking into using the dpsk module in our wifi setup.
After some issues we finally got it working.
Thanks, @hshimomura for the [perl script](https://github.com/hshimomura/cisco_dpsk_rlm_perl).

The biggest issue was, that we configured EysyPSK on our Cisco WLAN Controller (Catalyst 9800;  v17.12.x) with FT enabled.
The screenshot in the official Cisco documentation shows that Fast Transition is "Adaptive Enabled".
https://www.cisco.com/c/en/us/td/docs/wireless/controller/9800/technical-reference/easy-psk-deployment-guide.html

But we figured out that this is not compatible with the dpsk module.
If Fast Transition is enabled the EAPol-Message is too big.
Even after we ignored [this length check,](https://github.com/hshimomura/cisco_dpsk_rlm_perl/blob/0d3079b81bbde3eb5081b89d3609f974eed06709/cisco_easy_psk_perl.pl#L100) it still did not work.
As it turned out, there seems to be a difference in the crypto calculation between PSK and FT-PSK.
(At least that is what ChatGPT told us.)

After disabling FT is worked. 👍

So maybe the module (and perl script) could be enhanced to support Fast Transition.

Or add an info in the documentation that FT needs to be disable.

hshimomura:
Thanks @CWempe, I will check on my lab if FT-PSK is not usable with dpsk module.

hshimomura:
I have tested using 17.18.4a with FT-Adaptive and FT-Enabled. My test device (intel BE200 with Windows 11) is non-FT device, so I could not find the issue. I will find FT supported device and will test using it later this week or next week. Thanks.

hshimomura:
I confirmed that the current implement doesn't support Adaptive FT for FT capable device.

PR #5921, opened 2026-08-15
Add Cisco EasyPSK preprocessing support

hshimomura:
## Summary

This change adds opt-in Cisco Catalyst 9800 EasyPSK request
normalization to `rlm_preprocess`.

When `with_cisco_easy_psk_hack = yes` is configured, the module
decodes the binary Cisco EasyPSK `Cisco-AVPair` values and stores
them in temporary request attributes:

- `cisco-wlan-ssid` -> `Tmp-String-0`
- `cisco-bssid` -> `Tmp-Octets-0`
- `cisco-anonce` -> `Tmp-Octets-1`
- `cisco-8021x-data` -> `Tmp-Octets-2`

Local policy can then map these temporary attributes to the generic
`rlm_dpsk` request attributes:

```text
cisco_easy_psk {
	if (&request:Tmp-String-0 &&
	    &request:Tmp-Octets-0 &&
	    &request:Tmp-Octets-1 &&
	    &request:Tmp-Octets-2) {
		update request {
			&Called-Station-SSID := &request:Tmp-String-0
			&Called-Station-MAC := &request:Tmp-Octets-0
			&FreeRADIUS-802.1X-Anonce := &request:Tmp-Octets-1
			&FreeRADIUS-802.1X-EAPoL-Key-Msg := &request:Tmp-Octets-2
		}
		updated
	}
	else {
		noop
	}
}
```

The conversion is implemented in C and does not require `rlm_perl`.
The new option is disabled by default.

## Motivation

The generic `rlm_dpsk` policy can normally be populated using text-based attribute manipulation. Cisco IOS XE is different: Catalyst 9800 carries the BSSID, ANonce, and EAPOL-Key data inside `Cisco-AVPair` attributes as raw binary octets rather than textual values.

Those values cannot be handled safely as strings because embedded binary data may be truncated or corrupted. This optional `rlm_preprocess` path is therefore needed to parse and copy the IOS XE values using binary-safe C operations, then expose them as the generic octet attributes consumed by `rlm_dpsk`. No Perl adapter is required.

## Relationship to PR #5830

This is a follow-up to #5830, but it does not depend on the branch
from that pull request.

The changes proposed in #5830 were integrated into the current
`v3.2.x` branch by the maintainer with some adjustments. Those
changes kept `rlm_dpsk` vendor-neutral and made generic DPSK reply
attributes available to local vendor-specific policy.

This pull request complements that design by handling only the
Cisco-specific request normalization in `rlm_preprocess`. It does
not add Cisco-specific processing to `rlm_dpsk`.

The resulting flow is:

```text
Cisco-AVPair
    -> rlm_preprocess
    -> temporary attributes
    -> local policy
    -> generic rlm_dpsk request attributes
    -> rlm_dpsk
    -> generic reply attributes
    -> local Cisco reply policy
```

## Fast Transition limitation

This implementation supports non-FT PSK handshakes only.

With Cisco Adaptive FT, non-FT stations may continue to use the
normal PSK handshake, but FT-capable stations select FT-PSK. Those
requests contain fragmented EAPoL-Key data and require the IEEE
802.11r FT key hierarchy, which is not implemented here.

Multiple `cisco-8021x-data` fragments are therefore detected and
rejected explicitly instead of silently processing an incomplete
EAPoL-Key message.

Cisco Fast Transition must be configured as Disabled when using
this implementation with FT-capable stations.

FT-PSK support remains tracked in #5912.

## Validation

Tested from the current upstream `v3.2.x` branch.

Verified:

- `rlm_preprocess` builds successfully
- existing `dpsk-pmk` and `dpsk-psk` tests pass
- Cisco non-FT EasyPSK normalization test passes
- fragmented FT-PSK input is rejected by the regression test
- `git diff --check` passes
- Debian 13 package installation and FreeRADIUS startup were validated
- Cisco Catalyst 9800 EasyPSK authentication was validated with Fast
  Transition disabled
- PSKs containing leading zeroes work correctly

Related to #5830.
Related to #5912.


hshimomura:
@alandekok This is the Cisco IOS XE request-normalization follow-up to #5830. It keeps rlm_dpsk vendor-neutral and adds only binary-safe request normalization in rlm_preprocess. I would appreciate your review when convenient.

PR #5927, opened 2026-09-02
ldap: remove redundant talloc_free in retry paths

antoniotorresm:
These calls were correct when originally added, but were made redundant
in commit https://github.com/FreeRADIUS/freeradius-server/commit/1057ef6aa329aae05964f3ce02bc4294612b85c8, as `fr_connection_reconnect` now already frees the `extra`
string via `talloc_free_children`.

Remove these calls to prevent bugs related to double-frees.


alandekok:
Thanks.  I'll pull it into v3.0.x and v3.2.x

PR #5928, opened 2026-09-03
print.c: sync list of tokens from token.h

antoniotorresm:
Commit 92f79fea7e included the prepend operator `^=`, but forgot to include it in the list of tokens present in print.c, which needs to be in sync with token.h, as stated in its comment. This error may cause an incorrect operator string when using indices from 23-27 in the list, and an out-of-bounds read when the index is 28.


alandekok:
thanks.  I'll pull this into v3.0.x and v3.2.x.

PR #5929, opened 2026-09-03
fix legacy use of attribute names in filter policy

jimdigriz:
Found a number of upstream shipped statements still use the deprecated attribute referencing.


alandekok:
I don't know that this matters a lot?  It works fine in v3, and we're removing the `&` in v4.  :(

jimdigriz:
Only submitted it as in the logs it says "please change to `&Thingy`", so it was to reduce noise.

PR #5932, opened 2026-09-07
lib: fix integer64 to signed range check

steep8:
## Summary

Fix the range check when casting `PW_TYPE_INTEGER64` to
`PW_TYPE_SIGNED`.

The code checked `src->integer`, but subsequently converted
`src->integer64`. Values larger than `INT_MAX` could therefore bypass
the intended range check.

## Fix

Check `src->integer64` against `INT_MAX`.

## Tests

Added regression tests covering:

- `INT_MAX`
- `INT_MAX + 1`
- `1ULL << 32`
- `UINT64_MAX`

The test fails with the old range check and passes with this change.

### DIFF
--- a/src/lib/value.c
+++ b/src/lib/value.c
@@ -1289,7 +1289,7 @@ ssize_t value_data_cast(TALLOC_CTX *ctx, value_data_t *dst,
 			break;
 
 		case PW_TYPE_INTEGER64:
-			if (src->integer > INT_MAX) {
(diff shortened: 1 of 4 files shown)


alandekok:
I've fixed this, thanks.  We require commits to be PGP signed.

While tests are nice, it's better to add tests for many different scenarios.  We'll take a look at it.

PR #5933, opened 2026-09-08
multi-server: interim test env update for gperftools profiling

marc-casavant:
This PR produces gperftools profiling results along side the existing valgrind/callgrind results.  The multi-server tests that currently supports this are the accept, pap-auth, ldap and mysql profiling tests.

These changes focus on setting up the multi-server test environment, configs and test templates which sets us up nicely to profile FreeRADIUS with valgrind and gperftools once start_gperftools_profiling.sh has been fully implemented (in the next update).

Update highlights:

- New TOOL makefile env variable set to "valgrind" by default and allowing user to set it to valgrind or gperftools
- accept, pap-auth, ldap and mysql testcase templates updated to call start_valgrind_profiling.sh or start_gperftools_profiling.sh based on PROFILING_TOOL env variable passed into the container
- multi-server test dockerfile environments updated to pass in PROFILING_TOOL as a new env variable
- new start_gperftools_profiling.sh script which takes care of starting and stopping gperftools profiling during the test
- new start_freeradius.sh script that the testcase templates now runs, based on the tool profiling, start_valgrind_profiling.sh or start_gperftools_profiling.sh is then run
- Every run now writes to logs and listener subdirectories named after the profiler
  - Example: 
    - build/tests/multi-server/accept/short_ci/logs/valgrind/result.log
    - build/tests/multi-server/accept/short_ci/logs/gperftools/result.log 
- README updated based on latest changes

Verification:

Locally verified proper generation of profiling results using valgrind and gperftools for the accept, pap-auth, ldap and mysql multi-server tests.  All four profiling tests generated the callgrind.out.*, callgrind_report.txt, proof_report.txt and proof.out.*.bp.gz compressed gperftools results.

### DIFF
--- a/src/tests/multi-server/all.mk
+++ b/src/tests/multi-server/all.mk
@@ -3,25 +3,31 @@
 #
 # Makefile arguments:
 # - TEST_MULTI_SERVER_DEBUG=<0-2>   debug level for multi-server test framework
-# - TEST_MULTI_SERVER_VERBOSE=<0-4>  verbosity level
+# - TEST_MULTI_SERVER_VERBOSE=<0-4> verbosity level
 # - MODE=<service|profiling>        which FreeRADIUS image to drive the tests
 #                                   with. Default `service`. `profiling` swaps in
 #                                   freeradius4-profiling/<image>:<sha>, sets PROFILING=yes
 #                                   so the test template runs the server under
-#                                   valgrind/callgrind, and writes results to
-#                                   PROFILING_RESULT_PATH.
-# - PROFILING_RESULT_MODE=<ci|dev>      Profiling output layout (only meaningful when
+#                                   the profiler selected by TOOL, and writes
+#                                   results to PROFILING_RESULT_PATH.
+# - PROFILING_RESULT_MODE=<ci|dev>  Profiling output layout (only meaningful when
 #                                   MODE=profiling). Default `ci`.
 #                                     ci:  PROFILING_RESULT_ROOT/<suite>/<test>/<branch>/<commit>/<run-index>
 #                                     dev: PROFILING_RESULT_ROOT/<suite>/<test>  (flat)
+# - TOOL=<valgrind|gperftools>      Profiler to run under MODE=profiling. Default `valgrind`.
+#                                     valgrind:               only profiles with valgrind
+#                                     gperftools:             only profiles with gperftools
+#                                     "valgrind gperftools":  runs each test once per tool, in that order
 #
 # Usage:
-#   make -f src/tests/multi-server/all.mk test.multi-server                       # all suites, service image
-#   make -f src/tests/multi-server/all.mk test.multi-server.ci                    # CI subset, service image
-#   make -f src/tests/multi-server/all.mk test.multi-server.profiling             # all suites, profiling image
-#   make -f src/tests/multi-server/all.mk test.multi-server.profiling.ci          # CI subset, profiling image
-#   make -f src/tests/multi-server/all.mk test.multi-server.accept.short_ci       # single test
-#   make -f src/tests/multi-server/all.mk clean.test.multi-server                 # clean logs
+#   make -f src/tests/multi-server/all.mk test.multi-server                                          # all suites, service image
+#   make -f src/tests/multi-server/all.mk test.multi-server.ci                                       # CI subset, service image
+#   make -f src/tests/multi-server/all.mk test.multi-server.profiling                                # all suites, profiling image
+#   make -f src/tests/multi-server/all.mk test.multi-server.profiling.ci                             # CI subset, profiling image
+#   make -f src/tests/multi-server/all.mk test.multi-server.profiling.ci TOOL=gperftools             # CI subset, gperftools instead of valgrind
+#   make -f src/tests/multi-server/all.mk test.multi-server.profiling.ci TOOL="valgrind gperftools"  # CI subset, both profilers in turn
+#   make -f src/tests/multi-server/all.mk test.multi-server.accept.short_ci                          # single test
+#   make -f src/tests/multi-server/all.mk clean.test.multi-server                                    # clean logs
 #
 
 SHELL := /bin/bash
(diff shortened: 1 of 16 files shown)


src/tests/multi-server/scripts/profiling/start_gperftools_profiling.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+# Skeleton for the gperftools capture.
```
arr2036: remove this file?

src/tests/multi-server/tests/accept/template.yml.j2
```diff
@@ -31,7 +31,12 @@ states:
               printf "  MAX_REQUESTS:     %s\n" "$TEST_LOADGEN_MAX_REQUESTS"
 
               if [ "$PROFILING" = "yes" ]; then
-                source /etc/freeradius/start_valgrind_profiling.sh
+                START_SCRIPT="/etc/freeradius/start_$PROFILING_TOOL"_profiling.sh
```
arr2036: that's a really weird place to drop scripts, should be in /usr/local/bin

arr2036: and maybe PROFILING_SCRIPT?

src/tests/multi-server/all.mk
```diff
@@ -211,37 +250,45 @@ test.multi-server.${1}.${2}: $$(TEST_MULTI_SERVER_RENDERED.${1}.${2}) ${4}/start
...
-	                echo "-- line-type counts --"; \
-	                awk '{print $$$$1}' "$$$$f" | sort | uniq -c; \
-	                echo "-- last 200 lines --"; \
-	                ;; \
-	        esac; \
-	        tail -200 "$$$$f"; \
-	    done; \
-	    exit 1; \
-	}
+	for RUN in $(TEST_MULTI_SERVER_RUNS); do \
```
arr2036: I think this really wants to live in a separate profiling tool script.  Four dollars is my limit :/

arr2036: Also why do we source a startup script, and then have this launcher wrapper?

src/tests/multi-server/tests/accept/template.yml.j2
arr2036: @claude opinon?

src/tests/multi-server/scripts/profiling/start_gperftools_profiling.sh
arr2036: @claude any clue why this is here?

src/tests/multi-server/all.mk
arr2036: @claude explain?

arr2036: @claude So what's the point of this inline script if we have those launchers? just to handle the duplicated code?

src/tests/multi-server/tests/ldap/template.yml.j2
```diff
@@ -34,7 +34,12 @@ states:
               printf "  MAX_REQUESTS:     %s\n" "$TEST_LOADGEN_MAX_REQUESTS"
 
               if [ "$PROFILING" = "yes" ]; then
-                source /etc/freeradius/start_valgrind_profiling.sh
+                START_SCRIPT="/etc/freeradius/start_$PROFILING_TOOL"_profiling.sh
+                if [ ! -r "$START_SCRIPT" ]; then
+                  echo "Profiling start script not found: $START_SCRIPT"
+                  exit 1
+                fi
+                source "$START_SCRIPT"
```
arr2036: @claude Any reason why we  source "$START_SCRIPT"?

src/tests/multi-server/all.mk
arr2036: @claude So what's the point of this inline script if we have those launchers? just to handle the duplicated code?

arr2036: @claude can you make commits on a PR branch like copilot?

src/tests/multi-server/all.mk
```diff
@@ -46,6 +52,28 @@ GIT_COMMIT         := $(or $(shell git -C $(top_srcdir) rev-parse --short HEAD 2
...
+
+#
+#  One radenv invocation per entry, sequentially, per test target. Service
+#  mode is a single run; profiling mode is one run per profiler.
+#
+TEST_MULTI_SERVER_RUNS := $(if $(filter profiling,$(MODE)),$(or $(TOOL),$(PROFILING_TOOLS)),service)
+
 #
 #  Image plumbing.
 #
```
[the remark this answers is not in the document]
marc-casavant: I agree, I'll check that TOOL env variable is not an empty string.

src/tests/multi-server/tests/ldap/template.yml.j2
marc-casavant: Prior to the latest change, the test's template would source the start_valgrind_profiling.sh script which took care of setting up the env and start/stop profiling with valgrind.

`if [ "$PROFILING" = "yes" ]; then
                source /etc/freeradius/start_valgrind_profiling.sh`

In this change, $START_SCRIPT is set to either be `/etc/freeradius/start_valgrind_profiling.sh` or `/etc/freeradius/start_gperftools_profiling.sh` based on which tool is profiling `FreeRADIUS.`

The initial intent behind this PR was to only make the necessary changes to the test configs and makefile targets that allows us to profile with both valgrind and gperftools, leaving the start_gperftools_profiling.sh as a template/placeholder, however, based on the review, I think it might be best that as part of this PR, I finish implementing start_gperftools_profiling.sh.

src/tests/multi-server/all.mk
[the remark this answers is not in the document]
marc-casavant: Logic updated to also check if TOOL is an empty string.

src/tests/multi-server/scripts/run_test.sh
```diff
@@ -0,0 +1,110 @@
...
+#  each capture script uses distinct file names, so the files do not collide.
+if [ "$MODE" = "profiling" ]; then
+  FREERADIUS_IMAGE=$FREERADIUS_PROFILING_IMAGE
+  PROFILING=yes
+  if [ "$PROFILING_RESULT_MODE" = "dev" ]; then
+    PROFILING_RESULT_PATH="${PROFILING_RESULT_ROOT}/${SUITE}/${TEST}"
+  else
+    RUN_BASE="${PROFILING_RESULT_ROOT}/${GIT_BRANCH}/${GIT_COMMIT}"
+    EXISTING=$(find "$RUN_BASE" -mindepth 1 -maxdepth 1 -type d 2>/dev/null | wc -l | tr -d ' ')
+    PROFILING_RESULT_PATH="${RUN_BASE}/$(( EXISTING + 1 ))/${SUITE}/${TEST}"
```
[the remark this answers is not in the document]
marc-casavant: The plan at the moment is to update the directory structure where the results are stored when PROFILING_RESULT_MODE=ci to be `PROFILING_RESULT_ROOT/<suite>/<test>` in a future update.  

Maybe this is something that should be done in the PR to ensure we don't run into issues instead of how I've planned this change in a later update.

arr2036:
Thanks @marc-casavant. The profiling work from this PR is now on master. I reworked the profiling work to fit the trigger-driven capture and the structure that uses `RADIUSD_COMMAND` and compose `extends`. I am closing this PR as superseded. The two commits in this PR map to the following commits.

The makefile, environment, and template updates (this PR's `8ed207ffdf`):
- `92e3920b31` drive callgrind from the `server.start` and `server.stop` triggers
- `408d20d475` take the profiling results directory from the environment
- `8853637ccc` share the freeradius compose mounts
- `7b0a06add3` rework the harness around compose `extends` and `RADIUSD_COMMAND`

The gperftools capture (this PR's `9e70b5e203`):
- `3666ca6c3c` add the `%callgrind.*` and `%gperftools.*` xlat control functions that the capture triggers call
- `bc4a29ae7d` add the gperftools capture and standardise the profiling output

Two follow-ups are on 'developer/arr2036' and reach master at the next merge:
- `934c72ae5e` add the symbolised `profile.pb.gz` and the README content from this PR
- `283e839a55` retire the unused callgrind report helper and align the profiling README

`RADIUSD_COMMAND` replaces `run_test.sh` and `start_freeradius.sh`. `RADIUSD_COMMAND` runs freeradius directly in service mode, and runs the capture script in profiling mode. Moving the profiling start and stop onto the `server.start` and `server.stop` triggers also shrank the harness. On the paths that this PR touched, the reworked version is +398/-518 against +688/-289 here. The reworked version adds about 42% fewer lines.

PR #5934, opened 2026-09-10
Fix IPv6 Ascend filter port decoding

steep8:
## Summary

Fix IPv6 Ascend Data Filter decoding to read source and destination port fields from the IPv6 filter structure.

The IPv6 decoding path was checking `srcPortComp` and `dstPortComp`, and reading the corresponding port values, through `filter->u.ip` instead of `filter->u.ipv6`.

As a result, IPv6 filters containing source or destination port conditions could be printed without those conditions.

## Fix

Use the IPv6 union member for:

* `srcPortComp`
* `srcport`
* `dstPortComp`
* `dstport`

## Tests

Added an Ascend Data Filter regression test with both source and destination IPv6 port conditions.

Before the fix the test fails with:

```text
got      : Ascend-Data-Filter = "?? in drop tcp"
expected : Ascend-Data-Filter = "?? in drop tcp srcport > 1023 dstport != 53"
```

With the fix, `make tests.unit` passes.

### DIFF
--- a/src/lib/filters.c
+++ b/src/lib/filters.c
@@ -1372,18 +1372,18 @@ void print_abinary(char *out, size_t outlen, uint8_t const *data, size_t len, in
 		p += i;
 		outlen -= i;
 
-		if (filter->u.ip.srcPortComp > RAD_NO_COMPARE) {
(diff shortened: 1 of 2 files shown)


alandekok:
thanks.  We require patches to be signed, so I've merged this in manually

PR #5935, opened 2026-09-10
unlock sock->mutex before freeing/destroying it

martinsta:
moving radius_update_listener() call out of tls_socket_close() to prevent destroying the sock mutex before unlocking it & thus avoiding segmentation fault


alandekok:
Thanks.  I've put a slightly different fix in, which cleans up the code a little bit, too.

PR #5936, opened 2026-09-11
Handle cf_section_dup allocation failure

steep8:
## Summary

Handle allocation failure in `cf_section_dup()`.

`cf_section_dup()` did not check the result of `cf_section_alloc()` before
dereferencing the returned pointer. If allocation failed, this could result
in a NULL pointer dereference.

## Fix

Check the result of `cf_section_alloc()` and return `NULL` when allocation
fails.

## Tests

Added an API regression test which calls `cf_section_dup()` with a NULL
primary section name, causing `cf_section_alloc()` to return `NULL`.

Before the fix, the test fails due to the NULL pointer dereference.

After the fix:

```text
make tests.api
```
passes.

### DIFF
--- a/src/main/conffile.c
+++ b/src/main/conffile.c
@@ -821,6 +821,7 @@ CONF_SECTION *cf_section_dup(CONF_SECTION *parent, CONF_SECTION const *cs,
 	CONF_ITEM *ci;
 
 	new = cf_section_alloc(parent, name1, name2);
+	if (!new) return NULL;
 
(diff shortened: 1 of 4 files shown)


alandekok:
On modern operating systems with over-commit, `malloc()` will essentially never fail.  If `malloc()` does fail, it means that the error is unrecoverable, and the program will likely crash anyways.

These kinds of issues are generally a low priority for us.  They have no real-world impact on running systems.

PR #5939, opened 2026-09-16
rlm_radius: don't require an update section in status_check

arr2036:
`status_check { type = Status-Server }` stops the server:

```
ERROR : mods-enabled/radius-cloud[373]: Missing update {} subsection
ERROR : mods-enabled/radius-cloud[359]: Failed evaluating configuration for module "rlm_radius"
```

RFC 5997 Section 3 does not require the administrator provide anything for a Status-Server. Message-Authenticator is added to every packet, NAS-Identifier and a NAS address are permitted, and other attributes SHOULD NOT be present. `status_check_alloc()` already supplies a NAS-Identifier when the map is empty, so a conformant packet needs no `update` section.

The change drops `CONF_FLAG_REQUIRED` from the rule and the `map_list_empty()` rejection in the parser.

`CONF_FLAG_OK_MISSING` is deliberately not set. It returns before the section is allocated, which would leave `status_check_map` uninitialised, whereas without any flag `cf_section_parse_init()` allocates an empty section and `status_check_update_parse()` still runs.

Found deploying 4.0~40585 over 4.0~39600, where a configuration that had worked for months stopped the server at config parse.


alandekok:
I'll add tests.  :(

